PR #21, opened 2015-03-10
Add policy config

yokoi-h:
added structs for containers that are defined in routing-policy.yang.
And removed the use of address-family-configuration and made use of global-address-family-configuration instead because of the conflict of list 'afi-safi'. 
This PR is also including modifications of a comment that indicates original container name on each structs.


h-naoto:

PR #22, opened 2015-03-10
support l3_ipvpn

YujiOshima:
To support L3 IPVPN, make table and destination for ipv4 vpn.


h-naoto:

fujita:
Looks like the change to .gitignore isn't necessary. There is no point to keep the commit for only go fmt stuff so I squashed two commits into one and applied. Thanks!

PR #68, opened 2015-04-15
docs: add cli command syntax document

yokoi-h:


fujita:
Let's do this with CLI code change.

PR #84, opened 2015-04-28
Fix a build issue, add an error handling on MPLS label serialization

soramichi:
The current master does not compile with the latest grpc due to its api change,
thus gobgp.pb.go is re-generated with the latest protobuf.

Also add an error handling for MPLS label serialization
when a label exceeds 3 octets, which violates RFC 7432.


h-naoto:

fujita:
Conflicted with the current head. Please rebase.

PR #109, opened 2015-05-13
packet: cosmetic

a16:
Clean up after other function style.


h-naoto:

PR #126, opened 2015-06-04
packet: fix labelDecode

soramichi:
The type of data[*] is byte, thus data[0]<<16 and data[1]<<8 become zero due to overflow no matter what values data[0] and data[1] have.


h-naoto:

fujita:
pushed, thanks!

PR #127, opened 2015-06-04
packet: fix IPAddressAS RD serialization

soramichi:
RouteDistinguisherIPAddressAS.Value.Admin is a net.IP, which is 16-byte long, while administrator sub field of an RD is 4 byte-long


h-naoto:

PR #130, opened 2015-06-06
Fix length parameters of BGP message structures

soramichi:
Length members of some structures are properly set in this fix.

These "hidden" bugs do not cause errors right now because these values are never accessed in the current gobgp (serialization methods calculate the lengths by themselves).

However they will be problematic for example in the following code:

    msg := NewBGPOpenMessage(...)
    l := msg.OptParamLen // Oops! msg.OptParamLen is 0 regardless of the contents of msg
    // do something using l


h-naoto:

PR #134, opened 2015-06-09
scenario_test: build gobgp only at the beginning of the policy scenar…

yokoi-h:
improved the speed of policy scenario test.


fujita:
thanks!

PR #175, opened 2015-07-03
server: fix race in fsm

ishidawataru:
When fsm state goes to idle from established, fsm.keepaliveTicker is set
to nil.
This can happen before <-h.t.Dying() case in sendMessageloop()


fujita:
This doesn't fix the race (still the race between L731 and L732). Fix the race with commit 72d1fc4cedd233861df725a9eac5f18022f92125.

PR #201, opened 2015-07-25
proposing a split function for use with bufio

ramrunner:
This split function allows the use of bufio.Scanner to read a set of MRT messages. are you guys ok?


h-naoto:

yokoi-h:
add to whitelist

fujita:
Looks like a nice addition. Pushed, thanks!
My trivial comment is that always using len(data) might be better than using cap() too.

ramrunner:
thanks fujita! your comment is true. will keep it in mind for future usage!
thanks again guys.

PR #286, opened 2015-08-27
Mrt time helper func

ramrunner:
Just a small helper func to give you golang time from the MRT header timestamp.


fujita:
I've just rebased and pushed, thanks!

PR #290, opened 2015-08-28
packet: update reference to 'draft-haas-idr-flowspec-redirect-rt-bis'

j-kato:
The draft was promoted to IETF WG item.

It is already enqueued to RFC editors in Aug. 2015. It's expected to be published as RFC soon.

Signed-off-by: Jun-ya Kato <kato.junya@lab.ntt.co.jp>


h-naoto:

PR #314, opened 2015-09-04
allows to configure a local address for each bgp session

tamihiro:
iBGP sessions are in most cases initiated with a loopback address of each router.
This patch allows to configure a specific LocalAddress in each Neighbors.NeighborList.NeighborConfig stanza, which to be used in connect() function.

### DIFF
--- a/config/bgp_configs.go
+++ b/config/bgp_configs.go
@@ -719,6 +719,8 @@ type NeighborConfig struct {
 	// original -> bgp:neighbor-address
 	//bgp:neighbor-address's original type is inet:ip-address
 	NeighborAddress net.IP
+	// local address of BGP session, typically a loopback address, might be configured in some cases.
+	LocalAddress net.IP
 }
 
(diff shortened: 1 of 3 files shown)


h-naoto:

ishidawataru:
Hello @tamihiro 

Thank you for the PR!

Could you use https://github.com/osrg/gobgp/blob/master/config/bgp_configs.go#L527 instead of https://github.com/osrg/gobgp/pull/314/files#diff-b4479553130fb5ef91581f2f08da5ca8R723 ?

Since `bgp_configs.go` is generated from OpenConfig yang model by this tool (https://github.com/osrg/gobgp/tree/master/tools/pyang_plugins), we must avoid
modifying `bgp_configs.go` directly.

fujita:
add to whitelist

tamihiro:
My apologies... should have taken a better look at the relevant code before submitting a request.
I'll submit another request shortly after revoking this one.

PR #317, opened 2015-09-04
allows to configure a local address for each bgp session (second try)

tamihiro:
if Neighbors.NeighborList.Transport.TransportConfig.LocalAddress is configured, use this address as the source of BGP session.


fujita:
Thanks!
This implements only active connection side. I implemented passive connection side and pushed it with your patch. Can you try the latest code?

PR #321, opened 2015-09-04
Fix some issues which broke C library compilation with C++ compiler

pavel-odintsov:
So I'm C++ guy and I have found some compatibility issues with g++. Please review code and merge it. 

Thanks!

### DIFF
--- a/gobgp/lib/lib.go
+++ b/gobgp/lib/lib.go
@@ -34,7 +34,7 @@ package main
 //	path* p;
 //	int cap = 32;
 //	p = (path*)malloc(sizeof(path));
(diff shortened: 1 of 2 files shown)


h-naoto:

fujita:
add to whitelist

ishidawataru:
Hi.
The signature of `free_path()` in the header file and the source file is different.
Could you fix it?

Also `fmt.Println()` can be deleted instead of comment out (my bad..)

pavel-odintsov:
Hello!

Thanks for feedback. Fixed!

fujita:

PR #365, opened 2015-09-23
tools: change for renaming api-name of package/service to gobgpapi

ttsubo:
Needs to change it because of renaming name from Grpc to GobgpApi.


h-naoto:

fujita:
add to whitelist

fujita:
Pushed, thanks!

PR #377, opened 2015-09-30
Update C++ API examples

pavel-odintsov:
Hello!

I have enhanced C++ client library. Now we could call any important functions for IPv4 unicast and IPv4 Flow Spec. I have updated code for new service / packages names.


fujita:
Thanks! You might want to fix the author of the commit?

Author: root <root@ubuntu-14.04-x86_64.test.com>

PR #383, opened 2015-10-01
C++ client library update

pavel-odintsov:
Hello!

I have enhanced C++ client library. Now we could call any important functions for IPv4 unicast and IPv4 Flow Spec. I have updated code for new service / packages names.

# Second try


fujita:
Pushed, thanks!

pavel-odintsov:
Awesome! Thanks :)

PR #418, opened 2015-10-14
*: fix Errorf call with missing format verbs

a16:


h-naoto:

fujita:
add to whitelist

fujita:
Pushed, thanks!

PR #428, opened 2015-10-16
server: fix the number of accepted routes

fujita:
We need to update path.filtered before path is inserted into adjin.


fujita:
test this please

PR #454, opened 2015-10-23
policy: use radix-tree for prefix-list matching

ishidawataru:
with simple performance test https://gist.github.com/ishidawataru/fe79cd690a4d97030ae6

before 
./inject  0.20s user 0.12s system 1% cpu 19.146 total

after 
./inject  0.25s user 0.09s system 44% cpu 0.788 total

### DIFF
--- a/table/policy.go
+++ b/table/policy.go
@@ -16,6 +16,7 @@
 package table
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


ishidawataru:
test this please

PR #456, opened 2015-10-24
speed up policy matching

ishidawataru:
1. use radix-tree for prefix matching
2. stop using regexp for simple as-path-set matching


ishidawataru:
test this please

PR #490, opened 2015-11-11
doc: correct the example of rib add command

hzhou8:
Network prefix is required rather than just IP.


h-naoto:

fujita:
Thanks!

PR #540, opened 2015-12-10
make it easy to use gobgp just as a library 

ishidawataru:
several improvements to make it easy to use gobgp as a go native bgp library.
will add some example code to show how to use.


ishidawataru:
If do so, we have to specify 179 for the port number every time when we want to enable listening.
I think most of the case is listen on 179 or don't listen, so having the flag to disable listening may be more clear than rely on port number value.

fujita:
Why? Only when an user explicitly set the port to zero, gobgpd doesn't listen. If an user doesn't set, gobgpd listens on 179.

ishidawataru:
we can do that by configuration file but not through grpc api, because the default value of uint32 is 0.

fujita:
Oops. Then, why we can't use a negative like -1 to disable listening?

PR #625, opened 2016-01-15
rpki: fix serial notify handling

fujita:


fujita:
please test this

PR #685, opened 2016-02-11
rpki: use io instead of bufio

fujita:
Seems that sometimes it fails to get the proper data from a socket.


fujita:
test this please

PR #696, opened 2016-02-16
mrt: ability to skip entries when injecting

vincentbernat:
When we want to import a partial view from the same full view on two
different routers, it can be useful to have a different view for each
router, hence the ability to skip the first records.

Signed-off-by: Vincent Bernat <vincent@bernat.im>


h-naoto:

fujita:
add to whitelist

fujita:
Looks ok. Pushed, thanks!

PR #723, opened 2016-02-28
support active connection with md5 protection

ishidawataru:


ishidawataru:
this doesn't work with other bgp daemon(quagga).
will investigate and respin

Issue #705, opened 2016-02-20
Invalid order of BGP attributes in update message


pavel-odintsov:
Hello, folks!

According to RFC https://www.ietf.org/rfc/rfc4271.txt all BGP attributes in BGP message should be ordered in ascending order of attribute types:
```bash
   The sender of an UPDATE message SHOULD order path attributes within
   the UPDATE message in ascending order of attribute type.  The
   receiver of an UPDATE message MUST be prepared to handle path
   attributes within UPDATE messages that are out of order.
```

But actually you've order they other way.

I've GoBGP dynamic library and call it from C++.

```C
 path* path_c_struct = serialize_path(ipv4_unicast_route_family, (char*)"11.22.33.44/22 nexthop 11.22.33.44");

    std::cout << "GoBGP's serialize_path output" << std::endl;
    for (int path_attribute_number = 0; path_attribute_number < path_c_struct->path_attributes_len; path_attribute_number++) {
        // We store attribute type in second byte of attribute
        std::cout
            << "Attribute number " << path_attribute_number
            << " attribute length " << path_c_struct->path_attributes[path_attribute_number]->len
            << " attribute type " << uint32_t(path_c_struct->path_attributes[path_attribute_number]->value[1]) << std::endl;
    }   
```

I could share output of this piece of code:
```C
./gobgp_native_api_client_test 
GoBGP's serialize_path output
Attribute number 0 attribute length 7 attribute type 3
Attribute number 1 attribute length 4 attribute type 1
```

As you can see attributes ordered in wrong order. Attribute with type 1 (BGP_ATTRIBUTE_ORIGIN) should be placed before BGP attribute with type 3 (BGP_ATTRIBUTE_NEXT_HOP).

What do you think about this?

pavel-odintsov:
Actually, it's not critical and well written BGP clients could handle unordered attributes. But I prefer do everything according to standard rules :)

pavel-odintsov:
I have checked BGP message's attribute sequence for Bird and Juniper (srx240):
<img width="1027" alt="juniper announce to bird" src="https://cloud.githubusercontent.com/assets/2744166/13197362/4f7015c8-d7fd-11e5-90fa-2ac6ab06eceb.png">
<img width="976" alt="intel announce to juniper" src="https://cloud.githubusercontent.com/assets/2744166/13197363/4f741100-d7fd-11e5-83c8-f8b5fc8aa0c9.png">

They both uses ascending attribute order.

[...]

PR #727, opened 2016-03-01
Doc

ishidawataru:


fujita:
Looks like a regression https://github.com/osrg/gobgp/issues/705

PR #743, opened 2016-03-04
packet: fix FlowSpecComponent Len()

fujita:
We don't need to serialize to get the length of FlowSpecComponent.

Also we need to fix FlowSpecComponent()'s serialize bug that it
wrongly assumes that the length can be calculated by it's value.


fujita:
test this please

Issue #756, opened 2016-03-09
Ability to change Local Settings via gRPC


shevchenkodenis:
Hello.
As far as I understand, currently we can change Config.As and Config.RouterId only by changing config file and sending SIGHUP. 
It would be pretty useful to have an ability to do it via GRPC request "REQ_MOD_GLOBAL_CONFIG". Maybe by applying "api.Operation_REPLACE" operation.

ishidawataru:
Hi,
Can you explain about the use case?
Currently, GoBGP doesn't support Config.As and Config.RouterId modification even by changing config file. If you want to change that, you have to restart GoBGP.

fujita:
From the perspective of BGP, you can't change them dynamically (that is, with established peers)? Other peers use them for best path selection. So if you change them, you need to make sure that gobgpd shuts down all the established peers. I think that restarting gobgpd is simpler. We could add a new API for such.

shevchenkodenis:
You can see the attached pcap in pull request. Gobgp sends notification message to old peers and gracefully disconnects them.
Anyway if you think that this method is not fit enough, could you please add grpc api for editing local as and router id and then restarting server?

shevchenkodenis:
Or just restart server to non-configured state.

shevchenkodenis:
closed #768

PR #758, opened 2016-03-09
grpc: edit local AS and Router Id by OPERATION_REPLACE request

shevchenkodenis:
Solution for #756 
Was successfully tested on three virtual machines by switching one peer between remaining two peers without server restarting. Pcap in attachment.

[grpc-mod-locals.zip](https://github.com/osrg/gobgp/files/165381/grpc-mod-locals.zip)


shevchenkodenis:
closed via #768

PR #780, opened 2016-03-17
Remove offending argument to avoid ValueError

tamihiro:
The relevant field of Arguments message has been renamed to "family" which makes this method call cause ValueError. However, the "family" is not a required field, and therefore removing this argument seems like a better fix.


fujita:
Pushed, thanks!

PR #783, opened 2016-03-18
*: support Opaque Signaling

ishidawataru:
see https://tools.ietf.org/html/draft-lapukhov-bgp-opaque-signaling-01

using 16397 for AFI, 241 for SAFI, 41 for Opaque Path Attribute
temporarily.

$ gobgp global rib add -a opaque key hello value world


ishidawataru:
test this please

PR #802, opened 2016-03-29
Modify packet directory to bgp same as it's package name,make godef happy

gonotes:
Currently, in the packet directory where is actually package bgp.
First of all, it make code confusing, then it make  editor like vim(with vim-go plugin) which use godef cannot find the symbol in package bgp, not convenient for code reading.
So this PR wants change directory name and it's refers.


fujita:
Thanks for the suggestion!

I think that "packet" is a commonly used name for on-wire format, e.g., gopacket. Having "bgp" as a top directory is also confusing because after all most of code is about bgp, I think.

How about moving *.go in packet/ into packet/bgp/? It makes godef happy and makes clear that it contains the files for on-wire format. Here's a PR for that:

https://github.com/osrg/gobgp/pull/813

gonotes:
Great solution, make code structure stay clear and godef happy!

PR #804, opened 2016-03-30
Errors while compile gRPC client for C++

Anito23:
Correction of errors during compilation of gRPC client for C++.


pavel-odintsov:
LGTM

fujita:
Pushed, thanks!

PR #833, opened 2016-04-11
Add neighbor-port-number to bgp neighbor config

TakeshiTseng:
Allow user to setup neighbor BGP TCP port (Default is 179)


fujita:
Looks good. Thanks a lot!

PR #845, opened 2016-04-18
config: add set-next-hop option

tamihiro:
next-hop path attribute can be added/deleted on a per policy basis.


fujita:
Pushed, thanks!

PR #847, opened 2016-04-18
config: fix community-set match all evaluation

tamihiro:
Under the following condition:
```
[policy-definitions.statements.conditions.bgp-conditions.match-community-set]
  community-set = cs1
  match-set-options = "all"
```
match evaluation should return true if all the elements of cs1 are included in the community-set path attr.

For instance, with the aforementioned condition if cs1 = [c0, c1, c2] and the actual community path attr = [c0], it should be evaluated as false. However the current code without this patch returns true.

It was necessary to revert the previous commit 76525a9, which would cause erroneous regexp evaluations, for instance matching ”65000:100" against "65000:1" as true. 

Changes in gobgp/cmd/policy.go are meant to maintain legibility in cli output.


tamihiro:
closing this PR, please merge #859 instead... sorry for any confusion.

PR #907, opened 2016-05-11
zebra: support of debug log for zapi

ttsubo:
"PANIC=runtime error: index out of range", due to defining "ifindex" field.


fujita:
Pushed, thanks!

PR #922, opened 2016-05-13
grpc: fix python grpc client to incorporate gobgp.proto updates

tamihiro:
Example client scripts should catch up with the changes in [f9a9c41](https://github.com/osrg/gobgp/commit/f9a9c410c3b221b515bad9fcf154990be7bb70fa).
This PR is only good for the Python example :)
Also included is a timeout error handling.


fujita:
Thanks a lot!

PR #969, opened 2016-06-07
collector: ping InfluxDB and create database at startup

nlewo:
Ping InfluxDB to verify the connection at collector initialization and
create the database.


fujita:
Nice, thanks!
Very surprised that someone already uses this undocumented feature!

nlewo:
Yes, I was also surprised to see this feature since we were thinking to dump routes to an elasticsearch! Perfect timing;)

Thanks for the merge.

Issue #975, opened 2016-06-10
Dump routes to InfluxDB


nlewo:
### Context
We want to dump L3VPN routes to the InfluxDB, and more generally dump
any kind of routes.

### Proposal
To dump routes to the table and be able to query this table, we could
create a flat representation of routes, ie a structured representation
without nesteed data structures. Basically, the flat representation of
a route could consist of a map where key and value could be strings.

A partial example of a L3VPN path Flat representation:
```
{'prefix':` '192.168.1.2', 'nlri': '10.0.0.2:3:192.168.1.2', 'prefixLen': 32, 'routeTarget': '60:80'}
```

### Why another representation
Considering the Path data structure, there is currently a String
representation and a JSON representation. The string representation is
not useful since we can not easily store it in several columns. The
JSON representation is too complex: for instance, it quite hard to
store a list of dict in a column... and be able to query it!
We then need a third representation which could be this Flat
representation.

### Other usages
This representation could also be used by the CLI. The user could
then easily creates custom outputs by selecting a subset of keys. For
instance, suppose the user is only interested in prefix and nlri, she could do 
```
gobgp global rib --key nlri --key prefix
```

### Potential Roadmap

1. Add the Flat method to AddPrefixInterface and provide implementations for unicast and L3VPN (PR #971);
2. Use the Flat method to dump path in the collector (PR #971);
3. Add Flat method to other objects (Path, Route).

### Potential Bonus

1. Add a Flat json output to CLI
2. Add arguments to the CLI to build custom output formats.

fujita:
We finished this. Please reopen or create new one if not.

PR #980, opened 2016-06-14
Collector dumps ECs 

nlewo:
Related to #975

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -7086,10 +7086,16 @@ func (e *MessageError) Error() string {
 }
 
 func (e *TwoOctetAsSpecificExtended) Flat() map[string]string {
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks!
- I'm not sure we can "Flat" all the types of ECs. But let's see.
- I might like to move FlatUpdate() to collector.go later until we find another user of the function.

nlewo:
@fujita, concerning your first point, are you thinking to a particular EC?

Concerning your second point, FlatUpdate is used by a method in the bgp package. But it could be moved in an common util package.

fujita:
About the first point, just a hunch. No idea now.
About the second one, as you suggested, I thought about sorta utility package. The current code is fine by me now. Let's leave it alone until we hit a problem.

PR #986, opened 2016-06-17
collector: empty aspath when route is withdrawn

### LINKED ISSUE #985: Collector error when route is withdrawn
[signal 0xb code=0x1 addr=0x28 pc=0x696fe9]

goroutine 26 [running]:
panic(0xc717a0, 0xc82000e170)
        /usr/lib/go/src/runtime/panic.go:464 +0x3e6
github.com/osrg/gobgp/packet/bgp.(*PathAttributeAsPath).String(0x0, 0x0, 0x0)
        /home/aeiche/gocode/src/github.com/osrg/gobgp/packet/bgp/bgp.go:4231 +0x39
github.com/osrg/gobgp/server.path2data(0xc8202d86c0, 0x0, 0x1)
        /home/aeiche/gocode/src/github.com/osrg/gobgp/server/collector.go:95 +0x5c
github.com/osrg/gobgp/server.(*Collector).writeUpdate(0xc82019ed20, 0xc82029ae70, 0x0, 0x0)
        /home/aeiche/gocode/src/github.com/osrg/gobgp/server/collector.go:140 +0x170
github.com/osrg/gobgp/server.(*Collector).loop(0xc82019ed20)
        /home/aeiche/gocode/src/github.com/osrg/gobgp/server/collector.go:187 +0x312
created by github.com/osrg/gobgp/server.NewCollector
        /home/aeiche/gocode/src/github.com/osrg/gobgp/server/collector.go:231 +0x4fc

nlewo:
Fixes #985

### DIFF
--- a/server/collector.go
+++ b/server/collector.go
@@ -91,9 +91,14 @@ func (c *Collector) writePeer(msg *watcherEventStateChangedMsg) error {
 }
 
 func path2data(path *table.Path) (map[string]interface{}, map[string]string) {
-	fields := map[string]interface{}{
-		"ASPath": path.GetAsPath().String(),
+	fields := map[string]interface{}{}
+
+	if asPath := path.GetAsPath(); asPath == nil {
+		fields["ASPath"] = ""
+	} else {
+		fields["ASPath"] = asPath.String()
 	}
+
 	if origin, err := path.GetOrigin(); err == nil {
 		typ := "-"
 		switch origin {


nlewo:
Don't know if I should handle nil pointer inPathAttributeAsPath.String method, or if we generally consider String method should never be called on nil pointer.

What is the best practice?

server/collector.go
```diff
@@ -91,9 +91,14 @@ func (c *Collector) writePeer(msg *watcherEventStateChangedMsg) error {
 }
 
 func path2data(path *table.Path) (map[string]interface{}, map[string]string) {
-	fields := map[string]interface{}{
-		"ASPath": path.GetAsPath().String(),
+	fields := map[string]interface{}{}
+
+	if asPath := path.GetAsPath(); asPath == nil {
+		fields["ASPath"] = ""
```
fujita: How about removing the above line? IOW, if path.GetASPath() is nil, do nothing.

nlewo:
We need to write at least one field to Influxdb. Currently the update table contains 2 fields (ASPath, Origin)  and the field Origin can be omitted.  So if we also omit ASPath, the insert is not valid.

To always have at least one field, I then write an empty string.

fujita:
Sorry about the delay,

How about using RouterID for a field, which always exists and isn't commonly-queried metadata:

https://github.com/osrg/gobgp/pull/988/commits/fb8b7fb3973e6995f7610f03a501a2d0bdb1b2db

fujita:
Pushed the fix.

nlewo:
Thanks for the fix, better like this.

Issue #991, opened 2016-06-27
panic: interface conversion: bgp.AddrPrefixInterface is *bgp.FlowSpecIPv4VPN, not *bgp.FlowSpecIPv6VPN


tmmorin:
I have the following crash:

root@l-n4c-osbgpadm:~# gobgpd -f gobgp.conf -l debug
{"level":"info","msg":"gobgpd started","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"finished reading the config file","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Peer 127.0.0.1 is added","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Add a peer configuration for 127.0.0.1","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Peer 192.168.80.38 is added","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Add a peer configuration for 192.168.80.38","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Peer 192.168.80.37 is added","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Add a peer configuration for 192.168.80.37","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Peer 192.168.80.1 is added","time":"2016-06-27T09:58:15+02:00"}
{"level":"info","msg":"Add a peer configuration for 192.168.80.1","time":"2016-06-27T09:58:15+02:00"}
{"Duration":0,"Key":"192.168.80.38","Topic":"Peer","level":"debug","msg":"IdleHoldTimer expired","time":"2016-06-27T09:58:15+02:00"}
{"Duration":0,"Key":"192.168.80.37","Topic":"Peer","level":"debug","msg":"IdleHoldTimer expired","time":"2016-06-27T09:58:15+02:00"}
{"Key":"192.168.80.37","Topic":"Peer","level":"debug","msg":"state changed","new":"BGP_FSM_ACTIVE","old":"BGP_FSM_IDLE","reason":"idle-hold-timer-expired","time":"2016-06-27T09:58:15+02:00"}
{"Duration":0,"Key":"127.0.0.1","Topic":"Peer","level":"debug","msg":"IdleHoldTimer expired","time":"2016-06-27T09:58:15+02:00"}
[...]

PR #994, opened 2016-06-27
Fix inversion between v4 and v6 in ValidateAttribute

tmmorin:
This fixes a trivial inversion of v4/v6 calls
in ValidateAttribute.

Bug-Url: #991


fujita:
Pushed, thanks!

PR #1004, opened 2016-07-01
bgp: fix bug of flowspec vpn parser/serializer

### LINKED ISSUE #997: missing RD field in FlowSpec VPN code
RFC5575 says:

> The NLRI format for this address family consists of a fixed-length Route Distinguisher field (8 bytes) followed by a flow specification, following the encoding defined in this document.  The NLRI length field shall include both the 8 bytes of the Route Distinguisher as well as the subsequent flow specification.

However, GoBGP currently parses a FlowSpecIPv4VPN NLRI like a FlowSpecIPv4 route, not parsing the 8 first bytes as a Route Distinguisher. Same for creating/encoding.  

(The same applies to IPv6)

ishidawataru:
close #997


tmmorin:
Hi Wataru!
I have done minimal testing and with this fix gobgp now successfully parses and route-reflects flow-vpnv4 routes generated by ExaBGP (instead of failing to parse and resetting the session).
So I think it is worth merging this fix.

Thanks a lot.

Issue #1006, opened 2016-07-01
global rib <ip> longer-prefixes functionality


benagricola:
I've been testing out the 'longer-prefixes' route lookup today to implement a feature on my app that looks for more specific routes, but I've found some weirdness with how I was expecting the functionality to work and how it actually works in practice.

Taking an example from JunOS for example, the use of 'orlonger' on a routing policy prefix-list would mean the following:

`1.2.0.0/19 orlonger` would encompass:

- `1.2.0.0/20`
- `1.2.9.0/21`
- ...
- `1.2.2.0/23`
- `1.2.3.0/24`

And so on. Essentially any query for `1.2.0.0/19 orlonger` would return any subnet range which fits wholly inside the given search, if specified as a min and max integer.

I've tested this in GoBGP in the following manner, by adding the below routes to the global rib:

```
    Network             Next Hop             AS_PATH              Age        Attrs
*>  1.2.0.0/19          1.1.1.2              xxxxx               00:06:32   [{Origin: i} {Communities: xxxxx:100}]
*>  1.2.0.0/20          1.1.1.2              xxxxx               00:06:44   [{Origin: i} {Communities: xxxxx:100}]
*>  1.2.0.0/21          1.1.1.2              xxxxx               00:06:48   [{Origin: i} {Communities: xxxxx:100}]
*>  1.2.0.0/22          1.1.1.2              xxxxx               00:07:45   [{Origin: i} {Communities: xxxxx:100}]
*>  1.2.2.0/23          1.1.1.1              xxxxx               01:41:22   [{Origin: ?} {Communities: xxxxx:100}]
```

I then proceed to query for all routes longer than `1.2.0.0/19`:

```
$../gobgp -p 50051 global rib 1.2.0.0/19 longer-prefixes
    Network             Next Hop             AS_PATH              Age        Attrs
*>  1.2.0.0/19          1.1.1.2              xxxxx               00:15:16   [{Origin: i} {Communities: xxxxx:100}]
*>  1.2.0.0/20          1.1.1.2              xxxxx               00:15:28   [{Origin: i} {Communities: xxxxx:100}]
*>  1.2.0.0/21          1.1.1.2              xxxxx               00:15:32   [{Origin: i} {Communities: xxxxx:100}]
*>  1.2.0.0/22          1.1.1.2              xxxxx               00:16:29   [{Origin: i} {Communities: xxxxx:100}]
```

[...]

PR #1008, opened 2016-07-11
Fix longer-prefix search using radix trie walk

benagricola:
Bug-Url: #1006


fujita:
Pushed, thanks! Your another commit can be applied cleanly with this commit. Can you update it against the current master? If you have two related commits, probably easier for you to create one pull request including two commits.

PR #1009, opened 2016-07-11
Add shorter prefix search

benagricola:
Adds a shorter-prefixes search mode that finds any identical-or-less-specific routes
than the input values.

Bug-Url: #1006

Signed-off-by: Ben Agricola <bagricola@squiz.co.uk>


fujita:
Squashed commits and pushed, thanks a lot!

PR #1011, opened 2016-07-12
mrt: skip MRT errors instead of just dying

vincentbernat:
When trying to inject a full view from the router views project, it
seems that some entries have an unexpected flag (transitive flag
missing). Just skip those entries instead of dying.


fujita:
Pushed, thanks!

PR #1012, opened 2016-07-12
mrt: add an option to only select the best path

vincentbernat:
Not to be merged! This is a feature request with an incomplete implementation.

In fact, we assume the best path is the first one. There is no such
guarantee but since the main goal is to use less memory, this may work
for some people. Ideally, we should reuse what is done in
destination/path.go but the types are not compatible. Unsure about the
best way forward here. There is an obvious optimisation to not loop over
all entries if we need only the first, but the goal would be sort
correctly and this optimisation wouldn't hold.

The other problem is that we assume that each MRT segment will carry a
new prefix. This is usually true for MRT dumps. I don't think this is
important.

### DIFF
--- a/gobgp/cmd/common.go
+++ b/gobgp/cmd/common.go
@@ -99,8 +99,7 @@ var actionOpts struct {
 }
 
 var mrtOpts struct {
-	OutputDir  string
-	FileFormat string
+	BestPath bool `short:"b" long:"best-path" description:"only keep best path routes"`
 }
 
 func formatTimedelta(d int64) string {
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks!

fujita:
Oops, I should have closed another one. What's the blocker to use table/*.go ?

vincentbernat:
From my understanding, this is not the same tables. So you have to convert from one struct to another. Maybe there is something smarter to do?

fujita:
Added the feature. Please let me know if it works well for you.

https://github.com/osrg/gobgp/commit/55f473d7fe07a72dbddcc424d4d6bce76e609e1b

vincentbernat:
Thanks!

PR #1020, opened 2016-07-28
Everything uses log.WithFields

ramrunner:
Hi there. all modules now uses log.WithFields. i also fixed some capitalization issues, provided
some clean error messages and corrected the usage on the formatting variants (like Errorf Debugf) in places they were not needed.


ramrunner:
able to merge now with osrg/master HEAD

PR #1038, opened 2016-08-01
Change graceful restart tuple (afi/safi/flags) parsing to use capability lenth

### LINKED ISSUE #1037: Graceful restart capability is parsed incorrectly causing malformed BGP OPEN messages with BMP
BGP OPEN capabilities can be encoded as a single optional parameter with an array of capabilities or as individual optional parameters.   When using the latter, the optional parameter value is the size of the a capability.  This can enable bad practice with parsing based on parameter length instead of capability length because it would work if only one capability was encoded per optional parameter.   When multiple capabilities are encoded in a single optional parameter, capability length must be used to correctly parse capabilities.   

In the case of graceful restart, the goBGP is using the remaining optional parameter data length size instead of the capability length.  This results in malformed BMP received OPEN messages (PEER_UP), as well as gRPC/gobgp neighbor details, because the serialization of the graceful restart capability erroneously has a tuple size of more than it should.  

*I have a fix for this and will submit a pull request shortly.*

### To reproduce:

Send the below OPEN message:

> IOS XR does not have this issue because IOS XR encodes using multiple optional parameters.  We found this issue with Arista peering, where one optional parameter is used to convey the capabilities.  Below is an OPEN message from Arista. 

```
0000000: ffff ffff ffff ffff ffff ffff ffff ffff  ................
0000010: 0037 0104 fde8 0005 ac14 00c0 1a02 1801  .7..............
0000020: 0400 0100 0102 0040 0201 2c41 0400 00fd  .......@..,A....
0000030: e845 0400 0101 01
```

The above ```40 0201 2c``` is the graceful restart with a length of 2, no AFI/SAFI/Flags included.  The current code will attempt to decode this with a length of 14 because it's looking at the remaining data length of the optional parameter.  I've tested this with AFI/SAFI/Flags included and the same problem results as you would expect.   

### Examples

**Below shows incorrect parsing where no afi/safi/flags were encoded but three show up.**

```json
  {
        "conf": {
            "PrefixLimits": null,
            "id": "172.20.0.192",
            "local-as": 65000,
            "local_cap": [
                {
                    "code": 2
                },
                {
                    "code": 1,
                    "value": 65537
                },
                {
                    "code": 65,
                    "value": 65000
                }
            ],
            "remote_as": 65000,
            "remote_cap": [
                {
                    "code": 1,
                    "value": 65537
                },
                {
                    "code": 2
                },
                {
                    "code": 64,
                    "flags": 0,
                    "time": 300,
                    "tuples": [
                        {
                            "flags": 0,
                            "route_family": 1090781184
                        },
                        {
                            "flags": 4,
                            "route_family": 4259840069
                        },
                        {
                            "flags": 1,
                            "route_family": 65537
                        }
                    ]
                },
                {
                    "code": 65,
                    "value": 65000
                },
                {
                    "code": 69,
                    "mode": 1,
                    "value": 65537
                }
            ],
            "remote_ip": "10.20.254.251"
        }
```

**Below shows the capablity correctly with the pull request/fix applied**
```json
[
    {
        "conf": {
            "PrefixLimits": null,
            "id": "172.20.0.192",
            "local-as": 65000,
            "local_cap": [
                {
                    "code": 2
                },
                {
                    "code": 1,
                    "value": 65537
                },
                {
                    "code": 65,
                    "value": 65000
                }
            ],
            "remote_as": 65000,
            "remote_cap": [
                {
                    "code": 64,
                    "flags": 0,
                    "time": 300,
                    "tuples": null
                },
                {
                    "code": 65,
                    "value": 65000
                },
                {
                    "code": 69,
                    "mode": 1,
                    "value": 65537
                },
                {
                    "code": 1,
                    "value": 65537
                },
                {
                    "code": 2
                }
            ],
            "remote_ip": "10.20.254.251"
        },
        "conf": {
            "PrefixLimits": null,
            "id": "10.1.1.254",
            "local-as": 65000,
            "local_cap": [
                {
                    "code": 2
                },
                {
                    "code": 1,
                    "value": 65537
                },
                {
                    "code": 65,
                    "value": 65000
                }
            ],
            "remote_as": 65000,
            "remote_cap": [
                {
                    "code": 65,
                    "value": 65000
                },
                {
                    "code": 64,
                    "flags": 0,
                    "time": 400,
                    "tuples": [
                        {
                            "flags": 0,
                            "route_family": 65537
                        }
                    ]
                },
                {
                    "code": 69,
                    "mode": 3,
                    "value": 65537
                },
                {
                    "code": 1,
                    "value": 65537
                },
                {
                    "code": 128
                },
                {
                    "code": 2
                }
            ],
            "remote_ip": "172.29.50.198"
        }
   }
]
```

TimEvens:
Change graceful restart tuple (afi/safi/flags) parsing to use capability length instead of using the remaining parameter (data) length. 

Fixes #1037

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -362,12 +362,18 @@ func (c *CapGracefulRestart) DecodeFromBytes(data []byte) error {
 	c.Flags = uint8(restart >> 12)
 	c.Time = restart & 0xfff
 	data = data[2:]
-	c.Tuples = make([]*CapGracefulRestartTuple, 0, len(data)/4)
-	for len(data) >= 4 {
-		t := &CapGracefulRestartTuple{binary.BigEndian.Uint16(data[0:2]),
-			data[2], data[3]}
-		c.Tuples = append(c.Tuples, t)
-		data = data[4:]
+ 
+	var valueLen int = int(c.CapLen) - 2
+
+	if valueLen >= 4 && len(data) >= valueLen { 
+		c.Tuples = make([]*CapGracefulRestartTuple, 0, valueLen/4)
+
+		for i := valueLen; i >= 4; i -= 4  {
+			t := &CapGracefulRestartTuple{binary.BigEndian.Uint16(data[0:2]),
+				data[2], data[3]}
+			c.Tuples = append(c.Tuples, t)
+			data = data[4:]
+		}
 	}
 	return nil
 }
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -362,13 +362,13 @@ func (c *CapGracefulRestart) DecodeFromBytes(data []byte) error {
 	c.Flags = uint8(restart >> 12)
 	c.Time = restart & 0xfff
 	data = data[2:]
- 
-	var valueLen int = int(c.CapLen) - 2
 
-	if valueLen >= 4 && len(data) >= valueLen { 
+	valueLen := int(c.CapLen) - 2
+
+	if valueLen >= 4 && len(data) >= valueLen {
 		c.Tuples = make([]*CapGracefulRestartTuple, 0, valueLen/4)
 
-		for i := valueLen; i >= 4; i -= 4  {
+		for i := valueLen; i >= 4; i -= 4 {
 			t := &CapGracefulRestartTuple{binary.BigEndian.Uint16(data[0:2]),
 				data[2], data[3]}
 			c.Tuples = append(c.Tuples, t)


TimEvens:
odd...  All tests pass when using the run all tests script. 

```go version go1.5 linux/amd64```

```
.
----------------------------------------------------------------------
Ran 1 test in 29.766s

OK
.
----------------------------------------------------------------------
Ran 1 test in 27.059s

OK
all tests passed successfully
```

fujita:
Thanks for the fix! Looks like the format needs to be fixed. Please try `go fmt`.

https://travis-ci.org/osrg/gobgp/jobs/148983818

The command "if [[ "$TRAVIS_GO_VERSION" == "1.5" ]]; then test -z "$(go fmt ./...)"; fi && bash test/scenario_test/ci-scripts/travis-build-script.sh $SCENARIO" exited with 1.

fujita:
Squashed and pushed. Thank you so much!

PR #1042, opened 2016-08-03
Modify MPLS labels on EVPN RouteType.

Y-Iwami:
Bottom of Stack on MPLS labels are not treated in EVPN RouteTypes,
and gobgp users, using cli commands, are forced to convert
binary MPLS value including Botton of Stack into decimal.
So, using "MPLSLabelStack" for Labels, like "LabelledIPPrefix",
is a more better way.


ishidawataru:
Hi, thanks for the PR!

Yes, I also wanted to use `MPLSLabelStack`.
But the field is also used by VXLAN VNI.
( please see [this draft](https://tools.ietf.org/html/draft-ietf-bess-evpn-overlay-04 ) )

For VNI, we can't use `MPLSLabelStack` since there is no bottom of stack.

Y-Iwami:
Thaks for quick response.
I understood the reason of implementation.
Thank you!

tmmorin:
You could maybe provide, as an additional field, a "parsed mpls" value corresponding the 20 higher-level order bits ?

PR #1047, opened 2016-08-07
server: add missing default policy setting in (*BgpServer).Start()

ishidawataru:
policy.Reset() won't initialize global policy unless we pass a map
with key table.GLOBAL_RIB_NAME

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1173,8 +1173,7 @@ func (s *BgpServer) Start(c *config.Global) (err error) {
 
 		rfs, _ := config.AfiSafis(c.AfiSafis).ToRfList()
 		s.globalRib = table.NewTableManager(rfs, c.MplsLabelRange.MinLabel, c.MplsLabelRange.MaxLabel)
-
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -1173,8 +1173,9 @@ func (s *BgpServer) Start(c *config.Global) (err error) {
 
 		rfs, _ := config.AfiSafis(c.AfiSafis).ToRfList()
 		s.globalRib = table.NewTableManager(rfs, c.MplsLabelRange.MinLabel, c.MplsLabelRange.MaxLabel)
-
-		if err = s.policy.Reset(&config.RoutingPolicy{}, map[string]config.ApplyPolicy{}); err != nil {
```
fujita: How about simple one liner like:

if err = s.policy.Reset(&config.RoutingPolicy{}, map[string]config.ApplyPolicy{table.GLOBAL_RIB_NAME: c
.ApplyPolicy}); err != nil {

PR #1095, opened 2016-09-08
server: portability fix for setsockopt() on Linux

vincentbernat:
On some architecture, setsockopt() is a multiplexed syscall and is not
available directly (for example, on i386). Instead of invoking the
syscall directly, use syscall.SetsockoptString() which is safe as strings
are not null-terminated with Go.

On BSD, use syscall.SetsockoptInt(), but I don't know if there are
architectures with the same limitations as for Linux. I didn't test, but the change seems trivial.

Signed-off-by: Vincent Bernat <vincent@bernat.im>


fujita:
So sorry about the delay, it took time to fix the CI system. Now it's back. I've pushed this.
Thanks a lot!

PR #1116, opened 2016-09-27
test: Adopt to the updated Ryu BGP Packet lib API

iwaseyusuke:
From Ryu v4.6, the Packet library can parse the entire BGP packet
via packet.Packet() API.
This patch fixes to adopt bgp_router_test.py to the updated API,
and reverts the Ryu version in pip-requires.txt to the latest.

Note: This patch requires  the following patch should be merged
into the upstream of Ryu. (It might be done at the next release.)
https://sourceforge.net/p/ryu/mailman/message/35393517/

### DIFF
--- a/test/pip-requires.txt
+++ b/test/pip-requires.txt
@@ -5,4 +5,4 @@ fabric
 netaddr
 nsenter
 docker-py
-ryu==4.5
+ryu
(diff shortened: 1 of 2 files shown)


fujita:
Worked with Ryu 4.7. Thanks!

PR #1119, opened 2016-09-28
Support of ZAPI version 3 (handles VRF ID).

kishiguro:
This is change for supporting ZAPI version 3 (which handles VRF ID in the message).


fujita:
Thanks!

PR #1131, opened 2016-10-04
Unstrict config parsing

mrevilme:
Gives a developer who uses gobgp as a lib the possibility to add own config structures to the configuration file.

Will not break backwards comparability since i reimplemented the ReadConfigfileServe with the same arguments as before but its just calling the new ReadConfigfileServeExtensive.

### DIFF
--- a/config/serve.go
+++ b/config/serve.go
@@ -57,6 +57,9 @@ func ReadConfigfileServeExtensive(path, format string, configCh chan *BgpConfigS
 		}
 		cnt++
 		configCh <- c
+		if viperCh != nil {
+			viperCh <- v
+		}
 		goto NEXT
 	ERROR:
 		if cnt == 0 {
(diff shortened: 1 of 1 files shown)


mrevilme:
(Extract of example how to use it, stolen most of the code from gobgpd/main.go)

type Foobar struct {
    Bar         string      `mapstructure:"bar"`
}

type TestConfigSet struct {
   Foo          string
   Foobar       []Foobar    `mapstructure:"foobar"`
}

viperCh := make(chan *viper.Viper)
go config.ReadConfigfileServeExtensive(opts.ConfigFile, opts.ConfigType, configCh, false, viperCh)
var t *TestConfigSet = nil
    for {
        select {
        case newViper := <-viperCh:
            newViper.UnmarshalKey("test", &t)
            log.Infoln(t.Foo)
        }
    }
}

Presumes that your config looks like:
[test]
foo = "bar"
[[test.foobar]]
bar = "foo"

PR #1139, opened 2016-10-08
table: ensure compatibility with archs where int == int32

vincentbernat:
MED actions are expected to be int64 and AS are expected to be
uint32. Use ParseInt/ParseUint instead of Atoi for those until the tests
pass on ARM.

This is not an exhaustive audit, just the minimal modifications to make all tests pass and gobgp/gobgpd build correctly on both amd64 and arm platforms.


fujita:
Looks good. Thanks a lot!

PR #1160, opened 2016-11-12
Fix

ishidawataru:


fujita:
The commit for quagga is necessary? Looks like the CI works fine without the commit.

PR #1170, opened 2016-11-19
Fix client.AddNeighbor() API does not properly configure AFI/SAFI bug.

kishiguro:
When we use gRPC API to add neighbor.  Even if we specify, config.Neighbor.AfiSafis parameter, it is ignored in NewPeerFromConfigStruct().  Adding to that even if we fix the issue, NewNeighborFromAPIStruct() count Families twice.  So resulting neighbor has duplicated Families configuration.  This patch fixed both problems.

### DIFF
--- a/api/grpc_server.go
+++ b/api/grpc_server.go
@@ -89,6 +89,12 @@ func (s *Server) Serve() error {
 }
 
 func NewPeerFromConfigStruct(pconf *config.Neighbor) *Peer {
+	var families []uint32
+	for _, f := range pconf.AfiSafis {
+		if family, ok := bgp.AddressFamilyValueMap[string(f.Config.AfiSafiName)]; ok {
+			families = append(families, uint32(family))
+		}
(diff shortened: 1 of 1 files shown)


fujita:
Thanks for the fix!
The change for NewPeerFromConfigStruct() looks fine.
Looks like the change for NewNeighborFromAPIStruct() introduces a bug that ignores the prefix limit config. I think that better to remove the latter duplicated code to set up families https://github.com/osrg/gobgp/blob/master/api/grpc_server.go#L886
We don't need to consider the case of a.Conf != nil:
https://github.com/osrg/gobgp/blob/master/api/grpc_server.go#L786
I'll fix NewNeighborFromAPIStruct to return an error for the case after your fix.

kishiguro:
Thanks for the check.  I've removed latter duplicated code.

fujita:
Squashed and pushed, thanks!

PR #1172, opened 2016-11-21
Update the grpc server to accept the local address and remote port

cstockton:
Update the grpc server to accept the local address and remote port from the transport config. This allows using the gobgp/client to send a transport config with a port number.


fujita:
Thanks, pushed!

PR #1180, opened 2016-11-29
Fix Open Message optional parameter length == 1 GoBGP crashes issues.

kishiguro:
When Open Message optional parameter == 1, there is no packet length check for load paramtype and paramlen.  The patch fix the crash issues.


fujita:
Nice catch, thanks!

PR #1181, opened 2016-11-30
gRPC client API AddNeighbor did not include apply-policy information …

kishiguro:
…in API call.
So even if we set "apply-policy" parameter in "neighbor" config, it will not send to server.
This patch fix the issue.


fujita:
Pushed, thanks!

PR #1183, opened 2016-12-02
Allow creating a GoBGPClient using a user specific GRPC connection and client.

cstockton:
I touched more on this on Slack, but this allows wrapping the dialer, implementing your own wrapped GobgpApiClient or dialing with a context for cancelation on unreliable endpoints.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -39,27 +39,36 @@ func defaultGRPCOptions() []grpc.DialOption {
 	return []grpc.DialOption{grpc.WithTimeout(time.Second), grpc.WithBlock(), grpc.WithInsecure()}
 }
 
-// NewClient returns a new GoBGPClient, using the given conn and cli for the
-// underlying connection. The given grpc.ClientConn connection is expected to be
-// initialized and paired with the api client. See NewGoBGPClient to have the
-// connection dialed for you.
-func NewClient(conn *grpc.ClientConn, cli api.GobgpApiClient) *GoBGPClient {
-	return &GoBGPClient{conn: conn, cli: cli}
+// New returns a new GoBGPClient using the given target and options for dialing
+// to the grpc server. If an error occurs during dialing it will be returned and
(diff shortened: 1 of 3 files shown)


cstockton:
I updated my pull request to illustrate the discussion in slack, I'll leave nomenclature for you guys to decide but I think responsible use can be assumed here for anyone who wishes to provide their own connection. Limiting the ability to do so greatly narrows the usability for library use (which is the primary intent of this client). Thanks!

ishidawataru:
@cstockton Thanks, looks good to me.

client/client.go
```diff
@@ -39,21 +39,38 @@ func defaultGRPCOptions() []grpc.DialOption {
...
 	if err != nil {
 		return nil, err
 	}
 	cli := api.NewGobgpApiClient(conn)
 	return &GoBGPClient{conn: conn, cli: cli}, nil
 }
 
+// NewFrom returns a new GoBGPClient, using the given conn and cli for the
+// underlying connection. The given grpc.ClientConn connection is expected to be
+// initialized and paired with the api client. See NewGoBGPClient to have the
```
ishidawataru: `NewGoBGPClient` should be `New` ?

cstockton: I think New would be a solid name, but didn't want to break BC. I can update the pull req with the name as New if you want?

ishidawataru: I meant the word `NewGoBGPClient`  in the comment here should be `New`.

I agree that `New()` is better than `NewGoBGPClient()`

cstockton: Hah, good eye, I didn't even notice. Fixed.

fujita:
Squashed and pushed, thanks!

cstockton:
@fujita @ishidawataru Since it's early enough that you guys are taking BC breaking changes- want me to pull request a rename of client.GoBGPClient to client.Client to make it consistent with client.New, server.Server, etc?

PR #1188, opened 2016-12-07
policy: ext-community and large-community can be included in conditions

yokoi-h:
Currently gobgp seems to validate ext-community and large-community condition as community condition.

```
% gobgp global as 65000 router-id 10.10.0.1
% gobgp policy ext-community add ecs1 RT:46334:2605868060
% gobgp policy statement add import-policy-statement
% gobgp policy statement import-policy-statement add condition ext-community ecs1 any
rpc error: code = 2 desc = not found community set ecs1
```

This patch fix the issue.

Signed-off-by: Hiroshi Yokoi <yokoi.hiroshi@po.ntts.co.jp>


fujita:
Thanks!

PR #1189, opened 2016-12-07
policy: allow using invert condition in statement

yokoi-h:
Currently we get error when we use "invert" in statement condition like below.

```
% gobgp global as 65000 router-id 10.10.0.1
% gobgp policy neighbor add n1 172.16.132.11
% gobgp policy statement add statement1
% gobgp policy statement statement1 add condition neighbor n1 invert
rpc error: code = 2 desc = invalid match type
```

This patch fixes the issue.

Signed-off-by: Hiroshi Yokoi <yokoi.hiroshi@po.ntts.co.jp>

### DIFF
--- a/api/grpc_server.go
+++ b/api/grpc_server.go
@@ -1209,14 +1209,16 @@ func NewAPIStatementFromTableStruct(t *table.Statement) *Statement {
 func toStatementApi(s *config.Statement) *Statement {
 	cs := &Conditions{}
 	if s.Conditions.MatchPrefixSet.PrefixSet != "" {
+		o, _ := table.NewMatchOption(s.Conditions.MatchPrefixSet.MatchSetOptions)
 		cs.PrefixSet = &MatchSet{
-			Type: MatchType(s.Conditions.MatchPrefixSet.MatchSetOptions.ToInt()),
+			Type: MatchType(o),
 			Name: s.Conditions.MatchPrefixSet.PrefixSet,
 		}
 	}
 	if s.Conditions.MatchNeighborSet.NeighborSet != "" {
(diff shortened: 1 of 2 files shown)


ishidawataru:
@yokoi-h Since `bgp_configs.go` is generated by using `pyang`, we can't modify it manually.

Instead, we need to fix here.

https://github.com/osrg/gobgp/blob/master/table/policy.go#L2369

Currently we are doing wrong translation using `config.IntToMatchSetOptionsRestrictedTypeMap`.
We need a translation function to convert `table.MatchOption` to `config.MatchSetOptionsRestrictedType`

yokoi-h:
@ishidawataru  I added function to convert config.MatchSetOptionsRestrictedType.
https://github.com/yokoi-h/gobgp/blob/fix-condition-invert/table/policy.go#L113

However I needed to add another function to convert to gobgpapi.MatchType and use it grpc_server side. I think current modification is not so clean. If you have any comments for improvements, please let me know.

ishidawataru:
@yokoi-h How about using `table.NewMatchOption()` instead of creating another translation function?

https://github.com/osrg/gobgp/blob/master/table/policy.go#L159

yokoi-h:
@ishidawataru I appreciate your comment. It can avoid another convert function.
 I changed the code to use `table.NewMatchOption`.
If it doesn't have any problems, I will apply the same modification to `config.MatchPrefixSet`.

fujita:
Thanks, can you squash the commits into one commit?

yokoi-h:
@fujita  I fixed invert for `PrefixSet` as well and squashed commits into a single commit.

fujita:
Thanks!

PR #1201, opened 2017-01-10
Packet bgp confirm IPv4 to IPv6 nexthop address conversion when AFI_IP6

iwaseyusuke:
Currently, GoBGP will miss-convert IPv4 nexthop address to IPv6
address when advertising AFI_IP6 routes with IPv4 peering.
For example, if the local address of GoBGP is "172.20.0.1", the
advertised nexthop will be miss-converted into "AC14:1::".
In this case, the nexthop address should be "::ffff:172.20.0.1".
The following patches fix this problem.

Also, including some improvements detected by code inspection
tools (e.g., IDE, golint, ...).

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -3787,9 +3787,9 @@ func NewPrefixFromRouteFamily(afi uint16, safi uint8) (prefix AddrPrefixInterfac
 	case RF_OPAQUE:
 		prefix = &OpaqueNLRI{}
 	default:
-		return nil, fmt.Errorf("unknown route family. AFI: %d, SAFI: %d", afi, safi)
+		err = fmt.Errorf("unknown route family. AFI: %d, SAFI: %d", afi, safi)
 	}
-	return prefix, nil
+	return prefix, err
 }
 
 type BGPAttrFlag uint8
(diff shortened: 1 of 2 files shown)


packet/bgp/bgp.go
```diff
@@ -5103,25 +5098,31 @@ func (p *PathAttributeMpReachNLRI) Serialize() ([]byte, error) {
...
 	buf := make([]byte, 4+nexthoplen)
 	binary.BigEndian.PutUint16(buf[0:], afi)
 	buf[2] = safi
 	buf[3] = uint8(nexthoplen)
-	copy(buf[4+offset:], p.Nexthop)
-	if p.LinkLocalNexthop != nil {
-		copy(buf[4+offset+len(p.Nexthop):], p.LinkLocalNexthop)
+	if nexthoplen != 0 {
+		if afi == AFI_IP6 {
+			copy(buf[4+offset:], p.Nexthop.To16())
```
ishidawataru: Does this commit also fix other issues which are not stated in the commit message?

I think only this part (using `To16()` when `afi == AFI_IP6`) is sufficient to fix the issue stated in the message.

iwaseyusuke: Thank you for reviewing my patch.

Exactly.
I didn't mention enough.

To fix only conversion of IPv4 to IPv6, using `To16()` is sufficient.
But when testing it, I found that `nexthoplen` calculation is also need to parse and serialize NextHop field, otherwise "mpreach nexthop length is incorrect" at GoBGP or "Invalid NextHop length" at  BGP peers.

Should I separate this patch at this point? Or need to mention more clearly?

ishidawataru: Please add few lines to describe more.
Also, it would be really helpful if you can add a unit test in `bgp_test.go` for what you fixed.

iwaseyusuke: Sure.

First, I will update the commit comment on my patch,
and add unit test into `bgp_test.go`, later.
(Sorry, I'm new to golang, so I would like to add unit test in another PR...)

fujita:
Pushed, thanks!

PR #1202, opened 2017-01-11
Fix incorrect validation

a16:
The pseudo code in RFC6811 says:
```
   result = BGP_PFXV_STATE_NOT_FOUND;

   //Iterate through all the Covering entries in the local VRP
   //database, pfx_validate_table.
   entry = next_lookup_result(pfx_validate_table, route_prefix);

   while (entry != NULL) {
     prefix_exists = TRUE;

     if (route_prefix_length <= entry->max_length) {
       if (route_origin_as != NONE
           && entry->origin_as != 0
           && route_origin_as == entry->origin_as) {
         result = BGP_PFXV_STATE_VALID;
         return (result);
       }
     }
     entry = next_lookup_result(pfx_validate_table, input.prefix);
   }

   //If one or more VRP entries Covered the route prefix, but
   //none Matched, return "Invalid" validation state.
   if (prefix_exists == TRUE) {
     result = BGP_PFXV_STATE_INVALID;
   }

   return (result);
```

It means that it looks up the received route with ROA table recursively.
On the otherhand,  It looks validatePath() validates  only once.

Let's say, ROA table has:
```
10.0.0.0/24-24 AS65000
10.0.0.0/16-24 AS65001
```
and received:
```
10.0.0.0/24 AS65001
```

In this case, the current gobgp code should determine "Invalid".
But I believe that it should determine "Valid".
(I've confirmed with RIPE RPKI Validator in the similar situation.)

I've added the examples and fixed code.
What do you think?

### DIFF
--- a/server/rpki.go
+++ b/server/rpki.go
@@ -513,21 +513,25 @@ func validatePath(ownAs uint32, tree *radix.Tree, cidr string, asPath *bgp.PathA
 	_, n, _ := net.ParseCIDR(cidr)
 	ones, _ := n.Mask.Size()
 	prefixLen := uint8(ones)
-	_, b, _ := tree.LongestPrefix(table.IpToRadixkey(n.IP, prefixLen))
+	key := table.IpToRadixkey(n.IP, prefixLen)
+	_, b, _ := tree.LongestPrefix(key)
 	if b == nil {
 		return config.RPKI_VALIDATION_RESULT_TYPE_NOT_FOUND
 	}
 
-	bucket, _ := b.(*roaBucket)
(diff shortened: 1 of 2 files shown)


ishidawataru:
Thanks. I revised the code. https://github.com/osrg/gobgp/commit/9b2fc19bbcf39dd0937c45930f2a72c3c409c72c
Please check this doesn't break anything.

If it seems ok, please cherry pick the commit I made onto your [branch](https://github.com/a16/gobgp/tree/fix_incorrect_validation)

fujita:
Thanks a lot!

PR #1210, opened 2017-01-16
config: Detect config file type automatically

iwaseyusuke:
This patch enables to automatically detect the type of config file
by retrieving the extension from the config file name.
e.g.) "gobgpd.toml" will be assumed as TOML formatted file.

Note: This feature will override the format specified with the cli
"--config-type" option.


fujita:
Nice, thanks a lot!

PR #1213, opened 2017-01-17
Want to use server/rpki.go for library usage

a16:
I want to use server/rpki.go for library on my application.
Furthermore I've added a feature to return the matched RoaBucket when ValidatePath() is executed.
Because we can't see which ROA the result was validated with.


fujita:
Thanks a lot!

Issue #1200, opened 2017-01-06
support for Administrative Shutdown Communication


job:
The draft-ietf-idr-shutdown document specifies a mechanism to transmit a short freeform UTF-8 [RFC3629] message as part of a Cease NOTIFICATION message [RFC4486] to inform the peer why the BGP session is being shutdown. I believe this would be a useful feature for people who use GoBGP in Route Server scenarios.

https://tools.ietf.org/html/draft-ietf-idr-shutdown

job:
For a full overview of all feature considerations that we applied to OpenBGPD - please review the issues list at https://github.com/job/openbgpd/issues?utf8=%E2%9C%93&q=is%3Aissue

iwaseyusuke:
@job 
Hi,

GoBGP have already has the feature to send Cease NOTIFICATION messages with the following command.
(it seems to be deprecated, though...)
```
# with subcode "Administrative Shutdown"
$ gobgp neighbor <NEIGHBOR_IP> shutdown
WARNING: command `shutdown` is deprecated. use `disable` instead
```
You mean the feature to transmit an arbitrary UTF-8 message, right?
If so, I'm going to add "--message" option to specify the transmitted message like:
```
gobgp neighbor -m "some message" <NEIGHBOR_IP> shutdown
```
What do you think?

Thanks

job:
Yes indeed I mean the feature to transmit (or receive) the UTF8 message at shutdown time. A "--message" feature would be most welcome!

job:
@iwaseyusuke will you also add support so a received shutdown communication is printed into the appropriate places (terminal, syslog, etc)?

iwaseyusuke:
@job 
Thank you for your confirmation!

Currently, I have implemented "to send message" only.

I found that GoBGP already can log the received NOTIFICATION message body, but as a binary like `"c2h1dGRvd24gVGh1IEphbiAxOSAwMDoyMzoxMSBVVEMgMjAxNw=="`(base64?).

Though I don't know whether we can "always" decode the NOTIFICATION message body as "a character string", if we can, I will implement it as following.
What do you think about if we should decode as "a character string"?

Command:
```
$ gobgp neighbor -m "shutdown by administrator" 172.17.0.3 shutdown
```
Sent side:
```
{"Key":"172.17.0.3","Topic":"Operation","level":"info","msg":"Neighbor shutdown","time":"2017-01-19T00:51:52Z"}
[...]

PR #1218, opened 2017-01-23
Support for Administrative Shutdown Communication

iwaseyusuke:
These patch enable to send and log "Administrative Shutdown Communication".
issues: #1200 
Reference: https://tools.ietf.org/html/draft-ietf-idr-shutdown-03


server/fsm.go
```diff
@@ -696,13 +696,25 @@ func (h *FSMHandler) recvMessageWithError() (*FsmMsg, error) {
...
-					"Subcode": body.ErrorSubcode,
-					"Data":    body.Data,
-				}).Warn("received notification")
+				if body.ErrorCode == bgp.BGP_ERROR_CEASE && body.ErrorSubcode == bgp.BGP_ERROR_SUB_ADMINISTRATIVE_SHUTDOWN {
+					log.WithFields(log.Fields{
+						"Topic":         "Peer",
+						"Key":           h.fsm.pConf.Config.NeighborAddress,
+						"Code":          body.ErrorCode,
+						"Subcode":       body.ErrorSubcode,
+						"Lenght":        body.Data[0],
```
job: Possible typo? Maybe `Lenght` should be `Length` ?

iwaseyusuke: Thanks!
Yes, It is a typo...

Also, Travis-CI says my patch collapses some other features,
and I'm investigating why.

iwaseyusuke:
I corrected typos and fixed to pass Travis-CI.

job:
what commands should I type in to test this branch locally?

job:
I have tested it by running these commands:

```

cd $GOPATH
go get github.com/osrg/gobgp/gobgpd
go get github.com/osrg/gobgp/gobgp
cd src/github.com/osrg/gobgp/
git remote add iwaseyusuke https://github.com/iwaseyusuke/gobgp.git
git fetch iwaseyusuke
git checkout fd6b9260aab98fcebb99bbb5665dbb51c9c972c9
cd gobgpd
go build main.go
```

```
{"Code":6,"Communication":" 💩 🦄","Key":"165.254.255.26","Length":10,"Subcode":2,"Topic":"Peer","level":"warning","msg":"received notification","time":"2017-01-24T10:36:29Z"}
{"Key":"165.254.255.26","Reason":"notification-received code 6(cease) subcode 2(administrative shutdown)","State":"BGP_FSM_ESTABLISHED","Topic":"Peer","level":"info","msg":"Peer Down","time":"2017-01-24T10:36:29Z"}
```

A few remarks:

* The shutdown communication should also be send if the `disable` command is used instead of the `shutdown` command. 
* There should be an error when the CLI `--communication` string is longer then 128 bytes. Right now it seems to fail silently
* I think I liked `--message` better then `--communication`
* UTF-8 support seems to work well

iwaseyusuke:
@job 
Thank you for testing my patch!

The commands you have run is correct.
For sending shutdown, you need also build "./gobgp/main.go", I guess.

For your comments:

- Yes, `disable` command should also have this feature, but I found it is difficult with the current architecture. In case of `disable` command, the message sending loop is working on another Gogoutine from the command received Goroutine, so we need to extend the communication channel. Then, I added the feature into only `shutdown` command for the first step.

- "fail silently" means GoBGPd will miss logging the communication field when "Administrative Shutdown" message from another BGP daemon, right? If message sending peer is also GoBGPd (this is my test environment), GoBGPd shows the communication message correctly.
  ```
  # Command
  $ gobgp neighbor --communication "130 length messaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaase" 172.17.0.3 shutdown
  WARNING: command `shutdown` is deprecated. use `disable` instead

  # Sent side
  $  gobgpd -f gobgpd.toml
  ... (snip)
  {"Key":"172.17.0.3","Topic":"Operation","level":"info","msg":"Neighbor shutdown","time":"2017-01-25T05:41:37Z"}
  {"Code":6,"Communication":"130 length messaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa","Key":"172.17.0.3","Length":128,"State":"BGP_FSM_ESTABLISHED","Subcode":2,"Topic":"Peer","level":"warning","msg":"sent notification","time":"2017-01-25T05:41:37Z"}
  {"Key":"172.17.0.3","Reason":"notification-sent code 6(cease) subcode 2(administrative shutdown)","State":"BGP_FSM_ESTABLISHED","Topic":"Peer","level":"info","msg":"Peer Down","time":"2017-01-25T05:41:37Z"}
  ... (snip)

  # Received side
  $  gobgpd -f gobgpd.toml
  ... (snip)
  {"Code":6,"Communication":"130 length messaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa","Key":"172.17.0.2","Length":128,"Subcode":2,"Topic":"Peer","level":"warning","msg":"received notification","time":"2017-01-25T05:41:37Z"}
  {"Key":"172.17.0.2","Reason":"notification-received code 6(cease) subcode 2(administrative shutdown)","State":"BGP_FSM_ESTABLISHED","Topic":"Peer","level":"info","msg":"Peer Down","time":"2017-01-25T05:41:37Z"}
  ... (snip)
  ```
  This is because the sent side cuts the longer message into 128 bytes size, though.
  It seems to be required the check if the peer sending the longer message than 128 bytes.

- Thanks! `--reason` looks better for me! I will update the option name.

- Thank you for your confirmation!

job:
In the case of a 130 length I propose an error is presented to the user and the command is rejected (not executed). A shutdown communication longer than 128 is invalid.

```
# Command
$ gobgp neighbor --reason "130 length messaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaase" 172.17.0.3 shutdown
ERROR: shutdown communication is too long (max 128 bytes)
```

Perhaps the `WARNING` about shutdown being deprecated should be removed as well if `disable` does not support the shutdown communication?

iwaseyusuke:
About rejecting the longer message, I don't have strong will.
Okay, I will update to reject it.

About removing "deprecated" message, the reason I keep it is that I don't know why `shutdown` command is deprecated exactly.
From source code, `shutdown` command just sends NOTIFICATION message only and will reconnect to the neighbor, later. OTOH, `disable` command sends NOTIFICATION message and stops to connect the neighbor (literally "disable").
So, I guess `shutdown` command was deprecated.

job:
In that case I recommend the following:

Move the 'shutdown communication' functionality to work with just `disable`, and not with `shutdown`.

iwaseyusuke:
Yes, this feature should be also added into `disable` command, but [the reason I mentioned before](https://github.com/osrg/gobgp/pull/1218#issuecomment-275027099), I found it difficult.
So I implement this feature into only `shutdown` command for **the first step**.
For the next step, I will implement it into `disable` command.

ishidawataru:
@iwaseyusuke @job Please refer https://github.com/osrg/gobgp/commit/17f7cc4a44823a02153a158a0bfab9b4e396d9a9 for the reason why `shutdown` is deprecated

I think that it would be better to make `shutdown` command an alias of  `disable` command since there is no reason to keep both implementations. ( which means remove [here](https://github.com/osrg/gobgp/blob/master/server/server.go#L1784-L1788) )

@iwaseyusuke For implementing this feature to `disable` command, how about modifing here https://github.com/osrg/gobgp/blob/master/server/fsm.go#L118 to something like

```go
type AdminStateOperation struct {
   State AdminState
   Reason string
}

type FSM struct {
  // ..snip..
  adminStateCh chan AdminStateOperation
  // ..snip..
}
```

Then modify here https://github.com/osrg/gobgp/blob/master/server/fsm.go#L1228-L1236
to handle the reason.

iwaseyusuke:
@ishidawataru 
Thank you for your advice!
I will try to extend the current patch to support it.

job:
Version `-04` of the draft specifies that the Shutdown Communication is also allowed with an `Administrative Reset`.

I propose the shutdown communication is also allowed through this command:

`gobgp neighbor 10.0.255.1 --reason "resetting session for a reason" reset`

https://tools.ietf.org/html/draft-ietf-idr-shutdown-04

iwaseyusuke:
Thanks for letting me know.

BTW, "Communication" field is still named with "Shutdown", even if extended to "Reset"?

job:
Yes, because from a protocol perspective the 'Reset' event is a
shutdown. We also thought it was best to minimize the changes since we
are in Working Group Last Call, and impact on recognizability of this
concept.

iwaseyusuke:
Thank you for letting me know that reason.

I have updated my patch to support the Shutdown Communication in `disable` command in the 3rd commit.

job:
I tested this and this looks good!

```
Jan 31 09:55:13 kiera bgpd[68914]: neighbor 165.254.255.17: received notification: Cease, administratively down
Jan 31 09:55:13 kiera bgpd[68914]: neighbor 165.254.255.17: received shutdown reason: "some messages"


Jan 31 09:57:12 kiera bgpd[68914]: neighbor 165.254.255.17: received notification: Cease, administrative reset
Jan 31 09:57:12 kiera bgpd[68914]: neighbor 165.254.255.17: received shutdown reason: "some messages"
```

iwaseyusuke:
Thank you for testing my patches!

No, I guess there are more reasonable formats, but I had no good idea.
Many thanks for your advice!
I will update my patches to output with `Communicated-Reason` and `Data`.

iwaseyusuke:
I have updated my patches.
With this patch, the log messages would like
```
# Command
$ gobgp neighbor 172.17.0.3 --reason "resetting session for a reason" disable

# Sent side
{"Key":"172.17.0.3","State":"BGP_FSM_ESTABLISHED","Topic":"Peer","level":"info","msg":"Administrative shutdown","time":"2017-02-01T08:07:22Z"}
{"Code":6,"Communicated-Reason":"resetting session for a reason","Data":"","Key":"172.17.0.3","State":"BGP_FSM_ESTABLISHED","Subcode":2,"Topic":"Peer","level":"warning","msg":"sent notification","time":"2017-02-01T08:07:22Z"}
{"Key":"172.17.0.3","Reason":"notification-sent code 6(cease) subcode 2(administrative shutdown)","State":"BGP_FSM_ESTABLISHED","Topic":"Peer","level":"info","msg":"Peer Down","time":"2017-02-01T08:07:22Z"}

# Received side
{"Code":6,"Communicated-Reason":"resetting session for a reason","Data":"","Key":"172.17.0.2","Subcode":2,"Topic":"Peer","level":"warning","msg":"received notification","time":"2017-02-01T08:07:22Z"}
{"Key":"172.17.0.2","Reason":"notification-received code 6(cease) subcode 2(administrative shutdown)","State":"BGP_FSM_ESTABLISHED","Topic":"Peer","level":"info","msg":"Peer Down","time":"2017-02-01T08:07:22Z"}
```

job:
Excellent!

iwaseyusuke:
Many thanks for reviewing my patch!

fujita:
Pushed, thanks guys!

PR #1220, opened 2017-01-27
Fix End of RIB handling

a16:
- All of EoR except IPv4 unicast have MP_UNREACH only. (RFC4724)
    - length of the MP_UNREACH attr. should be 3bytes.
    - length of the value should be 0bytes.
  - Should send EoR marker in initial exchange of BMP Route Monitoring. (RFC7854)


fujita:
Thanks a lot!

PR #1222, opened 2017-01-31
Fix rpki server status

a16:
`gobgp rpki servers` command for the first time after gobgpd started seems to return "Down" even though the RTR session has been established. From the second time on, it seems to work correctly. I guess `client.state.Up = true` is missing in the case of `CONNECTED` in `HandleROAEvent()`.


fujita:
Thanks for the fix!

But even with this fix, we hit the same problem with the record numbers. Also we hit the same problem when a rpki server is down.

The root cause is that GetServers() uses the values before updating them. The following commit should fix the bug:
https://github.com/osrg/gobgp/commit/a7456d5aab36b04163da30ceb91ff69d6f223426

PR #1228, opened 2017-02-08
Tcpflags parser

Matt-Texier:
This patch proposes a new way to configure BGP flowspec TCP flags rules

It allows to comply with RFC 5575 by defining flags like this '=SA =A' or '!SA' or '=SA&=!U'
= means match, ! means not, & means and, all TCP flags are identified by their first charater S for SYN A for Ack ...

I did my own parser as using regexp was not an easy task considering the various lexical possibilities.

If you have a look at this draft: https://tools.ietf.org/html/draft-hr-idr-rfc5575bis-02, this patch comply with it.

Thanks Matt.

here under a couple of examples:

```
root@matt-dev-box:/home/matthieu# /home/matthieu/go-work/bin/gobgp global rib -a ipv6-flowspec add match destination 2001::1/128 port '=80' tcp-flags CEUAPRSF then discard
root@matt-dev-box:/home/matthieu# /home/matthieu/go-work/bin/gobgp global rib -a ipv6-flowspec
    Network                                                    Next Hop             AS_PATH              Age        Attrs
*>  [destination:2001::1/128/0][port: =80][tcp-flags:FSRPAUCE ]fictitious                                00:00:03   [{Origin: ?} {Extcomms: [discard]}]
root@matt-dev-box:/home/matthieu# /home/matthieu/go-work/bin/gobgp global rib -a ipv6-flowspec add match destination 2001::2/128 port '=80' tcp-flags =!SA&=!U then discard
bash: !SA: event not found
root@matt-dev-box:/home/matthieu# /home/matthieu/go-work/bin/gobgp global rib -a ipv6-flowspec add match destination 2001::2/128 port '=80' tcp-flags '=!SA&=!U' then discard
root@matt-dev-box:/home/matthieu# /home/matthieu/go-work/bin/gobgp global rib -a ipv6-flowspec
    Network                                                    Next Hop             AS_PATH              Age        Attrs
*>  [destination:2001::2/128/0][port: =80][tcp-flags:=!SA&=!U ]fictitious                                00:00:04   [{Origin: ?} {Extcomms: [discard]}]
*>  [destination:2001::1/128/0][port: =80][tcp-flags:UCEFSRPA ]fictitious                                00:02:02   [{Origin: ?} {Extcomms: [discard]}]
```

### DIFF
--- a/docs/sources/flowspec.md
+++ b/docs/sources/flowspec.md
@@ -45,7 +45,7 @@ CLI syntax to add ipv4/ipv6 flowspec rule is
 ```shell
 % global rib add match <MATCH_EXPR> then <THEN_EXPR> -a [ipv4-flowspec|ipv6-flowspec]
    <MATCH_EXPR> : { destination <PREFIX> [<OFFSET>] | source <PREFIX> [<OFFSET>] |
-                    protocol <PROTO>... | fragment <FRAGMENT_TYPE> | tcp-flags [not] [match] <TCPFLAG>... |
+                    protocol <PROTO>... | fragment <FRAGMENT_TYPE> | tcp-flags [!] [=] <TCPFLAG>... |
                     { port | destination-port | source-port | icmp-type | icmp-code | packet-length | dscp | label } <ITEM>... }...
    <PROTO> : ospf, pim, igp, udp, igmp, tcp, egp, rsvp, gre, ipip, unknown, icmp, sctp, <VALUE>
    <FRAGMENT_TYPE> : dont-fragment, is-fragment, first-fragment, last-fragment, not-a-fragment
--- a/docs/sources/flowspec.md
+++ b/docs/sources/flowspec.md
@@ -49,7 +49,7 @@ CLI syntax to add ipv4/ipv6 flowspec rule is
                     { port | destination-port | source-port | icmp-type | icmp-code | packet-length | dscp | label } <ITEM>... }...
    <PROTO> : ospf, pim, igp, udp, igmp, tcp, egp, rsvp, gre, ipip, unknown, icmp, sctp, <VALUE>
    <FRAGMENT_TYPE> : dont-fragment, is-fragment, first-fragment, last-fragment, not-a-fragment
-   <TCPFLAG> : rst, push, ack, urgent, fin, syn
+   <TCPFLAG> : U, C, E, F, S, R, P, A
    <ITEM> : &?{<|>|=}<value>
    <THEN_EXPR> : { accept | discard | rate-limit <value> | redirect <RT> | mark <value> | action { sample | terminal | sample-terminal } | rt <RT>... }...
    <RT> : xxx:yyy, xx.xx.xx.xx:yyy, xxx.xxx:yyy
@@ -71,6 +71,9 @@ that for l2vpn flowspec rule is
 ```shell
 # add a flowspec rule which redirect flows with dst 10.0.0.0/24 and src 20.0.0.0/24 to VRF with RT 10:10
 % gobgp global rib -a ipv4-flowspec add match destination 10.0.0.0/24 source 20.0.0.0/24 then redirect 10:10
+
+# add a flowspec rule wich discard flows with dst 2001::2/128 and port equals 80 and with TCP flags not match SA (SYN/ACK) and not match U (URG)
+% gobgp global rib -a ipv6-flowspec add match destination 2001::2/128 port '=80' tcp-flags '=!SA&=!U' then discard
 
 # show flowspec table
 % gobgp global rib -a ipv4-flowspec
--- a/gobgp/cmd/global.go
+++ b/gobgp/cmd/global.go
@@ -863,7 +863,7 @@ usage: %s rib %s%%smatch <MATCH_EXPR> then <THEN_EXPR> -a %%s
 			ExtCommNameMap[RATE], ExtCommNameMap[REDIRECT],
 			ExtCommNameMap[MARK], ExtCommNameMap[ACTION], ExtCommNameMap[RT])
 		ipFsMatchExpr := fmt.Sprintf(`   <MATCH_EXPR> : { %s <PREFIX> [<OFFSET>] | %s <PREFIX> [<OFFSET>] |
-                    %s <PROTO>... | %s <FRAGMENT_TYPE> | %s [not] [match] <TCPFLAG>... |
+                    %s <PROTO>... | %s <FRAGMENT_TYPE> | %s [!] [=] <TCPFLAG>... |
                     { %s | %s | %s | %s | %s | %s | %s | %s } <ITEM>... }...
    <PROTO> : %s
    <FRAGMENT_TYPE> : dont-fragment, is-fragment, first-fragment, last-fragment, not-a-fragment
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -2531,7 +2531,6 @@ func parseTcpFlagCmd(myCmd string) ([][2]int, error) {
 				tcpOperatorsFlagsValues = append(tcpOperatorsFlagsValues, operatorValue)
 				operatorValue[0] = 0
 				operatorValue[1] = 0
-				// resetAllFlagsToFalse(tcpFlagBitMap, tcpFlagOpBitMap)
 				index++
 			} else {
 				err := fmt.Errorf("AND or OR (space) operator appears multiple time")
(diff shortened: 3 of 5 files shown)


ishidawataru:
@Matt-Texier Thanks for your work! Two travis tests are failing. One is checking for `go fmt` and the other is a scenario test for flowspec.

As for `go fmt`, please see [here](https://blog.golang.org/go-fmt-your-code), and run `go fmt`.

As for flowspec test, we need to update [here](https://github.com/osrg/gobgp/blob/master/test/scenario_test/flow_spec_test.py#L47-L49) to the new syntax. 
Also, please update the [doc](https://github.com/osrg/gobgp/blob/master/docs/sources/flowspec.md)

Matt-Texier:
@ishidawataru Happy to contribute and hopefully community will like this proposal. I did the two first changes and I will work on doc asap ...

Matt-Texier:
I Just committed a proposal to update the documentation ... let me know if it looks good for you.
Thanks Matthieu.

packet/bgp/bgp.go
```diff
@@ -2489,32 +2489,139 @@ func flowSpecIpProtoParser(rf RouteFamily, args []string) (FlowSpecComponentInte
...
+	for index := 0; index < len(flags); index++ {
+		items = append(items, NewFlowSpecComponentItem(ops[index], flags[index]))
 	}
-	if elems[1] != "" {
-		op |= 0x2
+	return NewFlowSpecComponent(FLOW_SPEC_TYPE_TCP_FLAG, items), nil
+}
+
+func parseTcpFlagCmd(myCmd string) (error, []int, []int) {
+	var tcpFlagBitMap = map[string]bool{
```
ishidawataru: We can just put `var tcpFlagBitMap = map[string]bool{}` here, since when the key doesn't exist, `map[string]bool` returns `false`.

ishidawataru: Or, why not just use `int` like below?

```go

tcpFlagsValue := 0
tcpFlowSpecOp := 0

for index < len(myCmd) {
  myCmdChar := myCmd[index : index+1]
  switch myCmdChar {
    case TCPFlagOpNameMap[TCP_FLAG_OP_MATCH]:
      if bit := tcpFlagOpBitMap[myCmdChar]; bit & tcpFlagsValue == 0 {
          tcpFlagsValue |= bit
      }
//...(snip)...
  }
}
```

packet/bgp/bgp.go
```diff
@@ -2489,32 +2489,139 @@ func flowSpecIpProtoParser(rf RouteFamily, args []string) (FlowSpecComponentInte
...
-		op |= 0x1
+	for index := 0; index < len(flags); index++ {
+		items = append(items, NewFlowSpecComponentItem(ops[index], flags[index]))
 	}
-	if elems[1] != "" {
-		op |= 0x2
+	return NewFlowSpecComponent(FLOW_SPEC_TYPE_TCP_FLAG, items), nil
+}
+
+func parseTcpFlagCmd(myCmd string) (error, []int, []int) {
```
ishidawataru: How about changing the signature to below?

```go
func parseTcpFlagCmd(myCmd string) ([][2]int, error) {}
```

Then we can make above more simple like

```go
elems, err := parseTcpFlagCmd(fullCmd)
// ...(snip)...
for _, elem := range elems {
       items = append(items, NewFlowSpecComponentItem(elem[0], elem[1]))
}
```

packet/bgp/bgp.go
```diff
@@ -2489,32 +2489,139 @@ func flowSpecIpProtoParser(rf RouteFamily, args []string) (FlowSpecComponentInte
...
+	err, flags, ops := parseTcpFlagCmd(fullCmd)
+	if err != nil {
+		return nil, err
 	}
 	items := make([]*FlowSpecComponentItem, 0)
-	op := 0
-	if elems[2] != "" {
-		op |= 0x1
+	for index := 0; index < len(flags); index++ {
+		items = append(items, NewFlowSpecComponentItem(ops[index], flags[index]))
```
ishidawataru: Please see the comment about the signature of `parseTcpFlagCmd` below

packet/bgp/bgp.go
Matt-Texier: Ok, if I get it, you mean I get rid of the boolean map and test/update and create a bitmap with an int directly ... right ?

packet/bgp/bgp.go
Matt-Texier: yep makes sense ... I'll update the code according to your advices. Thanks for getting through it. I'll push a new version today of the week after as I am getting off for vacation next week.
Matt.

Matt-Texier:
Hi, this new version is taking into account all comments ... please let me know if anything is still unclear / to be improved.

As per you last remark, I finally found out that you need to have tow variables when you want to check if a value exists in a map. The map returns two things: the value found (0 if not found) and a boolean (if map exists).

```
if bit, isPresent := TCPFlagValueMap[myLoopChar]; isPresent && (bit&TCPFlag(operatorValue[1]) == 0) {
					operatorValue[1] |= int(bit) 
```


Thanks Matt.

docs/sources/flowspec.md
```diff
@@ -49,7 +49,7 @@ CLI syntax to add ipv4/ipv6 flowspec rule is
                     { port | destination-port | source-port | icmp-type | icmp-code | packet-length | dscp | label } <ITEM>... }...
    <PROTO> : ospf, pim, igp, udp, igmp, tcp, egp, rsvp, gre, ipip, unknown, icmp, sctp, <VALUE>
    <FRAGMENT_TYPE> : dont-fragment, is-fragment, first-fragment, last-fragment, not-a-fragment
-   <TCPFLAG> : rst, push, ack, urgent, fin, syn
+   <TCPFLAG> : U, C, E, F, S, R, P, A
```
ishidawataru: Please update [ `tcp-flags [not] [match] <TCPFLAG> ...`](https://github.com/osrg/gobgp/pull/1228/files#diff-e647e3b1d45f1144a2b744811179ce30L48) too.

Also, [here](https://github.com/osrg/gobgp/blob/master/gobgp/cmd/global.go#L866) needs the same update.

ishidawataru:
@Matt-Texier Thanks! I added two more comments. Could you reflect them?

packet/bgp/bgp.go
```diff
@@ -2489,32 +2489,85 @@ func flowSpecIpProtoParser(rf RouteFamily, args []string) (FlowSpecComponentInte
...
+				err := fmt.Errorf("Not flag appears multiple time")
+				return nil, err
+			}
+		case TCPFlagOpNameMap[TCP_FLAG_OP_AND], TCPFlagOpNameMap[TCP_FLAG_OP_OR]:
+			if bit := TCPFlagOpValueMap[myCmdChar]; bit&TCPFlagOp(operatorValue[0]) == 0 {
+				operatorValue[0] |= int(bit)
+				tcpOperatorsFlagsValues = append(tcpOperatorsFlagsValues, operatorValue)
+				operatorValue[0] = 0
+				operatorValue[1] = 0
+				// resetAllFlagsToFalse(tcpFlagBitMap, tcpFlagOpBitMap)
```
Matt-Texier: Argh ... Sorry I forgot this one !

docs/sources/flowspec.md
Matt-Texier: It is not very explicit but I will just mention = and ! in this section ... people can refer to flowspec.md examples to have a more detailed explanation of the syntax.

Matt-Texier:
Hopefully this one will be fine :) !  Sorry for the multiple commits but I getting better to it now ...
Thanks Matt.

gobgp/cmd/global.go
```diff
@@ -863,7 +863,7 @@ usage: %s rib %s%%smatch <MATCH_EXPR> then <THEN_EXPR> -a %%s
 			ExtCommNameMap[RATE], ExtCommNameMap[REDIRECT],
 			ExtCommNameMap[MARK], ExtCommNameMap[ACTION], ExtCommNameMap[RT])
 		ipFsMatchExpr := fmt.Sprintf(`   <MATCH_EXPR> : { %s <PREFIX> [<OFFSET>] | %s <PREFIX> [<OFFSET>] |
-                    %s <PROTO>... | %s <FRAGMENT_TYPE> | %s [not] [match] <TCPFLAG>... |
+                    %s <PROTO>... | %s <FRAGMENT_TYPE> | %s [!] [=] <TCPFLAG>... |
```
ishidawataru: @Matt-Texier Please do the same update in `flowspec.md`

Matt-Texier:
One more commit ... not yet as good as I was expecting to be :-) !

fujita:

Matt-Texier:
Thank you ! if you believe it could make sense I will investigate to do something similar for decimal values

PR #1230, opened 2017-02-10
zebra: Process per-route MTU zebra message

tatsushid:
Since Quagga 1.0.20160309, it implements per-route MTU handling which
adds MTU attribute to every ZEBRA_IPV4_ROUTE_ADD and
ZEBRA_IPV6_ROUTE_ADD messages. It causes "message length invalid" error
when GoBGP receives messages from Zebra and as the result, routes from
Zebra are no longer added to GoBGP.

This fixes the issue by decoding/encoding MTU attribute properly.

Now parsed MTU attribute is not used anywhere, just kept in an internal
structure.


fujita:
Thanks a lot!

PR #1232, opened 2017-02-11
gobgpd: Support for forwarding logs to fluentd

a16:
Command line options:
 --fluent `host:port` enables to forward logs to fluentd
 --fluent-tag `tag` sets `tag` on forwarded logs (Default tag is `gobgpd`)

Ex.)
gobgpd command
```bash
gobgpd -f gobgpd.conf --fluent :24224
```

fluentd config
```bash
<source>
  @type forward
  port 24224
</source>

<match gobgpd>
  @type stdout
</match>
```

### DIFF
--- a/gobgpd/main.go
+++ b/gobgpd/main.go
@@ -16,6 +16,14 @@
 package main
 
 import (
+	"io/ioutil"
+	"net/http"
+	_ "net/http/pprof"
+	"os"
+	"os/signal"
+	"runtime"
+	"syscall"
+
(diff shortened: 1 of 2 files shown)


ishidawataru:
How about using this?
http://docs.fluentd.org/v0.12/articles/in_syslog

I think it'd be better to keep gobgpd's cmd line options simple as much as possible.

fujita:
Please reopen this if the above approach doesn't work well. I'll push.

PR #1237, opened 2017-02-14
zebra/zapi: Support some messages for BGP Nexthop Tracking

iwaseyusuke:
These patches add the missing message types which defined in Quagga version 1.1.1 and also add support for NEXTHOP_UPDATE, NEXTHOP_REGISTER and NEXTHOP_UNREGISTER messages.
These messages are required to implement BGP Nexthop Tracking features.


fujita:
Thanks!

PR #1258, opened 2017-02-27
This Pull proposes CLI to comply with new draft RFC5575bis normalized…

Matt-Texier:
This pull allows to support all the following operators with decimal values used for port, destination ports, source ports, protocol number:

```
            +----+----+----+----------------------------------+
            | lt | gt | eq | Resulting operation              |
            +----+----+----+----------------------------------+
            | 0  | 0  | 0  | true (independent of the value)  |
            | 0  | 0  | 1  | == (equal)                       |
            | 0  | 1  | 0  | > (greater than)                 |
            | 0  | 1  | 1  | >= (greater than or equal)       |
            | 1  | 0  | 0  | < (less than)                    |
            | 1  | 0  | 1  | <= (less than or equal)          |
            | 1  | 1  | 0  | != (not equal value)             |
            | 1  | 1  | 1  | false (independent of the value) |
            +----+----+----+----------------------------------+
```

### DIFF
--- a/docs/sources/flowspec.md
+++ b/docs/sources/flowspec.md
@@ -50,7 +50,7 @@ CLI syntax to add ipv4/ipv6 flowspec rule is
    <PROTO> : ospf, pim, igp, udp, igmp, tcp, egp, rsvp, gre, ipip, unknown, icmp, sctp, <VALUE>
    <FRAGMENT_TYPE> : dont-fragment, is-fragment, first-fragment, last-fragment, not-a-fragment
    <TCPFLAG> : U, C, E, F, S, R, P, A
-   <ITEM> : &?{<|>|=}<value>
+   <ITEM> : &?{<|>|>=|<=|==|!=}<value>
    <THEN_EXPR> : { accept | discard | rate-limit <value> | redirect <RT> | mark <value> | action { sample | terminal | sample-terminal } | rt <RT>... }...
    <RT> : xxx:yyy, xx.xx.xx.xx:yyy, xxx.xxx:yyy
 ```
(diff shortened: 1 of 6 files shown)


ishidawataru:
Tested `parseDecValuesCmd` with follwing inputs.

```go
func Test_ParseDecValuesCmd(t *testing.T) {
  t.Log(parseDecValuesCmd("true", func(int) error { return nil }))
  t.Log(parseDecValuesCmd("100", func(int) error { return nil }))
  t.Log(parseDecValuesCmd("==100", func(int) error { return nil }))
}
```

```bash
$ go test -run ParseDec -v
=== RUN   Test_ParseDecValuesCmd
--- PASS: Test_ParseDecValuesCmd (0.00s)
        bgp_test.go:744: [[128 0]] <nil>
        bgp_test.go:745: [[128 100]] <nil>
        bgp_test.go:746: [[129 100]] <nil>
PASS
```

The result of the second test is expected?
What about setting `==` with bare values? 
Then we don't need to change [here](https://github.com/osrg/gobgp/pull/1258/files#diff-efb83e5a030203af28a46045cb041c16R47)

Matt-Texier:
Hi,

Yep I see ... good catch, there is a BUG here for bare values, 

Agree I'll work on setting `==` as default for bare value.

Working on it today ...

Thanks Matt.

ishidawataru:
https://travis-ci.org/osrg/gobgp/jobs/206091610

Please go fmt.

Matt-Texier:
Sorry for that ...

Matt-Texier:
btw, looks like travis sucks those days :-) !

ishidawataru:
thanks, LGTM.

you can run the same CI for your own forked repository before opening PR as I'm doing.

https://travis-ci.org/ishidawataru/gobgp

fujita:
Thanks, pushed!

PR #1259, opened 2017-03-02
zebra/zapi: Utility functions to decode/serialize nexthops

iwaseyusuke:
This patch implements the utility functions for decoding/serializing the nexthops list contained in some Zebra messages and fixes to use these functions to remove duplicated codes.


fujita:
Nice, thanks!

PR #1266, opened 2017-03-07
scenario_test: Remove "sudo" command prefix for running tests

iwaseyusuke:
To separate Python development environment from OS's Python environment, we can use Python virtualenv.
But to enable the root privilege on virtualenv, we need to specify Python interpreter on its virtualenv.
e.g.)
  $ sudo -E .venv/bin/python <script>

Currently, scenario_test requires the root privilege (for Docker execution?), so it is required some modification to scenario_test on virtualenv.
We need to modify all Python interpreter into virtualenv's interpreter in "scenario_test/run_all_tests.sh" for example. 

To solve this problem, this patch removes "sudo" command when running tests.

Note: To enable to execute Docker without the root privilege, please add your local user into "docker" group. With this setting, Docker does not require the root privilege.

### DIFF
--- a/test/scenario_test/bgp_router_test.py
+++ b/test/scenario_test/bgp_router_test.py
@@ -458,9 +458,6 @@ class GoBGPTestBase(unittest.TestCase):
 
 
 if __name__ == '__main__':
-    if os.geteuid() is not 0:
-        print "you are not root."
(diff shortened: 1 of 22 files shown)


fujita:
Looks fine, thanks.

PR #1270, opened 2017-03-16
Code-Inspecttion: pep8 and pylint for scenario_test

iwaseyusuke:
The following patches improve the result of pep8 and pylint for the Python scripts of scenario_test.

Also, the last patch fixes the error of "test/scenario_test/zapi_v3_test.py" on Travis-Ci which was cased by the update of "osrg/quagga:v1.0" Docker image. (Please see the commit message for the detail of this problem.)


fujita:
Thanks, pushed! I don't worship pep8 stuff much but not a bad thing, I guess.

iwaseyusuke:
I agree.

Many of editors (IDE) and other Python projects use pep8/pylint for evaluating codes based on "third party" perspective.
So I applied them to scenario_test, but I guess this PR should be just improvements, and should not be forced against other codes.
(First, I introduce pep8 check into Travis-CI, but it might be too much...)

PR #1271, opened 2017-03-17
scenario_test: Add test for Nexthop Tracking

iwaseyusuke:
The following patches implement the simple test case for the Nexthop Tracking feature which provided with Quagga/Zebra integration.


fujita:
Nice, pushed. Thanks!

PR #1278, opened 2017-03-22
server/zclient: Retry zebra message version negotiation

iwaseyusuke:
Currently, the Zebra message version used by ZClient is configurable (default 2), but if the given version is miss-matched with that of Zebra daemon, ZCient will fail to connect.

This patch fixes ZClient to retry the version negotiation.
For example, if failed with the version 2, retry with the version 3.

Note: In order to receive the first message from Zebra daemon when instantiating ZClient, this patch fixes ZClient to send HELLO and ROUTER_ID_ADD messages automatically.

Also the following fixes the default value for Zebra.Config.Version.


fujita:
Nice, thanks a lot!

PR #1287, opened 2017-03-30
zebra/zapi: Fix offset calculation in IPRouteBody

iwaseyusuke:
Currently, if the received IP route message has not Distance field, zebra/zapi will fail to parse message, because "offset" is incremented *before* parsing each field and it cases index error.
("offset" should be incremented *after* parsing each field.)
This patch fixes this problem.


fujita:
Thanks!

PR #1290, opened 2017-04-01
cli: don't calculate # of advertised routes when not necessary

ishidawataru:
Signed-off-by: Wataru Ishida <ishida.wataru@lab.ntt.co.jp>


fujita:
Thanks!

PR #1292, opened 2017-04-04
scenario tests: Use local GoBGP executables

iwaseyusuke:
Currently, to reflect the modification of the local source code or to switch the base image for GoBGP container, it is required to re-build the GoBGP container image, and this take a long time to test or debug with the scenario tests.

This patch fixes to use the local GoBGP executables (gobgp and gobgpd) in container, and enables to reflect the changes without re-building container image.

Note: according to this change, the following command is no longer required and removed, but it is required to build GoBGP executables before running the scenario tests.
Removed command: `$ sudo fab -f ./test/lib/base.py make_gobgp_ctn --set tag=gobgp`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ script: test/scenario_test/ci-scripts/travis-build-script.sh
 
 env:
   global:
-    - DOCKER_IMAGE=gobgp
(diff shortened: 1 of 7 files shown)


fujita:
Thanks, pushed!

Issue #1281, opened 2017-03-24
Feature Request: Support for Local RIB in BGP Monitoring Protocol (BMP)


mstrother:
[https://tools.ietf.org/html/draft-evens-grow-bmp-local-rib-00.html](https://tools.ietf.org/html/draft-evens-grow-bmp-local-rib-00.html)

The above draft defines a new peer type, Peer Type = 3: Loc-RIB Instance Peer. The peer type adds access to to the BGP Local-RIB (RFC 4271) after the local BGP speaker's decisions process. 

Currently monitoring the Loc-RIB requires two routers and a BGP session between them.

iwaseyusuke:
Hi @mstrother 

That sounds great feature for me!
Just to confirm... in this case, GoBGP should be an BMP client, but is there any BGP Sever supporting this feature?
I wonder wow we should test the implementation? Do you have any idea or some implementations?

mstrother:
BMP sends all routes in the Adj-RIB-In of every peer but does not provide a Local RIB view of the routes after best path selection. This is overly complex for simple monitoring. It also results in a very large amount of unwanted data. For example, with 5 peers and full routing tables it results in millions messages  upon starting GoBGP. In addition it doesn't allow monitoring of your own announcements (Adj-RIB-Out). The monitoring application has to reproduce the best path selection process.

The new peer type would add a third view to BMP.

Adj-RIB - per peer (pre-policy)
Adj-RIB - per peer (post-policy)
_Local-RIB - (after best path selection)_

A monitoring application may wish to collect all one or more of the views. For example, my default configuration may be Local-RIB but to debug a problem with a peer I may need to see their announcements even if they don't make it past best path selection and into the Local RIB.

This would need to be configurable:

[[bmp-servers]]
  [bmp-servers.config]
    address = "127.0.0.1"
    port=11019
    route-monitoring-policy = [ "pre-policy", "post-policy", "local" ]

iwaseyusuke:
Thank you for your reply!

It makes it clear for me!
I understand (roughly) that this draft adds the type of view in which the peer is interested, and enable to monitor the routes after the best path selection.

Thanks!

mstrother:
Yes. The same information I would see with `gobgp global rib`

mstrother:
There should be a way to monitor pre, post, and local-rib policy. Having more than two route monitoring policies also makes the configuration setting route-monitoring-policy = "local-rib" unclear.

[...]

PR #1297, opened 2017-04-06
BMP: Support for "bmp-draft-evens-grow-bmp-local-rib"

iwaseyusuke:
These patches enable to send BMP Route Monitoring message for Local RIB routes described in "bmp-draft-evens-grow-bmp-local-rib".

Configuration Example: gobgpd.toml
  ...
  [[bmp-servers]]
    [bmp-servers.config]
      address = "127.0.0.1"
      port=11019
      route-monitoring-policy = "local-rib"
  ...

issues #1281

Note: As @mstrother said, I guess we should have "route-monitoring-policy" to be configurable as a list of the interesting route types, but currently we need to configure it as a string value.
Could we accept the both types (list or string) as route-monitoring-policy"? How to implement it?
Or, should we have a new key like "route-monitoring-**policies**" or so?

### DIFF
--- a/config/bgp_configs.go
+++ b/config/bgp_configs.go
@@ -858,12 +858,14 @@ const (
 	BMP_ROUTE_MONITORING_POLICY_TYPE_PRE_POLICY  BmpRouteMonitoringPolicyType = "pre-policy"
 	BMP_ROUTE_MONITORING_POLICY_TYPE_POST_POLICY BmpRouteMonitoringPolicyType = "post-policy"
 	BMP_ROUTE_MONITORING_POLICY_TYPE_BOTH        BmpRouteMonitoringPolicyType = "both"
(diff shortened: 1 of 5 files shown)


mstrother:
As monitoring more than 1 route type is a likely scenario it should be configurable. Changing route-monitoring-policy to a list type would be easiest but (I think) would be a breaking change.

fujita:
Thanks, pushed! Sorry about the delay.

mstrother:
There should be a way to monitor pre, post, and local-rib policy. Having more than two route monitoring policies also makes the configuration setting `route-monitoring-policy = "local-rib"` unclear.

As stated in "bmp-draft-evens-grow-bmp-local-rib" :

For example, in order to validate if multi-path prefixes are installed as expected for all advertising peers, the Adj-RIB-In Post-Policy and Loc-RIB needs to be compared.

Thoughts?

Issue #1301, opened 2017-04-18
Default route issue when looking up prefixes


lspgn:
Hello,
I was trying to setup a default route and lookup a subnet with "shorter-prefixes" but it does not seem to work.

```
$ ./gobgp global rib add 0.0.0.0/0
$./gobgp global rib
    Network             Next Hop             AS_PATH              Age        Attrs
*>  0.0.0.0/0           0.0.0.0                                   00:00:05   [{Origin: ?}]
$./gobgp global rib 1.2.3.4/32 shorter-prefixes
Network not in table
```

```
$ ./gobgp global rib add 0.0.0.0/1
$./gobgp global rib
    Network             Next Hop             AS_PATH              Age        Attrs
*>  0.0.0.0/1           0.0.0.0                                   00:00:23   [{Origin: ?}]
$./gobgp global rib 1.2.3.4/32 shorter-prefixes
*>  0.0.0.0/1           0.0.0.0                                   00:00:36   [{Origin: ?}]
```

I believe the issue is [there (server.go#L1473)](https://github.com/osrg/gobgp/blob/master/table/table.go#L379):
`for i := ones; i > 0; i-- {`
(thus missing the default)
instead of
`for i := ones; i >= 0; i-- {`

Thank you

ishidawataru:
Thanks, yes, you are right.
Could you open a PR to fix the issue?

lspgn:
Done, you can see [PR-1305](https://github.com/osrg/gobgp/issues/1305).
Thanks.

PR #1305, opened 2017-04-20
table: Fixed bug when looking up prefixes with a default route.

lspgn:
For issue [1301](https://github.com/osrg/gobgp/issues/1301).
The last iteration of the loop was 1 so the mask calculation stopped at /1 thus leaving blank any lookup of a prefix when a default route is present.


fujita:
Pushed, thanks a lot!

PR #1324, opened 2017-05-12
bgp: Advertise IPv4 routes with IPv6 Next Hop

iwaseyusuke:
RFC5549 allows IPv6 Next Hop address for the advertisement of IPv4 related NLRIs for <AFI/SAFI> of <1/1>, <1/2>, <1/4> and <1/128>.
Currently, the advertisement using the MP_REACH_NLRI is supported, but IPv4 routes with IPv6 Next Hop is not enough.

This patch enable to advertise IPv4 routes for <AFI/SAFI> of <1/1> through GoBGP CLI command.
e.g.)
  $ gobgp global rib add -a ipv4 10.2.1.0/24 nexthop 2001:2::1

### DIFF
--- a/api/grpc_server.go
+++ b/api/grpc_server.go
@@ -568,7 +568,7 @@ func (s *Server) api2PathList(resource Resource, ApiPathList []*Path) ([]*table.
 			return nil, fmt.Errorf("not found nlri or nexthop")
 		}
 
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks!

PR #1325, opened 2017-05-12
bmp: Enable to specify all route monitoring policies

iwaseyusuke:
This patch adds the new key "all" for specifying the all BMP route monitoring policies.

Configuration Example:
```
  [[bmp-servers]]
    [bmp-servers.config]
      address = "127.0.0.1"
      port=11019
      route-monitoring-policy = "all"
```


fujita:
Can you remove "both" option?

iwaseyusuke:
@fujita 

Yes, we can remove "both" option I guess.
I left it just for the backward compatibility.

If we remove "both" option completely, the warning message like "both option is obsoleted" should be required?

fujita:
Pushed, thanks!

PR #1327, opened 2017-05-15
tools: Update usage of pyang_plugins

iwaseyusuke:
The current document refers Python Yang package (pyang), but does not explain how to install it.
This patch update to provide an additional explanation for pyang.
Also, this patch updates the format of this document.

### DIFF
--- a/tools/pyang_plugins/README.rst
+++ b/tools/pyang_plugins/README.rst
@@ -3,21 +3,38 @@ What's this ?
 This is a pyang plugin to generate config/bgp_configs.go from
 openconfig yang files (see https://github.com/openconfig/public).
 
+Prerequisites
+=============
+Please confirm $GOPATH is configured before the following steps.
+
 How to use
 ==========
-::
+Set the environment variables for this tool::
 
+   $ GOBGP_PATH=$GOPATH/src/github.com/osrg/gobgp/
+
+Clone the required resources by using Git::
+
+   $ cd $HOME
    $ git clone https://github.com/osrg/public
    $ git clone https://github.com/YangModels/yang
-   $ YANG_DIR=`pwd`
-   $ cd $PYANG_INSTALL_DIR
+   $ git clone https://github.com/mbj4668/pyang
+
+Setup environments for pyang::
+
+   $ cd $HOME/pyang
    $ source ./env.sh
-   $ PYTHONPATH=. ./bin/pyang --plugindir $GOBGP_PATH/tools/pyang_plugins \
-   -p $YANG_DIR/yang/standard/ietf/RFC \
-   -p $YANG_DIR/public/release/models \
-   -p $YANG_DIR/public/release/models/bgp \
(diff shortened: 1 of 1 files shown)


ishidawataru:
it would be helpful to add which commit of openconfig we should check out to generate golang config since the latest model doesn't work anymore with the plugin.

https://github.com/openconfig/public/commit/92b36c2adacd2ed16dbd8609419b91bc32556829

I'm checking out this commit, but not sure this is the latest one which we can successfully generate the config.

iwaseyusuke:
@ishidawataru 

Thank you for reviewing my patch!

Yes, I also found this plugin does not work with the latest `openconfig/public`.
So I guess this doc instructs to checkout [`osrg/public`](https://github.com/osrg/public).
With the latest `osrg/public`, this plugin should work.

But...
There are still some error messages.
```
$ PYTHONPATH=. ./bin/pyang --plugindir $GOBGP_PATH/tools/pyang_plugins -p $HOME/yang/standard/ietf/RFC -p $HOME/public/release/models -p $HOME/public/release/models/bgp -p $HOME/public/release/models/policy -f golang $HOME/public/release/models/bgp/openconfig-bgp.yang $HOME/public/release/models/policy/openconfig-routing-policy.yang $GOBGP_PATH/tools/pyang_plugins/gobgp.yang | gofmt > $GOBGP_PATH/config/bgp_configs.go
/home/iwase/public/release/models/bgp/openconfig-bgp.yang:428: error: restriction require-instance not allowed for this base type
/home/iwase/public/release/models/bgp/openconfig-bgp-policy.yang:140: error: restriction require-instance not allowed for this base type
/home/iwase/public/release/models/bgp/openconfig-bgp-policy.yang:163: error: restriction require-instance not allowed for this base type
/home/iwase/public/release/models/bgp/openconfig-bgp-policy.yang:184: error: restriction require-instance not allowed for this base type
/home/iwase/public/release/models/bgp/openconfig-bgp-policy.yang:443: error: restriction require-instance not allowed for this base type
/home/iwase/public/release/models/bgp/openconfig-bgp-policy.yang:492: error: restriction require-instance not allowed for this base type
warning bgp-types:percentage: percentage has already been emitted from openconfig-types:percentage.
```
I guess changing the yang version ("1" -> "1.1") is required to fix this problem, but I don't know whether it is appropriate or not.

ishidawataru:
Oops my bad, I overlooked that the repository used in the script is `osrg/public`.

> but I don't know whether it is appropriate or not.

me neither.

https://github.com/openconfig/public/issues/6

It seems the issue is solved in the latest openconfig model.

iwaseyusuke:
@ishidawataru 
I tried with the commit addressing this issue, but had no luck.
This plugin seems not to work with that commit...

fujita:
Pushed, thanks.
I guess that we'll use the newer openconfig and yang someday. Even with some errors, the current versions work for us so let's use them for now.

PR #1328, opened 2017-05-16
packet/bgp: Support Redirect-IPv6 Flow Spec Traffic Filtering Action

fujimoto-s:
This patch supports Redirect-IPv6 Flow Spec Traffic Filtering Action (draft-ietf-idr-flow-spec-v6-08).
Command example:
    `gobgp global rib add -a ipv6-flowspec match protocol udp then redirect 2001:db8::68:1000`

### DIFF
--- a/gobgp/cmd/global.go
+++ b/gobgp/cmd/global.go
@@ -108,17 +108,19 @@ func redirectParser(args []string) ([]bgp.ExtendedCommunityInterface, error) {
 	if err != nil {
 		return nil, err
 	}
-	t, _ := rt.GetTypes()
-	switch t {
-	case bgp.EC_TYPE_TRANSITIVE_TWO_OCTET_AS_SPECIFIC:
+	switch rt.(type) {
+	case *bgp.TwoOctetAsSpecificExtended:
 		r := rt.(*bgp.TwoOctetAsSpecificExtended)
 		return []bgp.ExtendedCommunityInterface{bgp.NewRedirectTwoOctetAsSpecificExtended(r.AS, r.LocalAdmin)}, nil
-	case bgp.EC_TYPE_TRANSITIVE_IP4_SPECIFIC:
+	case *bgp.IPv4AddressSpecificExtended:
 		r := rt.(*bgp.IPv4AddressSpecificExtended)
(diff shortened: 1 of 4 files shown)


gobgp/cmd/global.go
```diff
@@ -809,8 +811,21 @@ func ParsePath(rf bgp.RouteFamily, args []string) (*table.Path, error) {
 		if err != nil {
 			return nil, err
 		}
-		p := bgp.NewPathAttributeExtendedCommunities(extcomms)
+		normalextcomms := make([]bgp.ExtendedCommunityInterface, 0)
+		ipv6extcomms := make([]bgp.ExtendedCommunityInterface, 0)
+		for _, com := range extcomms {
+			bstring, _ := com.Serialize()
+			switch len(bstring) {
+			case 20:
```
ishidawataru: How about using `com.GetTypes()` to check if we need to put this community as an IPv6 extended community path attribute?

gobgp/cmd/global.go
```diff
@@ -809,8 +811,21 @@ func ParsePath(rf bgp.RouteFamily, args []string) (*table.Path, error) {
...
+				ipv6extcomms = append(ipv6extcomms, com)
+			default:
+				normalextcomms = append(normalextcomms, com)
+			}
+		}
+		p := bgp.NewPathAttributeExtendedCommunities(normalextcomms)
 		attrs = append(attrs, p)
+		ip6p := bgp.NewPathAttributeIP6ExtendedCommunities(ipv6extcomms)
```
ishidawataru: This always adds an IPv6 extended community path attribute even if `len(ipv6extcomms) == 0`. We need to check both `len(normalextcomms)` and `len(ipv6extcomms)`.

gobgp/cmd/global.go
fujimoto-s: In that case, IPv6 Address Specific Extended Community cannot be distinguished from the existing community, Two-octet AS Specific Extended Community, because both have similar type value(0x00XX).
So, here I use length instead of type value.
Do you have any idea?

ishidawataru: Hmm...I see. Then how about using type switch?

fujimoto-s: OK, I will modify this code with type switch. Thanks!

fujita:
Thanks!

PR #1330, opened 2017-05-17
server/fsm_test.go: fix race condition during tests

cstockton:
Add sync.Mutex to guard access to MockConnection to prevent tests ran with -race from failing. In addition change the calls to the fmt package to use t.Log show they only show with the -v flag to prevent unwanted noise between test runs. It has the added benefit of providing a file and line number as well.

### DIFF
--- a/server/fsm_test.go
+++ b/server/fsm_test.go
@@ -16,30 +16,35 @@
 package server
 
 import (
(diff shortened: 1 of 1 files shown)


fujita:
Pushed, thanks!

PR #1332, opened 2017-05-18
bmp: Stats reports for Adj-RIBs-In and Loc-RIB routes

iwaseyusuke:
This PR enables to send BMP statistics reports for the number of routes in Adj-RIBs-In and Loc-RIB.
Also includes the improvements of BMP statistics reports serializer and the updates of documents for BMP configurations.

### DIFF
--- a/server/bmp.go
+++ b/server/bmp.go
@@ -113,22 +113,31 @@ func (b *bmpClient) loop() {
 
 		if func() bool {
 			ops := []WatchOption{WatchPeerState(true)}
-			if b.typ == config.BMP_ROUTE_MONITORING_POLICY_TYPE_BOTH {
+			if b.c.RouteMonitoringPolicy == config.BMP_ROUTE_MONITORING_POLICY_TYPE_BOTH {
 				log.WithFields(
 					log.Fields{"Topic": "bmp"},
 				).Warn("both option for route-monitoring-policy is obsoleted")
 			}
-			if b.typ == config.BMP_ROUTE_MONITORING_POLICY_TYPE_PRE_POLICY || b.typ == config.BMP_ROUTE_MONITORING_POLICY_TYPE_ALL {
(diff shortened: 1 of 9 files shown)


server/bmp.go
```diff
@@ -113,16 +113,27 @@ func (b *bmpClient) loop() {
...
+			} else if b.c.RouteMonitoringPolicy == config.BMP_ROUTE_MONITORING_POLICY_TYPE_LOCAL_RIB {
 				ops = append(ops, WatchBestPath(true))
 			}
 			w := b.s.Watch(ops...)
 			defer w.Stop()
 
+			var tickerCh <-chan time.Time
+			if b.c.StatisticsTimeout <= 0 {
+				log.WithFields(log.Fields{"Topic": "bmp"}).Debug("statistics reports disabled")
+			} else if 15 <= b.c.StatisticsTimeout && b.c.StatisticsTimeout <= 65535 {
```
ishidawataru: What about do the validation here?

https://github.com/osrg/gobgp/blob/master/config/default.go#L218-L234

tools/pyang_plugins/gobgp.yang
```diff
@@ -549,6 +549,11 @@ module gobgp {
       type bmp-route-monitoring-policy-type;
       default PRE-POLICY;
     }
+    leaf statistics-timeout {
```
ishidawataru: `statistics-interval` is a better name?

server/bmp.go
iwaseyusuke: Thanks!
I will move this validation to `config/default.go`.

tools/pyang_plugins/gobgp.yang
iwaseyusuke: It just follows Junos OS configuration name [`statistics-timeout`](https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/statistics-timeout-edit-routing-options.html).
Also I picked the range for this value from Junos OS.
(Sorry I don't have strong will...)

ishidawataru: ah, I see. then I think we can keep it as is.

fujita:
Thanks!

PR #1333, opened 2017-05-18
table: reduce allocs in NewDestination by a factor of 3(ipv4)-5(ipv6)

cstockton:
In NewDestination we add additional path to AddrToRadixkey to check
the concrete type of nlri. If it is one of IP{,v6}AddrPrefix we will
use the existing Prefix and Length to avoid 2 allocs in ParseCidr
and 1 alloc in WriteString. 2 additional allocations are saved for
ipv6 due the larger RadixKey size the buffer ends up growing, since
it also escapes both need heap allocated. I bypass this by creating
a single buffer and passing 8 byte chunks of the slice header to
byteToBase2Buf.

Notes:
 - This was tested via [bench_test.go](https://gist.github.com/cstockton/ce806b81f3fcf0af190bebd20ecdfdae) - reduced from my own GoBGP tests but included for reference.
 - There may be other cases in the AddrToRadixkey switch - I am still learning my way around the code base so I wanted to only provide these two main paths and leave it the experts to extend to less common cases.
 - I made sure all the tests in table and server pass as, but it could use some more rigorous testing.
 - I made sure to maintain the old behavior for IPv4 RadixKey's of omitting the last byte, I'm sure it's expected but annotating here in case the current behavior is an oversight. I.E.:
   IpToRadixkey(net.ParseIP("10.3.32.0").To4(), 32)
       == IpToRadixkey(net.ParseIP("10.3.32.100").To4(), 32)
```

// Old:
// BenchmarkBasic/IPv4-8 100 13966201 ns/op 3838460 B/op 112012 allocs/op
// BenchmarkBasic/IPv6-8 50 26316389 ns/op 5884964 B/op 220010 allocs/op
//
// New:
// BenchmarkBasic/IPv4-8 200 5973462 ns/op 2816023 B/op 40251 all
// BenchmarkBasic/IPv6-8 200 6977079 ns/op 3551822 B/op 40250 all
```

### DIFF
--- a/table/destination.go
+++ b/table/destination.go
@@ -50,18 +50,51 @@ const (
 	BPR_NON_LLGR_STALE     BestPathReason = "no LLGR Stale"
 )
 
-func IpToRadixkey(b []byte, max uint8) string {
-	var buffer bytes.Buffer
-	for i := 0; i < len(b) && i < int(max); i++ {
-		buffer.WriteString(fmt.Sprintf("%08b", b[i]))
+func byteToBase2Buf(byt byte, buf []byte) {
+	i := len(buf)
+	for byt >= 2 {
+		i--
+		buf[i] = byte('0' + byt&1)
+		byt >>= 1
 	}
-	return buffer.String()[:max]
+	i--
+	buf[i] = byte('0' + byt)
+	for i > 0 {
+		i--
+		buf[i] = '0'
+	}
+}
+
+func IpToRadixkey(b []byte, max uint8) string {
+	buf := make([]byte, max)
+	for y, i := 8, 0; y <= len(buf); y, i = y+8, i+1 {
+		byteToBase2Buf(b[i], buf[y-8:y])
+	}
+	return string(buf[:max])
 }
 
 func CidrToRadixkey(cidr string) string {
 	_, n, _ := net.ParseCIDR(cidr)
 	ones, _ := n.Mask.Size()
 	return IpToRadixkey(n.IP, uint8(ones))
+}
+
+func AddrToRadixkey(addr bgp.AddrPrefixInterface) string {
+	var (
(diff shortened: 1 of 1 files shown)


ishidawataru:
Thanks for the PR!

> I made sure to maintain the old behavior for IPv4 RadixKey's of omitting the last byte, I'm sure it's expected but annotating here in case the current behavior is an oversight. I.E.:
IpToRadixkey(net.ParseIP("10.3.32.0").To4(), 32)
== IpToRadixkey(net.ParseIP("10.3.32.100").To4(), 32)

I couldn't reproduce the behavior.
Could you explain a little bit more how to reproduce the behavior?

I tried below

```go
package main

import (
        "fmt"
        "net"

        "github.com/osrg/gobgp/table"
)

func main() {
        fmt.Println(table.IpToRadixkey(net.ParseIP("10.3.32.0").To4(), 32))
        fmt.Println(table.IpToRadixkey(net.ParseIP("10.3.32.100").To4(), 32))
}
```

and get

```
vagrant@vagrant ~ $ go run test.go
00001010000000110010000000000000
00001010000000110010000001100100
```

---

What about simplifying like below?

```go
func byteToBase2Buf(byt byte, buf []byte) {
        i := len(buf)
        for i > 0 {
                i--
                buf[i] = byte('0' + byt&1)
                byt >>= 1
        }
}
```

cstockton:
Thanks for the comments, the note about the radix key was a annotation I made in my notes when I started that should have been omitted (before I tested it) sorry.

I simplified the byteToBase2Buf func similar to how you described, I also added the test I was using to ensure the output was identical, some of this pull req overlaps with [1334](https://github.com/osrg/gobgp/pull/1334) so I updated that one, I'll close this.

PR #1334, opened 2017-05-18
table: reduce allocations in destination selection by an additional factor of 5

cstockton:
This round of changes is slightly more aggressive than the last but
provides a flat reduction in allocations for the Select paths by an
additional factor of 5.

In table/path.go we provide a new reset method akin to clone, see
inline comments.
In table/destination.go we recalculate the RadixKey we can now derive
our new route destinations with the prior changeset in place to prevent
from the previous, skipping this expensive step. We then reuse the
existing paths at line 1012 that have already been created calling the
newly added reset() method mentioned above. **note:** it's unclear
if applying a empty filter is actually necessary, because it returns the
zero value in all the cases I saw resulting in the same behavior. If this
skip can be skipped it should, since it surprisingly accounts for around
25-35% of the cpu time.

This brings us to 2x allocs per row, which is fair. Anything beyond this will come down to rules around ownership of memory, we could reduce this to 1x allocation in some cases but would forbid callers from mutating the results. It could be technically be reduced to 0 allocations if a separate path was provided specifically for serving requests but I think from 220k to 8k is a good start :- )

```Go
BenchmarkBasic/IPv4-8         	     300	   3923225 ns/op	  959523 B/op	    8249 allocs/op
BenchmarkBasic/IPv6-8         	     500	   3697597 ns/op	  959641 B/op	    8250 allocs/op
PASS
```

### DIFF
--- a/table/destination.go
+++ b/table/destination.go
@@ -50,27 +50,23 @@ const (
 	BPR_NON_LLGR_STALE     BestPathReason = "no LLGR Stale"
 )
 
-func byteToBase2Buf(byt byte, buf []byte) {
-	i := len(buf)
-	for byt >= 2 {
-		i--
-		buf[i] = byte('0' + byt&1)
-		byt >>= 1
-	}
-	i--
-	buf[i] = byte('0' + byt)
-	for i > 0 {
-		i--
-		buf[i] = '0'
-	}
-}
-
 func IpToRadixkey(b []byte, max uint8) string {
-	buf := make([]byte, max)
+	var backing [128]byte
+	buf := backing[:max]
 	for y, i := 8, 0; y <= len(buf); y, i = y+8, i+1 {
-		byteToBase2Buf(b[i], buf[y-8:y])
+		n, byt, cur := 8, b[i], buf[y-8:y]
+		for byt > 0 {
+			n--
+			cur[n] = byte('0' + byt&1)
+			byt >>= 1
+		}
 	}
-	return string(buf[:max])
+	for i := 0; i < len(buf); i++ {
+		if buf[i] == 0 {
+			buf[i] = '0'
+		}
+	}
+	return string(buf)
 }
 
 func CidrToRadixkey(cidr string) string {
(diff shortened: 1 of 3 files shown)


cstockton:
I'm not sure why the build is failing- I'll look this weekend at some point to see how to get the build tests it's running configured locally so I can troubleshoot.

cstockton:
@ishidawataru This passes without the change to destinations, I'll look into why that change failed after this round of improvements is approved.

fujita:
Thanks a lot for the pr!

I like AddrToRadixkey() idea, however, I want to avoid the change to IpToRadixkey(). It's similar to the standard library's format code. I want to keep such code inside the standard library. Of course, I'll push it if that change gives the notable performance boost with real workload (such as injecting the full routes) rather than micro benchmarks.

I agree that we should avoid unnecessary memory allocations... as long as the code is easily understandable, IMHO. So I pushed https://github.com/osrg/gobgp/commit/b4831ce8e91e6f27e3b241efd0c3fdfd8feca43a

Thanks!

cstockton:
It's discouraging to see my change that reduced allocations by 27x overall increasing the throughput by 5x for ipv4 and 10x for ipv6 be replaced with a commit that leaves the same performance and only reduces the allocs by 1.4x while omitting credit to the original contributor (me) who hunted down the issue.

I don't agree that this is a micro benchmark, I started in the highest density area I could find in the call graphs I generated from the code so my work would provide the most benefit. It lives in the Select(...) path which is heavily used for route selection. Let me know if you want me to make changes or feel free to close this I guess and just copy whatever you do like rather than have me fix it.

cstockton:
I updated the pull req as requested via slack, before you decide against the changed to IpToRadixKey figured I would provide a simple benchmark:

The benefits at the cost of a more unsightly function:

- 1 allocation for ipv4 instead of the 6 in HEAD
- 1 allocation for ipv6 instead of the 19 in HEAD
- 6x faster than HEAD for ipv4
- 10x faster than HEAD for ipv6

Performance usually isn't worth readability, but a 6x and 19x increase in allocations is probably worth it in such a hot spot of code given the current memory usage and gc pressure already. But I understand if you opt against it. Gist and benchmark below in case you want to test:

from: https://gist.github.com/cstockton/78a1371c1c388dfb88c35d1bea8fc211
```
// IPv4Standard-8         	 2000000	       781 ns/op	     148 B/op	       6 allocs/op
// IPv6Standard-8         	  500000	      2980 ns/op	     400 B/op	      19 allocs/op
// IPv4StandardFast-8     	10000000	       130 ns/op	      32 B/op	       1 allocs/op
// IPv6StandardFast-8     	 5000000	       288 ns/op	     128 B/op	       1 allocs/op
// PASS
```

fujita:
Squashed and pushed, thanks!

PR #1337, opened 2017-05-22
*: Support Peer-Group Configuration

fujimoto-s:
This patch enables to create peer-groups,
also supports dynamic configuration to peer-groups.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -92,27 +92,29 @@ type BgpServer struct {
 	fsmStateCh    chan *FsmMsg
 	acceptCh      chan *net.TCPConn
 
-	mgmtCh      chan *mgmtOp
-	policy      *table.RoutingPolicy
-	listeners   []*TCPListener
-	neighborMap map[string]*Peer
-	globalRib   *table.TableManager
-	roaManager  *roaManager
-	shutdown    bool
-	watcherMap  map[WatchEventType][]*Watcher
-	zclient     *zebraClient
-	bmpManager  *bmpClientManager
-	mrtManager  *mrtManager
+	mgmtCh       chan *mgmtOp
+	policy       *table.RoutingPolicy
+	listeners    []*TCPListener
+	neighborMap  map[string]*Peer
+	peerGroupMap map[string]*PeerGroup
+	globalRib    *table.TableManager
+	roaManager   *roaManager
+	shutdown     bool
+	watcherMap   map[WatchEventType][]*Watcher
+	zclient      *zebraClient
+	bmpManager   *bmpClientManager
+	mrtManager   *mrtManager
 }
 
 func NewBgpServer() *BgpServer {
 	roaManager, _ := NewROAManager(0)
 	s := &BgpServer{
-		neighborMap: make(map[string]*Peer),
-		policy:      table.NewRoutingPolicy(),
-		roaManager:  roaManager,
-		mgmtCh:      make(chan *mgmtOp, 1),
-		watcherMap:  make(map[WatchEventType][]*Watcher),
+		neighborMap:  make(map[string]*Peer),
+		peerGroupMap: make(map[string]*PeerGroup),
+		policy:       table.NewRoutingPolicy(),
+		roaManager:   roaManager,
+		mgmtCh:       make(chan *mgmtOp, 1),
+		watcherMap:   make(map[WatchEventType][]*Watcher),
 	}
 	s.bmpManager = newBmpClientManager(s)
 	s.mrtManager = newMrtManager(s)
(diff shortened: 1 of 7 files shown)


server/server.go
```diff
@@ -1711,14 +1753,35 @@ func (server *BgpServer) deleteNeighbor(c *config.Neighbor, code, subcode uint8)
...
+	}, true)
+}
+
 func (s *BgpServer) DeleteNeighbor(c *config.Neighbor) error {
 	return s.mgmtOperation(func() error {
 		return s.deleteNeighbor(c, bgp.BGP_ERROR_CEASE, bgp.BGP_ERROR_SUB_PEER_DECONFIGURED)
 	}, true)
 }
 
+func (s *BgpServer) UpdatePeerGroup(pg *config.PeerGroup) error {
```
ishidawataru: Do we also need to update the configuration of neighbors who are using the peer group we are updating here?

I'm not sure about the behavior of other implementations.

server/server.go
```diff
@@ -1711,14 +1753,35 @@ func (server *BgpServer) deleteNeighbor(c *config.Neighbor, code, subcode uint8)
 	return nil
 }
 
+func (s *BgpServer) DeletePeerGroup(c *config.PeerGroup) error {
+	return s.mgmtOperation(func() error {
```
ishidawataru: I think we should not delete a peer group when some neighbor is using it as `(*BgpServer).DeleteVrf()` doing.

server/server.go
fujimoto-s: Yes, we need.

In my patch, in `config.UpdateNeighborConfig()`, neighbors are checked whether the peer-group which it belongs is updated.
And those neighbors will be updated later.

server/server.go
fujimoto-s: Thanks!
I will fix it.

server/server.go
ishidawataru: I think it is better to move this logic in `(*BgpServer).UpdatePeerGroup` to support `peer-group` feature for native go library users. Also, when we support `peer-group` configuration via gRPC API, it would be easier to implement it if the logic only remains in `BgpServer`.

fujimoto-s: Thanks for your review!
I agree with it. I will move my logic to `(*BgpServer).UpdatePeerGroup`.

server/server.go
```diff
@@ -1711,14 +1753,49 @@ func (server *BgpServer) deleteNeighbor(c *config.Neighbor, code, subcode uint8)
...
+	}, true)
+}
+
 func (s *BgpServer) DeleteNeighbor(c *config.Neighbor) error {
 	return s.mgmtOperation(func() error {
 		return s.deleteNeighbor(c, bgp.BGP_ERROR_CEASE, bgp.BGP_ERROR_SUB_PEER_DECONFIGURED)
 	}, true)
 }
 
+func (s *BgpServer) UpdatePeerGroup(pg *config.PeerGroup) (needsSoftResetIn bool, err error) {
```
ishidawataru: We need to wrap here by `s.mgmtOperation()` 
and create a method `updateNeighbor()`  which does what
`UpdateNeighbor()` is doing without `s.mgmtOperation()` .
See `DeleteNeigbor()` as an example.

fujimoto-s: I see.
I will wrap here, and call updateNeighbor(), not UpdateNeighbor()

ishidawataru:
@fujimoto-s  

`s.UpdatePeerGroup` is not working correctly.

Could you look into [this commit](https://github.com/ishidawataru/gobgp/commit/df6a7c441086bd04e50b14f6a05895b14e74b70f)?

I added a simple test and a fix.

We need to store a copy of neighbor config unless configuration comparison in `updateNeighbor()` doesn't work.

ishidawataru:
```go

        g := &config.PeerGroup{
                Config: config.PeerGroupConfig{
                        PeerAs:        10,
                        PeerGroupName: "g",
                },
        }
        if err := s.AddPeerGroup(g); err != nil {
                log.Fatal(err)
        }
        n := &config.Neighbor{
                Config: config.NeighborConfig{
                        NeighborAddress: "127.0.0.1",
                        PeerGroup:       "g",
                        PeerAs:          2,
                },
                Transport: config.Transport{
                        Config: config.TransportConfig{
                                PassiveMode: true,
                        },
                },
        }
        if err := s.AddNeighbor(n); err != nil {
                log.Fatal(err)
        }
```

I expected that this results in configuring a neighbor with peer-as 2, but it didn't. 
The neighbor is configured with peer-as 10.

http://www.cisco.com/c/en/us/td/docs/ios/12_2/iproute/command/reference/fiprrp_r/1rfbgp2.html

> If you specify a BGP peer group by using the peer-group-name argument, all the members of the peer group will inherit the characteristic configured with this command. Specifying the command with an IP address will override the value inherited from the peer group.

Could you fix the behavior?

fujimoto-s:
Thanks for your review!

> We need to store a copy of neighbor config unless configuration comparison in updateNeighbor() doesn't work.

I will fix it.

> I expected that this results in configuring a neighbor with peer-as 2, but it didn't.
The neighbor is configured with peer-as 10.

I saw the article, but I understood as there is a few item which is allowed to override the value inherited from the peer group.
I will investigate, but I think remote-as cannot be override.

fujimoto-s:
http://www.cisco.com/c/en/us/td/docs/ios/12_2/iproute/command/reference/fiprrp_r/1rfbgp2.html
I've read the above article, and I noticed that I had overlooked the important thing:

>By default, members of the peer group inherit all the configuration options of the peer group. Members also can be configured to override the options that do not affect outbound updates.
>
>Peer group members will always inherit the following configuration options: remote-as (if configured), version, update-source, out-route-map, out-filter-list, out-dist-list, minimum-advertisement-interval, and next-hop-self. All the peer group members will inherit changes made to the peer group. 

If we follow the Cisco's behivior, we should keep some neighbor's options except above-mentioned options. I will fix my code.

And, if an AS number is specified to a peer-group, the member doesn't seem to be able to configured with an individual AS number. Please check.

ishidawataru:
Thanks, I checked the behavior of FRR.
As you said, I couldn't override the remote-as. 

```
bgpd(config-router)# neighbor test remote-as 10
bgpd(config-router)# nei
bgpd(config-router)# neighbor 10.0.0.1 peer
bgpd(config-router)# neighbor 10.0.0.1 peer-group test
bgpd(config-router)# neighbor 10.0.0.1 remote
bgpd(config-router)# neighbor 10.0.0.1 remote-as 20
% Peer-group AS 10. Cannot configure remote-as for member
bgpd(config-router)#
```

Could you add an error message to let users knofw we can't change the remote-as for neighbor associated with a peer-group?

fujimoto-s:
Thanks for your comment!

I pushed the commit including:
* making `s.UpdatePeerGroup` works correctly
* overwriting neighbor's configs except manually configured fields and some special fields("remote-as" and so on)
* outputting an error message to notify we can't change the remote-as if peer-group is configured with remote-as

fujita:
Pushed, thanks!

PR #1339, opened 2017-05-23
bmp: Implement Information TLV on Initiation and Termination messages

iwaseyusuke:
Currently, the Information TLV fields are implemented as BMPTLV, but not enough decoded and required to be constructed in binary format.

This PR introduces BMPInfoTLV/BMPTermTLV and makes easy to handle these TLV fields.


fujita:
Pushed, thanks!

PR #1341, opened 2017-05-24
bmp: Support number of withdraw updates and prefixes

iwaseyusuke:
This PR enables to send BMP statistics reports for the following types;
- Stat Type = 11: (32-bit Counter) Number of updates subjected to treat-as-withdraw treatment.
- Stat Type = 12: (32-bit Counter) Number of prefixes subjected to treat-as-withdraw treatment.

Note: Currently, this implementation considers only updates or prefixes received from neighbors, but not enough to follow the handling process described in RFC7606.

### DIFF
--- a/config/bgp_configs.go
+++ b/config/bgp_configs.go
@@ -2168,6 +2168,10 @@ type Received struct {
 	Keepalive uint64 `mapstructure:"keepalive" json:"keepalive,omitempty"`
 	// original -> gobgp:DYNAMIC-CAP
 	DynamicCap uint64 `mapstructure:"dynamic-cap" json:"dynamic-cap,omitempty"`
+	// original -> gobgp:WITHDRAW-UPDATE
+	WithdrawUpdate uint32 `mapstructure:"withdraw-update" json:"withdraw-update,omitempty"`
+	// original -> gobgp:WITHDRAW-PREFIX
+	WithdrawPrefix uint32 `mapstructure:"withdraw-prefix" json:"withdraw-prefix,omitempty"`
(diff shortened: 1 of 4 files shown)


fujita:
How about adding these new counters to gobgp-message-counter instead of creating bmp-counter? They are not bmp specific; they could be exported via grpc api or such.

iwaseyusuke:
Thank you for your reviewing.

Yes, I also tried to implement it as your advise, but I found that `gobgp-message-counter` stores stats info  base on BGP message types (e.g., OPEN, UPDATE, ...), not suitable for "reason" or "situation" based counter (Withdrawn, Invalid, Rejected, ...).
So I introduce the new counter field. I agree this counter is not BMP specific.

fujita:
gobgp-message-counter already abuses bgp-counters-message-types_common by adding 'DynamicCap', 'Discarded', and 'Total'; none of them is BGP message type. If you prefer to create new struct, then how about 'MessageHandling'? Neighbor struct has 'ErrorHandling'

iwaseyusuke:
Thanks,
I overlooked "DynamicCap" field. It seems to be very similar to my case.

BTW, "DynamicCap" doesn't seems to be used anywhere... Is there any right way to increment this counter?

fujita:
I think that we can remove 'DynamicCap'.

iwaseyusuke:
Thanks, I will remove "DynamicCap" field.
And I will update my PR to add BMP stats counter into "gobgp-message-counter".

fujita:
Pushed, thanks!

PR #1344, opened 2017-05-30
Fix RouteDistinguisher parsing and VPNV6 VRF rib parsing

jlamanna:
RouteDistinguisher Parsing is off by 1 when examining the regular expression.

VPNV6 Path parsing is assigning to a v4 rather than a v6 structure.


fujita:
Squashed and pushed, thanks!

PR #1346, opened 2017-05-31
bmp: Enable to send BMP Route Mirroring messages

iwaseyusuke:
This PR enables to sent BMP Route Mirroring messages which described in [RFC7854](https://tools.ietf.org/html/rfc7854).
With these commits, it is possible to monitor all received UPDATE messages before the BGP state comparison.

Note: Currently, this feature notifies UPDATE message only, please refer to the note of the first commit `server: Implement event for BGP message watcher` for this constraint.


fujita:
Pushed, thanks!

PR #1347, opened 2017-05-31
packet/mrt: BGP Additional Path Extensions (RFC8050)

iwaseyusuke:
This PR enables to decode/encode MRT format with BGP Additional Path Extensions which described in [RFC8050](https://tools.ietf.org/html/rfc8050).

Note: This PR includes only decoder/encoder of MRT format, does not include the implementation for advertising routes with Add-Path extensions.


fujita:
Pushed, thanks!

PR #1348, opened 2017-05-31
server: close conn when graceful restart timer expired

ishidawataru:
Signed-off-by: Wataru Ishida <ishida.wataru@lab.ntt.co.jp>

PR #1350, opened 2017-06-02
server_test: Test for Graceful Restart timer expired

iwaseyusuke:
This PR includes and updates PR https://github.com/osrg/gobgp/pull/1348 , and also include improvements for server/server_test.go

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -31,13 +31,16 @@ func TestModPolicyAssign(t *testing.T) {
 	assert := assert.New(t)
 	s := NewBgpServer()
 	go s.Serve()
-	s.Start(&config.Global{
+	err := s.Start(&config.Global{
 		Config: config.GlobalConfig{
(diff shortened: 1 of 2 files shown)


server/server_test.go
```diff
@@ -35,6 +35,7 @@ func TestModPolicyAssign(t *testing.T) {
 		Config: config.GlobalConfig{
 			As:       1,
 			RouterId: "1.1.1.1",
+			Port:     10179,
```
fujita: This test doesn't need to listen any so why not using '-1'?

iwaseyusuke: Thanks, '-1' is better. I will update it.

fujita:
Pushed, thanks. I modified the commit message:
https://github.com/osrg/gobgp/commit/d47c818cb0b8803f62e956940533e82070ff2cf5

Issue #1296, opened 2017-04-05
long lived graceful restart does not work when we have 2 neighbor


kishiguro:
We configure one RR with two iBGP neighbors.

<pre>
RR <-> iBGP1
   <-> iBGP2
</pre>

With this setup, RR is configured as graceful restart's Restarting Router.  First convergence worked fine but when RR's gobgpd is restarted with -r option, iBGP1 can establish the peering but iBGP2 can't establish the peering.

The iBGP2 which can't establish the connection said: "restart flag is not set".   The behavior is not consistent.  Sometimes, iBGP2 can establish the connection but iBGP1 can't.  Among 20 tries, one time both peering can be established but is is very rare.

When this occurs, iBGP2 can't establish the peering forever.  There is no way to operationally fix the status unless we restart iBGP2's gobgpd.

Right now we simply comment out the code in server/fsm.go

 <pre>
						// RFC 4724 4.1
						// To re-establish the session with its peer, the Restarting Speaker
						// MUST set the "Restart State" bit in the Graceful Restart Capability
						// of the OPEN message.
						// if fsm.pConf.GracefulRestart.State.PeerRestarting && cap.Flags&0x08 == 0 {
						// 	log.WithFields(log.Fields{
						// 		"Topic": "Peer",
						// 		"Key":   fsm.pConf.Config.NeighborAddress,
						// 		"State": fsm.state.String(),
						// 	}).Warn("restart flag is not set")
						// 	// send notification?
						// 	h.conn.Close()
						// 	return bgp.BGP_FSM_IDLE, FSM_INVALID_MSG
						// }
						if fsm.pConf.GracefulRestart.Config.NotificationEnabled && cap.Flags&0x04 > 0 {
							fsm.pConf.GracefulRestart.State.NotificationEnabled = true
						}
</pre>

RR gobgpd.conf
<pre>
[global.config]
  as = 1
  router-id = "192.168.55.1"

[[neighbors]]
  [neighbors.config]
    neighbor-address = "192.168.55.2"
    peer-as = 1
  [neighbors.route-reflector.config]
    route-reflector-client = true
    route-reflector-cluster-id = "192.168.55.1"

  [neighbors.graceful-restart.config]
    enabled = true
    long-lived-enabled = true
  [[neighbors.afi-safis]]
    [neighbors.afi-safis.config]
    afi-safi-name = "ipv4-unicast"
    [neighbors.afi-safis.mp-graceful-restart.config]
        enabled = true
[...]

PR #1353, opened 2017-06-08
test: Test for Graceful Restart with multiple peers

fujimoto-s:
This patch adds tests for graceful restart with multiple peers (to cover bugs like #1296).

### DIFF
--- a/test/scenario_test/graceful_restart_test.py
+++ b/test/scenario_test/graceful_restart_test.py
@@ -26,6 +26,7 @@ from lib.noseplugin import OptionParser, parser_option
 
 from lib import base
 from lib.base import (
+    BGP_FSM_IDLE,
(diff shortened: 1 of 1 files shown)


fujita:
Why not adding this to https://github.com/osrg/gobgp/blob/master/test/scenario_test/graceful_restart_test.py

fujimoto-s:
Sorry, there is no reason to divide this test class.
I will update it.

fujita:
Pushed, thanks!
A easy way to reproduce a race condition of https://github.com/osrg/gobgp/issues/1296 ?

fujimoto-s:
Without this patch, #1296 is reproduced unless the restarting speaker connects to its helpers at exactly the same time, but it rarely happens.
So, I think we can reproduce #1296 in one or two trials.

PR #1355, opened 2017-06-12
api: add graceful restart support to AddNeighbor API

fujita:


pslzym:
Hi, I use the graceful restart code,but it doesn't work. https://github.com/osrg/gobgp/blob/master/docs/sources/graceful-restart.md
[[neighbors.afi-safis]]
    [neighbors.afi-safis.config]
    [neighbors.afi-safis.mp-graceful-restart.config]
        enabled = true
graceful restart must conf  afi-safi-name and mp-graceful-restart.config, but grpc addnei Can not be configured。

Issue #1268, opened 2017-03-14
Ability to lock TTL for iBGP


vincentbernat:
Hey!

It's possible to set TTL for eBGP peers. Would it be possible to set it also for iBGP peers? Additionnaly, being able to specify a minimal value would be great (`IP_MINTTL`). This is a security measure to set it to 255 to ensure only directly connected hosts are able to speak with the peer.

iwaseyusuke:
Hi @vincentbernat 

For `IP_MINTTL`, you mean TTL security, right?
I've just create a PR related to this feature.
Could you review my PR?

For 'ebgp-multihop' feature in iBGP, I guess implementation is not so difficult, but this feature should not be necessary in iBGP peering.
This topic seems to be discussed below;
https://supportforums.cisco.com/discussion/10714216/multihop-ibgp

vincentbernat:
I'll test your PR. For `ebgp-multihop`, this is just a matter of being able to set the TTL to a value high enough to pass the TTL security check. On Linux, the default TTL is 64. Maybe setting a minimum TTL also bump the outgoing TTL. I'll check today.

fujita:
@vincentbernat Can you try the latest master?

vincentbernat:
I already tested the PR, so all good :)

PR #1357, opened 2017-06-14
server: Implement TTL security

iwaseyusuke:
This patch enable to configure Generalized TTL Security Mechanism (GTSM).

TTL security enables to specify the minimum TTL(IPv4)/HOPLIMIT(IPv6) which can be accepted.
For example, "ttl-min = 255" means that only directly connected peer is acceptable.

Note: TTL Security feature is mututally exclusive with "neighbors.ebgp-multihop.config".

fixes https://github.com/osrg/gobgp/issues/1268

### DIFF
--- a/config/bgp_configs.go
+++ b/config/bgp_configs.go
@@ -2073,6 +2073,8 @@ type TransportConfig struct {
 	// original -> gobgp:remote-port
 	//gobgp:remote-port's original type is inet:port-number
 	RemotePort uint16 `mapstructure:"remote-port" json:"remote-port,omitempty"`
+	// original -> gobgp:ttl
+	Ttl uint8 `mapstructure:"ttl" json:"ttl,omitempty"`
 }
 
 func (lhs *TransportConfig) Equal(rhs *TransportConfig) bool {
@@ -2092,6 +2094,9 @@ func (lhs *TransportConfig) Equal(rhs *TransportConfig) bool {
 		return false
 	}
 	if lhs.RemotePort != rhs.RemotePort {
+		return false
+	}
+	if lhs.Ttl != rhs.Ttl {
 		return false
 	}
 	return true
--- a/config/bgp_configs.go
+++ b/config/bgp_configs.go
@@ -1565,6 +1565,8 @@ type PeerGroup struct {
 	UseMultiplePaths UseMultiplePaths `mapstructure:"use-multiple-paths" json:"use-multiple-paths,omitempty"`
 	// original -> gobgp:route-server
 	RouteServer RouteServer `mapstructure:"route-server" json:"route-server,omitempty"`
+	// original -> gobgp:ttl-security
+	TtlSecurity TtlSecurity `mapstructure:"ttl-security" json:"ttl-security,omitempty"`
 }
 
 func (lhs *PeerGroup) Equal(rhs *PeerGroup) bool {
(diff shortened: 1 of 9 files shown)


vincentbernat:
I have just tried with the following configuration:

```
global:
  config:
    as: 65000
    router-id: 1.0.0.3
    local-address-list:
      - 10.234.78.66
neighbors:
  - config:
      neighbor-address: 10.234.78.65
      peer-as: 65000
    ttl-security:
      config:
        enabled: true
        ttl-min: 255
```

And a symmetric configuration on the peer.

The TTL used for outgoing connections is 64. Despite that, the connection is established. Looking at your code, it seems you restrict the ability to use ttl-security to eBGP.

If I try again with eBGP, outgoing TTL is correctly set to 255. If I decrease the TTL with iptables, the connection cannot be established, as expected.

iwaseyusuke:
Hi @vincentbernat 

Thank you for testing my PR!

Yes, that is expected behavior. As "ebgp-multihop", TTL security seems to be supported in only eBGP peering on other BGP implementations.
http://www.cisco.com/c/en/us/td/docs/ios/12_2s/feature/guide/fs_btsh.html#wp1043334

vincentbernat:
It is supported by Quagga (`ttl-security hops`) and BIRD (`ttl security yes`) on both iBGP and eBGP. Juniper doesn't have such a settings (but this can be done with firewall filters). RFC 3682 says that such a setting is generally not used for intra-domain protocols (hence the Cisco choice to not support it for iBGP), but doesn't give more details on why not use it. This seems a rather arbitrary limitation to only support eBGP.

iwaseyusuke:
Oh, I didn't test it on Quagga and BIRD, thanks.

For Juniper, you mean the following?
https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/ttl-edit-protocols-bgp-mulithop.html
Hmmm... it seems that Juniper supports TTL security on iBGP peering...

vincentbernat:
For Juniper, it only configures outgoing TTL. Complete implementation requires the use of some firewall rules (I see an example is provided in the documentation you linked).

iwaseyusuke:
Thanks, I got it!

I guess it is better to enable TTL security on eBGP peering and also, like Juniper, enable to configure static TTL value at [config.TransportConfig](https://github.com/osrg/gobgp/blob/master/config/bgp_configs.go#L1959).

I will update my patch!

iwaseyusuke:
I've pushed my updates to enable TTL security on iBGP peering and configure static TTL for BGP packets.

fujita:
This breaks non Linux build (OSX, *BSD, etc) because of SetTcpMinTTLSockopts(). Please add non-functional SetTcpMinTTLSockopts() to non Linux platforms.

iwaseyusuke:
Thank you for pointing out.
I added `SetTcpMinTTLSockopts()` for non Linux platforms. They seems to use the same constants value of `IP_MINTTL` and `IPV6_MINHOPCOUNT`, and I reused the implementation on Linux.

fujita:
Pushed, thanks guys!

vincentbernat:
In fact, this works fine for outgoing connection, but this doesn't work for incoming ones. On Linux, the syn/ack will be sent with a TTL of 64 instead of 255. Moreover, the connection can be established with a TTL < 255. The option should be set on the listening socket as well. I don't know if it should be another option or if TTL security should be enforced to be the same for everybody and set on the listening when adding a neighbor.

iwaseyusuke:
Hi @vincentbernat 

To make your question more clear, may I confirm some points?
You mean the incoming TCP connection can be established even if TTL security is enabled, right?
I guess if TCP connection is established, but the BGP session should not be established.

I've tested with the following configuration and could not establish the BGP session.

Router1:
```toml
[global.config]
  as = 65000
  router-id = "10.0.0.1"

[[neighbors]]
  [neighbors.config]
    neighbor-address = "10.0.0.2"
    peer-as = 65000
  [neighbors.transport.config]
     passive-mode = true  # to make incoming connection only.
  [neighbors.ttl-security.config]
    enabled = true
    ttl-min = 255
  [[neighbors.afi-safis]]
    [neighbors.afi-safis.config]
  [[neighbors.afi-safis]]
    [neighbors.afi-safis.config]
      afi-safi-name = "ipv6-unicast"
```

Router2:
```toml
[global.config]
  as = 65000
  router-id = "10.0.0.2"

[[neighbors]]
  [neighbors.config]
    neighbor-address = "10.0.0.1"
    peer-as = 65000
#  [neighbors.ttl-security.config]  # TTL security disabled -> send BGP packets with TTL=64.
#    enabled = true
#    ttl-min = 255
  [[neighbors.afi-safis]]
    [neighbors.afi-safis.config]
  [[neighbors.afi-safis]]
    [neighbors.afi-safis.config]
      afi-safi-name = "ipv6-unicast"
```

As you said, currently, GoBGP does not set TTL in SYN/ACK packet, because TTL security configuration   is not enabled when the TCP connecting.
The current implementation does:
1. accept the TCP connection.
2. search the neighbor's configuration based the sender IP address.
3. set TTL value if the neighbor's configuration has TTL security.

To enforce the MINTTL for incoming ones, it is required to set TTL value into a socket (local IP address + port) and it takes effect for all neighbors which GoBGP listening on that local address. In this case, if no local address is configured, GoBGP will bind sockets with addresses "0.0.0.0" and "::", and TTL settings will take effect for all neighbors, I guess...

vincentbernat:
My scenario is with interacting with other implementations that will not accept the syn/ack with TTL 64. Suppose I configure GoBGP with dynamic neighbors. BIRD won't be able to connect to it because syn/ack with TTL 64 will be rejected by the kernel. I have also tried with Juniper vMX and ran into the same problem.

I didn't try two GoBGP talking with each other and each one having the setting enabled. Does they accept the syn/ack with TTL 64?

iwaseyusuke:
I also tried TTL security feature with Cisco router, and ran into the same problem you mentioned.

Currently, GoBGP accepts the SYN/ACK packets with TTL=64 because of the reason I mentioned in the previous comment.
But we need to fix this problem to ensure the interoperability with other BGP implementations...
I will investigate how to solve this problem.

iwaseyusuke:
@vincentbernat 
Sorry for the delay.
I've just create a new PR for this issue; https://github.com/osrg/gobgp/pull/1377
Could you review this?

PR #1289, opened 2017-04-01
[WIP] support dynamic neighbor

ishidawataru:
this implements https://github.com/osrg/gobgp/issues/1283

PR #1358, opened 2017-06-16
Support Dynamic Neighbor

fujimoto-s:
This patch adds the Dynamic Neighbor feature.
The configuration syntax is like following:
```
[[dynamic-neighbors]]
  [dynamic-neighbors.config]
    prefix = "10.0.0.0/24"
    peer-group = "my-peer-group"
```

(This patch updates #1289)

### DIFF
--- a/api/grpc_server.go
+++ b/api/grpc_server.go
@@ -862,7 +862,6 @@ func (s *Server) DeleteVrf(ctx context.Context, arg *DeleteVrfRequest) (*DeleteV
 func NewNeighborFromAPIStruct(a *Peer) (*config.Neighbor, error) {
 	pconf := &config.Neighbor{}
 	if a.Conf != nil {
-		pconf.Config.NeighborAddress = a.Conf.NeighborAddress
 		pconf.Config.PeerAs = a.Conf.PeerAs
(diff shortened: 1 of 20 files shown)


ishidawataru:
As we discussed here https://github.com/osrg/gobgp/pull/1289, 
can we add dynamic neighbor configuration in `PeerGroup` instead of
putting new struct `DynamicNeighbor` under `Bgp`?

fujimoto-s:
I think it is possible to add the configuratin in `PeerGroup`.
However, OpenConfig defines `Dynamic Neighbor` using a new struct.
https://github.com/openconfig/public/blob/master/release/models/bgp/openconfig-bgp-global.yang#L129

In #1327, it is said that we will use newer openconfig someday, so I think the configuration structure for Dynamic Neighbor should be similar to OpenConfig's one.

ishidawataru:
I see. then let's keep it as it is.

Can you add some comments in `tools/pyang_plugins/gobgp.yang` that says we are taking the model from the latest openconfig?

fujimoto-s:
Sure, I will write the comment.

fujita:
Pushed, thanks a lot!

Issue #1356, opened 2017-06-13
graceful restart api doesn't work


pslzym:
Hi, I'm happy to see the graceful restart code update.https://github.com/osrg/gobgp/pull/1355.
I test the code , but it  doesn't work. https://github.com/osrg/gobgp/blob/master/docs/sources/graceful-restart.md
[[neighbors.afi-safis]]
[neighbors.afi-safis.config]
afi-safi-name = "ipv4-unicast"
[neighbors.afi-safis.mp-graceful-restart.config]
enabled = true
graceful restart need to be configured mp-graceful-restart.config and afi-safi-name,but grpc  addnei can not be configured it.

iwaseyusuke:
Hi @pslzym 
Just an idea, how about adding a list of afi-safi values named "families" into "GracefulRestart" message structure?
```
 message GracefulRestart {
   bool enabled = 1;
   uint32 restart_time = 2;
   bool helper_only = 3;
   uint32 deferral_time = 4;
   bool notification_enabled = 5;
   bool longlived_enabled = 6;
+  repeated uint32 families = 7;
 }
```

iwaseyusuke:
Hi,
I re-open a new PR for this issue; https://github.com/osrg/gobgp/pull/1367
And it seems to be merged to upstream.
Could test the latest?

fujita:
@pslzym please reopen this if the current master doesn't work.

PR #1359, opened 2017-06-16
api: Support mp-graceful-restart config in AddNeighbor

iwaseyusuke:
This patch extends GracefulRestart message structure in AddNeighbor API to configure "mp-graceful-restart.config".

Also, this patch includes an improvement for reducing for loop against "config.Neighbor.AfiSafis".

fixes https://github.com/osrg/gobgp/issues/1356


fujita:
How about LongLivedGracefulRestart parameters per AfiSafi?

iwaseyusuke:
Thank you for pointing out.
For `LongLivedGracefulRestart`, in addition to a parameter for `enabled` or not, we need to a parameter for `restart-time`.
I will update to add a new message structure `LongLivedGracefulRestart` into `api/gobgp.proto`.

fujita:
Better to be consistent with openconfig model? Adding the list of a structure like FamilyInfo (or something better) to Peer structure. FamilyInfo includes gracefull restart parameters.

iwaseyusuke:
Yes, I think it is better and I tried to do so at first.
But I found Peer structure already has

iwaseyusuke:
Yes, I think it is better and I tried to do so at first.
But I found Peer structure already has [`repeated uint32 families = 1;`](https://github.com/osrg/gobgp/blob/83684d5e62fb14596aabf03747a8b395543f91bb/api/gobgp.proto#L547), I stopped to do it.
If consistency with OpenConfig is more reasonable, I will update my PR.

iwaseyusuke:
Sorry, I missed to delete this branch.
I will re-open another PR.

PR #1362, opened 2017-06-20
Fix Fails in Initializing Dynamic Peer

### LINKED ISSUE #1283: Dynamic neighbors
Hey!

When GoBGP acts as a route reflector, it would be convenient if it accepted any neighbors (or any neighbors in some subnet). This way, neighbors don't have to be pre-declared.

fujimoto-s:
If State.NeighborAddress is not set, setting default configs
for the dynamic peers will fail. This patch sets State.NeighborAddress
before setting default configs.
And this avoids overwriting State.NeighborAddress if it is set.

This fixes #1283.

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -130,8 +130,11 @@ func setDefaultNeighborConfigValuesWithViper(v *viper.Viper, n *Neighbor, asn ui
 		}
 	}
 
+	if n.State.NeighborAddress == "" {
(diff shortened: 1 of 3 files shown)


vincentbernat:
This works fine for me. Both IPv4 and IPv6. Thanks!

fujita:
Thanks, guys!

Issue #1361, opened 2017-06-18
GoBGP and issues with match field for fragmentation flags


pavel-odintsov:
Hello, guys!

Some time ago I slightly changed my flow spec announces for BGP fragmentation and TCP flags.  I decided to set match bit for them because Huawei requires it.

As you know, I craft flow spec NLRI messages from my own code. 

Then I found that GoBGP displays TCP flags correct (it print it as "(match)") but forgot to show "match" for fragmentation flags:
```
/opt/fastnetmon/libraries/gobgp_1_4_0_git/gobgp global rib -a ipv4-flow
    Network                                                                                                                                                                                             Next Hop             AS_PATH              Age        Attrs
*>  [destination:127.0.0.0/24][source:4.0.0.0/24][protocol: tcp][destination-port: =80][source-port: =53 =5353][tcp-flags: (match)syn][packet-length: =777 =1122][fragment: =is-fragment =dont-fragment]fictitious                                00:00:04   [{Origin: ?} {Extcomms: [rate: 1024.000000]}]
```

According to BGP Flow Spec RFC 5575 they suggest using "bitmask operand format" for fragmentation flags and TCP flags. 

But looks like GoBGP supports them only for TCP Flags and forgot about fragmentation flags.

Thank you!

pavel-odintsov:
Btw, I checked twice and could confirm that we definitely set match bit for outgoing announce:
```
flags: end of list: 1 and_bit: 0 bit_shift_len: 0 reserved1: 0 reserved2: 0 not_bit: 0 match_bit: 1
```

fujimoto-s:
First, you seem to use an older version of GoBGP.
The latest version solves the difference between "tcp-flags" and "fragment".

In the latest version, "(match)" is not used for "tcp-flags", "=" has been used instead.
And the output about "tcp-flags" and "fragment" is like follows:

```
$gobgp global rib -a ipv4-flowspec
    Network                                               Next Hop             AS_PATH              Age        Attrs
*>  [tcp-flags:=S ][fragment: =is-fragment =dont-fragment]fictitious                                00:00:04   [{Origin: i} {LocalPref: 100} {Extcomms: [remark: 24]}]

```

So, the problem which you mentioned had been solved in the latest version.
Could you consider to use the latest version of GoBGP?

pavel-odintsov:
Hello!

Thank you for fast response! I will check it soon!

fujita:
Please reopen if the current master doesn't work.

PR #1364, opened 2017-06-20
flowspec: Comply with RFC 5575 for TCP Flags and Fragment

fujimoto-s:
For "fragment" field,  RFC 5575 suggests using "bitmask operand format", but GoBGP does not have an interface to configure it.
This PR introduces the way to configure bitmask operands for"fragment" field.
(This fixes the problem which was pointed out in #1361.)

And for "tcp-flags" field, if the field is specified from CLI command, GoBGP sets 'AND' bit to the invalid operand.
This PR also fixes this problem to comply with RFC 5575.


fujita:
Pushed, thanks!

PR #1369, opened 2017-06-30
Update gRPC document and sample codes

fujimoto-s:
This PR updates gRPC document and sample codes to work with the latest GoBGP, gRPC and ProtoBuf and so on.


fujita:
Great, thanks!

PR #1370, opened 2017-06-30
config: Remove Equal() method from config structures

iwaseyusuke:
To simplify, this patch removes `Equal()` from the structures in `bgp_config.go` and uses `reflect.DeepEqual()` instead.

### DIFF
--- a/config/bgp_configs.go
+++ b/config/bgp_configs.go
@@ -981,19 +981,6 @@ type DynamicNeighborConfig struct {
 	PeerGroup string `mapstructure:"peer-group" json:"peer-group,omitempty"`
 }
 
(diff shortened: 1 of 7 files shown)


fujita:
We used DeepEqual() in the past. IIRC, one of reasons that we replaced DeepEqual is that it's too slow to reload a config file in the production (about 1,000,000 lines). Can you measure the performance with such large config file?

iwaseyusuke:
Sure, I will create dummy neighbor configs and measure the performance.

fujita:
Thanks! The config file includes hundreds of peers and lots of policies.

iwaseyusuke:
I compared the performance with the existing `Equal()` and `reflect.DeepEqual()`, the latter one was about 5 times "slower".
Better to keep `Equal()` in each structure, so I withdraw this PR.

Issue #574, opened 2015-12-21
Detailed information about RPKI status for prefix


pavel-odintsov:
Hello, folks!

I'm hardworking on RPKI deployment and have yet another suggestion about it.

I haven't found any ways to check details about RPKI invalid prefixes. Sign "I*>" is not enough for me.

I need something like RIPE RPKI validator REST API output (please look at description field):
curl http://127.0.0.1:8090/api/v1/validity/AS198068/159.253.16.0/24

``` bash
{
  "validated_route":{
    "route":{
      "origin_asn":"xx",
      "prefix":"xx/24"
    },
    "validity":{
      "state":"Invalid",
      "reason":"as",
      "description":"At least one VRP Covers the Route Prefix, but no VRP ASN matches the route origin ASN",
      "VRPs":{
        "matched":[],
        "unmatched_as":[{
          "asn":"xxx",
          "prefix":"xxx/20",
          "max_length":20
        },{
          "asn":"xx",
          "prefix":"xxx/20",
          "max_length":20
        },{
          "asn":"xx",
          "prefix":"xxx/20",
          "max_length":20
        },{
          "asn":"xxx",
          "prefix":"xxxx/20",
          "max_length":20
        }],
        "unmatched_length":[]
      }
    }
  }
}
```

Will be fine to get new command to get details about RPKI failed state.

fujita:
Thanks for the proposal. I've been working on a stream API to provide the similar information:
https://github.com/osrg/gobgp/blob/master/api/gobgp.proto#L474

The API enables you to continuously monitor the suspicious state changes (e.g. you get notified when a peer advertises an invalid route). Once I finish the feature, I'll work on this.

pavel-odintsov:
That's great! Thanks!

fujita:
Please reopen this or create a new one if the master doesn't work.

PR #1376, opened 2017-07-11
rpki: Add command to get detailed information

fujimoto-s:
This PR adds the command to get detailed information about RPKI validation.
(This implements #574.)

Example:
```bash
$ gobgp rpki detail 2.1.0.0/16 65001
Target Prefix: 2.1.0.0/16, AS: 65001
  This route is invalid  reason: as
  No VRP ASN matches the route origin ASN.

  Matched VRPs:
    No Entry
  Unmatched AS VRPs:
    Network            AS     MaxLen
    2.0.0.0/12         3215   16
    2.1.0.0/16         3215   16
  Unmatched Length VRPs:
    No Entry
```


fujita:
Can we integrate this new API into the existing ValidateRib API?

fujimoto-s:
So, the syntax should be like `gobgp rpki validate 2.1.0.0/16 3215`?
I think it is Yes, I'll fix it soon.

fujita:
On the second thought, might be better to integrate into rib command. Junos does such.

https://www.juniper.net/documentation/en_US/junos12.2/topics/topic-map/bgp-origin-as-validation.html

fujimoto-s:
Sorry, I could not find that Junos outputs a detailed information about RPKI in the page.

But, I think it is good to integrate it into RIB command. 
So, `gobgp global rib validation 2.1.0.0/16 3215` is good?
If so, I will implement as such.

fujimoto-s:
I will fix this command like below:
`gobgp neighbor 125.0.0.1 local validation 2.1.0.0/16`

* use `neighbor` command to show detailed information
* not specify origin_as
  * respond with all routes which match to the specified prefix

fujita:
gobgp neighbor 125.0.0.1 local validation 2.1.0.0/16

this can be used with only route server use-case.

gobgp neighbor 125.0.0.1 adj-in validation 2.1.0.0/16

?

fujimoto-s:
Sorry, I did not intended to limit it to `local` only.

More precisely, the command syntax will be like below:
`gobgp neighbor <neighbor-address> {adj-in|local|adj-out} validaton <prefix>`
Does it match to your image?

fujita:
Let's start with 'adj-in'. If someone requests the rests (I doubt), let's implement them.

fujimoto-s:
I see, I will do that.

fujita:
I suspect that the CI failed due to logrus package name breakage. Can you rebase against the current master and run the CI again?

fujimoto-s:
Thanks, I rebased & pushed!

docs/sources/cli-command-syntax.md
```diff
@@ -129,6 +129,8 @@ The following options can be specified in the global subcommand:
 % gobgp neighbor <neighbor address> [local|adj-in|adj-out] [<prefix>|<host>] [longer-prefixes|shorter-prefixes] [-a <address family>]
 # show table summary
 % gobgp neighbor <neighbor address> [local|adj-in|adj-out] summary [-a <address family>]
+# show RPKI detailed information in adj-in table
+% gobgp neighbor >neighbor address> adj-in <prefix> validation
```
fujita: >neighbor
typo?

fujimoto-s: Sorry, this is typo.
I will fix it.

fujita:
Thanks!

PR #1377, opened 2017-07-18
server: Set TTL before dialing for interoperability of TTL security

iwaseyusuke:
For the interoperability of TTL security, we need to set TTL before dialing and accepting connection to/from neighbors.
This patch fixes to set the default TTL(=255) for incoming connection, and set the given TTL for outgoing connection.

### DIFF
--- a/server/sockopt_bsd.go
+++ b/server/sockopt_bsd.go
@@ -36,13 +36,8 @@ func setsockoptTcpMD5Sig(fd int, address string, key string) error {
 }
 
 func SetTcpMD5SigSockopt(l *net.TCPListener, address string, key string) error {
-	fi, err := l.File()
+	fi, _, err := extractFileAndFamilyFromTCPListener(l)
 	defer fi.Close()
-	if err != nil {
-		return err
-	}
-	fl, err := net.FileListener(fi)
-	defer fl.Close()
 	if err != nil {
 		return err
 	}
@@ -57,6 +52,15 @@ func setsockoptIpTtl(fd int, family int, value int) error {
 		name = syscall.IPV6_UNICAST_HOPS
 	}
 	return os.NewSyscallError("setsockopt", syscall.SetsockoptInt(fd, level, name, value))
+}
+
+func SetListenTcpTTLSockopt(l *net.TCPListener, ttl int) error {
+	fi, family, err := extractFileAndFamilyFromTCPListener(l)
+	defer fi.Close()
+	if err != nil {
+		return err
+	}
+	return setsockoptIpTtl(int(fi.Fd()), family, ttl)
 }
 
 func SetTcpTTLSockopt(conn *net.TCPConn, ttl int) error {
--- a/server/sockopt_bsd.go
+++ b/server/sockopt_bsd.go
@@ -20,8 +20,9 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"strings"
 	"syscall"
+
+	log "github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


server/sockopt_bsd.go
```diff
@@ -20,68 +20,119 @@ import (
...
 const (
 	TCP_MD5SIG       = 0x10 // TCP MD5 Signature (RFC2385)
 	IPV6_MINHOPCOUNT = 73   // Generalized TTL Security Mechanism (RFC5082)
 )
 
-func SetTcpMD5SigSockopts(l *net.TCPListener, address string, key string) error {
+func setsockoptTcpMD5Sig(fd int, address string, key string) error {
+	// always enable and assumes that the configuration is done by setkey()
+	return os.NewSyscallError("setsockopt", syscall.SetsockoptInt(fd, syscall.IPPROTO_TCP, TCP_MD5SIG, 1))
+}
```
fujita: Hmm, you make  SetTcpMD5SigSockopt() not to work on FreeBSD?

iwaseyusuke: No, `os.NewSyscallError` SyscallError `error` only if some errors occur.
```go
// NewSyscallError returns, as an error, a new SyscallError
// with the given system call name and error details.
// As a convenience, if err is nil, NewSyscallError returns nil.
func NewSyscallError(syscall string, err error) error {
	if err == nil {
		return nil
	}
	return &SyscallError{syscall, err}
}
```

vincentbernat:
Works fine for me, with both IPv4 and IPv6.

iwaseyusuke:
@vincentbernat 
Thank you for testing my patch!

fujita:
Thanks guys!

fujita:
Now I got the following warning messages on OSX even with no ttl configuraiton. Can we try to set  ttl only if ttl is configured? 

`{"level":"info","msg":"gobgpd started","time":"2017-07-21T08:57:08+09:00"}
{"Topic":"Config","level":"info","msg":"Finished reading the config file","time":"2017-07-21T08:57:08+09:00"}
{"Key":{"IP":"0.0.0.0","Port":179,"Zone":""},"Topic":"Peer","level":"warning","msg":"cannot set TTL(=255) for TCPLisnter: setting ttl is not supported","time":"2017-07-21T08:57:08+09:00"}
{"Key":{"IP":"::","Port":179,"Zone":""},"Topic":"Peer","level":"warning","msg":"cannot set TTL(=255) for TCPLisnter: setting ttl is not supported","time":"2017-07-21T08:57:08+09:00"}
{"level":"info","msg":"Peer 10.0.255.1 is added","time":"2017-07-21T08:57:08+09:00"}
{"Topic":"Peer","level":"info","msg":"Add a peer configuration for:10.0.255.1","time":"2017-07-21T08:57:08+09:00"}
{"Key":"10.0.255.1","Topic":"Peer","level":"warning","msg":"setting ttl for active connection is not supported","time":"2017-07-21T08:57:19+09:00"}`

iwaseyusuke:
For the message "cannot set TTL(=255) for TCPLisnter: setting ttl is not supported", the current implementation will always try to set TTL=255 for TTL Security [at here](https://github.com/iwaseyusuke/gobgp/blob/c3d98dac4df7d55fc6aceb4558707f252bbbd765/server/server.go#L69-L76), but setting TTL feature seems not to be supported on OSX environment.

For the message "setting ttl for active connection is not supported", TTL should be set only if needed at ["server/fsm.go"](https://github.com/iwaseyusuke/gobgp/blob/178e65f8f113da82f8dfcc56b5fa7682da1bd932/server/fsm.go#L348-L357).
Are you trying to establish eBGP session? If so, GoBGP will set TTL=1 on this connection, and will fail to set TTL because the implementation for OSX does not support it.

We need to "+build darwin" on "server/sockopt_*.go"?

fujita:
We could add socketopt support to OSX. But why gobgp needs to play with ttl when I don't configure ttl at all?

iwaseyusuke:
The reason we need to set TTL=255 on TCPLisnter is TTL Security.
If we don't set TTL=255 on TCPLisnter, we can not accept incoming connection with TTL Security.
For example:
```
R1 (with TTL Security)                  R2 (no TTL setting on listen socket)
| ------------ SYN (TTL=255) ---------->|
| <--------- SYN+ACK (TTL=64) ----------|
X Reject packets
  (MINTTL > 64)
```
So BIRD and frr setup listen sockets with TTL=255.
BIRD:
https://github.com/BIRD/bird/blob/7a855725f2ffde508da0c7ee01dc1bcd6e0a5d93/proto/bgp/bgp.c#L903
frr:
https://github.com/FRRouting/frr/blob/c14777c6bfd0a446c85243d3a9835054a259c276/bgpd/bgp_network.c#L736

PR #1379, opened 2017-07-20
scenario_test/addpath_test: Test cases with non add-path router

iwaseyusuke:
This patch adds the test cases for the mixed situation of add-path enabled routers and non add-path enabled routers and advertising/withdrawing routes in such situation.


fujita:
Thanks

PR #1381, opened 2017-07-21
docs: Update config example

iwaseyusuke:
Currently, "docs/sources/configuration.md" has some obsoleted config parameters and some syntax errors.
This PR fixes these problems and introduces UT for validating syntax of the configuration example.

### DIFF
--- a/config/bgp_configs_test.go
+++ b/config/bgp_configs_test.go
@@ -16,8 +16,15 @@
 package config
 
 import (
-	"github.com/stretchr/testify/assert"
+	"bufio"
+	"os"
+	"path"
+	"runtime"
+	"strings"
 	"testing"
+
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestEqual(t *testing.T) {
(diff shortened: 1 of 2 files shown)


iwaseyusuke:
That commit enables the config example to be loaded as the "tested config file" by UT which added in  `config/bgp_configs_test: Add test for config validation`.
If not separated, difficult to load it as a test target file.
Also, if example file is separated from configuration.md, easy to add other format examples like .yaml or .json.

fujita:
Making tests easier is fine by me. However, I prefer not to put the other formats here to avoid the code maintenance burden. That is, toml is the only format officially supported.

iwaseyusuke:
> That is, toml is the only format officially supported.

Sorry, I didn't know that...
I guess if there is a tool for converting config format (e.g., toml -> yaml, toml -> json), it should not be the burden. We need to maintain only toml, and just convert it for other formats.

For example: `marshal_gobgpd_conf.go`
```go
package main

import (
	"encoding/json"
	"fmt"
	"os"
	"path/filepath"

	"github.com/BurntSushi/toml"
	"github.com/spf13/viper"
	"gopkg.in/yaml.v2"

	"github.com/osrg/gobgp/config"
)

func detectConfigFileType(path, def string) string {
	switch ext := filepath.Ext(path); ext {
	case ".toml":
		return "toml"
	case ".yaml", ".yml":
		return "yaml"
	case ".json":
		return "json"
	default:
		return def
	}
}

func main() {
	if len(os.Args) != 3 {
		os.Exit(1)
	}
	fileIn := os.Args[1]
	formatIn := "toml"
	formatIn = detectConfigFileType(fileIn, formatIn)
	fileOut := os.Args[2]
	formatOut := "yaml"
	formatOut = detectConfigFileType(fileOut, formatOut)

	c := &config.BgpConfigSet{}
	v := viper.New()
	v.SetConfigFile(fileIn)
	v.SetConfigType(formatIn)
	if err := v.ReadInConfig(); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
	if err := v.UnmarshalExact(c); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	f, err := os.Create(fileOut)
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
	defer f.Close()

	switch formatOut {
	case "toml":
		encoder := toml.NewEncoder(f)
		if err := encoder.Encode(c); err != nil {
			fmt.Println(err)
			os.Exit(1)
		}
	case "yaml":
		buf, err := yaml.Marshal(c)
		if err != nil {
			fmt.Println(err)
			os.Exit(1)
		}
		f.Write(buf)
	case "json":
		encoder := json.NewEncoder(f)
		encoder.SetIndent("", "    ")
		if err := encoder.Encode(c); err != nil {
			fmt.Println(err)
			os.Exit(1)
		}
	}
}
```

```
$ go run marshal_gobgpd_conf.go docs/sources/configuration.example.toml example.json
$ head example.json 
{
    "global": {
        "config": {
            "as": 1,
            "router-id": "1.1.1.1",
            "port": 1790
        },
        "state": {},
        "route-selection-options": {
            "config": {},
```

fujita:
I know that the library supports multiple formats. But it doesn't mean that you get multiple format support for free. Who debugs the bugs of multiple format support in the library? Always development resource is limited so the priority is toml now (might be changed in the future).

iwaseyusuke:
I understand this matter.

Well, to make tests easier, I want an example file as `.toml` file.
Should I update some on PR and need some messages like `"Currently, we officially supports only toml fomrat."` on `docs/sources/configuration.md`?

fujita:
I'm not sure. Currently, only toml is there so it's simple and obvious for me.

Could you do something like with configuration.md?

https://github.com/osrg/gobgp/blob/master/test/scenario_test/ci-scripts/build_embeded_go.py

iwaseyusuke:
Thanks, it is that I was looking for.
I will update my patch.

fujita:
Pushed, thanks! Why not running this test on travis-ci?

iwaseyusuke:
Thanks!

> Why not running this test on travis-ci?

I think this test should be run at [here](https://github.com/osrg/gobgp/blob/f48c4deaa19ca8b36d87a0d71776bc29b7e17ebe/.travis.yml#L22). Is that my misunderstanding?

fujita:
Oops, exactly.

iwaseyusuke:
Thank you for your confirmation.

Issue #1343, opened 2017-05-27
Please add support for RFC8050


tking:
Dear GoBGP Developers,

please add support for RFC8050 (https://tools.ietf.org/html/rfc8050): Multi-Threaded Routing Toolkit (MRT) Routing Information Export Format with BGP Additional Path Extensions.

Thanks in advance,
Thomas

iwaseyusuke:
Hi @tking 

The support for RFC8050 has been merged with by PR https://github.com/osrg/gobgp/pull/1347
But, please note this PR includes only decoder/encoder of MRT format, and does not include the implementation for advertising routes with Add-Path extensions.

For Add-Path extensions advertisement, it seems to be supported sometime soon as far as I've heard

Thanks,
Iwase

fujita:
@iwaseyusuke now the master includes addpath support. Can you update mrt support?

iwaseyusuke:
@fujita 
Sure, I will create another PR to support it.

fujita:
@iwaseyusuke can we close this?

iwaseyusuke:
@fujita Yes, now GoBGP supports exporting MRT files with Add-Paths extensions.

fujita:
@tking please reopen this if the new feature doesn't work as you expect.

PR #1390, opened 2017-07-27
mrt: Enable to dump/inject MRT routes with path_id

iwaseyusuke:
fixes https://github.com/osrg/gobgp/issues/1343

### DIFF
--- a/server/mrt.go
+++ b/server/mrt.go
@@ -86,18 +86,18 @@ func (m *mrtWriter) loop() error {
 	for {
 		serialize := func(ev WatchEvent) []*mrt.MRTMessage {
 			msg := make([]*mrt.MRTMessage, 0, 1)
-			switch m := ev.(type) {
+			switch e := ev.(type) {
 			case *WatchEventUpdate:
 				subtype := mrt.MESSAGE_AS4
-				mp := mrt.NewBGP4MPMessage(m.PeerAS, m.LocalAS, 0, m.PeerAddress.String(), m.LocalAddress.String(), m.FourBytesAs, nil)
-				mp.BGPMessagePayload = m.Payload
-				if m.FourBytesAs == false {
+				mp := mrt.NewBGP4MPMessage(e.PeerAS, e.LocalAS, 0, e.PeerAddress.String(), e.LocalAddress.String(), e.FourBytesAs, nil)
+				mp.BGPMessagePayload = e.Payload
+				if e.FourBytesAs == false {
 					subtype = mrt.MESSAGE
 				}
-				if bm, err := mrt.NewMRTMessage(uint32(m.Timestamp.Unix()), mrt.BGP4MP, subtype, mp); err != nil {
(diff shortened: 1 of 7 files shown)


fujita:
Do we need a new option to enable/disable this feature? Simply dump paths with id if a peer sends paths with id?

iwaseyusuke:
Thank you for reviewing.

Yes, I tried to implement it at first time, but in that case, the dumped MRT file should be treated as "malformed" by MRT readers which does not support RFC8050.
So I added this feature in order to control  and switch if path_id should be or not.

If this feature is redundant or impairs usability, I will fix to simply dump path_id if peer sent route with it.

server/mrt.go
```diff
@@ -88,11 +88,17 @@ func (m *mrtWriter) loop() error {
 			msg := make([]*mrt.MRTMessage, 0, 1)
 			switch e := ev.(type) {
 			case *WatchEventUpdate:
-				subtype := mrt.MESSAGE_AS4
 				mp := mrt.NewBGP4MPMessage(e.PeerAS, e.LocalAS, 0, e.PeerAddress.String(), e.LocalAddress.String(), e.FourBytesAs, nil)
 				mp.BGPMessagePayload = e.Payload
-				if e.FourBytesAs == false {
-					subtype = mrt.MESSAGE
+				isAddPaths := e.PathList[0].GetNlri().PathIdentifier() != 0
```
fujita: I think that zero is valid identifier.

fujita:
Tools that don't support RFC8050 should skip the entry with unknown type such as MESSAGE_AS4_ADDPATH?

server/mrt.go
iwaseyusuke: Thanks, I misunderstood. If zero is valid, there is no reserved path_id (can check path_id is advertised or not), I guess.
Instead of path_id, I will fix to check if the neighbor has add-paths feature enabled for the given family or not.

iwaseyusuke:
Yes, reader tools should be implemented to do so, but it is depending on its implementation, I guess.
If we suppose that reader tools will skip unknown entries, better to remove `add-paths-enabled` option.

fujita:
Let's remove the option. If tools can't handle it, it's a bug. Also if you disable add-path negotiation, then you can get a mrt without identifier.

iwaseyusuke:
Sure, I will remove the option.

server/mrt.go
```diff
@@ -46,6 +46,15 @@ func (m *mrtWriter) Stop() {
 	close(m.dead)
 }
 
+func (m *mrtWriter) isAddPathReceiveEnabled(address string, family bgp.RouteFamily) bool {
+	if p, ok := m.s.neighborMap[address]; !ok {
+		return false
```
fujita: the loop goroutine can't access to m.s.neighborMap; a race with the main goroutine?

iwaseyusuke:
I've updated PR to fix comments and rebase on current master.

fujita:
Pushed, thanks!

PR #1392, opened 2017-07-28
addpath-test: Test cases for send add-paths

iwaseyusuke:
This PR adds some test cases to test add-paths feature for sending routes and injecting routes via CLI.


fujita:
Thanks, pushed!

PR #1395, opened 2017-07-31
gobgpd: Avoid name collision

iwaseyusuke:
alias "p" for "github.com/kr/pretty" collides the local variable "p" for Peer structure.
This patch fixes to avoid this name collision and also sorts the import order with goimports.
  e.g.) goimports -w -local github.com/osrg/gobgp gobgpd/main.go


fujita:
Looks good, thanks.

PR #1405, opened 2017-08-04
table: fix handling of update msg which has both mpreach and mpunreach

ishidawataru:
don't put a mpunreach attribute in a path whose NLRI is taken from a mpreach
attribute.
also, don't put a mpreach attribute in a withdrawal path whose NLRI is
taken from a mpunreach attribute.

Signed-off-by: Wataru Ishida <ishida.wataru@lab.ntt.co.jp>


fujita:
A message carrying only the withdrawn routes doesn't need any attributes?

fujita:
Thanks!

PR #1409, opened 2017-08-09
Patch 1

### LINKED ISSUE #1408: PeerGroup.dynamicNeighors map values over written
If multiple dynamic neighbors are defined, all are the last read config.DynamicNeighor.

matthubb:
The temporary pointer from the range function is being passed to AddDynamicNeighbor which is the same for every iteration of the loop. So the PeerGroup.DynamicNeighbors all end up with the same config.DynamicNeighbor.

Fixes #1408


matthubb:
Quick fix, perhaps it would be better to pass values to AddDynamicNeighbor rather than pointers that stage?

fujita:
Agreed but all the similar APIs use a pointer so changing this api is confusing and I want to avoid the changes of exported APIs.

PR #1412, opened 2017-08-21
config: Introduce RouteFamily into AfiSafiState

iwaseyusuke:
To reduce the translations of "AfiSafiName" into "bgp.RouteFamily", this patch introduces "RouteFamily" field into "AfiSafiState" and stores the translated value.

Currently, this translations of "AfiSafiName" is not critical for the performance, but "AfiSafiName" will be translated at many place and many time, so this patch reduces the complication of handling translations (error checks or so).

### DIFF
--- a/api/grpc_server.go
+++ b/api/grpc_server.go
@@ -97,10 +97,10 @@ func NewMpGracefulRestartFromConfigStruct(c *config.MpGracefulRestart) *MpGracef
 	}
 }
 
-func NewAfiSafiConfigFromConfigStruct(c *config.AfiSafiConfig) *AfiSafiConfig {
+func NewAfiSafiConfigFromConfigStruct(c *config.AfiSafi) *AfiSafiConfig {
 	return &AfiSafiConfig{
-		Family:  uint32(bgp.AddressFamilyValueMap[string(c.AfiSafiName)]),
-		Enabled: c.Enabled,
+		Family:  uint32(c.State.Family),
+		Enabled: c.Config.Enabled,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


fujita:
The translations are in the performance-critical path? btw, RouteFamily is one of mistakes that we made earlier, IMHO. Just "Family" is more appropriate.

iwaseyusuke:
Thank you for reviewing.

> The translations are in the performance-critical path?

Most of translations are on constructing capabilities in OPEN message, so I guess it is not so critical if disconnection and re-connection occurs frequently.
But some of translations are on gRPC handling, it depending on how often APIs are called...

> btw, RouteFamily is one of mistakes that we made earlier, IMHO. Just "Family" is more appropriate.

This field is GoBGP implementation specific (not defined in OpenConfig/Public), so I avoided "widely used" name (e.g., "AddressFamily", "AfiSafi").
If it should be more appropriate name, I will update my match.

iwaseyusuke:
I've renamed "RouteFamily" to "Family".

fujita:
Why State instead of Config?

iwaseyusuke:
According to [OpenConfig Style Guide](http://www.openconfig.net/docs/style-guide/#grouping-naming),  "Config" should be configurable by user, but "State" should be read-only and stores "operational state".
In this case, "Family" is the translated value from "AfiSafiName" and never differ from the value which  "AfiSafiName" pointing to.
Then, I guess "Family" should be stored as "operational state" (read-only against user).

fujita:
I see, sounds reasonable. Pushed, thanks.

PR #1413, opened 2017-08-21
cli: Enable to specify AS_PATH

iwaseyusuke:
This patch enables to specify AS_PATH in "gobgp" client command.
The syntax is similar to the output of AS_PATH in "global rib" show command.


fujita:
Pushed, thanks!

Issue #1143, opened 2016-10-11
Duplicate Large Communities handling


pierky:
Hello,

I'm running some tests on these Large Communities draft's statements:

> Duplicate Large Communities SHOULD NOT be transmitted. A receiving speaker SHOULD silently remove duplicate Large Communities from a BGP UPDATE message.

As of my findings it seems that GoBGP is not removing duplicate communities from the received BGP UPDATEs and also it sends duplicate values:

https://github.com/pierky/large-bgp-communities-playground/blob/master/tests/03-DUPLICATE_COMMUNITIES.md#gobgp-announcing

pierky:
Hello,

I tested this new commit but I found that GoBGP still sends duplicate values if it is instructed via CLI.

This command `gobgp global rib add -a ipv4 203.0.113.24/32 large-community 65537:1:1,65537:1:1` produces the following:

```
    192.0.2.3.36189 > 192.0.2.103.179: Flags [P.], cksum 0x84dc (incorrect -> 0xcb9d), seq 132:207, ack 132, win 457, options [nop,nop,TS val 697398 ecr 691150], length 75: BGP, length: 75
        Update Message (2), length: 75
          Origin (1), length: 1, Flags [T]: Incomplete
          AS Path (2), length: 6, Flags [T]: 65537
          Next Hop (3), length: 4, Flags [T]: 192.0.2.3
          Unknown Attribute (32), length: 24, Flags [OT]:
            no Attribute 32 decoder
            0x0000:  0001 0001 0000 0001 0000 0001 0001 0001
            0x0010:  0000 0001 0000 0001
          Updated routes:
            203.0.113.24/32
```

(please note two identical values for attribute 32, `0001 0001 0000 0001 0000 0001`)

Now everything looks fine when it receives the duplicate values:

```
root@gobgpreceiver:/go# gobgp neighbor 192.0.2.3 adj-in
    Network             Next Hop             AS_PATH              Age        Attrs
    203.0.113.24/32     192.0.2.3            65537                00:10:18   [{Origin: ?} {LargeCommunity: [ 65537:1:1]}]

fujita:
@pierky This is the expected behavior. GoBPG sends what you want like exabgp. So you can use gobgp for sorta testings.

PR #1415, opened 2017-08-22
packet/bgp: Support Revised Error Handling (RFC7606)

### LINKED ISSUE #1147: Malformed large BGP communities attributes handling
I would like to ask your support to analyze a behaviour that seems unexpected to me. It seems GoBGP does not handle malformed large BGP communities attributes in the way I understand the draft wants to.

Draft:

> A BGP UPDATE message with a malformed Large Communities attribute SHALL be handled using the approach of "treat-as-withdraw" as described in section 2 [RFC7606].

RFC7606:

> Treat-as-withdraw: In this approach, the UPDATE message containing the path attribute in question MUST be treated as though all contained routes had been withdrawn just as if they had been listed in the WITHDRAWN ROUTES field (or in the MP_UNREACH_NLRI attribute if appropriate) of the UPDATE message, thus causing them to be removed from the Adj-RIB-In according to the procedures of [RFC4271].

I used an ExaBGP instance to send a malformed attribute; this is the output of tcpdump on the GoBGP host:

```
    192.0.2.2.40208 > 192.0.2.3.179: Flags [P.], cksum 0x8474 (incorrect -> 0x2f47), seq 350:422, ack 65, win 457, options [nop,nop,TS val 5124424 ecr 5123924], length 72: BGP, length: 72
        Update Message (2), length: 72
          Origin (1), length: 1, Flags [T]: IGP
          AS Path (2), length: 6, Flags [T]: 65536
          Next Hop (3), length: 4, Flags [T]: 192.0.2.2
          Unknown Attribute (30), length: 21, Flags [OTP]:
            no Attribute 30 decoder
            0x0000:  0000 0001 0000 0002 0000 0003 0000 0004
            0x0010:  0000 0005 ff
          Updated routes:
            203.0.113.12/32
    192.0.2.3.179 > 192.0.2.2.40208: Flags [P.], cksum 0x8441 (incorrect -> 0xf9e0), seq 65:86, ack 422, win 470, options [nop,nop,TS val 5124425 ecr 5124424], length 21: BGP, length: 21
        Notification Message (3), length: 21, UPDATE Message Error (3), subcode Attribute Length Error (5)
    192.0.2.3.179 > 192.0.2.2.40208: Flags [F.], cksum 0x842c (incorrect -> 0x0200), seq 86, ack 422, win 470, options [nop,nop,TS val 5124425 ecr 5124424], l
```

This is GoBGP's log:

```
WARN[3024] malformed BGP message  Key=192.0.2.2 State=BGP_FSM_ESTABLISHED Topic=Peer error=large communities length isn't correct
WARN[3024] sent notification      Data=&{Header:{Marker:[] Len:21 Type:3} Body:0xc4203f5040} Key=192.0.2.2 State=BGP_FSM_ESTABLISHED Topic=Peer
INFO[3024] Peer Down              Key=192.0.2.2 Reason=notification-sent code 3(update) subcode 5(attribute length error) State=BGP_FSM_ESTABLISHED Topic=Peer
```

It seems that GoBGP, instead of treating the message as a withdraw, drops the BGP session with the sending neighbor.

Thanks.

GoBGP config:

```
[global.config]
        as = 65537
        router-id = "192.0.2.3"

[[neighbors]]
        [neighbors.config]
                neighbor-address = "192.0.2.2"
                peer-as = 65536
```

fujimoto-s:
This patch enables GoBGP to keep the session established even if the received Update message contains some errors, and to handle these errors according to what defined in RFC7606.

This feature is enabled when 'treat-as-withdraw' in 'neighbors.error-handling.config' is specified to true in the GoBGP config file.

This PR fixes #1147.


packet/bgp/bgp.go
```diff
@@ -7870,20 +7871,46 @@ func (msg *BGPUpdate) DecodeFromBytes(data []byte, options ...*MarshallingOption
...
 		}
+
 		err = p.DecodeFromBytes(data, options...)
 		if err != nil {
-			return err
+			e = err.(*MessageError)
+			if e.(*MessageError).SubTypeCode == BGP_ERROR_SUB_ATTRIBUTE_FLAGS_ERROR {
+				e.(*MessageError).ErrorHandling = ERROR_HANDLING_TREAT_AS_WITHDRAW
+			} else {
+				e.(*MessageError).ErrorHandling = PathAttributeToErrorHandlingMap[p.GetType()]
```
fujita: Out of index with PathAttributeUnknown?

fujimoto-s: Oh, it will happen.
I'll fix it, thanks!

packet/bgp/bgp.go
```diff
@@ -7982,6 +7982,33 @@ func (msg *BGPUpdate) IsEndOfRib() (bool, RouteFamily) {
...
+	var unreach []AddrPrefixInterface
+
+	for _, p := range msg.PathAttributes {
+		switch nlri := p.(type) {
+		case *PathAttributeMpReachNLRI:
+			unreach = append(unreach, nlri.Value...)
+		case *PathAttributeMpUnreachNLRI:
+			unreach = append(unreach, nlri.Value...)
+		default:
+			withdraw.PathAttributes = append(withdraw.PathAttributes, p)
```
fujita: Why necessary? Withdraw message doesn't need any attributes.

fujita:
Can you add some tests for this?

packet/bgp/bgp.go
fujimoto-s: It is not necessary, sorry.
I will remove it.

fujimoto-s:
That would be fine. I will consider adding both unit tests and scenario tests.

fujimoto-s:
I've updated my PR with additional unit/scenario tests.

fujita:
Thanks!

fujita:
@fujimoto-s We can close this too?
https://github.com/osrg/gobgp/issues/1143#issuecomment-257372397

fujimoto-s:
No, it isn't.

RFC7606 says that the duplicate attribute should be discarded.
Thus, "handling the duplicate Large Communities attribute in an UPDATE message" is in the scope of RFC7606.
On the other hand, "handling the duplicate Large Communities __in one Large Communities attribute__" is out of scope.
I think we should make new PR to fix #1143.

fujita:
I see, can you make a new PR?

fujimoto-s:
Sure, I'll work on it.

PR #1416, opened 2017-08-23
BMP: Fix behavior for specifying pre-policy

a16:
- In the document, it says `route-monitoring policy` is `pre-policy` by default. However it doesn't work unless specifying `route-monitoring policy = "pre-policy"` in the config explicitly.
- When type `gobgp bmp add x.x.x.x:xxxx`, it works fine for pre-policy. But when type it with `pre`, it returns error.

This PR will fix those 2 issues.


fujita:

PR #1419, opened 2017-08-24
table/policy: fix ReplacePolicy with preserve=true to remove old statements

wydrych:
When policy is replaced, the **old** statements have to be checked for being still in use anywhere else, not the **new** ones

### DIFF
--- a/table/policy.go
+++ b/table/policy.go
@@ -3482,9 +3482,10 @@ func (r *RoutingPolicy) ReplacePolicy(x *Policy, refer, preserve bool) (err erro
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


wydrych:
It seems that the Travis failure is not related to the code changed, but is a general environment setup error. Is it possible to re-run PR check?

fujita:
Very sorry about the delay. You mean that "with preserve=false", right? With preserve=true, the fixed  code isn't executed.

wydrych:
Indeed, a typo.

fujita:
Pushed, thanks a lot!

PR #1420, opened 2017-08-28
server: export addROA()/delROA() for library usage

a16:
I'd like to use addROA()/delROA() in my code as library. But currently they are not exportable.
I am happy if you could merge this.


fujita:
Thanks!

PR #1429, opened 2017-09-06
table/policy: Enable to replace whole of the statement

fujimoto-s:
This PR enables ReplaceStatement() to replace everything in the statement by adding an argument "all" to ReplaceStatement().


Thoro:
Confirmed working as expected.

Thoro:
I just noticed a bug, ReplaceStatement does NOT work correctly for all Set items.

you need to validate the conditions before replaceing them

```
                        for _, c := range st.Conditions {
				if err = r.validateCondition(c); err != nil {
					return
				}
			}
```

fujimoto-s:
Please let me confirm.
"ReplaceStatement does NOT work correctly" means that, we can replace statements with NON-existing sets due to lack of validation, right?

If so, I've overlooked it (and it may be an existing bug).
Thank you for reporting! I'll fix it.

Thoro:
You can add existing and non-existing sets, both of them won't match anything, because they don't reference the real set, but are instead an empty set with that name.

fujimoto-s:
Thanks for the detail! I was able to clarify the problem.
And I updated this PR. Could you check this?

table/policy.go
```diff
@@ -3337,14 +3337,26 @@ func (r *RoutingPolicy) DeleteStatement(st *Statement, all bool) (err error) {
...
 }
 
-func (r *RoutingPolicy) ReplaceStatement(st *Statement) (err error) {
+func (r *RoutingPolicy) ReplaceStatement(st *Statement, all bool) (err error) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
+	for _, c := range st.Conditions {
+		if err = r.validateCondition(c); err != nil {
+			return
```
Thoro: I think we should return the error here

Thoro:
looks good and I think it could be merged

fujimoto-s:
Thanks! I fixed it and pushed.

fujita:
This is a bug fix? Please describe why this is necessary, not what this does.

fujimoto-s:
I modified only commit messages.

"Validate conditions in ReplaceStatement" is a bug fix. Currently, ReplaceStatement() doesn't work well, because it replaces condition sets with "empty" condition sets and it won't match anything.

"Enable to replace whole of the statement" is an enhancement. This patch adds an arguments to specify whether replace whole of the statement or not.

fujimoto-s:
I updated my patch to solve a conflict.

fujita:
I removed ReplaceStatement() API. Should create new statements instead of mutating the existing statements. If the existing policy objects really needs to be mutated, SetPolicies() API can be used.

PR #1434, opened 2017-09-12
server: Add GetAllPolicyAssignments API

### LINKED ISSUE #1427: GetPolicyAssignment should have a GetPolicyAssignments equivalent
Currently it's only possible to retrieve a single PolicyAssignment, but for management it would make sense to also provide a GetPolicyAssignments method to retrieve all set PolicyAssignments

fujimoto-s:
This fixes #1427.


fujita:
As discussed, this will not be merged.

PR #1436, opened 2017-09-13
server: Filter policies by name in GetPolicy API

### LINKED ISSUE #1431: GetPolicy returns ALL policies, should provide a filter like other Get methods
The GetPolicy method currently returns all policies within the system, other Get functions filter the results based on the passed name.

"" -> all
name -> just the policy

The GetPolicy API should behave the same as others.

fujimoto-s:
This resolves #1431.

__NOTE__: 
This will break backward compatibility because arguments for `server.GetPolicy()` and `client.GetPolicy()` are changed.
This will have an impact on people using GoBGP as a native library.


fujita:
fixed Policy and Statement too for the consistency.
https://github.com/osrg/gobgp/commit/f0b724878360618dd0a950364709b048486771b6

PR #1252, opened 2017-02-23
server/zclient: Support basic Nexthop Tracking features

iwaseyusuke:
These patches enable server/zclient to support the basic Nexthop Tracking features:
- registering the interested nexthop using the NEXTHOP_REGISTER messages
- updating the nexthop reachability state based on the NEXTHOP_UPDATE messages from Zebra

Note: The following features are still not implemented yet.
- unregistering the nexthop if no longer required
- calculating "penalty" like Cisco's router when the nexthop state is changed

Test: These patch was tested with the simple case which described on [Quagga's document about Nexthop Tracking](https://github.com/opensourcerouting/quagga/blob/master/doc/next-hop-tracking.txt#L300).

PR #1443, opened 2017-09-15
server: Avoid applying policy multiply when updating path

iwaseyusuke:
Currently, UpdatePath() of BgpServer will apply policy multiply, and this causes the unexpected result of policy configurations.

This patch avoids this multiple policy application and fixes an issue which mentioned on this commnet: https://github.com/osrg/gobgp/pull/1252#issuecomment-281984302

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -659,7 +659,7 @@ func (server *BgpServer) RSimportPaths(peer *Peer, pathList []*table.Path) []*ta
 func (server *BgpServer) propagateUpdate(peer *Peer, pathList []*table.Path) {
 	var dsts []*table.Destination
 
-	var gBestList, gOldList, bestList, oldList []*table.Path
+	var gBestList, gOldList []*table.Path
 	var mpathList [][]*table.Path
 
 	rib := server.globalRib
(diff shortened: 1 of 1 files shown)


fujita:
The current nexthop tracking code modifies paths in zclinet goroutine (that's not the main goroutine). I think that we need to think about how to fix it first.

iwaseyusuke:

Would you share me which code modifying paths without main goroutine?
zclient calls `server.UpdatePath` which wrapped with `mgmtOperation` to update nexthop sates;
https://github.com/osrg/gobgp/blob/ce258f05d2fd9c3f66b831cae1e40bc9958cef1f/server/zclient.go#L160-L165
https://github.com/osrg/gobgp/blob/ce258f05d2fd9c3f66b831cae1e40bc9958cef1f/server/server.go#L1320-L1330

iwaseyusuke:
Related to the races of goroutine, I guess the root cause of these problems is that the client side implementation (against `server.BgpServer`), `zclient` and `bmp` for example, can directly access to the resources of `server.BgpServer`.
If these client side implementations are moved from `server` package to `client` package, the resources of `server.BgpServer` should be invisible, and more safely implemented, I guess.

fujita:
Oops, I remember it wrong. Pushed, thanks.

PR #1446, opened 2017-09-20
zclient: Register all nexthops and unregister uninterested nexthops

iwaseyusuke:
This PR enables zclient to register all nexthops contained in RIBs and also unregister nexthop if there is no path bound for that nexthop.


iwaseyusuke:
Updated for rebasing on the current master.

fujita:
Thanks!

PR #1447, opened 2017-09-21
api: GetPath API alternative to GetRib

iwaseyusuke:
With the extremely large RIBs, GetRib gRPC API hits the limits of each message size, because this API is using simple request/response RPC.

This patch introduces a new API GetPath which using streaming RPC and can retrieve RIBs even if extremely large.

Note: GetRib can be replaced with GetPath and obsoleted in the future.


fujita:
Can the new API return stream Path instead of Destination? That's consistent with the name. Also I guess that we could remove Destination in the future?

iwaseyusuke:

As far as gRPC API, we can use `stream Path` instead of `stream Destination`, but `Client.getRIB()` in `client/client.go` returns `*table.Table`.
So in order to keep the compatibility, we need to convert the received list of `Path` into the grouped up lists of `*table.Path` (based on NLRI) and then into the lists of `*table Destination`, I guess.
However, if removing `Destination` is prefer, I will try it!

iwaseyusuke:
I've updated `GetPath` API to return `stream Path` instead of `stream Destination.

fujita:
Can GetPathRequest avoid using Destination? We could remove Destination without breaking the API in the future.

iwaseyusuke:
Well, `message Table` structure has some required query parameters...
I will sieve and sort out `GetPathRequest` parameters.

iwaseyusuke:
I've updated `GetPathRequest` to include query parameters instead of `Table` and `Destination`.

fujita:
Thanks!

PR #1448, opened 2017-09-21
Update broken link

MoBattah:


fujita:
Thanks for the finding! Looks like it's not a draft anymore. So how about using a link to the rfc like?
[[RFC7938]](https://tools.ietf.org/html/rfc7938)

MoBattah:
Good?

fujita:
Squashed and pushed, thanks!

PR #1450, opened 2017-09-30
server: Remove the wrong and verbose return value in validatePath()

a16:
The second return value (bucket) in validatePath() is wrong and verbose. bucket has the last result of the function(fn) only. However the first return value (validation) includes its correct value.  So I think it's enough and we can delete it.


fujita:
Thanks!

PR #1451, opened 2017-10-02
table: Implement ignore-as-path-length

iwaseyusuke:


iwaseyusuke:
Note: Currently, `advertise-inactive-routes`, `enable-aigp` and `ignore-next-hop-igp-metric` are also not evaluated on the best path calculation.
For `enable-aigp` and `ignore-next-hop-igp-metric`, I'm trying to implement it on other branches.
On the other hand, `advertise-inactive-routes` seems not to need to be implemented because of the architecture of GoBGP + Zebra integration.

fujita:
Pushed, thanks.

PR #1452, opened 2017-10-03
config/default: Set treat-as-withdraw to true as default

fujimoto-s:
RFC7606 updates some other RFC and says that BGP speakers
should not send NOTIFICATION for some errors in received UPDATE.
Thus, treat-as-withdraw should be true as a default value.


fujita:
Pushed, thanks.

PR #1455, opened 2017-10-07
Fix spelling of "occurred" and "length"

vincentbernat:


fujita:
Thanks!
@iwaseyusuke can we run a spell checking tool on travis-ci?

iwaseyusuke:
Purpose of spell checking is rejecting all typos? or just generating reports (not treat as errors on travis-ci)?
I guess we can run some checks on travis-ci (`aspell`, `scspell`, or IDE code inspection via CLI), but currently we don't have allowed words dictionary, we need to set `allow_failures` for a while.

iwaseyusuke:
Just an idea, target only typos in log messages:

```bash
# Extract lines outputting log messages
$ find ./ -type f -name *.go | xargs grep Infof\(\" > tmp.log

# Run spell checking
$ scspell --use-builtin-base-dict --override-dictionary dictionary.txt --report-only tmp.log
Continuing with empty natural dictionary

tmp.log:9: 'skiped' not found in dictionary (from token 'skiped')
```

With this approach, we need to maintain user-dictionary "dictionary.txt" though.

PR #1458, opened 2017-10-10
*: Support BGP Confederations (RFC 5065) 

fujimoto-s:
This PR adds the BGP confederations feature.

The configuration syntax is as follows:
```
[global.confederation.config]
  enabled = true
  identifier = 30
  member-as-list = [65001, 65002]
```


fujita:
Pushed, thanks.

PR #1459, opened 2017-10-10
*: Enhancement for "Remove Duplicate Large Communities"

fujimoto-s:
Currently, GoBGP does not remove duplicate large communities values
if it is instructed via CLI.
And the removal logic is written in packet/bgp/validate.go,
but it is not validation (RFC 8092 says that the packet which has
duplicate values shall not be considered as malformed).

This commit moves the removal logic to table/path.go,
and removes duplicate values even if it is instructed via CLI.

### DIFF
--- a/packet/bgp/validate.go
+++ b/packet/bgp/validate.go
@@ -191,22 +191,6 @@ func ValidateAttribute(a PathAttributeInterface, rfs map[RouteFamily]BGPAddPathM
 				}
 			}
 		}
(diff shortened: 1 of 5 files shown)


fujita:
Very sorry, I misunderstood.

https://github.com/osrg/gobgp/issues/1143#issuecomment-335681025

I think that the current behavior should not change.

fujimoto-s:
I see.
I close this PR.

PR #1460, opened 2017-10-13
 tools: Add script for spell checking

iwaseyusuke:
This patch introduces the spell checker for GoBGP's source codes.
See the discussion of https://github.com/osrg/gobgp/pull/1455#issuecomment-334984244


fujita:
Thanks, let's see if it works well.

PR #1461, opened 2017-10-13
Allow filtering MRT routes by IP version and overriding next-hop

XANi:
As title says, it adds few handy options to import.

I'm not really sure what use `mrtOpts` was supposed to have, as only a `Best` bool was used out of it, so I repurposed it to type usedpass arguments to injectMrt

### DIFF
--- a/gobgp/cmd/common.go
+++ b/gobgp/cmd/common.go
@@ -111,9 +111,10 @@ var actionOpts struct {
 type mrtOpts struct {
 	OutputDir  string
 	FileFormat string
-	Best       bool `long:"only-best" description:"only keep best path routes"`
-	SkipV4     bool `long:"no-ipv4" description:"Skip importing IPv4 routes"`
-	SkipV6     bool `long:"no-ipv4" description:"Skip importing IPv6 routes"`
+	Best       bool   `long:"only-best" description:"only keep best path routes"`
+	SkipV4     bool   `long:"no-ipv4" description:"Skip importing IPv4 routes"`
+	SkipV6     bool   `long:"no-ipv4" description:"Skip importing IPv6 routes"`
+	NextHop    net.IP `long:"nexthop" description:"Rewrite nexthop"`
 }
 
 func formatTimedelta(d int64) string {
--- a/gobgp/cmd/common.go
+++ b/gobgp/cmd/common.go
@@ -109,15 +109,16 @@ var actionOpts struct {
 }
 
 var mrtOpts struct {
-	OutputDir  string
-	FileFormat string
-	Filename   string `long:"filename" description:"MRT file name"`
+	OutputDir   string
+	FileFormat  string
+	Filename    string `long:"filename" description:"MRT file name"`
 	RecordCount int    `long:"count" description:"Number of records to inject"`
 	RecordSkip  int    `long:"skip" description:"Number of records to skip before injecting"`
-	Best       bool   `long:"only-best" description:"only keep best path routes"`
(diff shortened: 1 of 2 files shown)


gobgp/cmd/common.go
```diff
@@ -108,10 +108,12 @@ var actionOpts struct {
 	NexthopAction       string `long:"next-hop" description:"specifying a next-hop action of policy"`
 }
 
-var mrtOpts struct {
+type mrtOpts struct {
```
fujita: I prefer to keep 'var' for consistency. You can use 'mrtOpts' variant globally. The way to use neighborsOpts in gobgp/cmd/neighbor.go would be helpful.

gobgp/cmd/mrt.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func injectMrt(filename string, count int, skip int, onlyBest bool) error {
+func injectMrt(filename string, count int, skip int, opts *mrtOpts) error {
```
fujita: You don't need to pass opts *mrtOpts if you keep 'var' above.

gobgp/cmd/mrt.go
```diff
@@ -107,7 +109,7 @@ func injectMrt(filename string, count int, skip int, onlyBest bool) error {
 					paths = append(paths, table.NewPath(source, nlri, false, e.PathAttributes, t, false))
 				}
 
-				if onlyBest {
+				if opts.Best {
```
fujita: mrtOpts.Best can be used here.

gobgp/cmd/mrt.go
```diff
@@ -152,6 +154,7 @@ func injectMrt(filename string, count int, skip int, onlyBest bool) error {
 }
 
 func NewMrtCmd() *cobra.Command {
+	var opts mrtOpts
```
fujita: You don't need it.

gobgp/cmd/mrt.go
```diff
@@ -174,7 +177,7 @@ func NewMrtCmd() *cobra.Command {
 					}
 				}
 			}
-			err := injectMrt(filename, count, skip, mrtOpts.Best)
+			err := injectMrt(filename, count, skip, &opts)
```
fujita: I guess that the original code confused you. You don't need to pass anything here. injectMrt() can access to mrtOpts variant.

XANi: I thought it would be more readable if all parameters (well, all related to cli input) that change function is acting on would be explictly passed to it.

I changed it in the first place because originally my code was just
  ```go  
    func injectMrt(filename string, count int, skip int, skipV4 bool, skipV6, onlyBest bool)
```
but then I thought if there will be some other filter added in the future then that would be pretty ugly way to pass extra option so I moved it into mrtOpts

gobgp/cmd/common.go
XANi: Sure. Want me to add rest of the options there (filename, count, skip) ? Having half of options passed and half of it taken from some package variable would seem kinda hard to follow IMO.

fujita: Surely, let's add them there and handle in the same way. Thnaks!

XANi:
Done, I've also added option to change queue size and dropped it to some manageable value.

at current value of  it can eat 15 GB of RAM if importing [full view](http://data.ris.ripe.net/rrc21/) without `--only-best`

fujita:
Sorry about the delay, pushed, thanks a lot!
Having changing 'var' to 'type' and then revering in the history doesn't make sense so I squashed the first three commits.

PR #1465, opened 2017-10-17
Adding Com Hem to companies who uses gobgp in production

mrevilme:


fujita:
Thanks a lot, pushed (in the order of deployment)!
Always great to hear about production deployment. This is the use case?
https://www.peering-forum.eu/system/documents/157/original/Emil_Palm_BGPFlowspec.pdf

PR #1466, opened 2017-10-18
server: Avoid local paths left on iBGP peers

iwaseyusuke:
Currently, with iBGP peering, the CLI injected path will not be withdrawn
when the deletion if any other path exists on RIB, then the CLI injected
path will be left on iBGP peers even if the route is deleted on the
originator router.
This problem can cause routing loops in some case.

This patch fixes to withdraw the old best path when it was the CLI (or
gRPC) injected path.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -337,7 +337,7 @@ func sendFsmOutgoingMsg(peer *Peer, paths []*table.Path, notification *bgp.BGPMe
 
 func isASLoop(peer *Peer, path *table.Path) bool {
 	for _, as := range path.GetAsList() {
-		if as == peer.fsm.pConf.State.PeerAs {
+		if as == peer.AS() {
 			return true
 		}
 	}
@@ -378,7 +378,7 @@ func filterpath(peer *Peer, path, old *table.Path) *table.Path {
 			ignore = true
 			info := path.GetSource()
 			//if the path comes from eBGP peer
-			if info.AS != peer.fsm.pConf.State.PeerAs {
+			if info.AS != peer.AS() {
 				ignore = false
 			}
 			// RFC4456 8. Avoiding Routing Information Loops
@@ -416,13 +416,19 @@ func filterpath(peer *Peer, path, old *table.Path) *table.Path {
 		}
 
 		if ignore {
-			if !path.IsWithdraw && old != nil && old.GetSource().Address.String() != peer.ID() && old.GetSource().AS != peer.fsm.pConf.State.PeerAs {
-				// we advertise a route from ebgp,
-				// which is the old best. We got the
-				// new best from ibgp. We don't
-				// advertise the new best and need to
-				// withdraw the old.
-				return old.Clone(true)
+			if !path.IsWithdraw && old != nil {
+				oldSource := old.GetSource()
+				if old.IsLocal() || oldSource.Address.String() != peer.ID() && oldSource.AS != peer.AS() {
+					// In this case, we suppose this peer has the same prefix
+					// received from another iBGP peer.
+					// So we withdraw the old best which was injected locally
+					// (from CLI or gRPC for example) in order to avoid the
+					// old best left on peers.
+					// Also, we withdraw the eBGP route which is the old best.
+					// When we got the new best from iBGP, we don't advertise
+					// the new best and need to withdraw the old best.
+					return old.Clone(true)
+				}
 			}
 			log.WithFields(log.Fields{
 				"Topic": "Peer",
--- a/server/peer.go
+++ b/server/peer.go
@@ -120,6 +120,10 @@ func NewPeer(g *config.Global, conf *config.Neighbor, loc *table.TableManager, p
 	return peer
 }
 
+func (peer *Peer) AS() uint32 {
+	return peer.fsm.pConf.State.PeerAs
+}
+
 func (peer *Peer) ID() string {
 	return peer.fsm.pConf.State.NeighborAddress
 }


iwaseyusuke:
For the details of this problem, please refer to the following.

Topology:

```
+----+            +----+
| R1 |---(iBGP)---| R2 |
+----+            +----+
   |     +----+     |
 (iBGP)--| R3 |--(iBGP)
         +----+
```

When R1 and R2 advertised the same prefix "192.168.1.0/24", R3 has two paths from R1 and R2.

```
# On R1
r1> gobgp global rib -a ipv4 add 192.168.1.0/24
r1> gobgp global rib -a ipv4 
*> 192.168.1.0/24       0.0.0.0                                   hh:mm:ss   [{Metric: 0} {Origin: ?}]
*  192.168.1.0/24       10.0.0.2                                  hh:mm:ss   [{Metric: 0} {Origin: ?} {LocalPref: 100}]


# On R2
r2> gobgp global rib -a ipv4 add 192.168.1.0/24
r2> gobgp global rib -a ipv4
*> 192.168.1.0/24       0.0.0.0                                   hh:mm:ss   [{Metric: 0} {Origin: ?}]
*  192.168.1.0/24       10.0.0.1                                  hh:mm:ss   [{Metric: 0} {Origin: ?} {LocalPref: 100}]

# On R3
r3> gobgp global rib -a ipv4
*> 192.168.1.0/24       10.0.0.1                                  hh:mm:ss   [{Metric: 0} {Origin: ?} {LocalPref: 100}]
*  192.168.1.0/24       10.0.0.2                                  hh:mm:ss   [{Metric: 0} {Origin: ?} {LocalPref: 100}]
```

On this situation, even if the injected path is deleted on R1, R1 does not send the witdraw for that path, then the deleted path is left on R2 and R3.

```
# On R1
r1> gobgp global rib -a ipv4 del 192.168.1.0/24
r1> gobgp global rib -a ipv4 
*> 192.168.1.0/24       10.0.0.2                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]

# On R2
r2> gobgp global rib -a ipv4
*> 192.168.1.0/24       0.0.0.0                                   hh:mm:ss   [{Origin: ?}]
*  192.168.1.0/24       10.0.0.1                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]

# On R3
r3> gobgp global rib -a ipv4
*> 192.168.1.0/24       10.0.0.1                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]
*  192.168.1.0/24       10.0.0.2                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]
```

Also, even if R2 deletes the injected path, R1 has R2's path and R3 has both. This can cause routing loops.

```
# On R2
r2> gobgp global rib -a ipv4 del 192.168.1.0/24
r2> gobgp global rib -a ipv4
*> 192.168.1.0/24       10.0.0.1                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]

# On R1
r1> gobgp global rib -a ipv4 
*> 192.168.1.0/24       10.0.0.2                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]

# On R3
r3> gobgp global rib -a ipv4
*> 192.168.1.0/24       10.0.0.1                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]
*  192.168.1.0/24       10.0.0.2                                  hh:mm:ss   [{Origin: ?} {LocalPref: 100}]
```

This patch fixes this problem.

server/server.go
```diff
@@ -416,13 +416,17 @@ func filterpath(peer *Peer, path, old *table.Path) *table.Path {
...
-				// new best from ibgp. We don't
-				// advertise the new best and need to
-				// withdraw the old.
-				return old.Clone(true)
+			if !path.IsWithdraw && old != nil {
+				oldSource := old.GetSource()
+				if oldSource.Address == nil || oldSource.Address.String() != peer.ID() && oldSource.AS != peer.AS() {
+					// We withdraw the old best which was injected locally
+					// (from CLI or gRPC for example) in order to avoid the
+					// old best left on peers.
```
fujita: This happens only if this peer has the same prefix routes from an ibgp peer. Can you improve the comment?

server/server.go
```diff
@@ -416,13 +416,17 @@ func filterpath(peer *Peer, path, old *table.Path) *table.Path {
...
-			if !path.IsWithdraw && old != nil && old.GetSource().Address.String() != peer.ID() && old.GetSource().AS != peer.fsm.pConf.State.PeerAs {
-				// we advertise a route from ebgp,
-				// which is the old best. We got the
-				// new best from ibgp. We don't
-				// advertise the new best and need to
-				// withdraw the old.
-				return old.Clone(true)
+			if !path.IsWithdraw && old != nil {
+				oldSource := old.GetSource()
+				if oldSource.Address == nil || oldSource.Address.String() != peer.ID() && oldSource.AS != peer.AS() {
```
fujita: I think that old.IsLocal() is better than "oldSource.Address == nil"

iwaseyusuke: Yes, I think `old.IsLocal()` is better to read code, but `old.IsLocal()` calls `old.GetSource()` internally and `GetSource()` can NOT be lightweight getter function, because `GetSource()` need to draw "root" instance of this path. So I wanted to reduce the calls of `GetSource()`.
If drawing root does not take cost so high, better to use `old.IsLocal()` as you said.

server/server.go
iwaseyusuke: Sure, I will add some descriptions for it.

server/server.go
fujita: I prefer the code readability until someone complains with the performance results showing that old.IsLocal() actually degrades the performance.

iwaseyusuke: I see, I will use `IsLocal()` instead.

iwaseyusuke:
I've pushed to use `isLocal()` and improve comments.

fujita:
Thanks

PR #1468, opened 2017-10-20
table/path: net.IP.IsUnspecified() instead of isZero()

iwaseyusuke:
Fixes to use the standard library of golang.


fujita:
Pushed, thanks.

Issue #1469, opened 2017-10-20
MRT output doesn't work correctly when multiple configs are specified


a16:
I'm using multiple `mrt-dump` in the config like below. It has been working correctly as of 9c03026.

```
[[mrt-dump]]
  [mrt-dump.config]
    dump-type = "updates"
    file-name = "/var/dump/updates/2006/01/02.1504.dump"
    rotation-interval = 60

[[mrt-dump]]
  [mrt-dump.config]
    dump-type = "table"
    file-name = "/var/dump/table/2006/01/2006-01-02-15-04.mrt"
    rotation-interval = 3600
```

After that, the latter one(`table`) looks fine, but the former one(`updates`) is never rotated and its content is not `updates` but `table`.

fujita:
Thanks for the report. Can you try the latest master?

fujita:
Please reopen if you still hit the problem.

PR #1470, opened 2017-10-20
mrt: Use copy of MrtConfig to avoid data race

iwaseyusuke:
Because MrtConfig is passed as pointer value, configuration can be overwritten by caller, then if multiple configurations for MRT dump, only last one is enable and the others are not enabled.

This patch fixes to create copy of MrtConfig and avoids this problem.

Fixes https://github.com/osrg/gobgp/issues/1469

### DIFF
--- a/server/mrt.go
+++ b/server/mrt.go
@@ -34,12 +34,10 @@ const (
 )
 
 type mrtWriter struct {
-	dead             chan struct{}
-	s                *BgpServer
-	c                *config.MrtConfig
-	file             *os.File
-	rotationInterval uint64
-	dumpInterval     uint64
+	dead chan struct{}
+	s    *BgpServer
+	c    *config.MrtConfig
+	file *os.File
 }
 
 func (m *mrtWriter) Stop() {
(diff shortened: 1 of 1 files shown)


a16:
Confirmed #1469 would be fixed.

fujita:
Hmm, I think that it's a caller problem. I thought that we fixed the similar bug and looks like we still have the same bugs.
https://github.com/osrg/gobgp/pull/1477

I guess that these functions had better to take a copy not a pointer.

iwaseyusuke:

I agree taking a copy instead of a pointer is better.

But... functions of BgpServer (e.g, EnableMrt) is exported function, I don't think it better to require callers to take care of it.

How about change type of argument at non-exported function?
```diff
diff --git a/server/mrt.go b/server/mrt.go
index fe61f7a..e7ea21b 100644
--- a/server/mrt.go
+++ b/server/mrt.go
@@ -338,7 +338,7 @@ type mrtManager struct {
        writer    map[string]*mrtWriter
 }
 
-func (m *mrtManager) enable(c *config.MrtConfig) error {
+func (m *mrtManager) enable(c config.MrtConfig) error {
        if _, ok := m.writer[c.FileName]; ok {
                return fmt.Errorf("%s already exists", c.FileName)
        }
@@ -377,7 +377,7 @@ func (m *mrtManager) enable(c *config.MrtConfig) error {
                setRotationMin()
        }
 
-       w, err := newMrtWriter(m.bgpServer, c, rInterval, dInterval)
+       w, err := newMrtWriter(m.bgpServer, &c, rInterval, dInterval)
        if err == nil {
                m.writer[c.FileName] = w
        }
diff --git a/server/server.go b/server/server.go
index 03b9cbd..3843729 100644
--- a/server/server.go
+++ b/server/server.go
@@ -2272,7 +2272,7 @@ func (s *BgpServer) ReplacePolicyAssignment(name string, dir table.PolicyDirecti
 
 func (s *BgpServer) EnableMrt(c *config.MrtConfig) error {
        return s.mgmtOperation(func() error {
-               return s.mrtManager.enable(c)
+               return s.mrtManager.enable(*c)
        }, false)
 }
 ```

The same can be applied for RPKI and BMP, I guess.

fujita:
You say,  the majority of exported functions like AddNeighbor needs to be updated in the same way?

fujita:
I've pushed the fix (because we got a bug report). Let's discuss the better API on a new issue.

iwaseyusuke:
Hmmm... Yes.

Many of exported functions has pairs of non-exported version function (in case of AddNeighbor, addNeighbor is that one).
If we fix this problem in the mentioned way, we need to change non-exported version of them...

PR #1472, opened 2017-10-24
config: Properly set config of PeerGroup member

fujimoto-s:
Currently, the config of PeerGroup members are not set properly.

For example, if LocalAs of the neighbor is not set in the config file,
LocalAs will be set to Global.As as a default value.
After this, however, if the neighbor is a member of a peer group,
LocalAs will be unexpectedly overwritten by the peer group config.

This commit fixes this by setting default values after
applying the peer group config.

Adding to this, this commit avoids unnecessary applying
the peer group config.

### DIFF
--- a/config/util.go
+++ b/config/util.go
@@ -40,6 +40,32 @@ func detectConfigFileType(path, def string) string {
 	}
 }
 
+func (b *BgpConfigSet) getPeerGroup(n string) (*PeerGroup, error) {
+	if n == "" {
+		return nil, nil
+	}
+	for _, pg := range b.PeerGroups {
+		if n == pg.Config.PeerGroupName {
+			return &pg, nil
+		}
+	}
+	return nil, fmt.Errorf("no such peer-group: %s", n)
+}
+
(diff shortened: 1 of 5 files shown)


config/util.go
```diff
@@ -40,6 +40,32 @@ func detectConfigFileType(path, def string) string {
...
+	}
+	for _, pg := range b.PeerGroups {
+		if n == pg.Config.PeerGroupName {
+			return &pg, nil
+		}
+	}
+	return nil, fmt.Errorf("no such peer-group: %s", n)
+}
+
+func (d *DynamicNeighbor) Validate(b *BgpConfigSet) error {
```
fujita: Why exported? Its name is too generic, I think.

fujimoto-s: Sorry, this function is called only from the same package. This should be unexported.
I'll make a fix.

fujimoto-s:
I updated my patch and rebased with the latest master.

fujita:
Thanks!

PR #1478, opened 2017-10-31
server: fix AddPath to return uuid properly

fujita:

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1285,6 +1285,7 @@ func (s *BgpServer) AddPath(vrfId string, pathList []*table.Path) (uuidBytes []b
 		}
 		if len(pathList) == 1 {
 			pathList[0].AssignNewUUID()
+			uuidBytes = pathList[0].UUID().Bytes()
 		}
 		s.propagateUpdate(nil, pathList)
 		return nil


nao4arale:
Thank you fujita.
Sorry, I also think  you say "Is non-1st path's uuid ?"
You fix is right.

And Is AddPath() received multi pathes in []pathList?
If this, should fix more....

Example, this snippet,
https://gist.github.com/nao4arale/a94c986684780cbdab8caff5f9bf538f

go run yours fixies,
```bash
[root@net-gobgp pulreq]# go run main.go 
Added a route's UUID is,  00000000-0000-0000-0000-000000000000
Added a route's UUID is,  78bf0204-bff7-443d-9c65-a4574cc5e04f
Show a route's Prefix is 10.10.0.0/24, UUID is 00000000-0000-0000-0000-000000000000
Show a route's Prefix is 10.20.0.0/24, UUID is 00000000-0000-0000-0000-000000000000
Show a route's Prefix is 10.30.0.0/24, UUID is 78bf0204-bff7-443d-9c65-a4574cc5e04f
```
It's simple problem :pensive:

...If you can retrun values,from []byte　is "uuid ary",
example...

```bash
func (s *BgpServer) AddPath(vrfId string, pathList []*table.Path) (uuidBytes []byte, err error) {
        err = s.mgmtOperation(func() error {
                if err := s.fixupApiPath(vrfId, pathList); err != nil {
                        return err
                }
                //if len(pathList) == 1 {
                //      pathList[0].AssignNewUUID()
                //}
                for _, p := range pathList {
                        p.AssignNewUUID()
                        uuidBytes = append(uuidBytes, p.UUID().Bytes()...)
                }

                s.propagateUpdate(nil, pathList)
                return nil
        }, true)
        return
}
```

It's return [[16]byte,[16]byte,[16]byte...]

So, Under go run snippet,(HaHa, It's snippet very dirty)
https://gist.github.com/nao4arale/c21c9a443d14ac93f2993267c884c07f

```
b1, _ := s.AddPath("", []*table.Path{path1, path2})
        c1 := len(b1) / 16
        for i := 0; i < c1; i++ {
                u1, _ := uuid.FromBytes(b1[i*16 : (i+1)*16])
                fmt.Println("1:Added a route's UUID is, ", u1.String())
        }
```

```bash
1:Added a route's UUID is,  c964c4bc-5caa-4113-9e5f-ec540a77e0ba
1:Added a route's UUID is,  63086fb0-7438-48c0-b243-6285451dfd9b
2:Added a route's UUID is,  12010822-3e45-455a-abff-0c4f965474ef
Show a route's Prefix is 10.10.0.0/24, UUID is c964c4bc-5caa-4113-9e5f-ec540a77e0ba
Show a route's Prefix is 10.20.0.0/24, UUID is 63086fb0-7438-48c0-b243-6285451dfd9b
Show a route's Prefix is 10.30.0.0/24, UUID is 12010822-3e45-455a-abff-0c4f965474ef
```


....Ummmmmm...It's not cool.
Do you have good idea?

fujita:
Not returning uuid with multiple paths is intentional. The asymmetry isn't nice though. Adding multiple paths was added for injecting mrt dump (like full routes). We go comments on the performance of injecting mrt dump, so the api tries to do nothing else adding paths.

nao4arale:
Thanks, I'm understanding.
It would appreciate if you could add comment out,
 "Only when received one route, this api return uuid".

Reason, I think intend to using this api with multi path & want to receive uuid, such as me, use

fujita:
I've just pushed the fix. Surely, I would be happy to push a patch to add such comment.

nao4arale:
Thank you for pushed.
Okay, I will. :thumbsup:

PR #1481, opened 2017-11-02
packet/bgp: Remove non route key from EVPN NLRI string representation

iwaseyusuke:
[RFC7432](https://tools.ietf.org/html/rfc7432) and [draft-ietf-bess-evpn-prefix-advertisement](https://tools.ietf.org/html/draft-ietf-bess-evpn-prefix-advertisement) says some fields in EVPN NLRIs (e.g., MPLS Label(s)) should no be treated as the route key.

This patch removes non route key fields from NLRI string and enable to display these removed fields as path attributes on "gobgp" command.


fujita:
Pushed, thanks!

PR #1482, opened 2017-11-02
travis: Update .travis.yml

iwaseyusuke:
For scenario_test, Python build environment is more suitable rather than
Go build environment, because scenario_test is written in Python and the
building source codes will be done on Docker containers.

This patch fixes scenario_test to use Python build environment and
reduces "sudo" executions in order to use interpreter of virtualenv.

Note: On Travis-Ci, the Python interpreter of sudoer is older version,
and when install packages with "sudo pip", InsecurePlatformWarning will
be displayed. So it is better to use the newer interpreter of virtualenv.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,39 @@
 language: go
 
-before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo sysctl -w net.ipv6.conf.all.disable_ipv6=0; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo sysctl -w net.ipv6.conf.default.disable_ipv6=0; fi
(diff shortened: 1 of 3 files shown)


fujita:
'Suitable' isn't clear. What's the advantage of using Python build env rather than go env except for reducing sudo (not a big deal)?

iwaseyusuke:
Currently, we have no critical problem to use Python interpreter on Go Build environment, but the official Travis-CI does not recommend to use sudoer's (System) Python interpreter.

> Travis-CI - Building a Python Project
> ["Travis CI Uses Isolated virtualenvs"](https://docs.travis-ci.com/user/languages/python/#Travis-CI-Uses-Isolated-virtualenvs)
> The CI Environment uses separate virtualenv instances for each Python version. System Python is not used and should not be relied on.

The above describes about Python environment, but the same can be true for Go environment. Indeed, Python interpreter on Go environment is the old version and raises InsecurePlatformWarning for the every pip installations during CI.

fujita:
I see. Pushed, thanks.

PR #1485, opened 2017-11-06
*: Fix problems pointed by 'go vet'

fujimoto-s:
In the near feature (likely in Go1.10),
'go test' will never work if 'go vet' fails.
(See: https://github.com/golang/go/issues/18084)

This commit is for dealing with such a situation
(and also for improving the quality of our code).


fujita:
Pushed, thanks!

PR #1491, opened 2017-11-13
api: Define constants for address family

iwaseyusuke:
To call some gRPC APIs, it is required to specify the address family value which is combined value of AFI and SAFI values, so users of gRPC APIs need to calculate this value.

This PR introduces the constants for the address families for the convenience.
Also, improves the convenience of AddNeighbor API with AfiSafi structure.


vasu-dasari:
Thanks. The constants definition will be quite helpful.

iwaseyusuke:
@vasu-dasari Thank you for your confirmation!

fujita:
Pushed, thanks.

PR #1495, opened 2017-11-16
cmd/neighbor: Avoid shifted RIB display 

fujimoto-s:
Currently, display of RIB may be shifted from route to route
because placement is calculated for each destination.

This PR avoides the shifted display by calculating the placement
with whole of the destinations.

Also this PR does some refactoring.


fujita:
Pushed, thanks.

PR #1499, opened 2017-11-20
packet/bgp: Rename snake case to camel case

iwaseyusuke:
This patch renames some variable names in snake case (e.g., "last_byte_value") to that in the camel case (e.g., "lastByte") because Go prefers the camel case for variable names.

This patch makes linter (a little) more happy.


fujita:
Thanks

PR #1504, opened 2017-11-22
Allow 0 value to roaManager.AS for library usage

a16:
Currently, 0 is not allowed as `roaManager.AS`. In the case of using gobgp's RPKI library for a 
 monitoring, it probably can't have AS. So I'd like to use `0` for special purpose for such kind of systems.


fujita:
Pushed, thanks.

PR #1506, opened 2017-11-22
packet/bgp: Improve FlowSpec CLI argument format and string representation

iwaseyusuke:
Currently, the parser for the string representation of FlowSpec rules splits the given string into characters and validates them character by character. So we need to handle the unexpected white spaces carefully.
This problem is related to https://github.com/osrg/gobgp/pull/1489.

This PR introduces the regular expressions to parse the FlowSpec rules and simplify the parser.
With these patches, the robustness for the unexpectedly inserted white spaces like "   &   ==   tcp   " will be improved.

Also, includes some improvements for the string representation of the FlowSpec rules.
Example:
- The NOT MATCH operand of "tcp-flags" is replace from "=!" to "!="
- The combined "fragment" value will be joined with "+" (e.g. "is-fragment+first-fragment")

FYI, I will update documents for the FlowSpec on another PR.

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -3364,39 +3364,46 @@ var flowSpecParserMap = map[BGPFlowSpecType]func(RouteFamily, []string) (FlowSpe
 	FLOW_SPEC_TYPE_INNER_COS:     flowSpecNumericParser,
 }
 
-func ParseFlowSpecComponents(rf RouteFamily, input string) ([]FlowSpecComponentInterface, error) {
-	idxs := make([]struct {
-		t BGPFlowSpecType
-		i int
-	}, 0, 8)
-	args := strings.Split(input, " ")
-	for idx, v := range args {
-		if t, ok := FlowSpecValueMap[v]; ok {
-			idxs = append(idxs, struct {
-				t BGPFlowSpecType
-				i int
-			}{t, idx})
-		}
-	}
-	if len(idxs) == 0 {
-		return nil, fmt.Errorf("failed to parse: %s", input)
-	}
-	cmps := make([]FlowSpecComponentInterface, 0, len(idxs))
-	for i, idx := range idxs {
-		var a []string
-		f := flowSpecParserMap[idx.t]
-		if i < len(idxs)-1 {
-			a = args[idx.i:idxs[i+1].i]
+func extractFlowSpecArgs(args []string) map[BGPFlowSpecType][]string {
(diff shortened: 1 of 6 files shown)


packet/bgp/bgp.go
```diff
@@ -3415,6 +3415,20 @@ type FlowSpecComponentInterface interface {
...
+}
+
+func (c FlowSpecComponents) Swap(i, j int) {
+	c[i], c[j] = c[j], c[i]
+}
+
+func (c FlowSpecComponents) Less(i, j int) bool {
+	return c[i].Type() < c[j].Type()
+}
+
```
fujita: This patch moves sorting the rules into the inside of the "packet"
module and sorts them when decoding binary and creating new NLRI.

What's the motivation of the above?

sort.Slice() makes the code simpler?
https://golang.org/pkg/sort/#Slice

iwaseyusuke: Thank you for reviewing my patch.

---

> sort.Slice() makes the code simpler?
> https://golang.org/pkg/sort/#Slice

Thanks, it makes the code simpler as you said. I will update like the following.
(I will use `SliceStable` instead because godoc says `The sort is not guaranteed to be stable. For a stable sort, use SliceStable.`)
``` 
sort.SliceStable(n.Value, func(i, j int) bool { return n.Value[i].Type() < n.Value[j].Type() })
```

---

> What's the motivation of the above?

Sorry, I didn't mentioned enough.

There are two reasons.

The first one is improving the usability of "packet" module.
Currently, the users are required to sort the FlowSpec NLRI "every time" for getting the consistent ordered NLRI on its own implementation like;
https://github.com/osrg/gobgp/blob/d3adb6d680a409d00791f95445f75689338f56af/gobgp/cmd/global.go#L341-L364
With this patch, users are no longer required to sort the rules.

The second is the difference of the interpretation of [RFC5575](https://tools.ietf.org/html/rfc5575#section-5.1).
RFC5575 says;

>  This ordering function must be such that it must not depend on the
>  arrival order of the flow specification's rules and must be constant
>  must be constant.

If we emphasize "must be constant must be constant", we might need to reject the un-ordered rules from peers and should not sort them when decoding.
But if emphasize "it must not depend on the arrival order", it is better to sort the received rules into the defined order to handle them consistently.
The current implementation seems to stand for the former, but this patch stands for the latter because RFC5575 does not clarify "the un-ordered rules should cause the error handling (e.g., reset of session)".

fujita:
Thanks!

PR #1512, opened 2017-11-27
test/lib: Allow unconfigured peer-as for Unnumbered BGP

fujimoto-s:
For Unnumbered BGP, neighbor AS number isn't needed
to be specified, so 'peer-as' should not be specified
in the scenario test for Unnumbered BGP.
However, currently 'peer-as' is always specified
in test/lib/gobgp.py

This PR solves this by not specifying `peer-as`
if 'neighbor-interface' is configured.

### DIFF
--- a/test/lib/gobgp.py
+++ b/test/lib/gobgp.py
@@ -366,15 +366,17 @@ class GoBGPContainer(BGPContainer):
 
             neigh_addr = None
             interface = None
+            peer_as = None
             if info['interface'] == '':
                 neigh_addr = info['neigh_addr'].split('/')[0]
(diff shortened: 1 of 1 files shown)


fujita:
Pushed, thanks.

Seems that the unnumbered BGP feature via CLI doesn't work:
https://github.com/osrg/gobgp/blob/master/docs/sources/unnumbered-bgp.md#configuration-via-cli
Can you take look?

fujimoto-s:
It seems to be that NeighborAddress is not configured if Unnumbered BGP is configured via CLI.
I made a PR(#1516) for this, could you review this?

PR #1513, opened 2017-11-27
Introduce NexthopState to store IGP Metric

iwaseyusuke:
Currently, when the nexthop state is changed, zclient updates only the paths which contained on RIBs at that time and can not update the paths which arrived after the nexthop state was changed.
This PR introduces the NexthopState structure and enables to synchronize the nexthop state among all paths even if the new path is arrived the zclient updated the nexthop state.
Also, with the NexthopState structure, this PR enables to calculate the Best Path based on the IGP costs to nexthop.


table/path.go
```diff
@@ -139,17 +139,16 @@ type Validation struct {
...
+	IsWithdraw   bool
+	pathAttrs    []bgp.PathAttributeInterface
+	attrsHash    uint32
+	reason       BestPathReason
+	parent       *Path
+	dels         []bgp.BGPAttrType
+	filtered     map[string]PolicyDirection
+	VrfIds       []uint16
+	nexthopState *NexthopState
 }
```
fujita: fattening Path struct is the last thing that I want to do. With millions of Paths, this hugely impacts the memory usage. This is only for zebra so is there any way not to fatten Path struct?

iwaseyusuke: Even if pointer?
Well, in that case, we need to access the `NexthopState` cache to calculate the best path at the following. (It seems to be difficult to access to zclient instance nor BgpServer instance though...);
https://github.com/osrg/gobgp/blob/a0de90619b8e09b90c03dcf2e96451cfc2940a08/table/destination.go#L538-L544
If the global variable for the `NexthopState` cache is okay in "table" module, I can remove `nexthopState *NexthopState` from the `Path` structure.
e.g.) the `NexthopState` cache is populated via TableManager and set as "table" module wide variable (accessible from `paths.Less()`).

PR #1516, opened 2017-11-28
config/util: Fix to add Unnumbered BGP neighbor via CLI

fujimoto-s:
Currently, Unnumbered BGP neighbors cannot be added via CLI
because link local address is not extracted before it is used.
This PR fixes it by returning the link local neighbor address
when extracting neighbor address.

### DIFF
--- a/config/util.go
+++ b/config/util.go
@@ -179,7 +179,7 @@ func ExtractNeighborAddress(c *Neighbor) (string, error) {
 	if addr == "" {
 		addr = c.Config.NeighborAddress
 		if addr == "" {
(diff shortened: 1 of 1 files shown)


fujita:
How about the following fix to be consistent with configuration via file
https://github.com/osrg/gobgp/pull/1519/commits/37b0e2a02cf2bfd679cad0cd8a682aae62b5b91d

fujimoto-s:
I think it is good.
I agree that the fix for problems with the CLI should be included in the functional part of the CLI.

Thanks!

PR #1517, opened 2017-11-29
Fix for Zebra configuration via GoBGP API

vasu-dasari:


config/bgp_configs.go
```diff
@@ -521,7 +521,7 @@ const (
 	INSTALL_PROTOCOL_TYPE_OSPF               InstallProtocolType = "ospf"
 	INSTALL_PROTOCOL_TYPE_OSPF3              InstallProtocolType = "ospf3"
 	INSTALL_PROTOCOL_TYPE_STATIC             InstallProtocolType = "static"
-	INSTALL_PROTOCOL_TYPE_DIRECTLY_CONNECTED InstallProtocolType = "directly-connected"
+	INSTALL_PROTOCOL_TYPE_DIRECTLY_CONNECTED InstallProtocolType = "connect"
```
fujita: "directly-connected" comes from OpenConfig:
https://github.com/osrg/public/blob/master/release/models/policy/openconfig-policy-types.yang#L175

Seems needs to fix routeTypeValueMap instead

vasu-dasari: Ok. Let me take a look at routeTypeValueMap. Thanks

fujita:
Fixed. The current master should work.
https://github.com/osrg/gobgp/commit/3145a05a84da7036def7996d76ae9e110cdc1a10

Issue #1509, opened 2017-11-26
Compilation breaks on x86


shadrick:
Commit 68d2478e2ea1d8607f85e6ab3d4a792e054ce6c2 introduced a change, in packet/bgp/bgp.go:3370 that overflows an int on 32bit architectures.
Compiler line:
`src/github.com/osrg/gobgp/packet/bgp/bgp.go:3370:18: constant 1099511627775 overflows int`

iwaseyusuke:
Hi @shadrick 

The following fixes this issue?
```diff
diff --git a/packet/bgp/bgp.go b/packet/bgp/bgp.go
index 86b4cfd..56e15a3 100644
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -3367,7 +3367,7 @@ func flowSpecSnapParser(rf RouteFamily, typ BGPFlowSpecType, args []string) (Flo
        args = normalizeFlowSpecOpValues(args)
 
        f := func(i int) error {
-               if 0 <= i && i <= 0xffffffffff { // 5 bytes
+               if 0 <= i && uint64(i) <= 0xffffffffff { // 5 bytes
                        return nil
                }
                return fmt.Errorf("%s range exceeded", typ.String())
```

fujita:
@iwaseyusuke 
Needs to fix the following and the argument type of validateFunc()?
https://github.com/osrg/gobgp/blob/master/packet/bgp/bgp.go#L3034

iwaseyusuke:
You mean `validateFunc()` should accept the value as `int64` for the case the value of `strconv.Atoi()` exceeds the range of `int`, right?
I guess `strconv.Atoi()` should raise error "value out of range" if the given value exceeds the range of `int` (`0x7fffffff` on the x86?), then the exceeded value should not be passed to `validateFunc()`.

But it is better to change `validationFunc func(int)` to `validationFunc func(int64)`, if the value can be larger than the max of `int32`.
If the above patch fixes this issue, I will update them.

fujita:
> But it is better to change validationFunc func(int) to validationFunc func(int64), if the value can be larger than the max of int32.

The value can't be larger than int32? Why 0xffffffffff is used here?

iwaseyusuke:
It means the the value can't be larger than `int32` "as far as on x86 env", but CAN be larger then `int32` if on 64-bit env.
[...]

PR #1520, opened 2017-12-05
travis: Cross-compile test on i386 arch

iwaseyusuke:
This patch introduces a new test for cross-compile on i386 arch in order to check out errors like; https://github.com/osrg/gobgp/issues/1509

Also, Travis-CI unexpectedly overrides the "go env" variables after running `gimme`, this patch fixes to use "before_script" to avoid this issue.


fujita:
Pushed, thanks.

PR #1522, opened 2017-12-05
doc: remove "gobgp global as" from documentation

vincentbernat:
This subcommand doesn't exist. Dunno if it was present at some point.

### DIFF
--- a/docs/sources/cli-command-syntax.md
+++ b/docs/sources/cli-command-syntax.md
@@ -20,8 +20,6 @@ gobgp has six subcommands.
 ### 1.1 Global Configuration
 #### syntax
 ```shell
-# configure global setting and start acting as bgp daemon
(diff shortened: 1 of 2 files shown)


fujita:
Hmm, it exists
`fujita$ ./gobgp global`
`AS:        0`
`Router-ID: `
`fujita$ ./gobgp global as 1 router-id 1.1.1.1`
`fujita$ ./gobgp global`
`AS:        1`
`Router-ID: 1.1.1.1`
`Listening Port: 179, Addresses: 0.0.0.0, ::`

vincentbernat:
Hummm, I don't know what happened. For some reason, the command failed for me and since it's not in `--help` and the code for it is not implemented as a subcommand, I assumed it didn't exist. Retrying the same commands today work just fine. 😖

PR #1523, opened 2017-12-06
packet/bgp: JSON encode of MP_UNREACH_NLRI attribute

vincentbernat:
MP_REACH_NLRI was already correctly encoded in JSON, while this was
not the case for MP_UNREACH_NLRI.

Before:

```json
"PathAttributes": [
  {
    "type": 15,
    "value": "AAIBQCABDbgAEwAA"
  }
]
```

After:

```json
"PathAttributes": [
  {
    "type": 15,
    "afi": 2,
    "value": [
      {
        "prefix": "2001:db8:8::/64"
      }
    ],
    "safi": 1
  }
]
```

This commit also adds two tests not directly related to JSON encoding
as I was first thinking the problem may be related to incorrect
parsing of the path attribute.

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -6042,6 +6042,20 @@ func (p *PathAttributeMpUnreachNLRI) Serialize(options ...*MarshallingOption) ([
 	return p.PathAttribute.Serialize(options...)
 }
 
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks!

PR #1524, opened 2017-12-07
test: use the current PYTHONPATH in run_all_tests.sh

fujita:
On some environments, it's necessary to use the current PYTHONPATH to run python.


imcom:
+1

PR #1525, opened 2017-12-07
Request to ignore common files created by IDEs and reports from py-tests

imcom:
IDE creates its workspace configs in the project dir, which shall not be tracked

I also find that the `test/scenario_test/*.xml` reports should be ignored in git repo


fujita:
I removed the black line and pushed, thanks.

PR #1526, opened 2017-12-07
For better cleanup after tests

imcom:
added container name prefix
added network label

so later on when we do cleanup, we can filter the containers and networks created by the test. So other setups would not be affected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ _cgo_export.*
 
 _testmain.go
 
-test/scenario_test/*.xml
+test/scenario_test/nosetest*.xml
 
 *.exe
 *.test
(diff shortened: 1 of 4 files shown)


fujita:
Thanks for the pr.

1. please use the latest master as a base.
2. please drop 9eb08a9 and 7fb3a05. Pushing them doesn't make sense at all.
3. please remove `test/scenario_test/cleanup.sh`. Simply put these command as an example to test/scenario_test/README.md. I don't like the hard-coded `osrg-gobgp-test_` in the script.
4. let's use 'osrg-gobgp-test' rather than 'gobgp-test'. It should be enough.

imcom:
Thanks for your feedback, let me make some adjustments

imcom:
Let me close this one and use `master` to create a new PR

PR #1529, opened 2017-12-08
Introduce a common label for containers and networks created in gobgp tests

imcom:
use a common prefix for containers for easy cleaning up;
use label for networks in order to avoid colliding with others

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ _cgo_export.*
 
 _testmain.go
 
-test/scenario_test/*.xml
+test/scenario_test/nosetest*.xml
 
 *.exe
 *.test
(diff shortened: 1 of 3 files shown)


iwaseyusuke:
Hi @imcom 
Looks good to me, thanks!
I think this change is very good idea!

Just wondering, how about using "--label" to run container at the following?
https://github.com/osrg/gobgp/blob/9da6e7c3a25cfaa1f33ed5ddcc0ad1699c430e17/test/lib/base.py#L211-L216

If adding label to container instances, I guess we can filter target containers without renaming.

imcom:
yea, sure. How come I missed that label filtering in `docker ps`, let me use label for filtering in containers instead

iwaseyusuke:
Great! Thanks!

fujita:
Squashed #2 and #3, pushed, thanks!

PR #1531, opened 2017-12-11
cli-command-syntax.md: Missing arg to show VRF RIB

iwaseyusuke:
This patch fixes the example on the doc for showing the VRF RIB where
the "rib" keyword is missing to execute the command.

Also, the "-a <adddress family>" is an option and this patch puts it
into brackets.


fujita:
Thanks

PR #1533, opened 2017-12-12
 table/policy: Support prefix representation in NeighborSet

fujimoto-s:
Currently, "neighbor-set" supports only IP address representation
and IP prefix representation(such as "192.168.0.0/24") is not supported.

This commit enables to accept the prefix representation for "neighbor-set"
to allow neighbors to be specified as range.


fujita:
Looks reasonable feature addition, thanks.

PR #1536, opened 2017-12-14
config: Refactor some functions

fujimoto-s:
For code readability, this commit make some arguments to receivers.

And default.go and serve.go has some functions which are not directly related to its main role.
This commit moves those functions to util.go, a more suitable place.


fujita:
Looks reasonable, thanks.

PR #1538, opened 2017-12-16
packet/bgp: use strconv.FormatInt() instead of strconv.Itoa() in tests

vincentbernat:
Itoa() takes an int which will only be 32-bit on i386 and similar
architectures. In tests, some constants are too big. Therefore, switch
all uses of strconv.Itoa() to strconv.FormatInt().

### DIFF
--- a/packet/bgp/bgp_test.go
+++ b/packet/bgp/bgp_test.go
@@ -1140,7 +1140,7 @@ func Test_ParseEthernetSegmentIdentifier(t *testing.T) {
 	}, esi)
 
 	// ESI_LACP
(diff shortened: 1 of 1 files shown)


vincentbernat:
The too big constant is `0xaabbccdd`.

fujita:
Makes sense to me. Thanks! Seems that the latest image on travis-ci doesn't work on some tests (probably ipv6 stuff) so tests fails. The following commit would help 
https://github.com/osrg/gobgp/commit/eb5e3b8ff21726f819e431751777b489ba7da546

vincentbernat:
I have rebased on master.

fujita:
Thanks!

PR #1543, opened 2017-12-26
cli: Support EVPN Default Gateway attribute

iwaseyusuke:
Also this PR adds description and examples for ESI syntax formats.


fujita:
Nice, thanks.

PR #1545, opened 2017-12-28
cli: Slash separated offset arg for FlowSpec Prefix

iwaseyusuke:
Currently, "gobgp" command supports only;
  `"<ipv6 prefix>/<len> <offset>"`  # space separated
format to specify the prefix offset for the FlowSpec destination/source rules, but the displayed format with "rib" command is;
  `"<ipv6 prefix>/<len>/<offset>"`  # slash separated
and the slash separated format is not allowed as command arguments.

This patch enables to specify the slash separated format for the FlowSpec destination/source rules arguments and resolve unmatch of the input/output formats.

Example:
```bash
$ gobgp global rib -a ipv6-flowspec add match destination 0:db8:1::1/64/16 then accept
$ gobgp global rib -a ipv6-flowspec
   Network                        Next Hop             AS_PATH              Age        Attrs
*> [destination: 0:db8:1::/64/16] fictitious                                00:00:00   [{Origin: ?}
```

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -3099,34 +3099,39 @@ func flowSpecPrefixParser(rf RouteFamily, typ BGPFlowSpecType, args []string) (F
 	// - IPv6 Prefix
 	//   args := []string{"2001:db8:1::/64"}
 	// - IPv6 Prefix with offset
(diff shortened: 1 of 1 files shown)


fujita:
Looks reasonable, thanks.

PR #1546, opened 2017-12-28
addpath_test: Wait for re-establishment with ExaBGP

iwaseyusuke:
When adding or deleting routes in ExaBGPContainer, it is required to restart ExaBGPContainer with the current implementation.

This patch fixes to wait for the re-establishment with ExaBGP and improves the stability of this test case.


fujita:
Great, thanks!

I guess that it would be nice to move this kinda workaround to the exabgp class:
https://github.com/osrg/gobgp/blob/master/test/lib/exabgp.py

PR #1548, opened 2018-01-05
travis: Run "dep ensure" before "go fmt" and "go vet"

iwaseyusuke:
Because "satori/go.uuid" changed some syntax of its API, "go fmt" and
"go vet" commands will fail with the latest codes.

This patch fixes to run dep ensure before running these tests and solves
this problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,10 +89,10 @@ matrix:
 #
 # Misc
 #
-    - <<: *_go
-      script: test -z "$(go fmt ./...)"
-    - <<: *_go
-      script: test -z "$(go vet ./...)"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -89,10 +89,10 @@ matrix:
 #
 # Misc
 #
-    - <<: *_go
```
fujita: We had better to remove _go section?

fujita: https://github.com/osrg/gobgp/pull/1550

iwaseyusuke: Thank you for your reviewing. Yes, I agree to remove `_go` section!

Issue #1549, opened 2018-01-05
gobgp and zebra integration doesn't work if there are IPv6 prefixes


dbarrosop:
If you enable zebra integration and you have IPv6 prefixes in the RIB zebra will crash because `gobgp` will try to insert those IPv6 prefixes using IPV4 commands. Enabling debug shows the following line:

```
time="2018-01-03T10:48:09Z" level=debug msg="send command to zebra" Body="type: ROUTE_BGP, instance: 0, flags: , message: 1, safi: SAFI_UNICAST, prefix: 2406:3003:2032::/48, src_prefix: <nil>/0, nexthops[0]: 2001:418:0:5000::e10, distance: 0, metric: 0, mtu: 0, tag: 0" Header="{8 255 3 0 IPV4_ROUTE_ADD}" Topic=Zebra
```

And `zebra` crashes with this error:

```
ZEBRA: Assertion `masklen >= 0 && masklen <= 32' failed in file prefix.c, line 566, function masklen2ip
```

I tested a patch provided by @iwaseyusuke that seemed to do the trick. Attaching the patch for reference:

```
func (c *Client) SendIPRoute(vrfId uint16, body *IPRouteBody, isWithdraw bool) error {
    command := IPV4_ROUTE_ADD
    if c.Version <= 3 {
        if body.Prefix.To4() != nil {
            if isWithdraw {
                command = IPV4_ROUTE_DELETE
            }
        } else {
            if isWithdraw {
                command = IPV6_ROUTE_DELETE
            } else {
                command = IPV6_ROUTE_ADD
            }
        }
    } else { // version >= 4
        if body.Prefix.To4() != nil {
            if isWithdraw {
                command = FRR_IPV4_ROUTE_DELETE
            } else {
                command = FRR_IPV4_ROUTE_ADD
            }
        } else {
            if isWithdraw {
                command = FRR_IPV6_ROUTE_DELETE
            } else {
                command = FRR_IPV6_ROUTE_ADD
            }
        }
    }
    return c.SendCommand(command, vrfId, body)
}
```

PR #1552, opened 2018-01-08
zebra/zapi: Fix Zebra API type when IPv6 routes

iwaseyusuke:
Fixes https://github.com/osrg/gobgp/issues/1549


iwaseyusuke:
Also I think we need to investigate why scenario tests didn't complain this issue...

fujita:
Thanks!

Issue #1547, opened 2018-01-04
Compilation breaks


shadrick:
Compiler output:
`osrg/gobgp/table/path.go:392:37: multiple-value uuid.NewV4() in single-value context`

If you have your plates full i can suggest in a PR
Thanks, guys

ddexterpark:
I've tried to build a gobgp docker container with both Ubuntu & Alpine both have the same issue:
 
```
/go/src/github.com/osrg/gobgp/table/path.go:392:37: multiple-value uuid.NewV4() in single-value context
The command '/bin/sh -c go get github.com/osrg/gobgp/gobgp' returned a non-zero code: 2
```

Any help would be appreciated, thanks!

iwaseyusuke:
Thanks for your reporting!

This issue seems to be caused from the following commit on `satori/go.uuid`.
This commit changes signature of `uuid.NewV4()` to return two value `(UUID, error)` instead of `UUID`.
https://github.com/satori/go.uuid/commit/0ef6afb2f6cdd6cdaeee3885a95099c63f18fc8c

The following fixes this issue?
```go
# osrg/gobgp/table/path.go:

func (path *Path) AssignNewUUID() {
	path.OriginInfo().uuid = uuid.Must(uuid.NewV4())
}
```

iwaseyusuke:
Or, how about using `dep ensure` before your compilation as [README](https://github.com/osrg/gobgp#to-start-developing-gobgp)?

fujita:
As @iwaseyusuke pointed out, please follow the instruction in README
https://github.com/osrg/gobgp#to-start-developing-gobgp

PR #1553, opened 2018-01-09
table: follow changes of go.uuid library

a16:
The return value of go.uuid library has been changed from single to multiple since https://github.com/satori/go.uuid/commit/0ef6afb2f6cdd6cdaeee3885a95099c63f18fc8c.
It seems to have to use uuid.Must() helper instead.


fujita:
You need to use 'dep', dependency management tool, as discussed at https://github.com/osrg/gobgp/issues/1547

Another option is updating the dependency with the latest satori/go.uuid

PR #1555, opened 2018-01-10
 server: Empty path list in initial WatchEventUpdate

iwaseyusuke:
Currently, MonitorRib API with ADJ_IN type and "current" flag returns the initial response with an empty path list when gRPC client connecting to GoBGP daemon.
This causes the client can not receive the current paths which GoBGP daemon already has.

This PR fixes to compose the initial MonitorRib response with the current paths.
Also, enhances the unit test cases for MonitorRib API.


fujita:
pushed, thanks.

PR #1556, opened 2018-01-11
Support "current" flag in MonitorPeerState API

iwaseyusuke:
Also, updates "docs/sources/cli-command-syntax.md" for "monitor" command.


fujita:
Oops, somehow I forgot to export the feature via the API.  Pushed, thanks.

PR #1557, opened 2018-01-11
packet/bgp: use strconv.ParseUint instead of strconv.Atoi()

vincentbernat:
Atoi() returns a signed int. On a 32-bit platform, this is not big
enough to fit an unsigned 32-bit int. Replace all occurrences of
Atoi() to ParseUint() with the appropriate size as a parameter.

This fix this failure:

```
--- FAIL: Test_ParseEthernetSegmentIdentifier (0.00s)
        Error Trace:    bgp_test.go:1181
        Error:          Expected nil, but got: &errors.errorString{s:"invalid esi values for type ESI_AS: [2864434397 287454020]"}

        Error Trace:    bgp_test.go:1182
        Error:          Not equal: bgp.EthernetSegmentIdentifier{Type:0x5, Value:[]uint8{0xaa, 0xbb, 0xcc, 0xdd, 0x11, 0x22, 0x33, 0x44, 0x0}} (expected)
                                != bgp.EthernetSegmentIdentifier{Type:0x5, Value:[]uint8{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}} (actual)

                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1,2 +1,2 @@
                        -(bgp.EthernetSegmentIdentifier) ESI_AS | as 2864434397, local discriminator 287454020
                        +(bgp.EthernetSegmentIdentifier) ESI_AS | as 0, local discriminator 0

FAIL
FAIL    github.com/osrg/gobgp/packet/bgp        0.003s
```

There are other occurrences of Atoi() that seems suspicious when parsing command-line. However, I didn't have time to look at them. Please pay attention to `fst` and `snd`. I believe they are 16 bits, right?

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -1392,17 +1392,17 @@ func ParseRouteDistinguisher(rd string) (RouteDistinguisherInterface, error) {
 	if err != nil {
 		return nil, err
 	}
-	assigned, _ := strconv.Atoi(elems[10])
+	assigned, _ := strconv.ParseUint(elems[10], 10, 32)
 	ip := net.ParseIP(elems[1])
 	switch {
 	case ip.To4() != nil:
 		return NewRouteDistinguisherIPAddressAS(elems[1], uint16(assigned)), nil
 	case elems[6] == "" && elems[7] == "":
-		asn, _ := strconv.Atoi(elems[8])
+		asn, _ := strconv.ParseUint(elems[8], 10, 16)
 		return NewRouteDistinguisherTwoOctetAS(uint16(asn), uint32(assigned)), nil
 	default:
-		fst, _ := strconv.Atoi(elems[7])
-		snd, _ := strconv.Atoi(elems[8])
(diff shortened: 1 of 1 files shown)


iwaseyusuke:
It looks good to me! Thanks!
I will replace the remaining `strconv.Atoi()` after this PR will be merged.

> Please pay attention to fst and snd. I believe they are 16 bits, right?

Yes, right. `fsn` is the upper 16 bits of `asn` and `snd` is lower 16 bits.

iwaseyusuke:
Also I think better to prohibit the use of `strconv.Atoi()` and to checkout them in Travis-CI.
It might be overdoing, but be safer for the 32-bit platform...

vincentbernat:
You can also run tests with `GOARCH=386`, that would catch the ones I fixed.

iwaseyusuke:
@vincentbernat Thanks, I will try it!

fujita:
nice, thanks.

PR #1559, opened 2018-01-14
travis: also run unittests for 386

vincentbernat:
This would catch some errors due to limited size of `int`.


fujita:
Thanks, pushed another pull request that you guys've worked on.

PR #1560, opened 2018-01-14
Use strconv.ParseUint instead of strconv.Atoi()

iwaseyusuke:


client/client.go
```diff
@@ -818,7 +818,7 @@ func (cli *Client) GetRPKI() ([]*config.RpkiServer, error) {
 	}
 	servers := make([]*config.RpkiServer, 0, len(rsp.Servers))
 	for _, s := range rsp.Servers {
-		port, err := strconv.Atoi(s.Conf.RemotePort)
+		port, err := strconv.ParseUint(s.Conf.RemotePort, 10, 32)
```
vincentbernat: A port is 16 bit.

gobgp/cmd/bmp.go
```diff
@@ -40,7 +41,7 @@ func modBmpServer(cmdType string, args []string) error {
 		address = args[0]
 	} else {
 		address = host
-		pn, _ := strconv.Atoi(p)
+		pn, _ := strconv.ParseUint(p, 10, 32)
```
vincentbernat: A port is 16 bits.

server/rpki.go
```diff
@@ -453,7 +454,7 @@ func (c *roaManager) GetServers() []*config.RpkiServer {
 		l = append(l, &config.RpkiServer{
 			Config: config.RpkiServerConfig{
 				Address: addr,
-				Port:    func() uint32 { p, _ := strconv.Atoi(port); return uint32(p) }(),
+				Port:    func() uint32 { p, _ := strconv.ParseUint(port, 10, 32); return uint32(p) }(),
```
vincentbernat: Port is 16 bits

client/client.go
iwaseyusuke: I think it so at first, but config.RpkiServer has uint32 width of port.
https://github.com/osrg/gobgp/pull/1560/files#diff-1cd41a2611da49ce63630d5a51bd82c5R828

vincentbernat: Because it is part of the RPC interface, it cannot be changed, right? So, OK.

iwaseyusuke: No, `config.RpkiServer` is a part of config structures and not directly bound to gRPC API structure.
https://github.com/osrg/gobgp/blob/c6b44ccd8a73bd0caf715bbabe7e7a8fd0685f39/tools/pyang_plugins/gobgp.yang#L489-L493

But... You are right, the width of port field in TCP/UDP is 16 bits length and it cannot be other than uint16...
The type definition of `config.RpkiServerConfig.Port` was not proper...

Thanks, I will update it!

iwaseyusuke:
I've updated this PR for @vincentbernat 's comments.

fujita:
I feel that banning strconv.Atoi is going too far but I've pushed. Thanks.

PR #1561, opened 2018-01-15
zebra.md: Add description for supported version of FRRouting

iwaseyusuke:
Also improves the result of Markdown linter.


fujita:
Pushed, thanks. Worth adding Markdown linter to tests on travis-ci?

iwaseyusuke:
Hmmm... I don't think so.
Indeed better to apply Markdown linter, but I'm afraid that it make contributors feel difficulties to write new documentations.
I guess applying lint on Markdown is not "so" important, which means that it has no significant differences on HTML page even if linter complains.
And I think it more important to create and update documents actively.
Of course, breaking Markdown syntax should be avoided, but it can be checked easily by previewing the files via PR.

PR #1562, opened 2018-01-15
flow_spec_test: Enhance test cases with YABGP

iwaseyusuke:
This PR extends test cases using YABGP;
 - Test "gobgp" command outputs representations with IPv4/IPv6
 - IPv4 FlowSpec rules advertisements and withdraws

Note: Because YABGP does not support the IPv6 and L2VPN FlowSpec, this PR does not extend test cases for these family. If ExaBGP releases the next version of v4.0.2 or YABGP supports these family, I will create a new PR.

### DIFF
--- a/test/lib/base.py
+++ b/test/lib/base.py
@@ -48,6 +48,33 @@ BGP_ATTR_TYPE_CLUSTER_LIST = 10
 BGP_ATTR_TYPE_MP_REACH_NLRI = 14
 BGP_ATTR_TYPE_EXTENDED_COMMUNITIES = 16
 
+FLOWSPEC_NAME_TO_TYPE = {
+    "destination": 1,
+    "source": 2,
+    "protocol": 3,
+    "port": 4,
+    "destination-port": 5,
+    "source-port": 6,
+    "icmp-type": 7,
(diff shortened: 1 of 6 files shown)


fujita:
You plan to drop this YABGP flowspec tests once ExaBGP 4.0.2 is released?

iwaseyusuke:
I'm not going to do so at this time, because I found another issue(*) on ExaBGP...
"exabgpcli" can be available on the next release (v4.0.3?), but this issue might not be fixed until this version.

(*): A rules combination like ">80&<90" will be composed into "&>80&<90", but RFC says the first operator should not have "&".

The FlowSpec rules combinations are very complex, I think it better to test with the multiple implementations until the FlowSpec is widely deployed and the detail of the specification(**) is more clarified.

(**): For example;
- The operators "true", "false" and "!=" can be used or not?
- A IPv6 prefix of "2001:db8::/64/16" should be composed with 64 bits of "2001:db8::" or 48 bits of "db8::"? (RFC says `"The encoded prefix contains enough octets for the bits used in matching (length minus offset bits)."`.

fujita:
pushed, thanks.

PR #1563, opened 2018-01-16
gobgp/cmd/mrt: Avoid if statement always true

iwaseyusuke:
A statement `if v != A || v != B` becomes always true regardless of the value `v`.
Golang compiler seems to detect this problem with the `tip` version on Travis-CI.


fujita:
Oops, thanks.

PR #1566, opened 2018-01-19
test/lib/{bird,gobgp,quagga,yabgp}: Wait for daemon boot

iwaseyusuke:
Currently, scenario test library does not wait for the daemons boot in
containers at starting up or after sending SIGHUP for reloading new
configurations.
So some test cases (e.g., graceful_restart_test.py) fails occasionally
to insert static routes into GoBGP after reloading new configurations.

This patch fixes to wait for the daemons boot and improves the stability
of scenario tests.

Note: This patch does not introduce these improvements to
BagpipeContainer and ExaBGPContainer, because;
- Docker image for BaGPipe, which used in scenario test library, is too
  old and BagpipeContainer does not seem to be used in the current test
  cases.
- The version of ExaBGP in "osrg/exabgp" Docker image has no way to ask
  the daemon status other than "ps" command ("ps" is already done, but
  not enough). If "exabgpcli" is available on ExaBGP container, which is
  required the next release of 4.0.2 or 3.4.21, we can use it to check
  whether ExaBGP is started up or not.

### DIFF
--- a/test/lib/bird.py
+++ b/test/lib/bird.py
@@ -25,6 +25,7 @@ from lib.base import (
     BGPContainer,
     CmdBuffer,
     try_several_times,
+    wait_for_completion,
 )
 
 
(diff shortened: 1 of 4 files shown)


fujita:
pushed, thanks.

PR #1572, opened 2018-01-25
Enable GoBGP to resolve next-hop with FRR/Zebra for VRF routes 

amanshaikh75:
Hi

I have updated how VRF ID is determined before  z.client.SendNexthopRegister() method is invoked. This is based on the suggestion made by @iwaseyusuke .  I have tested it in my test-bed and the fix works as expected (i.e., GoBGP is able to resolve next-hop of VRF routes using information provided by FRR/Zebra, and then is able to communicate these routes to FRR/Zebra for ultimate installation in Linux Kernel forwarding table of respective VRFs).

aman


iwaseyusuke:
@amanshaikh75 Thanks! It looks good to me!

fujita:
Squashed the commits and pushed, thanks.

PR #1573, opened 2018-01-26
test/lib/{quagga,gobgp}: Add/Del routes without reloading daemons

iwaseyusuke:
With the current implementation, when adding or deleting routes on QuaggaBGPContainer or GoBGPContainer, it is required to reloading or rebooting daemons because new routes are added via their configure file.
This causes the frequent reloading of daemons and makes the scenario tests unstable.

This patch fixes to use their CLI commands ("gobgp" command for GoBGP, "vtysh" for Quagga) and reduces the number of restating daemons, then improves the stability of the scenario tests.

### DIFF
--- a/test/lib/base.py
+++ b/test/lib/base.py
@@ -123,6 +123,18 @@ def try_several_times(f, t=3, s=1):
     raise e
 
 
+def assert_several_times(f, t=30, s=1):
+    e = AssertionError
+    for _ in range(t):
+        try:
+            f()
+        except AssertionError as e:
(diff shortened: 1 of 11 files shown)


fujita:
Sounds great, thanks.

iwaseyusuke:
FYI, the similar improvements can be applied to ExaBGPContainer, but we need the newer version of ExaBGP.
I will create patches for ExaBGPContainer, if the following image is merged and published; https://github.com/osrg/dockerfiles/pull/14

PR #1576, opened 2018-01-29
cli: Implement "neighbor update" command

iwaseyusuke:
Please note that "neighbor" commands support only a few options for neighbor's configuration.
If you need to configure more detail options, please consider to use [the `HUP` signal to update configuration file](https://github.com/osrg/gobgp/blob/master/docs/sources/cli-operations.md#adding-or-deleting-a-peer-dynamically).


nodakai:
> "neighbor" commands support only a few options

That's unfortunate 😞  especially when hold time configs are missing from CLI...

fujita:
If the CLI supports all the options, the syntax will be unmaintainable. Please use the GRPC API.

iwaseyusuke:
Updated to resolve some conflicts.

PR #1578, opened 2018-01-31
config: Option to disable best path selection

iwaseyusuke:
In some use-cases, the option to disable best path selection is useful to reduce the compute resource usage.

### DIFF
--- a/table/destination.go
+++ b/table/destination.go
@@ -36,6 +36,7 @@ type BestPathReason uint8
 
 const (
 	BPR_UNKNOWN BestPathReason = iota
+	BPR_DISABLED
 	BPR_ONLY_PATH
 	BPR_REACHABLE_NEXT_HOP
 	BPR_HIGHEST_WEIGHT
@@ -53,6 +54,7 @@ const (
 
 var BestPathReasonStringMap = map[BestPathReason]string{
 	BPR_UNKNOWN:            "Unknown",
+	BPR_DISABLED:           "Bestpath selection disabled",
 	BPR_ONLY_PATH:          "Only Path",
 	BPR_REACHABLE_NEXT_HOP: "Reachable Next Hop",
 	BPR_HIGHEST_WEIGHT:     "Highest Weight",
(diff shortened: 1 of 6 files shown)


table/destination.go
```diff
@@ -500,6 +502,14 @@ func (dest *Destination) computeKnownBestPath() (*Path, BestPathReason, error) {
...
 	}
 
+	if SelectionOptions.DisableBestPathSelection {
+		log.WithFields(log.Fields{
+			"Topic": "Table",
+		}).Debug("computeKnownBestPath skipped")
+		// Note: The older path gets preference, because "newPathList" was
+		// appended after "knownPathList".
+		return dest.knownPathList[0], BPR_DISABLED, nil
+	}
```
fujita: I think that the best path selection is disabled, nothing should be sent to peers because there is no best path. This is for a use-case that gobgp receives routes from peers and sends nothing, usually called, a route-collector.

iwaseyusuke:
I've updated my PR. With this update, computeKnownBestPath() will return `nil` as the best path and GoBGP will not send any route to peers if the best path selection disabled.

table/destination.go
```diff
@@ -522,6 +524,12 @@ func (dest *Destination) computeKnownBestPath() (*Path, BestPathReason, error) {
 		return nil, BPR_UNKNOWN, nil
 	}
 
+	if SelectionOptions.DisableBestPathSelection {
```
fujita: Better to place this above L512?

iwaseyusuke: Thank your for reviewing. L512 is typo of L520? If so, yes, better to place there because no longer needed to access to `dest.knownPathList[0]`. I will update my patch.

fujita: Oops, right, L520.

iwaseyusuke: Thank you for confirmation. I will fix it.

iwaseyusuke:
I've updated my PR to move a `if` statement.

fujita:
Pushed, thanks.

PR #1582, opened 2018-02-06
table: Change type of BestPathReason to uint8

iwaseyusuke:
The current type is `string`, which size is multiple of `byte(=uint8)`.


fujita:
Pushed, I would love anything to shrink the size of Path.

PR #1583, opened 2018-02-08
bgp: Remove binary length field in BGP messages

iwaseyusuke:
Some fields in BGPOpen and BGPUpdate are referred from only parser package, then this patch removes these field to slim down.


fujita:
Hmm, for what? They are allocated but freed soon when GoBGP receives messages. Unlike table.Path, millions of these objects are not allocated so there is little benefit to reduce the size of them for me. I also like that they represent the wire-format.

iwaseyusuke:
Thank you for reviewing. Hmmm... exactly as you said. It does not contribute reduce the size memory. I will withdraw this PR.

PR #1589, opened 2018-02-10
scenario_test: FlowSpec with ExaBGP

iwaseyusuke:
This PR enhances test cases of scenario_test/flow_spec_test.py with using ExaBGP version 4.0.5.

With ExaBGP version 4.0.5, we can use "exabgpcli" command which enables ExaBGPContainer to add/delete routes without restarting ExaBGP daemon and get adj-RIB-{in,out}. Then, this PR extends the APIs of ExaBGPContainer using "exabgpcli" command.

### DIFF
--- a/test/lib/exabgp.py
+++ b/test/lib/exabgp.py
@@ -15,66 +15,59 @@
 
 from __future__ import absolute_import
 
-from itertools import chain
-import time
-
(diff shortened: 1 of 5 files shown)


fujita:
Thanks!

PR #1590, opened 2018-02-13
packet/bgp: Remove raw binary field on PathAttribute struct

iwaseyusuke:
This PR removes "Value" field, which stores the raw decoded/serialized binary data, in order to reduce the memory usage.
Also implements some missing String()/MarshalJSON() functions.


fujita:
Great, thanks.

PR #1600, opened 2018-02-21
packet/bgp: Rename AigpTLV to AigpTLVInterface

iwaseyusuke:
The other interface types in "packet/bgp" package have "Interface" suffix.

This patch fixes to follow this naming convention.
Golang's naming convention recommends to use "er" suffix though.

Also, this patch moves the AIGP TLVs definitions to the near of the PathAttributeAigp structure and defines NewAigpTLVDefault function.


fujita:
looks reasonable, thanks.

PR #1605, opened 2018-02-23
zclient: Avoid directly access to RIB

iwaseyusuke:
In order to avoid races, zclient should not access to RIB without calling APIs of BgpServer.
But currently zclient directly accesses to the global RIB when handling NEXTHO_UPDATE messages.

This patch fixes this problem by calling BgpServer.GetRib().


fujita:
pushed, thanks.

PR #1612, opened 2018-03-07
packet/bgp: Reduce struct size of "IPAddrPrefix" struct

iwaseyusuke:
This patch refactors the inheritance (embedding) relationship of IPAddrPrefix and its derivatives and removes "addrlen" field. "addrlen" is used only to determine whether the given prefix is IPv4 or IPv6.

### DIFF
--- a/server/zclient.go
+++ b/server/zclient.go
@@ -231,18 +231,18 @@ func newIPRouteBody(dst pathList) (body *zebra.IPRouteBody, isWithdraw bool) {
 	switch path.GetRouteFamily() {
 	case bgp.RF_IPv4_UC, bgp.RF_IPv4_VPN:
 		if path.GetRouteFamily() == bgp.RF_IPv4_UC {
-			prefix = path.GetNlri().(*bgp.IPAddrPrefix).IPAddrPrefixDefault.Prefix.To4()
+			prefix = path.GetNlri().(*bgp.IPAddrPrefix).Prefix.To4()
 		} else {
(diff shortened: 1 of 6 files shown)


server/zclient.go
```diff
@@ -231,18 +231,18 @@ func newIPRouteBody(dst pathList) (body *zebra.IPRouteBody, isWithdraw bool) {
 	switch path.GetRouteFamily() {
 	case bgp.RF_IPv4_UC, bgp.RF_IPv4_VPN:
 		if path.GetRouteFamily() == bgp.RF_IPv4_UC {
-			prefix = path.GetNlri().(*bgp.IPAddrPrefix).IPAddrPrefixDefault.Prefix.To4()
+			prefix = path.GetNlri().(*bgp.IPAddrPrefix).Prefix.To4()
```
cstockton: This should be refactored to future proof the code from panics, even if currently this is highly unlikely, i.e.:

```Go
switch T := path.GetNlri().(type) {
case *bgp.IPAddrPrefix:
  prefix = T.Prefix.To4()
case *bgp.LabeledVPNIPAddrPrefix):
  ...
default:
  ... what is sane in this case? Right now that is a panic about concrete types, can we do better?
```

fujita:
Thanks for the patch, and sorry for the long delay.

The `addrlen` field was removed in master, together with the switch to `netip.Prefix`. So this is no longer needed. Closing.

PR #1616, opened 2018-03-16
cmd/global: Add and refine help messages for adding routes

fujimoto-s:
This PR contains the following fixes for help messages for adding routes via CLI:
* Show help messages when an user inputs an invalid L3VPN or MPLS routes
* Add a description to the parameter that was not sufficiently explained


fujita:
pushed, thanks.

PR #1618, opened 2018-03-21
Fix AddPath MarshallingOption bug.

kishiguro:
When neighbor had AddPath capability, then disable AddPath after that.  Old marshallingOptions remains.  So once neighbor enabled AddPath it can't be disabled with no AddPath capability option in Open Message.


fujita:
Nice catch, thanks!

PR #1623, opened 2018-03-27
server/server: Set rsRib to RS client in adding dynamic neighbor

### LINKED ISSUE #1622: Route server & dynamic neighbor issue
Hi,

I encountered an issue while deploying a route server with dynamic neighbor.
You can reproduce the issue with the following config

**Route server config**
```yaml
global:
  config:
    as: 65000
    router-id: 1.1.0.1
  use-multiple-paths:
    config:
      enabled: true

dynamic-neighbors:
  - config:
      prefix: 172.17.0.0/8
      peer-group: local

peer-groups:
  - config:
      peer-group-name: local
    route-server:
      config:
        route-server-client: true
    transport:
      config:
        passive-mode: true
```

**Route client 1**
```yaml
global:
  config:
    as: 65001
    router-id: 1.1.1.1
  use-multiple-paths:
    config:
      enabled: true

neighbors:
  - config:
      peer-as: 65000
      neighbor-address: 172.17.0.2
```

**Route client 2**
```yaml
global:
  config:
    as: 65002
    router-id: 1.1.2.1
  use-multiple-paths:
    config:
      enabled: true

neighbors:
  - config:
      peer-as: 65000
      neighbor-address: 172.17.0.2
```

Now if you run this config and add route, you'll see on the route server that each neighbor does not have adj out rib.
If you change the config for static IP and defined neighbor instead of dynamic one, everything works as expected.

This issue is probably related to how a newly connected neighbor config is parsed.

https://github.com/osrg/gobgp/blob/master/server/server.go#L1015

As the config does not include `NeighborAddress` the peer.taldId referenced in `peer.go` 
https://github.com/osrg/gobgp/blob/master/server/peer.go#L114
won't exists and will cause the following issue.

Cheers

fujimoto-s:
Currently, globalRib is always set to dynamic neighbors even if they are route server clients.
This patch fixes to set rsRib when the neighbor is a route server client.

This fixes #1622.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -251,7 +251,11 @@ func (server *BgpServer) Serve() {
 				log.WithFields(log.Fields{
 					"Topic": "Peer",
 				}).Debugf("Accepted a new dynamic neighbor from:%s", remoteAddr)
-				peer := newDynamicPeer(&server.bgpConfig.Global, remoteAddr, pg.Conf, server.globalRib, server.policy)
+				rib := server.globalRib
+				if pg.Conf.RouteServer.Config.RouteServerClient {
+					rib = server.rsRib
+				}
+				peer := newDynamicPeer(&server.bgpConfig.Global, remoteAddr, pg.Conf, rib, server.policy)
 				if peer == nil {
 					log.WithFields(log.Fields{
 						"Topic": "Peer",


fujita:
Looks fine, pushed, thanks.

PR #1624, opened 2018-03-27
cmd: Prevent accepting unexpected argument names

fujimoto-s:
In CLI operation, currently, unexpected argument names
(such as 'aspath' for 'gobgp neighbor add')
may pass the validations and return no errors.
This commit prevents accepting those argument names
by specifying the number of expected arguments for each argument names.


fujita:
Looks fine, pushed, thanks.

PR #1625, opened 2018-03-28
mrt: Enable to dump locally generated routes

iwaseyusuke:
For reloading locally generated routes by using MRT dump file, this patch enables to include locally generated routes into TABLE_DUMPv2 records.

For the backward compatibility, this patch adds a new configure option "include-local" into "mrt-dump.config" section and please enable this option to include locally generated routes.

Configuration Example:

```toml
[[mrt-dump]]
  [mrt-dump.config]
    dump-type = "table"
    file-name = "/tmp/table.dump"
    dump-interval = 60
    include-local = true
```

### DIFF
--- a/server/mrt.go
+++ b/server/mrt.go
@@ -392,12 +392,12 @@ func (m *mrtManager) enable(c *config.MrtConfig) error {
 }
 
 func (m *mrtManager) disable(c *config.MrtConfig) error {
-	if w, ok := m.writer[c.FileName]; !ok {
+	w, ok := m.writer[c.FileName]
+	if !ok {
 		return fmt.Errorf("%s doesn't exists", c.FileName)
-	} else {
-		w.Stop()
-		delete(m.writer, c.FileName)
 	}
+	w.Stop()
+	delete(m.writer, c.FileName)
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


fujita:
What 'the backward compatibility' means?

iwaseyusuke:
Thank you for your reviewing.

> What 'the backward compatibility' means?

The current implementation does not include locally generated routes. I afraid there are someone who expect this behavior, then I added `include-local` option. If we can say that the lack of locally generated routes is a bug, I think this option is not required.

fujita:
Can you check out how quagga handles this?

iwaseyusuke:
Sure, I will.

iwaseyusuke:
I've checked out the MRT dumped output of FRR, and by the default, FRR includes locally generated routes with the follosing peer info;

```json
  ...(snip)...
  "peer": {
    "count": 2,
    "entry": [
      {
        "ip": "0.0.0.0",  # for "local"
        "bgp_id": "0.0.0.0",
        "type": 2,
        "asn": "0"
      },
      {
        "ip": "10.0.0.2",
        "bgp_id": "10.0.0.2",
        "type": 2,
        "asn": "65002"
      }
    ],
  ...(snip)...
```

I think we can say that the lack of locally generated routes is a bug. I will update my patch.

iwaseyusuke:
I've updated to remove `"include-local"` option.

fujita:
Pushed, thanks a lot!

Issue #1614, opened 2018-03-15
bug: repetitive GetRib() requests can taint existing Global RIB entry


tamihiro:
- After gobgp daemon keeps responding to the gRPC GetRib() requests for the same ipv4-unicast network for an unspecified times, its rib entry would be overwritten with an incorrect NLRI.

- Please see https://github.com/tamihiro/taintgobgprib for a specific example and reproducing procedure.

fujita:
Thanks a lot for the report and code to trigger the bug! Really helpful. As discussed on the slack channel, I think that the following commit fixes this:
https://github.com/osrg/gobgp/commit/cda7c44afe723be74e39266823b54ba28d0ec019

tamihiro:
Thank you fujita-san, I have been running my code for hours now but not a single incorrect NLRI shows up yet. I deem your commit a definite fix.

fujita:
Thanks a lot for the confirmation!

PR #1626, opened 2018-03-29
packet/bgp: Avoid data race when serializing

iwaseyusuke:
This PR fixes `Serialize()` functions of parsers in `packet/bgp/bgp.go` in order to avoid data races when serializing BGP UPDATE messages.

similar to https://github.com/osrg/gobgp/issues/1614


fujita:
Great, thanks! We should run all the unit tests with 'race' option in the future.

PR #1629, opened 2018-04-02
*.md: Improvements for markdownlint

iwaseyusuke:
This PR improves the results of markdownlint and also adds a new UT to invoke markdownlint.

When using IDE (e.g. VSCode), we can easily check the results of markdownlint and it should not be a heavy burden to write docs.


fujita:
pushed, thanks.

PR #1634, opened 2018-04-04
server/rpki.go - Fixes #1633 - fixed typo

### LINKED ISSUE #1633: A typo in server/rpki.go causing logical error
Here's a typo in server/rpki.go that will cause the Less method on `roas` to be always ending up comparing the `AS` member. 

https://github.com/osrg/gobgp/blob/264bd0ca126bf721687764332a01315d56e5b671/server/rpki.go#L68

and 

https://github.com/osrg/gobgp/blob/264bd0ca126bf721687764332a01315d56e5b671/server/rpki.go#L70

I guess the fix is rather trivial or if the comparison is never meant to be done, then we should simply get rid of the code.

gabhijit:
The typo in function Less() was causing logical error during comparison
and the comparison would always return whether AS number was less or not
regardless of Maxlen

### DIFF
--- a/server/rpki.go
+++ b/server/rpki.go
@@ -65,9 +65,9 @@ func (r roas) Less(i, j int) bool {
 	r1 := r[i]
 	r2 := r[j]
 
-	if r1.MaxLen < r1.MaxLen {
+	if r1.MaxLen < r2.MaxLen {
 		return true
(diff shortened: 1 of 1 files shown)


fujita:
Pushed, thanks a lot!

PR #1637, opened 2018-04-06
policy: ImplementAsAppendAction

fujimoto-s:
In this PR, we implement "as-append" action which appends AS to the tail of AS_PATH.
Usage:
```
    $ gobgp policy statement <statement name> add action as-append <as num> <repeat num>
```
```
    [policy-definitions.statements.actions.bgp-actions.set-as-path-append]
    as = "6666"
    repeat-n = 3
```


mrosic:
Would it be possible to merge this?

PR #1641, opened 2018-04-09
server: Avoid infinite UPDATE loop of RTM NLRI

iwaseyusuke:
When GoBGP dropped adj-Rib-out per Peer, we fixed to send the same Route Target Membership (RTM) NLRI even if it is already sent. This can cause the infinite UPDATE loop when Route Reflector(RR) reflects RTM NLRI to its clients.

For example, the following situation causes the infinite UPDATE loop.

Topology:

```text
     +----- RR -----+
     |              |
  Client1        Client2
```

When Client1 has VRF with RT 65000:1 and sends a RTM NLRI to RR, then RR reflects the NLRI to Client2. If a new VRF with the same RT 65000:1 on Client2 is created, Client2 will notify it to RR, then RR calculates the best, but RR will send the NLRI from Client2 to Client1 even if it is not the best. Client1 receives the NLRI again, calculates the best and re-sends the best. Then, RR reflects the received NLRI ... (infinite loop).

This patch fixes to compare the candidate path to be sent with the old path and assume the given candidate path was already sent before if the candidate path and the old path is the same path. Then avoids the infinite UPDATE loop.

### DIFF
--- a/server/peer.go
+++ b/server/peer.go
@@ -21,10 +21,11 @@ import (
 	"time"
 
 	"github.com/eapache/channels"
+	log "github.com/sirupsen/logrus"
+
 	"github.com/osrg/gobgp/config"
 	"github.com/osrg/gobgp/packet/bgp"
 	"github.com/osrg/gobgp/table"
-	log "github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


iwaseyusuke:
Sorry for disturbing, I fixed the behavior when the deletion of VRF (fixed to send withdraw correctly).

kishiguro:
Seems infinite loop bug is fixed with the patch.  Thanks for the work!

PR #1644, opened 2018-04-13
server: Avoid sending same update multiple times

iwaseyusuke:
When the multiple address families are contained in the path lists "gBestList" or "gOldList" at "propagateUpdateToNeighbors()", this lists will be sent to neighbors repeatedly as many number as that of the included address families.

This problem can be occurred when a list of paths which contains multiple address families are given to "AddPath()" API of "BgpServer".

This patch fixes to re-calculate a list of path for each address family and send it to neighbors.


fujita:
I think that I fixed the above issue in a different way. Can you check out the latest mater?

iwaseyusuke:
Thank you. I confirmed this issue is fixed on the latest master and will close this PR.

PR #1646, opened 2018-04-13
RTC: Fix outgoing RT filter

iwaseyusuke:
Currently, even if RTM NLRIs are exchanged as expected, RT filter is not successfully applied on a Route Reflector. This patch fixes problem and registers the existing scenario test `rtc_test` to the Travis-CI tests.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -363,11 +363,10 @@ func filterpath(peer *Peer, path, old *table.Path) *table.Path {
 		if _, y := peer.fsm.rfMap[bgp.RF_RTC_UC]; y && path.GetRouteFamily() != bgp.RF_RTC_UC {
 			ignore = true
 			for _, ext := range path.GetExtCommunities() {
-				for _, path := range peer.adjRibIn.PathList([]bgp.RouteFamily{bgp.RF_RTC_UC}, true) {
-					rt := path.GetNlri().(*bgp.RouteTargetMembershipNLRI).RouteTarget
-					if rt == nil {
-						ignore = false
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -363,10 +363,10 @@ func filterpath(peer *Peer, path, old *table.Path) *table.Path {
 		if _, y := peer.fsm.rfMap[bgp.RF_RTC_UC]; y && path.GetRouteFamily() != bgp.RF_RTC_UC {
 			ignore = true
 			for _, ext := range path.GetExtCommunities() {
-				for _, path := range peer.adjRibIn.PathList([]bgp.RouteFamily{bgp.RF_RTC_UC}, true) {
-					rt := path.GetNlri().(*bgp.RouteTargetMembershipNLRI).RouteTarget
+				for _, p := range peer.adjRibIn.PathList([]bgp.RouteFamily{bgp.RF_RTC_UC}, true) {
+					rt := p.GetNlri().(*bgp.RouteTargetMembershipNLRI).RouteTarget
 					if rt == nil {
```
marcin-ptaszynski: I understand why this is skipped, but I don't understand, why this NLRI can even have nil RouteTarget.

iwaseyusuke: Thanks for your comment! When `nil` Route Target means the default route target which indicates to a peer the willingness to receive all VPN route advertisements.

Oops, in that case, we should not ignore the given route. I will fix my patch!

fujita:
pushed thanks!

PR #1649, opened 2018-04-19
RTC: Withdraw routes when received RTM withdrawal

iwaseyusuke:
Because the RTM is removed from adj-RIB-in before collecting the candidate routes to be withdrawn, the candidates are unexpectedly filtered before sending withdraw messages. Then on the peers, the VPN routes are left on.

This patch fixes to selects candidates from the filtered paths and send withdraw messages as expected. Also resolves the TODO in "test/scenario_test/rtc_test.py".


iwaseyusuke:
Updated to resolve some conflicts.

fujita:
Thanks a lot! I've thought about a cleaner fix but out of ideas. I've resolved the conflicts. I'll push once the ci finishes. 
https://github.com/osrg/gobgp/pull/1699

PR #1650, opened 2018-04-20
server/server: Restart peers when capabilities are changed

fujimoto-s:
If capabilities are added or removed by updating peer configs,
it is needed to resend Open Message contains the new capabilities.
But currently, GoBGP does not resend Open Message for
Add-Path and Graceful Restart capabilities.
This commit fixes it to resend Open Message for them.


fujita:
pushed, thanks.

PR #1658, opened 2018-04-27
Close incoming channel in zapi on receive error

shitz:
Currently, it is not possible for a client of the zebra library to react to receive errors, e.g., when the far end closes the connection.

This simple fix allows a client to be notified about a receive error by closing the `incoming` channel.


fujita:
Pushed, thanks! Looks like more work is necessary for handling errors properly but the patch does nothing wrong.

Issue #1660, opened 2018-05-02
Error: Missing mandatory attribute: 01


dbarrosop:
Hello,
if you try to install a path using the grpc interface and you forget to set the Origin field the grpc interface will install it successfully without the attribute set and will advertise that to neighbors causing the neighbors to drop the connection because a mandatory field is missing.

Regards.

greenpau:
> to install a path using the grpc interface and you forget to set the Origin field the grpc 

@dbarrosop , what is the command or API call did you use to accomplish the above?

dbarrosop:
Sure, should have added it to begin with. This is the function I was playing with (I commented out the origin attribute). Also, note I was trying to understand how to this at this point so I apologize if there is a much better and simpler way :P
  
```
func addPath(c pb.GobgpApiClient, vrf string, nexthop string, id uint32) {
        pattr := make([]bgp.PathAttributeInterface, 0)
        nlri := bgp.NewIPAddrPrefix(24, "10.0.0.0")
        pattr = append(pattr, bgp.NewPathAttributeNextHop(nexthop))
        // pattr = append(pattr, bgp.NewPathAttributeOrigin(0))
        nlri.SetPathIdentifier(id)
        path := table.NewPath(nil, nlri, false, pattr, time.Now(), false)

        r, err := c.AddPath(context.Background(), &pb.AddPathRequest{
                Resource: pb.Resource_VRF,
                VrfId:    vrf,
                Path:     pb.ToPathApi(path),
        })
        if err != nil {
                log.Fatal(err)
        }
        log.Println("path: %v", r)
}
```

greenpau:
@dbarrosop , that's why we are in this game :-) Please take a look at this snippet https://github.com/osrg/gobgp/blob/07196197675c07c712eba0c24a934dcbe20783e4/table/destination_test.go#L75 and compare it to the above function. My thinking is that in the above function, the Origin is initialized with zero, because of make allocation. See https://stackoverflow.com/questions/29164375/golang-correct-way-to-initialize-empty-slice?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

greenpau:
@dbarrosop , there is no check for mandatory attributes here: https://github.com/osrg/gobgp/blob/b05e13b4951858524ce81a532c0fa8e2a42b88ba/table/path.go#L156

dbarrosop:
Yeah, my point with this issue is that the grpc interface should do one of two things:

[...]

PR #1663, opened 2018-05-03
table/path: require mandatory BGP attributes for a proposed path

greenpau:
[RFC4271 Section 5.1](https://tools.ietf.org/html/rfc4271#section-5.1) lists
well-known mandatory attributes. They are `ORIGIN`, `AS_PATH`, `NEXT_HOP`.
With this commit, the `NewPath()` function has a check to ensure that the
attributes are present when the function is being called.

Resolves: #1660

Signed-off-by: Paul Greenberg <greenpau@outlook.com>

### DIFF
--- a/table/path.go
+++ b/table/path.go
@@ -161,6 +161,19 @@ func NewPath(source *PeerInfo, nlri bgp.AddrPrefixInterface, isWithdraw bool, pa
 		return nil
 	}
 
+	if !isWithdraw {
+		originFound := false
+		for _, a := range pattrs {
+			if a.GetType() == bgp.BGP_ATTR_TYPE_ORIGIN {
+				originFound = true
+				break
+			}
+		}
+		if !originFound {
+			pattrs = append(pattrs, bgp.NewPathAttributeOrigin(0))
+		}
+	}
+
 	return &Path{
 		info: &originInfo{
 			nlri:               nlri,


greenpau:
@fujita , @dbarrosop  , once mandatory BGP attributes check was introduced, the test function [Test_ParsePath](https://github.com/osrg/gobgp/blob/master/gobgp/cmd/global_test.go#L26-L38) has an issue.

The buffer `10.0.0.0/24 rt 100:100 med 10 nexthop 10.0.0.1 aigp metric 10 local-pref 100"` does not have `AS_PATH` in it. Further the [ParsePath](https://github.com/osrg/gobgp/blob/master/gobgp/cmd/global.go#L1080) does not set `AS_PATH`.

What should  `AS_PATH` default to and where do you think is the most appropriate place to do so?

greenpau:
Thinking out loud ...

The test failures are related to `AS_PATH` path attribute. Let's say we do not check for its existence, what are the consequences?

The `table.UpdatePathAttrs`  modifies the `AS_PATH` from `server/peer.go`. Importantly, filterpath happens prior to the `table.UpdatePathAttrs()` call.

```
    if path = filterpath(peer, path, old); path == nil {
        return nil
    }
```

The `ProcessMessage()` function makes a number of `NewPath()` calls. The one below, happens on delete, will not have path attributes:

```
    for _, nlri := range dels {
        p := NewPath(peerInfo, nlri, true, []bgp.PathAttributeInterface{}, timestamp, false)
        pathList = append(pathList, p)
    }
```

Perhaps, it is better to introduce extra function parameter, `isValidate bool`?

```
./table/table_manager.go:		p := NewPath(peerInfo, nlri, false, attrs, timestamp, false)
./table/table_manager.go:			p := NewPath(peerInfo, nlri, false, reachAttrs, timestamp, false)
./table/table_manager.go:		p := NewPath(peerInfo, nlri, true, []bgp.PathAttributeInterface{}, timestamp, false)
./table/table_manager.go:		msgs = append(msgs, NewPath(info, nlri, false, pattr, time.Now(), false))
./table/path.go:func NewPath(source *PeerInfo, nlri bgp.AddrPrefixInterface, isWithdraw bool, pattrs []bgp.PathAttributeInterface, timestamp time.Time, noImplicitWithdraw bool) *Path {
./table/path.go:	path := NewPath(p.OriginInfo().source, nlri, p.IsWithdraw, p.GetPathAttrs(), p.GetTimestamp(), false)
./table/path.go:	path := NewPath(p.OriginInfo().source, nlri, p.IsWithdraw, p.GetPathAttrs(), p.GetTimestamp(), false)
```

Then, if `isValidate` set to `true`, check for compliance.

fujita:
Closed since I pushed a different fix for the problem.

PR #1666, opened 2018-05-08
config: Enable to configure VRF via config file

iwaseyusuke:
This patch enables to configure VRF tables by using config file. This feature is asked frequently on Slack and mailing-list.

Example of Usage:

```toml
[[vrfs]]
  [vrfs.config]
    name = "vrf1"
    id = 1
    rd = "65000:100"
    both-rt-list = ["65000:100"]
```


fujita:
pushed, thanks.

Issue #1679, opened 2018-05-15
statement not work after a "statement replace"


lnever:
There is a case that statement does not work after a `statement replace`
But if I try `del` and `add`, it works

here is an example:

```
root@nx:/go# gobgp policy statement FROM_ROUTE_SERVER:community_INFRAPROTECT:10 del condition community TMS any
root@nx:/go# gobgp policy statement FROM_ROUTE_SERVER:community_INFRAPROTECT:10 add condition community TMS any
root@nx:/go# gobgp global rib add 1.1.1.1/32 community 888:142
root@nx:/go# gobgp global rib                                 
   Network              Next Hop             AS_PATH              Age        Attrs
*> 1.1.1.1/32           0.0.0.0                                   00:00:06   [{Origin: ?} {Communities: 888:142}]
root@nx:/go# gobgp policy statement FROM_ROUTE_SERVER:community_INFRAPROTECT:10 set condition community TMS any
root@nx:/go# gobgp global rib add 1.1.1.1/32 community 888:142
root@nx:/go# gobgp global rib
Network not in table
```

imcom:
same issue here

PR #1681, opened 2018-05-17
fix: missing validateCondition in ReplaceStatement

lnever:
cause #1679


iwaseyusuke:
@lnever Thanks for your PR! Looks good to me.

If adding information, I found why we need to call `validateCondition()` function before replacement. `validateCondition()` function (added in this commit) does not just validation, but also finds and fills condition statement for each type. So we need to call it, but missing. This patch solve this problem.
https://github.com/osrg/gobgp/blob/2c388e07576a8ea1d7f661387f5a7ba44bc6f01e/table/policy.go#L2999-L3008

fujita:
pushed, thanks a lot guys!

PR #1683, opened 2018-05-18
zclient: Zvoid flapping of nexthop reachability

iwaseyusuke:
Fixes an issue @agember reported on Slack. Please refer to the commit comment of "zclient: Avoid flapping of nexthop reachability" for more details.
Also includes some improvements for `test/scenario_test/bgp_zebra_nht_test.py` and `zebra/zapi.go`.

### DIFF
--- a/server/zclient.go
+++ b/server/zclient.go
@@ -210,7 +210,11 @@ func (m *nexthopTrackingManager) filterPathToRegister(paths pathList) pathList {
 func filterOutExternalPath(paths pathList) pathList {
 	filteredPaths := make(pathList, 0, len(paths))
 	for _, path := range paths {
-		if path == nil || path.IsFromExternal() {
+		// Here filters out:
+		// - Nil path
+		// - External path (advertised from Zebra) in order avoid sending back
+		// - Unreachable path because invalidated by Zebra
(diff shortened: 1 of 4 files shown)


agember:
This fix does not seem to address the problem I am experiencing. Attached is an example topology and all log entries pertaining to a prefix (3.2.0.0/24) that demonstrates the unexpected behavior. I've noted the odd behaviors with questions asking _why_ some things occur. 
[example.txt](https://github.com/osrg/gobgp/files/2024793/example.txt)

iwaseyusuke:
@agember Thank you for testing my patch! I think you are facing another problem due to the danpening feature of the Next-Hop Tracking.
The danpening feature delays the update of nexthop states in order to avoid the states are "extremely" frequent updated. I implemented this feature by queuing the updated path instances and then doing replace the old paths after sleeping for a while.
But on your situation, the old paths seems to be withdrawn while the danpening feature sleeping, then this feature unexpectedly inserts the withdrawn paths. If answer the comment on your attached file "There are again 2 paths: one via edge2_0 and one via edge2_1 -- WHY??", the danpening feature causes this problem.

I will open another issue for this problem and will make another PR.

fujita:
Pushed, thanks.

PR #1687, opened 2018-05-21
server/server: Broadcast BMPPeerDownMessage for peer deletion via CLI

fujimoto-s:
Currently, GoBGP does not send BMPPeerDownMessage when its peer is deleted
via CLI.

When a peer is deleted, the connection will be lost and FsmMsg will be sent
to BgpServer. Then, BgpServer checks FsmMsg whether the peer in FsmMsg exists
in BgpServer's neighbor map.
In this case, the peer has already deleted from neighborMap, so BgpServer
regards the FsmMsg as invalid and discards it.

This commit fixes it by holding the neighbor in neighborMap until receiving
FsmMsg from the neighbor.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -194,11 +194,25 @@ func (server *BgpServer) Serve() {
 				"Topic": "Peer",
 			}).Warnf("Can't find the neighbor %s", e.MsgSrc)
 			return
-		}
-		if e.Version != peer.fsm.version {
+		} else if e.Version != peer.fsm.version {
 			log.WithFields(log.Fields{
 				"Topic": "Peer",
 			}).Debug("FSM version inconsistent")
+			return
+		} else if peer.fsm.adminState == ADMIN_STATE_DELETED {
+			// The peer is requested to be deleted via API.
+			// When the connection is lost, we delete the peer from neighborMap
+			// and broadcast the message to watchers.
+			if e.MsgType == FSM_MSG_STATE_CHANGE &&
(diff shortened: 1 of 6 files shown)


server/server.go
```diff
@@ -194,12 +194,26 @@ func (server *BgpServer) Serve() {
...
+		} else if peer.fsm.adminState == ADMIN_STATE_DELETED {
+			// The peer is requested to be deleted via API.
+			// When the connection is lost, we delete the peer from neighborMap
+			// and broadcast the message to watchers.
+			if e.MsgType == FSM_MSG_STATE_CHANGE &&
+				e.MsgData.(bgp.FSMState) == bgp.BGP_FSM_IDLE {
+				delete(server.neighborMap, peer.fsm.pConf.Config.NeighborAddress)
+				peer.fsm.state = bgp.BGP_FSM_IDLE
+				server.broadcastPeerState(peer, bgp.BGP_FSM_ESTABLISHED, e)
+				return
```
fujita: Can we avoid adding another place to delete from neighborMap?

fujita:
the first three commits look straightforward and good. Can you create a new pr including them?

fujimoto-s:
Thanks for reviewing!
I've opened a new PR for the first three commits(#1700).
In accordance with this, I changed the title of this PR, describing last two commits.

fujita:
Too long overdue but finally fixed this bug:
https://github.com/osrg/gobgp/commit/04843ab5032cb6f8edebf1226acc9ba78c5ff1e2

Thanks for finding the bug and working on it.

PR #1693, opened 2018-05-22
server: use negotiated family instead configured

fujita:


kishiguro:
Thanks!  This is much cleaner fix.

fujita:
pushed, thanks!

Issue #1682, opened 2018-05-18
Broken Peer Session when RTC capability is advertised by single Peer and Graceful Restart is enabled


marcin-ptaszynski:
## Setup:
Having RR with config:
```
global:
  config:
    as: 64512
    router-id: 192.168.80.221
  apply-policy:
    config:
        import-policy-list: []
        default-import-policy: accept-route
        export-policy-list: []
        default-export-policy: accept-route

neighbors:
- config:
    neighbor-address: 172.19.0.1
    peer-as: 64512
  route-reflector:
    config:
      route-reflector-client: true
      route-reflector-cluster-id: 192.168.80.221
  afi-safis:
  - config:
      afi-safi-name: l3vpn-ipv4-unicast
      enabled: true
  - config:
      afi-safi-name: rtc
      enabled: true
- config:
    neighbor-address: 172.19.63.254
    peer-as: 64512
  graceful-restart:
    config:
      enabled: true
      long-lived-enabled: true
  route-reflector:
    config:
      route-reflector-client: true
      route-reflector-cluster-id: 192.168.80.221
  add-paths:
    config:
      send-max: 4
  afi-safis:
  - config:
      afi-safi-name: l3vpn-ipv4-unicast
      enabled: true
    long-lived-graceful-restart:
      config:
        enabled: true
        restart-time: 172800
    mp-graceful-restart:
      config:
        enabled: true
  - config:
      afi-safi-name: rtc
      enabled: true
- config:
    neighbor-address: 172.19.0.3
    peer-as: 64512
  graceful-restart:
    config:
      enabled: true
      long-lived-enabled: true
  route-reflector:
    config:
      route-reflector-client: true
      route-reflector-cluster-id: 192.168.80.221
  add-paths:
    config:
      send-max: 4
  afi-safis:
  - config:
      afi-safi-name: l3vpn-ipv4-unicast
      enabled: true
    long-lived-graceful-restart:
      config:
        enabled: true
        restart-time: 172800
    mp-graceful-restart:
      config:
        enabled: true
  - config:
      afi-safi-name: rtc
      enabled: true
```
and 2 connected Peers (172.19.63.254 RTC disabled, other enabled)

## Issue
[...]

PR #1697, opened 2018-05-23
test/rtc: Add test for conflict RTC config

fujimoto-s:
Add test for the issue reported on #1682.


fujita:
pushed, thanks a lot!

PR #1698, opened 2018-05-23
fix withdraw with addpath and rtc enabled

fujita:
Needs withdrawn paths that includes attributes. The attributes are
necessary because they are used with rtc table to check if the paths
were sent.

### DIFF
--- a/table/destination.go
+++ b/table/destination.go
@@ -557,6 +557,29 @@ func getMultiBestPath(id string, pathList []*Path) []*Path {
 	return list
 }
 
+func (u *Update) GetWithdrawnPath() []*Path {
+	if len(u.KnownPathList) == len(u.OldKnownPathList) {
+		return nil
+	}
+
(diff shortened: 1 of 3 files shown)


table/destination.go
```diff
@@ -557,6 +557,30 @@ func getMultiBestPath(id string, pathList []*Path) []*Path {
...
 
+func (u *Update) GetWithdrawnPath() []*Path {
+	if len(u.KnownPathList) == len(u.OldKnownPathList) {
+		return nil
+	}
+
+	l := make([]*Path, 0, len(u.OldKnownPathList))
+
+	for i, p := range u.OldKnownPathList {
+		fmt.Println(i, p.GetNlri().String())
```
marcin-ptaszynski: debug

fujita: oops, i'll fix. thanks.

fujita:
pushed.

Issue #1696, opened 2018-05-23
Withdrawn paths restored unexpectedly due to Next-Hop Tracking dampening


iwaseyusuke:
As @agember reported on https://github.com/osrg/gobgp/pull/1683, there are cases that the withdrawn paths will be restored unexpectedly due to the Next-Hop Tracking dampening feature.
I think we need to;
- Disable the dampening feature
- Or checking whether paths going to be updated are still on RIB or not before calling `BgpServer.propagateUpdate()`

iwaseyusuke:
@agember Sorry for the delay. I've just pushed a PR for the issue you reported. Could you try my PR (https://github.com/osrg/gobgp/pull/1708)?

PR #1708, opened 2018-05-29
zclient: Drop Next-Hop Tracking dampening feature

iwaseyusuke:
The dampening feature was implemented to delay the update of nexthop states in order to avoid the states are "extremely" frequent updated. But with this implementation, if a path is withdrawn by peer or via API
which the dampening feature delaying the update of that path, the withdrawn path can be restored unexpectedly again.

And currently Quagga and FRRouting does not support the dampening, this path drops the dampening feature to avoid this problem.

Fixes https://github.com/osrg/gobgp/issues/1696

### DIFF
--- a/server/zclient.go
+++ b/server/zclient.go
@@ -17,6 +17,7 @@ package server
 
 import (
 	"fmt"
+	"math"
 	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


fujita:
pushed, thanks.

Issue #1701, opened 2018-05-25
user-unfriendly gobgp errors


greenpau:
When running `gobgp` cli against a router that is down, a user may encounter "context deadline exceeded" error:

```bash
# gobgp neighbor
context deadline exceeded
```

It is not entirely intuitive to understand whether there are no neighbors or whether the router is down.

Thus, it makes sense to intercept the errors and, if necessary, provide user-friendly error message.

fujita:
Surely, pull requests for such improvement are very welcomed.

PR #1712, opened 2018-05-31
gobgp/cmd: add message gRPC connection error

greenpau:
Adds a user-friendly error message on gRPC connection error.

Resolves: #1701


fujita:
pushed, thanks.

Issue #1702, opened 2018-05-25
exec: "gcc": executable file not found 


greenpau:
Unless a user has `gcc`, the build will fail:

```bash
go build -o ./gobgp/gobgpd ./gobgpd/
# github.com/osrg/gobgp/vendor/github.com/vishvananda/netlink
exec: "gcc": executable file not found in $PATH
```

Perhaps, there could be a build option that would allow building without the `vishvananda/netlink`. For example, if there is no need to configure data plane, that library maybe unnecessary.

fujita:
GoBGP's way to use the netlink package isn't related with data plane at all. It used for getting some information from linux network stack to configure unnumbered bgp feature.
https://github.com/osrg/gobgp/blob/master/config/default_linux.go

So a build option for this doesn't sound appropriate to me. If you send a pr that replace the netlink package with something, I'm happy to push it. But I think that it's normal to assume that golang development environment has gcc.

greenpau:
> If you send a pr that replace the netlink package with something, I'm happy to push it. 

@fujita , 👍  will take a look. Perhaps, the few calls you are making could be done by Go's standard library https://golang.org/src/syscall/netlink_linux.go.

> But I think that it's normal to assume that golang development environment has gcc.

Your assumption is reasonable.

PR #1715, opened 2018-05-31
config: remove dependency on external netlink library

greenpau:
This commit removes the dependency on the external netlink library
"github.com/vishvananda/netlink" with stdlib syscall.

Resolves: #1702

Signed-off-by: Paul Greenberg <greenpau@outlook.com>


fujita:
Sorry, but I don't think that this is a good move. This just brings the complexity and details of netlink into GoBGP. I want to put such complexity into outside popular packages.

Issue #1628, opened 2018-03-29
Per-Neighbor apply-policy doesn't work in most cases


marcin-ptaszynski:
Based on used YANG schema and generated [config model](https://github.com/osrg/gobgp/blob/master/config/bgp_configs.go#L3736), `apply-policy` is possible to be set in under following nodes:
- Global
- PeerGroup
- Neighbor
- AfiSafi

But for general scenario, only `global` `apply-policy` setting is effective, while per-neighbor configuration only works for Route Server client mode.

PR #1717, opened 2018-06-02
don't process RTC Messages with Policies

marcin-ptaszynski:
Currently Policy mechanism has no way to distinguish RTC messages from
proper Paths, e.g. AfiSafi Match.

When using RTC with default import/export reject all RTC messages
are rejected as well.

This commit is simple workaround for this issue.

### DIFF
--- a/table/policy.go
+++ b/table/policy.go
@@ -2868,7 +2868,7 @@ func (r *RoutingPolicy) ApplyPolicy(id string, dir PolicyDirection, before *Path
 		return nil
 	}
 
-	if before.IsWithdraw {
+	if before.IsWithdraw || before.GetRouteFamily() == bgp.RF_RTC_UC {
 		return before
 	}
 	result := ROUTE_TYPE_NONE
--- a/test/scenario_test/rtc_test.py
+++ b/test/scenario_test/rtc_test.py
@@ -38,8 +38,8 @@ class GoBGPTestBase(unittest.TestCase):
 
     def assert_upd_count(self, src, dst, sent, received):
         messages = src.get_neighbor(dst)['state']['messages']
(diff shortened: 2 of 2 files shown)


fujita:
How other BGP implementations handle this?

marcin-ptaszynski:
@fujita , I've taken a look into juniper documentation: https://www.juniper.net/documentation/en_US/junos/topics/example/vpn-bgp-route-target-export-policy-configuring.html and it seems there are 2 major differences:
1. Ability to match route family, e.g. `set policy-options policy-statement filter-rtc from family route-target`.
2. RTF Prefix lists, e.g. `set policy-options rtf-prefix-list exclude-103 203:203:103/96`

It seems GoBGP doesn't support either - related #1628 for point 1.

marcin-ptaszynski:
@fujita , Actually https://github.com/openconfig/public/blob/master/release/models/bgp/openconfig-bgp-policy.yang#L409 is supported. I've also verified it in gobgp implementations. Sorry about that.

marcin-ptaszynski:
@fujita , previous comment was premature, `afi-safi-in` condition isn't supported. Test I've performed included positive check only, while condition for `afi-safi-in` was completely ignored, which resulted in accept-all. I think proper fix for this issue is to add support for `afi-safi-in`.

fujita:
Yeah, I know that it's not supported. :)

marcin-ptaszynski:
@fujita , ok, now we both know. For our use-case, we don't care so much about juniper equivalent of `rtf-prefixes`, but we need to be able to at least match RT for vpnv4 rf without breaking rtc.

Do you think we can merge this workaround before proper fix or should we work on proper support for `afi-safi-in`?

fujita:
Should go with the proper fix, I think.

PR #1720, opened 2018-06-04
table: fix potential panics in tableKey

cstockton:
It is possible for a call from other Table methods for an interface that does
not match the tables route family to reach the type assertion and panic. This
change better reflects the intention of the code by checking the concrete type
using a type switch instead.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -291,18 +291,16 @@ func (t *Table) setDestination(dst *Destination) {
 }
 
 func (t *Table) tableKey(nlri bgp.AddrPrefixInterface) string {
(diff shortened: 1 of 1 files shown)


cstockton:
If Go 1.10 is an option I really dislike the unsafe usage here- strings builder could be used in place to the same affect, i.e.:

```Go
var sb strings.Builder
sb.Grow(N)
// copy
return sb.String()
```

fujita:
I think that if someone hits a panic here, he misuses this package but the change does nothing wrong. So I pushed, thanks.

PR #1724, opened 2018-06-05
cmd/rpki_test: Unit test for showRPKITable

iwaseyusuke:
Adds an unit test for ed3f7b711c92c69bf231bc9e6a06312b23270db0


fujita:
pushed, thanks.

Issue #1719, opened 2018-06-04
Route-reflector: Why cluster-id need to be a valid IPv4 address


murali-reddy:
Whey for using Route-reflector, cluster-id need to be a valid IPv4 address?

https://github.com/osrg/gobgp/blob/master/config/default.go#L245

It could be just valid integer as well?

iwaseyusuke:
@murali-reddy Hi, an integer value might be enough when configuring `cluster-id` on files, but RFC4456 says;

```text
7.  Redundant RRs

   Usually, a cluster of clients will have a single RR.  In that case,
   the cluster will be identified by the BGP Identifier of the RR.
   However, this represents a single point of failure so to make it
   possible to have multiple RRs in the same cluster, all RRs in the
   same cluster can be configured with a 4-byte CLUSTER_ID so that an RR
   can discard routes from other RRs in the same cluster.
```

I think this implies the BGP Identifier (represented as IPv4 address) can be the default value of `cluster-id` and the representation of `cluster-id` should be the same with the BGP Identifier.

iwaseyusuke:
I checked the Cisco router's `bgp cluster-id` command, Cisco can accept `cluster-id` as an integer. i will create a patch to accept `cluster-id` as an 32-bit integer value.

murali-reddy:
thanks @iwaseyusuke

fujita:
@murali-reddy please try the latest master.

murali-reddy:
@iwaseyusuke @fujita thanks for a such a quick turnaround. Will try the latest master.

PR #1725, opened 2018-06-05
config: Accept CLUSTER_ID as an integer value

iwaseyusuke:
Currently, only IPv4 address is acceptable for the CLUSTER_ID setting, this patch enables to specify the CLUSTER_ID as a 32-bit unsigned integer.

With this expansion, "Config.RouteReflectorClusterId" stores the raw configured value for the CLUSTER_ID and "State.RouteReflectorClusterId" stores the value used for construct the CLUSTER_LIST attribute.

Fixes https://github.com/osrg/gobgp/issues/1719


fujita:
Looks fine, thanks.

Issue #1457, opened 2017-10-09
Better API for addPath


nemith:
I was able to use GoBGPDas a tutorial for a NANOG hackathon, however there were a couple shortfalls when using a gobgpd as a base for automation.  The largest issue was how difficult it was to add new routes using the GRPC API.  (minus the lack of package level documentation).

https://github.com/nemith/nanog71/blob/master/pkg/bgpctrl/main.go

This was difficult when using Go, but near impossible when using other client languages.

I would imagine this would take a lot more modeling of routes and attributes required inside protobuf.  Any may help here? https://developers.google.com/protocol-buffers/docs/proto3#any

fujita:
Agreed that we really need to improve the usability of AddPath API (and its document).
I don't see how proto.Any would help. Even with it, we still need to define a structure for an attribute? Can you elaborate on the idea?
Btw, thanks a lot for using GoBGP for the tutorial! The slides look great.

nemith:
I was thinking that you can define many different attribute structures (one for each supported attribute) and allow proto.Any to a repeated list of them.

I may give this a shot.

dylandreimerink:
A few days ago I stared on a project that aims to make a PHP wrapper for the gRPC client. So far I have only been able to decode flowspec routes. But my intentions are to cover the full API including inserting routes and using the monitor function ect. I also provides links to the RFC's I have used so far. I hope you find something you can use. https://github.com/dylandreimerink/gobgp-php

fujita:
@nemith 
something like this?
https://github.com/fujita/gobgp/blob/grpc/api/gobgp.proto#L1139-L1173

Then you can put any attribute like:
https://github.com/fujita/gobgp/blob/grpc/api/gobgp.proto#L1139-L1173

nemith:
Yes.  That is exactly what i was thinking!

iwaseyusuke:
Hi, I make a draft proposal of `.protobuf`: [`api/bgp/attribute.proto`](https://github.com/iwaseyusuke/gobgp/blob/f3f26b9e512e78c4e04f15ea146322b8ddac8df8/api/bgp/attribute.proto)

For the usage of this protobuf, please refer to this UT file: [`api/bgp/attribute_test.go`](https://github.com/iwaseyusuke/gobgp/blob/f3f26b9e512e78c4e04f15ea146322b8ddac8df8/api/bgp/attribute_test.go)

With this protobuf, `ModPath` API (not implemented yet) would be like:

```go
[...]

PR #1729, opened 2018-06-08
api: Define protobuf structures of BGP NLRI and Path Attributes

iwaseyusuke:
Currently, the gRPC client need to serialize the NLRI and Path Attributes into bytes representation when calling AddPath gRPC API, but this requires the client side applications have the implementation to encode/decode the BGP attributes. Then it can be difficult to implement gRPC client with the languages other than Go.

This PR defines the protobuf structures of BGP NLRI and Path Attributes and enable to call AddPath API with these structures.

For the discussion about this improvements and sample code of gRPC client written in Python, see https://github.com/osrg/gobgp/issues/1457.


fujita:
Great, can you add some tests of grpc with this new API? I'm thinking about using gogo/protobuf but concern of Any type a bit.

cstockton:
I think this change would be great. One thing to consider would be taking this opportunity to remove the `table` importing of `api`. This is because it forces anyone who wants to use the `api` to download all of the packages the sever requires, which is a lot of dependencies. If more functions like `ToPathApiInBin` are added to the `api` package now it will make migrating away more difficult in the future. The change itself would be fairly simple, remove ToNativePath and add `table.PathToAPI()` `table.APIToPath()` (or whatever names are preferred). Then in the future other dependencies could slowly be peeled away, which I would be happy to help with if the GoBGP authors would accept the change.

fujita:
I guess that we are on the same page. Once I push this pr, I want to make api/ include only code to access to the gRPC APIs and handy functions to handle the protobuf structures. That should remove lots of dependencies. I want to move grpc_server.go and util.go to somewhere else. Then I want to remove client/ and rewrite gobgp/*.go to use only api/.

iwaseyusuke:
I've updated to move `api/bgp/attribute*` into `api/` to build protobuf by single `protoc` command. (Also reworded a commit tile, sorry for disturbing)

fujita:
Thanks a lot, pushed. I think that it's better to remove the usage of logrus and just return an error. I'll take care of such minor issues.

PR #1731, opened 2018-06-08
table/message.go : Changed MP packer

denizaydin:
Changed MP packer interface to send single update for MP Reach NLRI that have same nexthop and attributes. This will decrease the convergence time with large number of MP routes like Internet in a VRF scenarios.


fujita:
Thanks for the patch, and sorry for the very long delay.

This is now implemented. Commit 150e672c ("Pack multiple NLRIs into MP-BGP UPDATE messages") groups paths by attributes and nexthop and packs as many NLRIs as fit into one MP_REACH, the same way this PR does. It also batches MP_UNREACH withdrawals, handles the ADD-PATH NLRI length, and respects the negotiated extended message length.

Closing this as done.

PR #1733, opened 2018-06-10
cmd: don't display uptime for a BGP session if no uptime information

vincentbernat:
Otherwise, we get something like this:

    BGP neighbor is 2001:db8:4::2, remote AS 65000
      BGP version 4, remote router ID unknown
      BGP state = active, up for 17692d 16:53:12


fujita:
looks good. pushed, thanks.

PR #1737, opened 2018-06-12
config: Change "RedistributeRouteTypeList" to []string

fujimoto-s:
RedistributeRouteTypeList is []InstallProtocolType,
but InstallProtocolType only defines limited number of protocols,

And currently, EnableZebra() in gRPC API validates its request
with the protocols defined by InstallProtocolType,
so the protocols such as "babel" or "lldp" could not be configured
to Zebra, via gRPC API.

This patch fixes RedistributeRouteTypeList to be []string,
and fixes EnableZebra() to validate with the protocols defined
in zapi.go.


fujita:
pushed, thanks.

PR #1742, opened 2018-06-13
 gobgpd: Use gRPC APIs Instead of BgpServer functions

fujimoto-s:
In order to integarate the setting via configuration files and via gRPC API,
This commit fixes gobgpd to use gRPC APIs
instead of using BgpServer functions directly, for consistency.


fujita:
Great, thanks! Long overdue but now everything can be configured via the GRPC APIs.

PR #1744, opened 2018-06-14
Pull regex into variables to prevent the regex to compile in the hotpath

jeffbean:
The MustCompile is useful since you can declare them as package variables since it will panic on load time. This also fixes implicit runtime panic conditions where these regex patterns are not tested and could have panicked. 

Move regex.MustCompile to package variables to avoid recompiling. This improves it by 10X since the benchmarked function is doing two compiles. 

Before: 
```
$ go test -bench=.                                                                  
goos: darwin
goarch: amd64
pkg: github.com/osrg/gobgp/packet/bgp
BenchmarkNormalizeFlowSpecOpValues-4       50000             22607 ns/op
PASS
ok      github.com/osrg/gobgp/packet/bgp        1.469s
```

After: 
```
$ go test -bench=.                                                                  
goos: darwin
goarch: amd64
pkg: github.com/osrg/gobgp/packet/bgp
BenchmarkNormalizeFlowSpecOpValues-4     1000000              1604 ns/op
PASS
ok      github.com/osrg/gobgp/packet/bgp        1.643s
```

### DIFF
--- a/packet/bgp/bgp.go
+++ b/packet/bgp/bgp.go
@@ -287,6 +287,51 @@ func (c BGPCapabilityCode) String() string {
 	return fmt.Sprintf("UnknownCapability(%d)", c)
 }
 
+var (
+	// Used parsing RouteDistinguisher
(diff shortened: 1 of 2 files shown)


fujita:
oh, looks great improvement. pushed, thanks!

PR #1748, opened 2018-06-14
simple staticcheck fixes

jeffbean:
Some staticcheck issues that were identified: 

```
~/gocode/src/github.com/osrg/gobgp/api/grpc_server.go:72:45: printf-style function with dynamic first argument and no further arguments should use print-style function instead
~/gocode/src/github.com/osrg/gobgp/gobgp/cmd/policy.go:68:34: printf-style function with dynamic first argument and no further arguments should use print-style function instead                         
~/gocode/src/github.com/osrg/gobgp/server/server_test.go:172:14: sleeping for 1 nanoseconds is probably a bug. Be explicit if it isn't: time.Sleep(time.Nanosecond)                                     
~/gocode/src/github.com/osrg/gobgp/server/sockopt_darwin.go:42:2: should check returned error before deferring fi.Close()                                                                    
~/src/github.com/osrg/gobgp/server/sockopt_darwin.go:51:2: should check returned error before deferring fi.Close()
```

### DIFF
--- a/api/grpc_server.go
+++ b/api/grpc_server.go
@@ -69,7 +69,7 @@ func (s *Server) Serve() error {
 
 	serve := func(host string) {
 		defer wg.Done()
-		lis, err := net.Listen("tcp", fmt.Sprintf(host))
+		lis, err := net.Listen("tcp", host)
 		if err != nil {
 			log.WithFields(log.Fields{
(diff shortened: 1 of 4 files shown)


fujita:
pushed, thanks.

PR #1750, opened 2018-06-19
api: Use attribute.proto struct in message Vrf and VRF commands

iwaseyusuke:
The current formats of the RD and import/export RTs in message Vrf are the binary type representation, this PR fixes to use the structures defined in attribute.proto file.

Also, removes "table/" package use in `gobgp vrf` commands and fixes to use "api/" package


fujita:
pushed, thanks.

PR #1753, opened 2018-06-21
api: Define protobuf for BGP capabilities

iwaseyusuke:
The current formats of the remote/local capabilities in `message PeerConf` are the binary type representation.

This PR defines the protobuf structures for BGP capabilities and fixes to use these structures in `message PeerConf`.


fujita:
pushed, thanks.

PR #1754, opened 2018-06-21
packet/bgp: Fix length calc when multiple RTM NLRIs

iwaseyusuke:
When multiple RTM NLRIs are composed in a single MP_REACH_NLRI, the current implementation will fail to decode those NLRIs because RouteTargetMembershipNLRI.DecodeFromBytes() tests the length of the given bytes even if including the subsequent bytes.

This patch fixes to cut the bytes by using the prefix length before testing the byte length.


fujita:
pushed, thanks.

Issue #1638, opened 2018-04-07
Don't use global logrus


phemmer:
This is a proposal to not use the global logrus logger, and to instead use an instanced logger (`logrus.New()`).
The reasons behind this are:
* Consumers of gobgp as a library might inadvertently trample over gobgp's log settings if they, or another library, are also using logrus. Such as calling `logrus.SetLevel(logrus.DebugLevel)`, or `logrus.AddHook()`.
* If a consumer wants to perform changes, like increasing gobgp's log level, they have to be aware that gobgp is using logrus, and call the logrus package directly to make settings. It would be better if the logging implementation were hidden, and a generic API were provided instead to perform operations such as changing the log level. Of course it would still be an option to expose the logrus instance via a method where a consumer could access it directly if they needed to.

Ref https://github.com/cloudnativelabs/kube-router/issues/363

fujita:
I could do that but I'm wondering what a golang library should do for logging. Not logging at all?

phemmer:
sorry, I somehow missed the notification for this. But to respond, i'm not sure I'm following the question. The library should log, but the application using each library should be able to control the logging of each library separately. If you want to turn up the log level on libfoo, you shouldn't be forced to turn up the log level on libbar as well.

dcarrier:
If no one else is working on this I can take a stab at introducing a per package logger with setLevel methods as described above.

fujita:
@dcarrier surely, thanks! I've been thinking about rewriting logging stuff (such as replace logrus, etc) but that's the long-term item in my todo list.

jeffbean:
Zap might have some properties you are looking for: https://github.com/uber-go/zap

Its designed for applications with sensitive performance characteristics especially around allocations. 

Another option is to provide `NewBgpServer` with the `Options` pattern allowing users to supply logging settings. It can expand to exposing and composing a nice API surface to `NewBgpServer` while not breaking the current API.

```go
// Option is a bgpserver option.
type Option func(*BGPServer)

// WithLogger sets a logger 
func WithLogger(logger *zap.Logger) Option {
	return func(b *BGPServer) {
		b.logger = logger
	}
}

func NewBgpServer(options ...Option) *BGPServer {
    server := &BgpServer{...}
    for _, option := range options {
         option(server)
    }
    return server
}
```

vincentbernat:
[...]

PR #1756, opened 2018-06-21
log: don't use global logger from logrus

vincentbernat:
Instead, use our own global logger. It is defined in config package
and users should be able to configure it this way. Example has been
updated to reflect this change.

Unfortunately, for the user, this is not totally a drop-in change:

 - Get the logger: `log := config.Logger`.
 - Remove import `github.com/sirupsen/logrus`, except when using
 - levels (like `logrus.DebugLevel`). In this case, don't name the
 - import as `log`, but keep `logrus`.
 - If using `log.SetOutput`, use `log.Output` instead.
 - If using `log.SetFormatter`, use `log.Formatter` instead.

With recent versions of logrus, no other changes are needed.

While this change is not as neat as to have a per-server logger, this
should fit most uses where we want to control how GoBGP do the logging
as a library.

See #1638.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -48,7 +48,10 @@
   name = "github.com/golang/protobuf"
   packages = [
     "proto",
-    "ptypes/any"
+    "ptypes",
+    "ptypes/any",
+    "ptypes/duration",
+    "ptypes/timestamp"
   ]
   revision = "748d386b5c1ea99658fd69fe9f03991ce86a90c1"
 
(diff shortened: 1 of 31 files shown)


vincentbernat:
I didn't base my branch on the latest master. I am fixing that.

jeffbean:
This still has the issue of being a global logger (just not the logrus global logger). What really needs to happen is a logger is passed to each call site and a constructor creates the logger.

Overall seems OK in the current state of affairs and is an incremental improvement.

fujita:
Sorry about the conflicts. If you really want to push this before the next release, I'll. I prefer to refactor logging wholly for 2.X after the next release though.

vincentbernat:
I don't mind waiting. However, I don't think it qualifies as a refactor. It's really just a very first step before being able to specify a per-instance logger. This is still helpful to have that in case you have several logrus users in the same application.

fujita:
Sorry about commenting on the issue created three years ago. Finally, I Introducing Logger interface to solve this logging issue.

Issue #1745, opened 2018-06-14
server.Shutdown should not os.Exit


jeffbean:
The BGPServer [Shutdown](https://github.com/osrg/gobgp/blob/master/server/server.go#L1424) should not `os.Exit`. Nothing besides a main package should `os.Exit`. 

Callers should handle any kind of exit or fatal condition.

PR #1757, opened 2018-06-21
server: Avoid calling os.Exit() in BgpServer.Shutdown()

iwaseyusuke:
In case that GoBGP is used as Go native library, calling os.Exit() can cause unexpected termination including application which uses GoBGP.

This patch removes calls of os.Exit() derived from BgpServer.Shutdown().

fixes https://github.com/osrg/gobgp/issues/1745

### DIFF
--- a/gobgpd/main.go
+++ b/gobgpd/main.go
@@ -169,134 +169,110 @@ func main() {
 		go config.ReadConfigfileServe(opts.ConfigFile, opts.ConfigType, configCh)
 	}
 
-	var c *config.BgpConfigSet
-	for {
-		select {
-		case newConfig := <-configCh:
-			var added, deleted, updated []config.Neighbor
-			var addedPg, deletedPg, updatedPg []config.PeerGroup
-			var updatePolicy bool
+	loop := func() {
+		var c *config.BgpConfigSet
+		for {
+			select {
+			case <-sigCh:
+				apiServer.Shutdown(context.Background(), &api.ShutdownRequest{})
+				return
+			case newConfig := <-configCh:
+				var added, deleted, updated []config.Neighbor
+				var addedPg, deletedPg, updatedPg []config.PeerGroup
+				var updatePolicy bool
 
-			if c == nil {
-				c = newConfig
-				if _, err := apiServer.StartServer(context.Background(), &api.StartServerRequest{
-					Global: api.NewGlobalFromConfigStruct(&c.Global),
-				}); err != nil {
-					log.Fatalf("failed to set global config: %s", err)
-				}
-
-				if newConfig.Zebra.Config.Enabled {
-					tps := c.Zebra.Config.RedistributeRouteTypeList
-					l := make([]string, 0, len(tps))
-					for _, t := range tps {
-						l = append(l, string(t))
-					}
-					if _, err := apiServer.EnableZebra(context.Background(), &api.EnableZebraRequest{
-						Url:                  c.Zebra.Config.Url,
-						RouteTypes:           l,
-						Version:              uint32(c.Zebra.Config.Version),
-						NexthopTriggerEnable: c.Zebra.Config.NexthopTriggerEnable,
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -1415,13 +1402,19 @@ func (s *BgpServer) DeleteBmp(c *config.BmpServerConfig) error {
...
+			Communication: nil,
+		}
 		for _, p := range s.neighborMap {
 			p.fsm.adminStateCh <- stateOp
 		}
-		// the main goroutine waits for peers' goroutines to stop but if no peer is configured, needs to die immediately.
-		if len(s.neighborMap) == 0 {
-			os.Exit(0)
+		// Waits for all goroutines per peer to stop
+		for _, p := range s.neighborMap {
```
jeffbean: A suggestion maybe as a followup would be to have the `Peer` have a `Close` method that can be used to signal the shutdown or server `Stop`.

gobgpd/main.go
```diff
@@ -307,139 +283,170 @@ func main() {
...
 				}
 			}
-		case <-sigCh:
-			apiServer.Shutdown(context.Background(), &api.ShutdownRequest{})
 		}
 	}
+
+	loop()
+
+	os.Exit(0)
```
jeffbean: shouldn't need an Exit here if you modify the signal being captured to `os.Interrupt`

server/server.go
```diff
@@ -1414,20 +1414,31 @@ func (s *BgpServer) DeleteBmp(c *config.BmpServerConfig) error {
...
+		s.shutdownWG = new(sync.WaitGroup)
+		s.shutdownWG.Add(1)
+		stateOp := AdminStateOperation{
+			State:         ADMIN_STATE_DOWN,
+			Communication: nil,
+		}
 		for _, p := range s.neighborMap {
 			p.fsm.adminStateCh <- stateOp
 		}
-		// the main goroutine waits for peers' goroutines to stop but if no peer is configured, needs to die immediately.
```
jeffbean: We can just do a `s.shutdownWG.Wait()` here since there is already a `Stop` that does not wait on these idle neighbors. This also eliminates the need for another method to run after a Shutdown. 

Imagine the `Shutdown` call being the "graceful" handling of `server.Stop` + wait for neighbors to be idle.

iwaseyusuke: Yeah, I know `Shutdown()` and `Stop()` are very similar, and I tried to combine shutting down and waiting features into a single `Shutdown()` function. But as I mentioned as https://github.com/osrg/gobgp/pull/1757#discussion_r197689335, if we do `s.shutdownWG.Wait()` within `s.mgmtOperation()` wrapper function, it causes the deadlock of the main goroutine.

So, how about putting `s.shutdownWG.Wait()` in `Shutdown()` and the outside of `s.mgmtOperation()` wrapper like;
```go
func (s *BgpServer) Shutdown() {
	s.mgmtOperation(func() error {
		// ...(snip)...
	}, false)

	// Waits for all goroutines per peer to stop.
	// Note: This should not be wrapped with s.mgmtOperation() in order to
	// avoid the deadlock in the main goroutine of BgpServer.
	if s.shutdownWG != nil {
		s.shutdownWG.Wait()
		s.shutdownWG = nil
	}
}
```

jeffbean: Perfect. That makes this a lot easier to read through and keeps the waiting within the Shutdown function. 

Looks great!

iwaseyusuke: Thunks for your confirmation!

fujita:
Pushed, thanks guys!

vincentbernat:
The documentation does not say how to properly stop a GoBGP instance. Should we call `Stop()` then `Shutdown()`? Or is `Shutdown()` enough? The example could be updated with this information.

vincentbernat:
Shutdown() could be done to remove the listeners and Stop() could be deprecated then.

For the deadlock, I'll do that in the coming days.

PR #1761, opened 2018-06-22
api: Remove table/ usage from cmd/

fujimoto-s:


gobgp/cmd/mrt.go
```diff
@@ -45,7 +44,7 @@ func injectMrt() error {
 		return fmt.Errorf("Specified queue size is smaller than 1, refusing to run with unbounded memory usage")
```
jeffbean: For errors like these we can make variables at the top of the file to make these more readable and testable.

```
var (
    ...
    errQueueSizeTooSmall = errors.New("refusing to run with unbounded memory usage, specified queue size is smaller than 1")
)
```

gobgp/cmd/mrt.go
```diff
@@ -107,47 +106,65 @@ func injectMrt() error {
...
+								attrs = append(attrs, bgp.NewPathAttributeMpReachNLRI(nexthop, []bgp.AddrPrefixInterface{nlri}))
 							}
 						}
-						paths = append(paths, table.NewPath(source, nlri, false, attrs, t, false))
 					}
-				}
-				if mrtOpts.NextHop != nil {
-					for _, p := range paths {
-						p.SetNexthop(mrtOpts.NextHop)
+					path := &api.Path{
```
jeffbean: you're not using your NewPath method here? Maybe its not needed? or needs a better function signature

gobgp/lib/path.go
```diff
@@ -57,12 +57,13 @@ func serialize_path(rf C.int, input *C.char) *C.path {
 		return nil
 	}
 	path := C.new_path()
-	if nlri := p.GetNlri(); nlri != nil {
+	if nlri, _ := p.GetNativeNlri(); nlri != nil {
```
jeffbean: Again handle the error.

table/policy.go
```diff
@@ -3619,8 +3627,14 @@ func (r *RoutingPolicy) GetAllPolicy() []*config.PolicyDefinition {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
```
jeffbean: Don't defer - place the unlock immediately after reading the map. Line 3634 in this case.

table/policy.go
```diff
@@ -3453,6 +3454,13 @@ func (r *RoutingPolicy) GetDefinedSet(typ DefinedType, name string) (*config.Def
 	if !ok {
 		return nil, fmt.Errorf("invalid defined-set type: %d", typ)
 	}
+
```
jeffbean: Again - unlock after copying the values over to `dl`

api/policy.go
```diff
@@ -0,0 +1,150 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/osrg/gobgp/packet/bgp"
+)
+
+func ParseCommunityRegexp(arg string) (*regexp.Regexp, error) {
+	i, err := strconv.ParseUint(arg, 10, 32)
+	if err == nil {
```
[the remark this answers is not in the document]
fujimoto-s: Here we use the error to determine if `arg` is a simple numeric value.
So, errors which appear in this place only mean `arg` is not a simple numeric value(such as `65000:100`), it don't means something unfavorable happened.

gobgp/cmd/monitor.go
```diff
@@ -39,29 +39,39 @@ func makeMonitorRouteArgs(p *table.Path, showIdentifier bgp.BGPAddPathMode) []in
...
-	if n := p.GetNexthop(); n != nil {
-		nexthop = p.GetNexthop().String()
+	if n := getNextHopFromPathAttributes(attrs); n != nil {
+		nexthop = n.String()
 	}
 	pathStr = append(pathStr, nexthop)
 
 	// AS_PATH
-	pathStr = append(pathStr, p.GetAsString())
+	aspathstr := func() string {
```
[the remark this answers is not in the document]
fujimoto-s: Thanks, I'll make this to be a standalone function.

gobgp/cmd/monitor.go
```diff
@@ -39,29 +39,39 @@ func makeMonitorRouteArgs(p *table.Path, showIdentifier bgp.BGPAddPathMode) []in
...
 	// If Add-Path required, append Path Identifier.
-	nlri := p.GetNlri()
+	nlri, _ := p.GetNativeNlri()
 	if showIdentifier != bgp.BGP_ADD_PATH_NONE {
-		pathStr = append(pathStr, nlri.PathIdentifier())
+		pathStr = append(pathStr, p.GetIdentifier())
 	}
 	pathStr = append(pathStr, nlri)
 
+	attrs, _ := p.GetNativePathAttributes()
```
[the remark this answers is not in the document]
fujimoto-s: Thanks, I'll adds the error handling.

gobgp/cmd/mrt.go
fujimoto-s: Thanks, however it will leads a lot of fixes over the whole code.
I think it is better to do it in another PR.

gobgp/cmd/mrt.go
fujimoto-s: Thanks, I'll use api.NewPath() in this place.

gobgp/lib/path.go
fujimoto-s: Thanks, I'll adds the error handling.

table/policy.go
fujimoto-s: Thanks, I'll unlock it after reading it.

table/policy.go
fujimoto-s: Thanks, I'll unlock it after reading it.

fujimoto-s:
> Overall its strange to be returning regexp objects in the api/policy.go file. These methods seem to be something only really needed to parse strings that should be done inside the library and exposed or returned as values of a struct.

As you said, `Parse*CommunityRegexp()` returns regexp.
By the way, these functions are used by table package, too. And the package has the struct which holds regexp structs.

https://github.com/osrg/gobgp/blob/5fbd0cb24324f1569377ce5c627bb2db97b76ce7/table/policy.go#L981

Also, api package uses these functions, but it is only interested whether the string can be properly parsed, and the regexp struct is discarded.

Then, I think that returning regexp is not a bad performance. What do you think?

Anyway, I think changing returning structs should be done in another PR.

fujimoto-s:
Thanks for reviews! I updated my PR.
Could you check this?

jeffbean:
Looks good to me.

As for the regex objects I'm not familiar enough with the library to have an opinion on this. I would say its not something we should have in the future but I'm not sure how to refactor it yet.

I would say its good for now and we just make sure to address the regex in some future work.

PR #1762, opened 2018-06-22
Adding travis config for Codecov 

jeffbean:
Codecov will need to be given access to the repo. This will help understand the state of the codebase over time.

The example report and system can be found here: https://codecov.io/gh/jeffbean/gobgp

Also adding `megacheck` to help cleanup and ensure diffs and code stay clean. It fails with many errors but should not block PRs until we fix the current issues (a followup PR).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ _unittest: &_unittest
   script:
     - go test $(go list ./... | grep -v '/vendor/') -timeout 120s
     - if [ "$(go env GOARCH)" = "amd64" ]; then go test -race github.com/osrg/gobgp/packet/bgp -run ^Test_RaceCondition$; else echo 'skip'; fi
-    - go build -o ./gobgp/gobgp   ./gobgp/
-    - go build -o ./gobgpd/gobgpd ./gobgpd/
 
 _build: &_build
   <<: *_dep_ensure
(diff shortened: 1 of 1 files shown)


fujita:
Can we keep 'go vet' and 'go fmt' until we run megacheck properly?

jeffbean:
> Can we keep 'go vet' and 'go fmt' until we run megacheck properly?

Yes. They are sill run here and I will update the description to reflect this.

fujita:
megacheck does what vet and fmt do (and more), right?

jeffbean:
> megacheck does what vet and fmt do (and more), right?

Great question. So `megacheck` is a tool that runs a number of static analysis rules against the source code. It will find things like unused variables or simplification of some codepaths. `vet` and `fmt` are more linting the code for syntax mistakes and literal formatting of the files. In this change I just run all three of these tools in one job since there is no reason to break them into seperate jobs (it take longer to build)

In reality this is a helper command that just runs a set of these tools all in one: https://github.com/dominikh/go-tools 

In a later diff we can add `structlayout-optimize` that will optimize structs for the least amount of padding (seems something this project wants)

jeffbean:
depends on #1767

fujita:
Pushed, thanks. Yeah, running all three makes sense to me.

> In a later diff we can add structlayout-optimize that will optimize structs for the least amount of padding (seems something this project wants)

I occasionally run structlayout-optimize for table.Path and table.Destination. Running it on travis is fine by me.

PR #1766, opened 2018-06-26
Zapi version 5

amanshaikh75:
This branch implements Zebra API version 5. Using this branch I have been able to add required routes to kernel for end-to-end forwarding in an BGP/MPLS L3-VPN scenario.


fujita:
As I said before, I'm not interested in adding another one the supported list unless dropping one in the list.

PR #1767, opened 2018-06-27
Fixing staticcheck errors from linux files

jeffbean:
Missed some of the static check errors from the linux specific files. 

```
server/sockopt_linux.go:69:2: should check returned error before deferring fi.Close() (SA5001)
server/sockopt_linux.go:88:2: should check returned error before deferring fi.Close() (SA5001)
server/sockopt_linux.go:97:2: should check returned error before deferring fi.Close() (SA5001)
server/sockopt_linux.go:116:2: should check returned error before deferring fi.Close() (SA5001)
server/sockopt_linux_test.go:45:5: should use bytes.Equal(buf1, buf2) instead (S1004)
server/sockopt_linux_test.go:65:5: should use bytes.Equal(buf1, buf2) instead (S1004)                   
server/sockopt_linux.go:35:2: field pad1 is unused (U1000)
server/sockopt_linux.go:37:2: field pad2 is unused (U1000)                                     
```


fujita:
pushed, thanks!

PR #1769, opened 2018-06-28
Exclude bgp.RF_RTC_UC from address farmily in soft reset in and out.

kishiguro:
When RTC and AddPath both are enabled.   Issuing softreset command to Route Reflector with many Client cause route update storm.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1762,6 +1762,16 @@ func (s *BgpServer) Stop() error {
 	}, true)
 }
 
+func filterFamilyList(families []bgp.RouteFamily, f bgp.RouteFamily) []bgp.RouteFamily {
+	l := make([]bgp.RouteFamily, 0, len(families))
+	for _, family := range families {
+		if family != f {
+			l = append(l, family)
+		}
+	}
+	return l
+}
+
 func (s *BgpServer) softResetIn(addr string, family bgp.RouteFamily) error {
 	peers, err := s.addrToPeers(addr)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -1762,6 +1762,16 @@ func (s *BgpServer) Stop() error {
 	}, true)
 }
 
+func filterFamilyList(families []bgp.RouteFamily, f bgp.RouteFamily) []bgp.RouteFamily {
```
jeffbean: please provide tests for this function.

kishiguro: sure.

fujita:
Hmm, with this change, it's impossible to send RF_RTC_UC address farmily in any cases? You can specify a family for softreset/out. Why it doesn't work?

kishiguro:
This is for suppressing "softreset" update storm when  both RTC and AddPath are enabled
When RTC is configured, RR will send RTC Membership update to it's client when "softreset" command is issued.  Client resend all of the routes when it receives RTC Membership update.  Since AddPath is enabled, the update are propagated to all of other peers.
Thus there are tremendous number of update will be generated especially when we have 1000 RR clients.
I think we can exclude RF_RTC_UC by default, but when it is explicitly specified, we can use it.

fujita:
Excluding RF_RTC_UC by default (when no family specified) is fine by me, however, your patch always excludes RF_RTC_UC even when RF_RTC_UC specified?

kishiguro:
Patch has beed updated to handle RF_RTC_UC is specified case.

fujita:
Would be better to test RF_RTC_UC specified case and remove the duplicated code?
How about the following?
https://github.com/osrg/gobgp/pull/1770

kishiguro:
It would be better.  Let me close this PR.

PR #1770, opened 2018-06-29
server: exclude RTC for softreset if not specified

fujita:


kishiguro:
Thanks.  LGFM.

fujita:
@kishiguro thanks, pushed.

Issue #1765, opened 2018-06-26
problems abount graceful-restart


qsqx:
1. fsm.pConf.GracefulRestart.State.PeerRestarting keeps true after restart time expired.
```
if fsm.pConf.GracefulRestart.State.PeerRestarting {
    log.WithFields(log.Fields{
        "Topic": "Peer",
        "Key":   fsm.pConf.State.NeighborAddress,
        "State": fsm.state.String(),
    }).Warn("graceful restart timer expired")
    return bgp.BGP_FSM_IDLE, NewFsmStateReason(FSM_RESTART_TIMER_EXPIRED, nil, nil)
}
```
2. gobgp server never connect to peers which are concidered being restart even if the peers are in passive mode.
```
if fsm.state == bgp.BGP_FSM_ACTIVE && !fsm.pConf.GracefulRestart.State.PeerRestarting {
    go connect()
}
```
3. neighbors objects in same peer group shares same AfiSafis array becuse of the code
```   
if !v.IsSet("neighbor.afi-safis") {
    c.AfiSafis = pg.AfiSafis
}
```
and as the moment one peer's peer.fsm.pConf.AfiSafis[i].MpGracefulRestart.State.EndOfRibReceived set true,  the attribute of all peers which has the same peer group become true, and this will cause 
```
allEnd := func() bool {
    for _, p := range server.neighborMap {
        if !p.recvedAllEOR() {
            return false
        }
    }
    return true
}()
```
return wrong results and affect the behavior of bgp server while restarting.

fujita:
Thanks a lot for the report! Can you send pull requests for these problems?

PR #1771, opened 2018-06-29
fsm: Connect to bgp peers regardless of whether the peers are restarting

qsqx:
https://github.com/osrg/gobgp/issues/1765

### DIFF
--- a/server/fsm.go
+++ b/server/fsm.go
@@ -466,7 +466,7 @@ func (fsm *FSM) connectLoop() error {
 			}).Debug("stop connect loop")
 			return nil
 		case <-timer.C:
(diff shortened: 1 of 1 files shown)


server/fsm.go
```diff
@@ -466,7 +466,7 @@ func (fsm *FSM) connectLoop() error {
 			}).Debug("stop connect loop")
 			return nil
 		case <-timer.C:
-			if fsm.state == bgp.BGP_FSM_ACTIVE && !fsm.pConf.GracefulRestart.State.PeerRestarting {
+			if fsm.state == bgp.BGP_FSM_ACTIVE {
```
fujita: Also needs to remove the same condition at L.455?

fujita:
pushed with the above additional change. Thanks!

PR #1773, opened 2018-06-29
config: Copy afi-safi objects from peer group to neighbor instead of …

qsqx:
…share the same array
https://github.com/osrg/gobgp/issues/1765

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -496,7 +496,9 @@ func OverwriteNeighborConfigWithPeerGroup(c *Neighbor, pg *PeerGroup) error {
 	overwriteConfig(&c.TtlSecurity.Config, &pg.TtlSecurity.Config, "neighbor.ttl-security.config", v)
 
 	if !v.IsSet("neighbor.afi-safis") {
-		c.AfiSafis = pg.AfiSafis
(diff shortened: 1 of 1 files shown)


config/default.go
```diff
@@ -496,7 +496,9 @@ func OverwriteNeighborConfigWithPeerGroup(c *Neighbor, pg *PeerGroup) error {
 	overwriteConfig(&c.TtlSecurity.Config, &pg.TtlSecurity.Config, "neighbor.ttl-security.config", v)
 
 	if !v.IsSet("neighbor.afi-safis") {
-		c.AfiSafis = pg.AfiSafis
+		for _, afiSafi := range pg.AfiSafis {
+			c.AfiSafis = append(c.AfiSafis, afiSafi)
+		}
```
fujita: Each peer has own slice but still shares State object so one peer modifies a State object then it affects the rest of peers?

qsqx: yes, an example is

qsqx: AfiSafi.State is not a pointer but a value and modification to one won't affect others.

fujita: oops. pushed, thanks.

PR #1774, opened 2018-07-08
Cleanup api and layout

fujita:
- api/ has only code generated by grpc with the IDL files (no package dependency except for grpc).
- follow the standard golang layout (you can see clearly what are private and public library code).


vincentbernat:
I have some concerns over this change. Go forbids use of an internal package. However, both `internal/pkg/config` and `internal/pkg/table` are needed when embedding GoBGP. To start a server, we need a config. To add a neighbor, we need a config. The table package is needed for the NewPath method in it. I can propose a PR to move these two packages out of internal.

fujita:
@vincentbernat thanks a lot for the comment!

As you said, this pr was incomplete but the issue was fixed. The following example gives some APIs for embedding GoBGP:
https://github.com/osrg/gobgp/blob/master/docs/sources/lib.md

Only API to be fixed is 'Watch' feature. I'm still investigating how 'Watch' API should be. Any input are appreciated!

vincentbernat:
Thanks for the pointer, that's helpful! I find this change globally positive as it's more obvious how the API should be used and what part of the code we don't have to read.

Previously, I was using structures from `packet/bgp` and functions from `table` to build attributes and paths. From your example, I now have to use a different way. However, this means using marshalling functions from protobuf. I understand this makes sense because the embedded BGP server works just like the remote one (through gRPC). The code however becomes more verbose. I didn't do the conversion in my project yet. I'll do that first and come back with some examples.

vincentbernat:
I have almost converted my code to use the exposed API only. Not having type-safety (as anything needs to be marshalled to any.Any) and having to handle protobuf stuff is a bit bothersome. However, there are some pieces I don't know how to do concisely. Notably, when parsing the result of `ListPath`, how to easily parse the result? There are large chunks of code for this purpose in `internal/pkg/apiutil` for this purpose.

vincentbernat:
I am interested in `AnyNlri` and `AnyPattrs`. The `cmd` package heavily relies on `apiutil` to parse them. In my case, even a string representation would help (as I currently only need them in tests).

fujita:
@vincentbernat sorry about the delay,
https://gist.github.com/fujita/34d2f27666c2ebf2533b815cfb036b27
You can use Any's String() method.

```
fujitanombp:gobgp fujita$ go run a.go 
type_url:"type.googleapis.com/gobgpapi.OriginAttribute" 
type_url:"type.googleapis.com/gobgpapi.NextHopAttribute" value:"\n\01010.0.0.1" 
```

vincentbernat:
The result is already difficult to read for simple routes, but when handling more complex routes like EVPN routes, I get stuff like this:

````
age:1537801812 best:true validation_detail:<> family:<afi:AFI_L2VPN safi:SAFI_EVPN > source_asn:100 source_id:\"1.1.1.1\" neighbor_ip:\"127.0.0.1\" local_identifier:1 any_nlri:<type_url:\"type.googleapis.com/gobgpapi.EVPNMACIPAdvertisementRoute\" value:\"\\nG\\n8type.googleapis.com/gobgpapi.RouteDistinguisherIPAddress\\022\\013\\n\\0071.1.1.1\\020\\004\\022\\013\\022\\t\\000\\000\\000\\000\\000\\000\\000\\000\\000\\\"\\0210a:00:27:00:00:00*\\005<nil>2\\004\\302\\303\\223\\007\" > any_pattrs:<type_url:\"type.googleapis.com/gobgpapi.OriginAttribute\" > any_pattrs:<type_url:\"type.googleapis.com/gobgpapi.AsPathAttribute\" > any_pattrs:<type_url:\"type.googleapis.com/gobgpapi.LocalPrefAttribute\" value:\"\\010d\" > any_pattrs:<type_url:\"type.googleapis.com/gobgpapi.ExtendedCommunitiesAttribute\" value:\"\\n0\\n*type.googleapis.com/gobgpapi.EncapExtended\\022\\002\\010\\010\\nG\\n7type.googleapis.com/gobgpapi.TwoOctetAsSpecificExtended\\022\\014\\010\\001\\020\\002\\030d \\302\\303\\223\\207\\001\" > any_pattrs:<type_url:\"type.googleapis.com/gobgpapi.MpReachNLRIAttribute\" value:\"\\n\\004\\010\\031\\020F\\022\\0071.1.1.1\\032\\262\\001\\n8type.googleapis.com/gobgpapi.EVPNMACIPAdvertisementRoute\\022v\\nG\\n8type.googleapis.com/gobgpapi.RouteDistinguisherIPAddress\\022\\013\\n\\0071.1.1.1\\020\\004\\022\\013\\022\\t\\000\\000\\000\\000\\000\\000\\000\\000\\000\\\"\\0210a:00:27:00:00:00*\\005<nil>2\\004\\302\\303\\223\\007\" >
```

Which is quite impractical to use. Even looking at it as a human, I can't know if this is correct.

Looking at the `cmd` package, there is a lot of use of internal packages (notably `apiutil` and `table`). This should hint the public API is too limited to query gobgpd from a third-party client. It's not a very constructive rant on my side but I don't have a concrete solution to propose for this problem. In my opinion, the protobuf stuff should be hidden from the user as it is not a practical representation and it's not type-safe. Of course, that would be a huge work and another API change.

fujita:
> In my opinion, the protobuf stuff should be hidden from the user as it is not a practical representation and it's not type-safe. Of course, that would be a huge work and another API change.

I'm not sure whats kinda of representation you propose. Can you show a prototype of your proposal?

vincentbernat:
For example, in the `table` package, we have `PathAttrs()`. This enables to do:

```go
        mpreach := packet.NewPathAttributeMpReachNLRI(
                c.config.IP.String(), []packet.AddrPrefixInterface{nlri})
        extcomms := []packet.ExtendedCommunityInterface{
                packet.NewEncapExtended(packet.TUNNEL_TYPE_VXLAN),
                packet.NewTwoOctetAsSpecificExtended(
                        packet.EC_SUBTYPE_ROUTE_TARGET,
                        c.config.ASN,
                        vni|evpnAutoRTVXLAN,
                        true),
        }
        attrs := table.PathAttrs(
                []packet.PathAttributeInterface{
                        packet.NewPathAttributeOrigin(0),
                        packet.NewPathAttributeLocalPref(100),
                        packet.NewPathAttributeExtendedCommunities(extcomms),
                        mpreach,
                })
        sort.Sort(attrs)
        return table.NewPath(nil, nlri, false, attrs, time.Now(), false)
```

While now, I have to do:

```go
	e1, err := ptypes.MarshalAny(&gobgpapi.EncapExtended{
		TunnelType: uint32(gobgppacket.TUNNEL_TYPE_VXLAN),
	})
	if err != nil {
		panic(err)
	}
	e2, err := ptypes.MarshalAny(&gobgpapi.TwoOctetAsSpecificExtended{
		IsTransitive: true,
		SubType:      uint32(gobgppacket.EC_SUBTYPE_ROUTE_TARGET),
		As:           uint32(c.config.ASN),
		LocalAdmin:   vni | evpnAutoRTVXLAN,
	})
	if err != nil {
		panic(err)
	}
	extcomms := []*any.Any{e1, e2}
	a0, err := ptypes.MarshalAny(&gobgpapi.NextHopAttribute{
		NextHop: net.IP(c.config.IP).String(),
	})
	if err != nil {
		panic(err)
	}
	a1, err := ptypes.MarshalAny(&gobgpapi.OriginAttribute{Origin: 0})
	if err != nil {
		panic(err)
	}
	a2, err := ptypes.MarshalAny(&gobgpapi.LocalPrefAttribute{LocalPref: 100})
	if err != nil {
		panic(err)
	}
	a3, err := ptypes.MarshalAny(&gobgpapi.ExtendedCommunitiesAttribute{
		Communities: extcomms,
	})
	if err != nil {
		panic(err)
	}
	pattrs := []*any.Any{a0, a1, a2, a3}
	return &gobgpapi.Path{
		Family: &gobgpapi.Family{
			Afi:  gobgpapi.Family_AFI_L2VPN,
			Safi: gobgpapi.Family_SAFI_EVPN},
		AnyNlri:   nlri,
		AnyPattrs: pattrs,
	}
```

This is far more verbose and I loose most type-safety (and in case of error, I'll have a panic at runtime). The API in `packet` and `table` is more convenient. However, I cannot use `table` since it's private and I cannot use `packet` as I have to marshal nested struct to protobuf first.

Note however, the direct initialization of structs instead of using `NewX` constructors doesn't matter to me.

Ideally, structures in `gobgpapi` shouldn't expose protobuf (but use Go native structures instead). But I understand this would be a very extensive and difficult change as `gobgpapi` package is autogenerated. If you look at the `cmd` package, there is almost no use of `any.Any` but the code extensively uses private packages (notably `apiutil`). You could unhide these packages, but I understand this will make the API a bit more confusing since it would essentially reverse to the previous situation.

However, as for my use, I have almost converted my code to the new API (I am still struggling with route inspection, but I'll eventually bite the bullet and copy/paste part of apiutils). So, I don't want to pressure you on any direction, I'll adapt my code.

PR #1775, opened 2018-07-09
cmd: implement ebgp-multihop-ttl option when add neighbor

chenhaiq:
usage:
gobgp neighbor add 192.168.1.2 as 65001 ebgp-multihop-ttl 100


fujita:
Pushed, thanks! I don't fancy adding an option to the CLI since I don't want the CLI to end up with all the crazy bgp options supported though.

PR #1776, opened 2018-07-09
cmd: fix error when delete policy statement

chenhaiq:
fixed bug: `gobgp policy statement del statement1` has no effect


fujita:
nice catch, pushed, thanks!

PR #1778, opened 2018-07-10
fix nil pointer error when create pmsitunnel

chenhaiq:
fix error:
# gobgp global rib -a evpn add multicast 172.31.9.12 etag 10 rd 1:10 rt 1:1000 encap vxlan nexthop 172.31.9.12 pmsi ingress-repl 10 172.31.9.12
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6190bf]

goroutine 1 [running]:
github.com/osrg/gobgp/packet/bgp.NewPathAttributePmsiTunnel(0x1010000, 0x0, 0x0, 0x0)


fujita:
Thanks for the report. I prefer to fix ParsePmsiTunnel() instead because we try to avoid mutating the existing object in such way.

https://github.com/fujita/gobgp/commit/662e2a0682761ec2d8340a53e46ce4186d392191

fujita:
fixed by https://github.com/osrg/gobgp/commit/3dc23d3075f13f2976afd255ef2fd9b1410acc77

PR #1779, opened 2018-07-10
enable ebgp-multihop if configured

tamihiro:
- neighbors.ebgp-multihop.config configuration is ignored without this patch, meaning ebgp multihop peer can't be established because ttl is set to 1.
- this bug seems to have crept in since commit [0c334f5](https://github.com/osrg/gobgp/commit/0c334f5b3f8d330ced71649316c894119b2432d3).


fujita:
pushed, thanks!

PR #1784, opened 2018-07-14
fix goreleaser to build static binaries

fujita:


fujita:
pushed.

Issue #1788, opened 2018-07-16
graceful restart: rtc paths are not send after deferral time expired


qsqx:
The problem was appeared after commit https://github.com/osrg/gobgp/commit/db6683bfbfee3a6e04e78d6af40bbdb9f4fcd317 

RTC paths should still be send after deferral time expired in a graceful restarting process ?

qsqx:
It's a better choice. Thanks a lot.

fujita:
thanks!

PR #1789, opened 2018-07-16
not exclude RTC for softreset during local restarting

qsqx:
https://github.com/osrg/gobgp/issues/1788


fujita:
fixed by https://github.com/osrg/gobgp/commit/c3e41d89620a4dafc53f22732f5eef054c01b39c

PR #1792, opened 2018-07-18
packet/bgp: Support SR TE Policy

caozhenglei:
This patch adds support for the Segment Routing Policy NLRI encoding and
the path attributes encoding defined in
"draft-ietf-idr-segment-routing-te-policy-01".


iwaseyusuke:
@caozhenglei Hi, thanks for having interest in my patch! But sorry I think this patch should be re-designed and fixed the CLI commands format (to support multiple segment lists for example).

fujita:
@caozhenglei can you split the patch? How about pushing the changes to packet/bgp first? Also you need to rebase against the current master. Please create a new pull request.

PR #1793, opened 2018-07-19
fix path to lib directory

tamihiro:
correct the path according to newly directory structure.


fujita:
pushed, thanks!
Do you use this feature? We added this because C easily can't build binary format of path attributes for the API but now we have protobuf structures for all the path attributes. So I guess that we can remove this.

Issue #1804, opened 2018-08-15
deletePath not sending withdraw when connected to RR


astone282:
Hello, 

Doing some experimentation. I have gobgp configured (54.54.54.54) to talk to a vendor route reflector (35.121.146.13). Using the gobgp CLI or Java GRPC, when I addPath (ipv4), goBgp correctly sends the route and it is received on the RR.

The RR then echo's back the route with the RR cluster ID back to gobgp. I can see this route come in via monitor and also in the global rib dump.

When I try to deletePath, via CLI or Java GRPC, i do NOT see the [DELROUTE] log message. The entry is deleted from the gobgp but a withdraw message is not received on the RR. Therefore the route is not withdrawn.

**Note: if i disable route-reflector on the other node, gobgp can successful add and withdraw the route. My guess is gobgp is not withdrawing the route it generated due to a collision with the reflected route ** 

Here's a step through of what's happening:

```

# Add the route in gobgp

# ./gobgp global rib -a ipv4 add 5.5.5.5/24 nexthop 10.10.10.10

```

```
# gobgp monitor outputs the new route :

[ROUTE] 5.5.5.0/24 via 10.10.10.10 aspath [] attrs [{Origin: ?}]
```

```
# RR receives the update

116 2018/08/15 17:50:13.443 UTC MINOR: DEBUG #2001 Base Peer 1: 172.10.54.2
"Peer 1: 172.10.54.2: UPDATE
Peer 1: 172.10.54.2 - Received BGP UPDATE:
    Withdrawn Length = 0
    Total Path Attr Length = 21
    Flag: 0x40 Type: 1 Len: 1 Origin: 2
    Flag: 0x40 Type: 2 Len: 0 AS Path:
    Flag: 0x40 Type: 3 Len: 4 Nexthop: 10.10.10.10
    Flag: 0x40 Type: 5 Len: 4 Local Preference: 100
    NLRI: Length = 4
        5.5.5.0/24
"

```

```
# RR sends a bgp update with the cluster is

117 2018/08/15 17:50:36.951 UTC MINOR: DEBUG #2001 Base Peer 1: 172.10.54.2
"Peer 1: 172.10.54.2: UPDATE
Peer 1: 172.10.54.2 - Send BGP UPDATE:
    Withdrawn Length = 0
    Total Path Attr Length = 35
    Flag: 0x40 Type: 1 Len: 1 Origin: 2
    Flag: 0x40 Type: 2 Len: 0 AS Path:
[...]

PR #1807, opened 2018-08-16
pkg/server: handleUpdate() should not inject echo route from RR

yicwang:
#1804 Draft code to fix the issue...

### DIFF
--- a/pkg/server/peer.go
+++ b/pkg/server/peer.go
@@ -491,6 +491,24 @@ func (peer *Peer) handleUpdate(e *FsmMsg) ([]*table.Path, []bgp.RouteFamily, *bg
 					continue
 				}
 			}
+			// RFC4456 8. Avoiding Routing Information Loops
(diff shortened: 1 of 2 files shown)


pkg/server/peer.go
```diff
@@ -491,6 +491,23 @@ func (peer *Peer) handleUpdate(e *FsmMsg) ([]*table.Path, []bgp.RouteFamily, *bg
 					continue
 				}
 			}
+			// RFC4456 8. Avoiding Routing Information Loops
+			// A router that recognizes the ORIGINATOR_ID attribute SHOULD
+			// ignore a route received with its BGP Identifier as the ORIGINATOR_ID.
+			if peer.isIBGPPeer() && !path.IsLocal() {
```
fujita: peer.isIBGPPeer() accesses to peer.fsm so peer.fsm.lock.RLock() must be called before that.

fujita: A local path isn't passed to handleUpdate() so IsLocal() is not necessary.

PR #1808, opened 2018-08-17
pkg/packet/bgp: Fix NewEVPNMacIPAdvertisementRoute() to use correct MAC length

yicwang:
According to the implementation, the MacAddressLength seems to be in bit, not byte. So fixing it...


fujita:
pushed, thanks!

PR #1813, opened 2018-08-21
scenario_test: Make BusyBox based container available

7310510:
This PR enable us to use container images which contain BusyBox for scenario test.

When we want to do scenario test with our own container image, for example which is based on alpine, we can specify own image like `GOBGP_IMAGE=my/image:tag ./run_all_tests.sh`.


fujita:
Looks fine. Pushed, thanks.

PR #1819, opened 2018-08-27
api: change Family

### LINKED ISSUE #1812: Inconsistent type for Family in api
The [Family type and constants][1] and [this value map][2] are int32 but many of the api structs, [such as AfiSafiConfig][3] require uint32. The result is an unexpected compile error like this:

```
root@lab1r1u05:# go build
./main.go:129:13: cannot use gobgpapi.Family_IPv6 (type gobgpapi.Family) as type uint32 in field value
```

I can cast it but that doesn't seem right. Why not use a consistent type?

[1]: https://github.com/osrg/gobgp/blob/d6a5e9f1572aff35d8ab47c896f71c06871c2b15/api/gobgp.pb.go#L271
[2]: https://github.com/osrg/gobgp/blob/d6a5e9f1572aff35d8ab47c896f71c06871c2b15/api/gobgp.pb.go#L322
[3]: https://github.com/osrg/gobgp/blob/d6a5e9f1572aff35d8ab47c896f71c06871c2b15/api/gobgp.pb.go#L4069

fujita:
fixes #1812

Avoid the cast and use strightforward data strcuture; two variables
for afi and safi instead of squashing two into one variable.


fujita:
@ecbaldwin thanks a lot for reviewing! As you pointed out, there should be converted too. I'll fix them and push this.

fujita:
squashed and pushed.

PR #1820, opened 2018-08-27
scenario_test: Make static ip address available

7310510:
This RP enable us to use static IP address in scenario test.

Basically, GoBGP's container based scenario test is executed with dynamic IP address allocation by docker. In some case, for example when we use static configuration file, static IP address is help us to keep a consistency of our test.

Usage:
```
br01 = Bridge(name='br01', subnet='192.168.0.0/24', self_ip=False)
g1 = GoBGPContainer(name='g1', asn=64512 router_id='10.0.0.1')
br01.addif(g1, ip_addr='192.168.0.11')
```

P.S. I'm going to send other PR that include a change for static configuration.


fujita:
Looks fine. Pushed, thanks.

Issue #1818, opened 2018-08-24
gobgpd hangs when adding many paths with graceful restart enabled


ecbaldwin:
The switch I'm working with can handle 512k host routes in its fib and so I was trying to push that many with gobgp. Without GR enabled, I can easily push this many routes. However, when I started using GR or LLGR, gobgpd would hang.

I tried three methods:

1. AddPathStream with many paths per AddPathStreamRequest.
1. AddPathStream with one Path per AddPathStreamRequest.
1. AddPath in a loop.

I observed the first one hanging before it hit 100 routes. The second seems to hang after pushing around 700 - 800 routes to the peer. The last one actually gets pretty far. I've seen it get over 300,000 before hanging. Of course, it is about 3 times slower that the other methods.

When gobgpd is hung, it does not respond to API requests nor a SIGTERM. A SIGINT will kill it.

Below is the script I used to generate the routes. You'll see all three methods in the add_paths function. Just some of them are commented out.

```
#!/root/python-gobgp/venv/bin/python

import attribute_pb2
import gobgp_pb2_grpc
import gobgp_pb2
from google.protobuf import any_pb2
import sys

import grpc
import netaddr
from grpc.framework.interfaces.face.face import ExpirationError

_TIMEOUT_SECONDS = 1


def wrap_any(thing):
    a = any_pb2.Any()
    a.Pack(thing)
    return a


def create_ipv4_path(prefix, prefix_len, next_hop):
    nlri = attribute_pb2.IPAddressPrefix(prefix=prefix, prefix_len=int(prefix_len))
    origin_attr = attribute_pb2.OriginAttribute(origin=0)
    next_hop_attr = attribute_pb2.NextHopAttribute(next_hop=next_hop)

    return gobgp_pb2.Path(family=gobgp_pb2.IPv4,
                          any_nlri=wrap_any(nlri),
                          any_pattrs=[wrap_any(origin_attr),
                                      wrap_any(next_hop_attr)])

def create_ipv6_path(prefix, prefix_len, next_hop):
    nlri = attribute_pb2.IPAddressPrefix(prefix=prefix, prefix_len=int(prefix_len))
    origin_attr = attribute_pb2.OriginAttribute(origin=0)
    mp_reach_nlri_attr = attribute_pb2.MpReachNLRIAttribute(family=gobgp_pb2.IPv6,
                                                            next_hops=[next_hop],
                                                            nlris=[wrap_any(nlri)])

[...]

Issue #1760, opened 2018-06-22
Proposal - Start a v2/dev branch to do refactoring, maybe API breaking


jeffbean:
After a few weeks of using, implementing, testing, and running in production it seems there is some work that is needed to be done to harden this library. There are also issues capturing the desire to enhance Logging and add Metrics.

I propose we fixup and add some core features that will allow this library to be run in large scale production environments with little operational burden.

There are a few options around how we might be able collaborate in the community.
1. We could make a dev, v2 or \<insert name here\> branch where folks could contribute and make changes together in this repository. This has a nice property of getting help and review from the community.
2. Another avenue is for myself to fork and work separately to then later propose a PR that would potentially have API breaking changes. This has a large con where work would be isolated until the end and could lead to a near impossible PR to merge from such a huge set of changes. Also I am no expert in BGP so it would probably not be a great outcome.
3. There is also the path of just doing refactoring over time in the current codebase as is just fixing the issues in the `master` branch. We can just make issues with all of the designs in the issues. This would be a nice option in terms of validation of changes during the process, but is also the hardest in terms of execution. This would also cost the most in terms of time.

---
Here are the areas that seem the most important, but not limited to:
- Restructure all the go routines with `context` and ensure the Server is not leaking and racy
	- Issue that captures this most: #1331
	- There are many major data races and go routine leaks
	- No more `time.Sleep` - this is a smell for both less than deal error and goroutine patterns
	- No more `GOTO` - its almost never needed in practice
- Refactor error usage pattern
	- without errors the library is harder to refactor since we need to know state throughout the server
	- Its hard to reason around the codebase when we have methods called "validate" that return nothing and only log messages.
- Logging and metrics get passed to call sites as needed
	- makes using and testing this library difficult, especially around production level observability
	- ties into refactoring the error patterns
- Stricter style guidelines - Comments on all public methods, etc...
	- The general style guides should be followed to be consistent with the Go ecosystem
- Add Makefile infrastructure
	- Providing easier and predictable builds, tests, lint, fmt, vet, etc...
- Add Full Docker testing suite.
	- we should be able to run everything (especially python) inside containers to make this whole testing thing easier.
- Restructure some files and structs - general cleanup
	- It seems memory is a concern we can ensure this is addressed
- More Tests!
	- Some of the most important code paths are left untested in Go.
	- Benchmark tests can also help maintain a level of performance guidelines.
- More Docs - always helps 
	- Collaboration Guide
---

[...]

PR #1821, opened 2018-08-27
Release read lock before calling llgrFamilies

ecbaldwin:
llgrFamilies attempts to obtain a read lock on its own but the caller already
has one. This results in deadlock.

Addresses #1818

### DIFF
--- a/pkg/server/peer.go
+++ b/pkg/server/peer.go
@@ -282,8 +282,9 @@ func (peer *Peer) llgrFamilies() ([]bgp.RouteFamily, []bgp.RouteFamily) {
 
 func (peer *Peer) isLLGREnabledFamily(family bgp.RouteFamily) bool {
 	peer.fsm.lock.RLock()
-	defer peer.fsm.lock.RUnlock()
-	if !peer.fsm.pConf.GracefulRestart.Config.LongLivedEnabled {
+	llgrEnabled := peer.fsm.pConf.GracefulRestart.Config.LongLivedEnabled
+	peer.fsm.lock.RUnlock()
+	if !llgrEnabled {
 		return false
 	}
 	fs, _ := peer.llgrFamilies()


pkg/server/peer.go
```diff
@@ -282,10 +282,10 @@ func (peer *Peer) llgrFamilies() ([]bgp.RouteFamily, []bgp.RouteFamily) {
 
 func (peer *Peer) isLLGREnabledFamily(family bgp.RouteFamily) bool {
 	peer.fsm.lock.RLock()
-	defer peer.fsm.lock.RUnlock()
 	if !peer.fsm.pConf.GracefulRestart.Config.LongLivedEnabled {
 		return false
 	}
+	peer.fsm.lock.RUnlock()
```
armando-migliaccio: probably it’s best to split the function and have `llgrFamilies` called as a start of a new function  and put the defer back. That way you don't end up keeping the lock indefinitely should you return at line 286. Have you thought of that?

armando-migliaccio: I guess splitting the function means you’ have to acquire the lock twice and if the entire code is mean to be atomic, then probably there's room for a more elaborated fix.

ecbaldwin: Good catch. Yes, a return on 286 would be bad. Let me go split the function and we'll see what it looks like.

pkg/server/peer.go
```diff
@@ -282,8 +282,9 @@ func (peer *Peer) llgrFamilies() ([]bgp.RouteFamily, []bgp.RouteFamily) {
 
 func (peer *Peer) isLLGREnabledFamily(family bgp.RouteFamily) bool {
 	peer.fsm.lock.RLock()
-	defer peer.fsm.lock.RUnlock()
-	if !peer.fsm.pConf.GracefulRestart.Config.LongLivedEnabled {
+	llgrEnabled := peer.fsm.pConf.GracefulRestart.Config.LongLivedEnabled
+	peer.fsm.lock.RUnlock()
```
ecbaldwin: There is significant precedent for this RLock/read config/RUnlock pattern in gobgp. This avoids the issue but leaves us open to inconsistent configuration because the two locks could read different states. Unfortunately, there is some precedent for that also in many other places in the code.

I think a better way to handle this would be to rearchitect the sharing of the fsm, fsm.pConf, and whatever else is under synchronization with this lock. I think that a single goroutine should be responsible for managing updates to fsm and should push copies of it to other goroutines through channels.

My observation is that updates to the fsm do not happen that frequently in the grand scheme of things. Also, I don't see any harm in another goroutine getting an update to the fsm a little bit late if it is in the middle of an operation. My opinion is that it would be more consistent for it to wait to get an update to the state by selecting on a channel.

fujita:
Great, thanks a lot!

There were discussion about redesigning the fsm, conf, etc, for example, 
https://github.com/osrg/gobgp/issues/1760

Once I finish the public APIs (protobuf), I'll start that work. Please add your comments.

ecbaldwin:
A quick look through #1760 doesn't show anything about getting rid of the semaphore around fsm, which is what I'd like to see. I don't think that we need a semaphore at all if we rearchitect things a little bit to send copies of the config and state through channels as it changes.

I'd be willing to find some time to spend on this if you'd like.

PR #1827, opened 2018-08-31
Expose function in the server to watch a config file

ecbaldwin:
This is entirely refactored from code that existed already in
cmd/gobgpd. It is useful for apps that want to embed gobgp and still
benefit from configuring the server through the regular config file
format.

### DIFF
--- a/cmd/gobgpd/main.go
+++ b/cmd/gobgpd/main.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -26,35 +27,18 @@ import (
 	"runtime"
 	"syscall"
 
-	"github.com/golang/protobuf/ptypes/any"
 	"github.com/jessevdk/go-flags"
 	"github.com/kr/pretty"
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
 	api "github.com/osrg/gobgp/api"
-	"github.com/osrg/gobgp/internal/pkg/apiutil"
-	"github.com/osrg/gobgp/internal/pkg/config"
-	"github.com/osrg/gobgp/internal/pkg/table"
-	"github.com/osrg/gobgp/pkg/packet/bgp"
+	"github.com/osrg/gobgp/pkg/config"
 	"github.com/osrg/gobgp/pkg/server"
 )
 
 var version = "master"
-
-func marshalRouteTargets(l []string) ([]*any.Any, error) {
-	rtList := make([]*any.Any, 0, len(l))
-	for _, rtString := range l {
-		rt, err := bgp.ParseRouteTarget(rtString)
-		if err != nil {
-			return nil, err
-		}
-		rtList = append(rtList, apiutil.MarshalRT(rt))
-	}
-	return rtList, nil
-}
 
 func main() {
 	sigCh := make(chan os.Signal, 1)
@@ -136,10 +120,8 @@ func main() {
 		log.SetFormatter(&log.JSONFormatter{})
 	}
 
-	configCh := make(chan *config.BgpConfigSet)
 	if opts.Dry {
-		go config.ReadConfigfileServe(opts.ConfigFile, opts.ConfigType, configCh)
-		c := <-configCh
+		c := config.ReadConfigFile(opts.ConfigFile, opts.ConfigType)
 		if opts.LogLevel == "debug" {
 			pretty.Println(c)
 		}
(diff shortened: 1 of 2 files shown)


ecbaldwin:
Let me know what you think of the approach and I can fix up the tests.

cmd/gobgpd/main.go
```diff
@@ -167,280 +149,8 @@ func main() {
...
 	if opts.ConfigFile != "" {
-		go config.ReadConfigfileServe(opts.ConfigFile, opts.ConfigType, configCh)
-	}
-
-	loop := func() {
-		var c *config.BgpConfigSet
-		for {
-			select {
-			case <-sigCh:
-				apiServer.StopBgp(context.Background(), &api.StopBgpRequest{})
```
ecbaldwin: I failed to keep this line. It should go after L155 in the new file.

pkg/server/config.go
```diff
@@ -0,0 +1,304 @@
...
+			Address:   "",
+			Direction: api.ResetPeerRequest_IN,
+			Soft:      true,
+		}); err != nil {
+			log.Warn(err)
+		}
+	}
+	return c
+}
+
```
ecbaldwin: I'll remove this blank line at the end.

pkg/server/config.go
```diff
@@ -0,0 +1,304 @@
...
+	return rtList, nil
+}
+
+func WatchConfig(configFile, configType string, apiServer *Server, isGracefulRestart bool) {
+	ch := make(chan *config.BgpConfigSet)
+	go config.ReadConfigfileServe(configFile, configType, ch)
+	configLoop := func() {
+		var c *config.BgpConfigSet
+		for {
+			log.Debug("Loading new configuration (new code)")
```
ecbaldwin: I probably don't need to keep this Debug line. I should either remove it or make it an Info level log statement because it is kind of useful to get acknowledgement that the configuration was reloaded.

fujita:
i'm reluctant but making the config package public is fine by me. I think that using protobuf structures for config file is better though. But I don't move the config file routine to the server package.
Conversely, the usage of the config package in the server package should be removed completely. Currently, the server package gets protobuf structures from the apis then converts them to config package structures. Then the server packages converts the config packages structures to the protobuf. These conversion doesn't make sense at all, due to the mess-up in the past.
The config structures generated from OpenConfig but the future of OpenConfig doesn't look bright. gobgpd will support the current config format but no plan to use the config package for other stuff in GoBGP.

ecbaldwin:
I have updated this PR to move the code to a new public config package instead of the server package.

I'm not sure that I understand your entire vision for using protobuf structures for the config file. It seems to work fine if you want to hard-code configuration into the executable but I don't see how it works as a configuration file external to the compiled code.

IMO, it is very important to allow executables that embed gobgp to be configured in the same way as the generic gobgpd daemon. This PR accomplishes that and so I hope you can come to accept some form of it. As for as the rest, I think we can consider it new scope for future work. Maybe you can explain your vision for how this should work in the future and we can work together to get there.

pkg/config/main.go
```diff
@@ -0,0 +1,338 @@
...
+			Name: pg.Config.PeerGroupName,
+		}); err != nil {
+			log.Warn(err)
+		}
+	}
+}
+
+func updatePeerGroup(apiServer *server.Server, updated []config.PeerGroup) bool {
+	updatePolicy := false
+	for _, pg := range updated {
```
ecbaldwin: This method appears to do some redundant work. This was not a mistake in refactoring. The original does the same redundant work. I only did pure refactoring in this PR and did not attempt to fix anything that I thought was incorrect. Decomposing this code into smaller functions helps to see this kind of thing.

fujita:
> I'm not sure that I understand your entire vision for using protobuf structures for the config file. It seems to work fine if you want to hard-code configuration into the executable but I don't see how it works as a configuration file external to the compiled code.

I simply meant that viper is used with structures generated by protobuf instead of structures generated from openconfig:
https://gist.github.com/fujita/f1f33c674a89ce8540b20361d15726ac

Then you don't need any structure conversion with this approach. But I believe that a db should be always used to manage configurations. Well, it's matter of taste, I guess. 

I prefer to simply move from internal/pkg/config to pkg/config and leave gobgpd alone.
The current watch code isn't for everyone. for example, using a signal for reloading a file. I'm sure that there are a developer who want to reload a file in a different way. Probably, to watch a config file, such would be better rather than gobgp handmade function:
https://godoc.org/github.com/spf13/viper#WatchConfig

fujita:
>Maybe you can explain your vision for how this should work in the future and we can work together to get there.

Surely, I should write up such. So overdue.

ecbaldwin:
I don't want to lose the work that I did to move the code from gobgpd to `pkg/config/main.go` in this PR. I agree that the `Watch` function may not be the right solution for everyone. However, the `applyInitialConfig` and `updateConfig` methods in that new file capture important details about how configuration data is applied to a server. These should be available publicly outside of the gobgpd package. IMO, there is nothing in the current `internal/config` package that should be available publicly. It is too low level.

I could move the code from the `Watch` function to back into the gobgpd file and make the two functions that apply configuration to the server public. That would offer others who embed gobgp code some flexibility.

a16:
Sorry to interrupt you, but I also want you to expose not only `internal/pkg/config` but also `internal/pkg/table`. Especially I'm using these files as libralies related with RPKI/BMP. 

```
internal/pkg/table/roa.go
internal/pkg/table/path.go
internal/pkg/table/message.go
internal/pkg/table/table_manager.go
internal/pkg/table/destination.go
internal/pkg/config/bgp_configs.go
```

ecbaldwin:
That should be filed as another issue or pr. This one is about allowing code loading gobgp as a library to load the same config files as the stand alone daemon. I don't wish to expand the scope.

fujita:
@a16 pkg/table is a different topic. What in pkg/config is necessary and for what?

cmd/gobgpd/main.go
ecbaldwin: Done

ecbaldwin:
Closing this in favor of #1858 which is pretty much just a pure refactor.

PR #1829, opened 2018-09-06
pyang_plugins: Fix an import path

iwaseyusuke:
This PR fixes a hard-coded Go import path to "packet/bgp" in `bgpyang2golang.py` as following.
NG: "github.com/osrg/gobgp/packet/bgp"
OK: "github.com/osrg/gobgp/pkg/packet/bgp"

Also improves pylint and pycodestyle results.


fujita:
pushed, thanks!

Issue #1834, opened 2018-09-18
dead lock on fsm.lock


qsqx:
1. pkg/server/server.go  
function BgpServer.softResetOut  
```
if restarting {
    peer.fsm.lock.Lock() // acquire write lock
    peer.fsm.pConf.GracefulRestart.State.LocalRestarting = false
    log.WithFields(log.Fields{
        "Topic":    "Peer",
        "Key":      peer.ID(), // acquire read lock
        "Families": families,
    }).Debug("deferral timer expired")
    peer.fsm.lock.Unlock()
}
```
pkg/server/peer.go
```
func (peer *Peer) ID() string {
       peer.fsm.lock.RLock()
       defer peer.fsm.lock.RUnlock()
       return peer.fsm.pConf.State.NeighborAddress
}
```
acquire read lock while holding a read lock,  case dead lock

2. pkg/server/server.go 
BgpServer.SetPolicies
```
for _, peer := range s.neighborMap {
    peer.fsm.lock.RLock()
    log.WithFields(log.Fields{
        "Topic": "Peer",
        "Key":   peer.fsm.pConf.State.NeighborAddress,
    }).Info("call set policy")
    a, err := getConfig(peer.ID())
    if err != nil {
        return err  // return and keep lock locked
    }
    ap[peer.ID()] = *a
    peer.fsm.lock.RUnlock()
}
```

3.pkg/server/fsm.go
opensent
```
fsm.lock.Lock()
fsm.pConf.State.PeerAs = peerAs
fsm.peerInfo.AS = peerAs
fsm.peerInfo.ID = body.ID
fsm.capMap, fsm.rfMap = open2Cap(body, fsm.pConf)

. . . . . . .

if fsm.pConf.GracefulRestart.State.PeerRestarting && cap.Flags&0x08 == 0 {
    log.WithFields(log.Fields{
        "Topic": "Peer",
        "Key":   fsm.pConf.State.NeighborAddress,
        "State": fsm.state.String(),
    }).Warn("restart flag is not set")
    // send notification?
    h.conn.Close()
    return bgp.BGP_FSM_IDLE, NewFsmStateReason(FSM_INVALID_MSG, nil, nil) // return and keep lock locked
}
```

PR #1835, opened 2018-09-18
fsm: get rid of dead lock

qsqx:
https://github.com/osrg/gobgp/issues/1834


fujita:
All look good, thanks! Really needs to refactor the locking...

PR #1839, opened 2018-09-20
fix: RoutePolicy PrefixSet Append nil type cast

imcom:
when add an empty prefixset twice, the gobgpd ends up crashing on 
```
panic: interface conversion: interface {} is nil, not []*table.Prefix

goroutine 10 [running]:
github.com/osrg/gobgp/table.(*PrefixSet).Append(0xc420438400, 0xcd83c0, 0xc421708660, 0x13, 0xc4201bcbd8)
    /home/fujita/go/src/github.com/osrg/gobgp/table/policy.go:397 +0x261
github.com/osrg/gobgp/table.(*RoutingPolicy).AddDefinedSet(0xc420030d80, 0xcd83c0, 0xc421708660, 0x0, 0x0)
    /home/fujita/go/src/github.com/osrg/gobgp/table/policy.go:3510 +0x142
github.com/osrg/gobgp/server.(*BgpServer).AddDefinedSet.func1(0xc4204063c0, 0xc421493d18)
    /home/fujita/go/src/github.com/osrg/gobgp/server/server.go:2468 +0x43
github.com/osrg/gobgp/server.(*BgpServer).handleMGMTOp(0xc4201a2d80, 0xc4217086a0)
    /home/fujita/go/src/github.com/osrg/gobgp/server/server.go:171 +0x36
github.com/osrg/gobgp/server.(*BgpServer).Serve(0xc4201a2d80)
    /home/fujita/go/src/github.com/osrg/gobgp/server/server.go:305 +0x579
created by main.main
    /home/fujita/go/src/github.com/osrg/gobgp/gobgpd/main.go:150 +0x2a4
```

The root cause is that after the `add` op, the lhs.tree can be still empty. I am kind of aware that this is rare or unexpected operation but I believe we should guard from crashing the BGP daemon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,12 @@ matrix:
       before_script:
         - export GOOS="darwin"
         - go env
+    - <<: *_build
+      env:
+        - GOARCH="386"
+      before_script:
+        - export GOARCH="386"
+        - go env
 #
 # Misc
 #


fujita:
Looks like the most of commits are pointless (adds some code and then delete them later). Please create a pr including a valid commit.
This bug happens only when lhs.tree and rhs.tree are empty, right? Then I think that we should return in such case at the head of this function.

imcom:
@fujita yea, when both lhs.tree and rhs.tree are empty. I will check that out and try to return in the very beginning. I just had a hard time to amend the git commits... we will see )

imcom:
@fujita please check out [#1850](https://github.com/osrg/gobgp/pull/1850) instead

PR #1840, opened 2018-09-29
api: add GwAddress to EVPNIPPrefixRoute

chenhaiq:
Fix error that GwAddress attr is missing in EVPNIPPrefixRoute.


fujita:
Please fix the code to pass the unit test.

PR #1841, opened 2018-09-29
api: add GwAddress to EVPNIPPrefixRoute

chenhaiq:
Fix error that GwAddress attr is missing in EVPNIPPrefixRoute.

Updated testcase.


fujita:
thanks.

PR #1844, opened 2018-10-08
table: compare etag and peer ip in mac mobility

qsqx:
Mac mobility allow a given host or end-station to move from one Ethernet segment to another, but not move cross broadcast domains. 
Ethernet Tag ID should be compared.

If the PE is the originator of the MAC route and it receives the same
MAC address with the same sequence number that it generated, it will
compare its own IP address with the IP address of the remote PE and
will select the lowest IP.  If its own route is not the best one, it
will withdraw the route.
Peer ip should be compared if the sequence number are same.


fujita:
Sorry about the delay, pushed. Thanks!

PR #1847, opened 2018-10-21
Supporting Zebra API version 6 which is used in FRRouting version 6

irino:
Supporting Zebra API version 6 which is used in FRRouting version 6.


fujita:
Great, thanks!

PR #1858, opened 2018-10-29
Refactor code that configures a server from config file

ecbaldwin:
Issue: https://github.com/osrg/gobgp/issues/2106

Straight-forward refactoring change which greatly simplifies `cmd/gobgpd/main.go` making it easier to read and follow. It refactors the config file processing code from gobgpd into smaller methods which are easier to understand and makes the flow clearer.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,323 @@
+package config
+
+import (
+	"github.com/golang/protobuf/ptypes/any"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/context"
+
+	api "github.com/osrg/gobgp/api"
+	"github.com/osrg/gobgp/internal/pkg/apiutil"
+	"github.com/osrg/gobgp/internal/pkg/config"
+	"github.com/osrg/gobgp/internal/pkg/table"
+	"github.com/osrg/gobgp/pkg/packet/bgp"
+	"github.com/osrg/gobgp/pkg/server"
+)
+
+func marshalRouteTargets(l []string) ([]*any.Any, error) {
+	rtList := make([]*any.Any, 0, len(l))
+	for _, rtString := range l {
+		rt, err := bgp.ParseRouteTarget(rtString)
+		if err != nil {
+			return nil, err
+		}
+		rtList = append(rtList, apiutil.MarshalRT(rt))
+	}
+	return rtList, nil
+}
+
+func assignGlobalpolicy(bgpServer *server.BgpServer, a *config.ApplyPolicyConfig) {
+	toDefaultTable := func(r config.DefaultPolicyType) table.RouteType {
+		var def table.RouteType
+		switch r {
+		case config.DEFAULT_POLICY_TYPE_ACCEPT_ROUTE:
+			def = table.ROUTE_TYPE_ACCEPT
+		case config.DEFAULT_POLICY_TYPE_REJECT_ROUTE:
+			def = table.ROUTE_TYPE_REJECT
+		}
+		return def
+	}
+	toPolicies := func(r []string) []*table.Policy {
+		p := make([]*table.Policy, 0, len(r))
+		for _, n := range r {
+			p = append(p, &table.Policy{
+				Name: n,
+			})
+		}
+		return p
+	}
+
+	def := toDefaultTable(a.DefaultImportPolicy)
+	ps := toPolicies(a.ImportPolicyList)
+	bgpServer.SetPolicyAssignment(context.Background(), &api.SetPolicyAssignmentRequest{
+		Assignment: table.NewAPIPolicyAssignmentFromTableStruct(&table.PolicyAssignment{
+			Name:     table.GLOBAL_RIB_NAME,
+			Type:     table.POLICY_DIRECTION_IMPORT,
+			Policies: ps,
+			Default:  def,
+		}),
+	})
+
+	def = toDefaultTable(a.DefaultExportPolicy)
+	ps = toPolicies(a.ExportPolicyList)
+	bgpServer.SetPolicyAssignment(context.Background(), &api.SetPolicyAssignmentRequest{
+		Assignment: table.NewAPIPolicyAssignmentFromTableStruct(&table.PolicyAssignment{
+			Name:     table.GLOBAL_RIB_NAME,
+			Type:     table.POLICY_DIRECTION_EXPORT,
+			Policies: ps,
+			Default:  def,
+		}),
+	})
+
+}
+
+func addPeerGroups(bgpServer *server.BgpServer, addedPg []config.PeerGroup) {
+	for _, pg := range addedPg {
+		log.Infof("PeerGroup %s is added", pg.Config.PeerGroupName)
+		if err := bgpServer.AddPeerGroup(context.Background(), &api.AddPeerGroupRequest{
+			PeerGroup: config.NewPeerGroupFromConfigStruct(&pg),
+		}); err != nil {
+			log.Warn(err)
+		}
+	}
+}
+
+func deletePeerGroups(bgpServer *server.BgpServer, deletedPg []config.PeerGroup) {
+	for _, pg := range deletedPg {
+		log.Infof("PeerGroup %s is deleted", pg.Config.PeerGroupName)
+		if err := bgpServer.DeletePeerGroup(context.Background(), &api.DeletePeerGroupRequest{
+			Name: pg.Config.PeerGroupName,
+		}); err != nil {
+			log.Warn(err)
+		}
+	}
+}
+
+func updatePeerGroups(bgpServer *server.BgpServer, updatedPg []config.PeerGroup) bool {
+	for _, pg := range updatedPg {
+		log.Infof("PeerGroup %s is updated", pg.Config.PeerGroupName)
+		if u, err := bgpServer.UpdatePeerGroup(context.Background(), &api.UpdatePeerGroupRequest{
+			PeerGroup: config.NewPeerGroupFromConfigStruct(&pg),
+		}); err != nil {
+			log.Warn(err)
+		} else {
+			return u.NeedsSoftResetIn
+		}
+	}
+	return false
+}
+
+func addDynamicNeighbors(bgpServer *server.BgpServer, dynamicNeighbors []config.DynamicNeighbor) {
+	for _, dn := range dynamicNeighbors {
+		log.Infof("Dynamic Neighbor %s is added to PeerGroup %s", dn.Config.Prefix, dn.Config.PeerGroup)
+		if err := bgpServer.AddDynamicNeighbor(context.Background(), &api.AddDynamicNeighborRequest{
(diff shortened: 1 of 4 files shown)


ecbaldwin:
I am working on the test failure here. Bare with me. I am having trouble reproducing the tests locally.

ecbaldwin:
Sorry for the delay. I had limited time to invest into debugging the issue with this PR. I believe that this is ready for review now. I appreciate your attention on this.

ecbaldwin:
There must be a problem with conflict resolution. There were many conflicts so I'm not surprised. I'll find where the issue is.

cstockton:
Would it make more sense to have applying config to the bgp server live in the bgpserver package? i.e. `(*server.BgpServer).UpdateConfig(newConfig *config.BgpConfigSet)` Also the name "main.go" is usually reserved for packages with a main() function.

ecbaldwin:
@cstockton Thanks for your review. I'll change the name of the main.go file.

Why not keep the config applying code in a separate package ? Afterall, it is a client of the server's API in the end. I felt like including it with the server would bloat that package unnecessarily and risk coupling them more than necessary. It might even be useful to create a stand-alone command to configure a running server over GRPC in a future PR.

cstockton:
@ecbaldwin No problem.

To clarify a little I think the new `config` package is a perfect place for anything related to encoding / conversion / validation. I also don't *strongly* object to the server import logic, I just think it would be a nice bonus to an already great patch. Reason being importing bgpserver essentially imports the entire GoBGP and seems to have maybe influenced a couple small design details (i.e. `context.Background()`).

Maybe a nice compromise here that would solve my minor concerns would be:
- Replace `*bgpserver.Server` with `api.GobgpApiServer` since I think the methods may be identical right? This would allow wrapping the config setup logic as well programmatically by overriding an interface method or two while also lightening up the imports.
- Accept & propagate `context.Context` in [ApplyInitialConfig](https://github.com/osrg/gobgp/pull/1858/files#diff-091ace02b49a46f55f2586cfec89a160R165) and all related blocking API calls.

The imports aren't a big deal, but would be nice to not use background context anywhere. Thanks for the changes, they look good otherwise.

ecbaldwin:
@cstockton Great ideas. I think both points could be very good for the code base but it does increase the scope from a straight-forward refactoring to a little bit more than that. I don't want to further increase the scope of this patch because it is already difficult enough to get eyes on it as it is. My thinking is that a straight-forward refactoring would be the easiest way for reviewers to consume a patch as large as this. They don't have to think about the effect of other, more subtle changes, hiding in here. My promise is that this patch merely moves existing code to expose useful functionality in a new config package.  It doesn't change the way anything works.

Wrt to changing to api.GobgpApiServer, there is an issue. While the methods on server.BgpServer are very similar, they are simplified. For example, [AddPeer][1] in BgpServer simplifies the return values as compared with the [api interface definitation][2]. The [GRPC server wraps these to account for it. So, unfortunately, it isn't quite a drop-in replacement and would take significant changes to the code.

Wrt to passing context, that also breaks the "straight-forward" refactoring goal that I was aiming for. However, it is a good idea and one that I would be willing to clean up very quickly when this patch merges.

[1]: https://github.com/osrg/gobgp/blob/8c90684b276d994d09b3dc0d5a030900bbb39929/pkg/server/server.go#L2830
[2]: https://github.com/osrg/gobgp/blob/8c90684b276d994d09b3dc0d5a030900bbb39929/api/gobgp.pb.go#L6527
[3]: https://github.com/osrg/gobgp/blob/8c90684b276d994d09b3dc0d5a030900bbb39929/pkg/server/grpc_server.go#L775

cstockton:
Fair enough points on the GobgpApiServer - but I think adding context.Context is a reasonable ask. A couple reasons:

 - Once the API is public it's harder to change function signatures
 - Adding context.Context does not change the scope or diff lines of the patch
 - The change itself is a straightforward copy and replace that won't make the current diff any larger

Here's a quick sed that should do most the work:
`sed -i 's/(bgpServer \*server/(ctx context.Context, bgpServer *server/g;s/context.Background()/ctx/g' config.go`

Then all you need to do is change the calls.Apply/Update to take a context.Background() (or the current context if there is one?). Hope we can get context added before making the API public, but if you're still against the change that's okay I'll defer to the maintainers. Thanks for getting back to me regardless @ecbaldwin

pkg/config/config.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/osrg/gobgp/pkg/packet/bgp"
+	"github.com/osrg/gobgp/pkg/server"
+)
+
+func ReadConfigFile(configFile, configType string) (*config.BgpConfigSet, error) {
+	return config.ReadConfigfile(configFile, configType)
+}
+
+func ReadConfigFileOnSighup(configFile, configType string) chan *config.BgpConfigSet {
+	ch := make(chan *config.BgpConfigSet)
```
fujita: Returning a channel is a preferable design of a public API? For example, who is responsible for closing?

ecbaldwin: That is a fair point. I had a difficult time deciding where to draw the line here, especially given how [this code][1] was being used by the `cmd/gobgpd/main.go` file [here][2] and [here][3]. However,, now that I look at this, the last commit in this PR refactors [it][1] to expose the `ReadConfigFile` method which is friendlier. Let me see if I can clean this up with on more commit...

[1]: https://github.com/osrg/gobgp/blob/2f5a4d59cb8d494bcf1a64d45d84bcd8fd4a7f06/internal/pkg/config/serve.go#L27-L76
[2]: https://github.com/osrg/gobgp/blob/2f5a4d59cb8d494bcf1a64d45d84bcd8fd4a7f06/cmd/gobgpd/main.go#L187
[3]: https://github.com/osrg/gobgp/blob/2f5a4d59cb8d494bcf1a64d45d84bcd8fd4a7f06/cmd/gobgpd/main.go#L187

ecbaldwin:
@fujita Thank you very much for your review. Your feedback led me to clean things up a little bit more and make the library interface cleaner.

ecbaldwin:
@cstockton Thanks for your feeback, I have add Context to the API signatures and passed it through.

pkg/config/config.go
```diff
@@ -0,0 +1,327 @@
...
+	api "github.com/osrg/gobgp/api"
+	"github.com/osrg/gobgp/internal/pkg/apiutil"
+	"github.com/osrg/gobgp/internal/pkg/config"
+	"github.com/osrg/gobgp/internal/pkg/table"
+	"github.com/osrg/gobgp/pkg/packet/bgp"
+	"github.com/osrg/gobgp/pkg/server"
+)
+
+func ReadConfigFile(configFile, configType string) (*config.BgpConfigSet, error) {
+	return config.ReadConfigfile(configFile, configType)
```
fujita: no context?

ecbaldwin: Hi, thank you very much for your review.

I did not include a context in this call because it isn't expected to make any API calls or spawn any threads that would want a context. Its job is to read a config file from disk and return the parsed representation of it. I don't think a consumer of this API would expect to pass a context to it.

Under the hood, this call makes a `filepath.Ext(path)` call to try to detect the file time and then it makes `viper` calls to read the config file and get a` BgpConfigSet` instance. Neither accepts a context.

fujita:
This is all you want to change about the config stuff? Or you want more changes on the top of this?

ecbaldwin:
I'm not aware of anything else at this time. This meets my needs.

Most of the activity on this PR has been rebasing and resolving conflicts. I also broke a few commits into more smaller commits as I was investigating the test failures. So, it may have appeared that I was adding commits. This PR could be squashed if you prefer fewer commits.

ecbaldwin:
By the way, it took some bisecting and digging to track down what was going on with the CI tests but I finally found the proverbial smoking gun. At one point, I had rebased and had to resolve a conflict and made a mistake. [This line got inadvertently removed][1] and so the code was not responding to SIGHUP. I was perplexed because my local testing of the gobgpd binary had been successful when I first posted the change.

[1]: https://github.com/osrg/gobgp/pull/1858/commits/490c87cc43cfb3a68b1b2695d55ff0a0ead160f4#diff-d421b587344fe2358aff3646e9926d70R171

pkg/config/config.go
```diff
@@ -0,0 +1,327 @@
...
+		}); err != nil {
+			log.Warn(err)
+		} else {
+			return u.NeedsSoftResetIn
+		}
+	}
+	return false
+}
+
+func ApplyInitialConfig(ctx context.Context, bgpServer *server.BgpServer, newConfig *config.BgpConfigSet, isGracefulRestart bool) *config.BgpConfigSet {
```
ecbaldwin: The one thing I might suggest is to add `error` to the list of return values to this function and to `UpdateConfig`. Then, error handling could be modified in the future without having to change the API.

fujita:
Can config.ApplyInitialConfig() be merged into config.UpdateConfig()? Some can be configured only once (e.g. zebra) but bmp and such should be updated after the initial configuration.

ecbaldwin:
Thank you very much for your attention on this. I'm happy to make the API call look the way you want it to. However, let me first make a case for the way that I have structured it here. I do feel strongly that this is the better way to go. So, please hear me out and let me know.

When I went through the old code, there were a couple of things that made the two paths distinct. The first time that you apply configuration, you pass a boolean to indicate whether graceful restart should be done and you don't pass in the old configuration. On subsequent updates, the old configuration needs to be passed and graceful restart is not relevant. Combining the two would lead to a slightly more bloated single API where certain arguments are ignored depending on the context. This is an indication that there should be two calls.

IMO, the [resulting code in main.go][1] is more straight-forward with these two separate calls. The first call unconditionally applies the initial configuration, passing the graceful restart boolean. It is clear that there in no need to pass in initial configuration because it doesn't exist yet. Then, it sets up the loop to reread configuration and update it as needed (waiting on SIGHUP in this case). At this point, it is clear that `opts.GracefulRestart` is no longer relevant because it no longer appears in the code. IMO, combining the two calls into one leads to a less natural code structure for the consumer.

Furthermore, combining the calls requires passing `nil` the first time it is called because there was no previous configuration. This leads to special case logic within the API call itself. A good portion of the code would be inside an if/then/else statement because there really are two distinct paths in the code to handle two separate use cases. That is another indication that they should be two calls.

Another thing to point out is that some apps may not use UpdateConfig at all. My own app is one of those. It is more convenient for such an app to use the `ApplyInitialConfig` call and forget about `UpdateConfig` and its extra parameters.

If you are willing to accept this, I will be happy to provide clear documentation to explain how they should be called and provide example code for the documentation.

I might also suggest renaming `ApplyInitialConfig` to just `InitialConfig` so that it looks more consistent with `UpdateConfig`.

[1]: https://github.com/osrg/gobgp/blob/333adfca8f621696ef44d619e05ed3f2f8cab7e7/cmd/gobgpd/main.go#L184-L204

ecbaldwin:
Looking into failing checks...

fujita:
Pushed, thanks!

PR #1859, opened 2018-10-29
Use assertEqual in python tests

ecbaldwin:
In trying to run the local tests, I found that changing these
assertTrue calls to assertEqual helped me out. With this, a failure
shows the actual and expected values rather than just saying "True is
not False" which is less helpful.


fujita:
Sounds good. Pushed, thanks!

PR #1861, opened 2018-10-30
WIP Remove fsm mutex

ecbaldwin:
I've seen a couple of deadlock bugs already due to problems with this mutex. The write locks and read locks in this code makes it very difficult to wrap one's head around. There are many places where we'd like to be under a lock for the entire method but the code is forced to release the lock to make a method call because that method, or one it calls, wants to acquire the same lock. It forces many tiny synchronization points over the course of what should be a single logical operation. This make consistency difficult.

The point of this thread is to explore simplifying the synchronization in fsm.go, peer.go, and server.go to make it simpler to think about and therefor easier to maintain and develop further. This will be a work in progress for a little while.


cstockton:
I Imagine this pull req was canceled? If not I think I needs a lot of work to be reviewed - for starters all usage of https://github.com/jtolds/gls needs to be removed.

ecbaldwin:
fwiw, the references to https://github.com/jtolds/gls were never meant to be permanent. I used it to help me understand which calls were locking the mutex from which threads. I put WIP in the title to signal that, yes, it "needs a lot of work to be reviewed".

PR #1862, opened 2018-11-05
api: fix regression that match set option has no default value

chenhaiq:
The commit 000589f3c "api: Implement UpdatePolicy()" introduced
a regression: if match-set-options is not defined in config file,
gobgpd starts with error: failed to get policy info: not found...

This fix set match-set-options to default when needed.

### DIFF
--- a/internal/pkg/config/bgp_configs.go
+++ b/internal/pkg/config/bgp_configs.go
@@ -55,14 +55,6 @@ var RemovePrivateAsOptionToIntMap = map[RemovePrivateAsOption]int{
 	REMOVE_PRIVATE_AS_OPTION_REPLACE: 1,
 }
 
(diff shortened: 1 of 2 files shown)


fujita:
I can start gobgpd without such error with the following config:
https://gist.github.com/fujita/8d6a1f6f0ce0ff23034235315a3b3de3

fujita:
Thanks, confirmed. Unfortunately, this file is automatically generated from yang files and should not edited by hand
https://github.com/osrg/gobgp/blob/master/internal/pkg/config/bgp_configs.go#L1-L2
Can you please fix the callers or the generator:
https://github.com/osrg/gobgp/blob/master/tools/pyang_plugins/bgpyang2golang.py

fujita:
Great, thanks a lot!

PR #1864, opened 2018-11-06
gobgp: delete c-shared-lib

fujita:
Now the gRPC API doesn't require the binary wire-format for path
attributes. No need for the library for C.


pavel-odintsov:
Hello!

Thanks! It was nice thing for first time but gRPC API is much better now!

cmd/gobgp/lib/lib.go
```diff
@@ -1,78 +0,0 @@
-// Copyright (C) 2015 Nippon Telegraph and Telephone Corporation.
```
jeyamb: These changes are not aligned with build.sh, gobgp_api_client.cc (sample). 

Still, the gobgp_api_client.cc sample file looking for "libgobgp.h" file and build.sh is pointing to the removed "lib" directory.  These files needs to be updated.

PR #1870, opened 2018-11-07
server: propagate local routes to neighbour with policy

chenhaiq:
This fix remove the limit of not sending local routes to neighbour
when it has policy and enabled router-server-client, in order to
be used as bgp speaker to inject route to specified neighbours.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1070,7 +1070,7 @@ func (server *BgpServer) propagateUpdateToNeighbors(source *peer, newPath *table
 	}
 	family := newPath.GetRouteFamily()
 	for _, targetPeer := range server.neighborMap {
-		if (source == nil && targetPeer.isRouteServerClient()) || (source != nil && source.isRouteServerClient() != targetPeer.isRouteServerClient()) {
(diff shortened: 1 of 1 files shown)


fujita:
Sending local routes to route server clients? That's confusing. Other BGP speakers like frr supports the feature? Policies using community enables you to send local routes to specified neighbors.

chenhaiq:
This is my user case:
1. one gobgp  instance connect to lots of TOR switches in clos topology. Each of TOR switch has an unique as number in the cluster.
2. vm runs in servers.

When vm got an ip address, eg: 192.168.0.2, I want the gobgp instance add route "192.168.0.2/32 nexthop <server ip>" to the TOR switches that connected to the vm's host server, in order to advise the route to the whole network.

The idea is to add a policy to each gobgp neighbor with community condition, and create routes with communities.

Without router-server-client=ture in the neighbor, policy has no effect. I am not sure whether this is a bug.

fujita:
It's not a bug. Policies are applied to a rib. There is only one global rib for non route server clients. Route server clients has the own (virtual) rib. With neighbor match condition, you can define policies per neighbor.

PR #1874, opened 2018-11-08
server: Fix some values are all off by one in MRT TABLEDUMP_V2

a16:
The values of Peer information and BGP Message in MRT TABLE_DUMP_V2 output are all off by one. Not sure, but I guess it is due to 04f36f09a17c45f09f103dc167caf0fbdb721578.


fujita:
pushed, thanks!

PR #1875, opened 2018-11-08
Added timestamp

yoichi-cloudwan:
"gobgp monitor" is a great command for debugging, but it does not output timestamp. This change adds it.


fujita:
pushed, thanks!

PR #1880, opened 2018-11-10
Avoid deadlock

a16:
Some deadlocks still remain. It'll occur when using multiple peers. This PR will fix them as workaround.


fujita:
Great, thanks!

PR #1883, opened 2018-11-12
Support AdminDown from config

a16:
Currently, `gobgp neighbor x.x.x.x enable/disable` command enables to control the peer.
However it doesn't work the same thing from config using `admin-down = true/false`. This PR will fix it.


fujita:
pushed, thanks!

PR #1884, opened 2018-11-12
Fix intervals for MRT

a16:
In spite of having 2 fields for MRT interval(dump-interval, rotation-interval), EnableMrtRequest() in api has only one field.

### DIFF
--- a/pkg/server/mrt.go
+++ b/pkg/server/mrt.go
@@ -282,15 +282,15 @@ func (m *mrtWriter) loop() error {
 	}
 }
 
-func mrtFileOpen(filename string, interval uint64) (*os.File, error) {
+func mrtFileOpen(filename string, rInterval uint64) (*os.File, error) {
 	realname := filename
-	if interval != 0 {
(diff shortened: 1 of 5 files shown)


pkg/server/mrt.go
```diff
@@ -282,15 +282,15 @@ func (m *mrtWriter) loop() error {
 	}
 }
 
-func mrtFileOpen(filename string, interval uint64) (*os.File, error) {
+func mrtFileOpen(filename string, dInterval uint64) (*os.File, error) {
 	realname := filename
```
fujita: Two callers of mrtFileOpen() uses rotationInterval for the interval argument?
https://github.com/osrg/gobgp/blob/master/pkg/server/mrt.go#L253
https://github.com/osrg/gobgp/blob/master/pkg/server/mrt.go#L328

a16: Thanks for your review. revised and pushed again.

fujita:
pushed, thanks!

Issue #1831, opened 2018-09-11
No route updates sent to BMP server


agember:
Starting with commit 840807a, gobgpd no longer sends route monitoring messages to the configured BMP server. The configured BMP route-monitoring-policy is `local-rib`.

fujita:
I can't reproduce this. Can you post your config file?

agember:
[Koper.gobgp.txt](https://github.com/osrg/gobgp/files/2407794/Koper.gobgp.txt) is the configuration for one of the devices. All other devices in the network have the same BMP configuration.

fujita:
Please let me know if 2.X versions don't work.

PR #1886, opened 2018-11-12
server: Fix wrong value of RouteMonitoringPolicy from grpc

a16:
`config.BmpRouteMonitoringPolicy(t)` returns "PRE" from grpc in the case of using "pre-policy".
and then it doesn't match `config.BMP_ROUTE_MONITORING_POLICY_TYPE_PRE_POLICY` in https://github.com/osrg/gobgp/blob/master/pkg/server/bmp.go#L122
As a result,  it couldn't send BMP Monitoring messages.
This commit will fix it. and #1831 might be fixed.


fujita:
Pushed, thanks!

PR #1887, opened 2018-11-13
add anyNlri toPathApi when path.IsWithdraw

jqklius:
add anyNlri toPathApi when path.IsWithdraw, then we can GetExtCommunities when IsWithdraw


fujita:
pushed, thanks.

PR #1890, opened 2018-11-13
Fix Statistics Report in BMP

a16:
BMP Statistics Report hasn't been sent since the code structure changed. This commit will fix it.


fujita:
pushed, thanks!

PR #1895, opened 2018-11-15
packet: fix NewPathAttributeNextHop() to handle ipv6 addresss

fujita:
NewPathAttributeNextHop struct can handle ipv6 address but
NewPathAttributeNextHop can't.


fujita:
@Arvinderpal pushed, thanks.

PR #1900, opened 2018-11-16
api: rename NlriAny in api.Path to Nlri

fujita:
Also rename PathAttrsAny to PathAttrs. Both are the first choice
rather than the binary format members.

Support SortType member to ListPathRequest to add an option to return
unordered paths.


iwaseyusuke:
Thank you so much! This is what I have left undone.

PR #1909, opened 2018-11-28
server: fix nil pointer error when apply policy

chenhaiq:
this error can be reproduced by:
gobgp p statement s1 add action as-prepend 65001 1
gobgp p add p1 s1
gobgp n 192.168.0.5 p import add p1
gobgp n 192.168.0.5 softresetin

then:


fujita:
nice, thanks!

PR #1910, opened 2018-11-29
server: revert softResetOut change to sent withdrawn routes

chenhaiq:
Fixed a regression form commit dbd0c21:
    server: fix handleRouteRefresh and softResetOut

When route server applied a new policy and softrestout, the rejected
routes must be withdrawn from peers.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -2181,20 +2181,16 @@ func (s *BgpServer) softResetOut(addr string, family bgp.RouteFamily, deferral b
 			}
 		}
 
-		pathList, _ := s.getBestFromLocal(peer, families)
+		pathList, filtered := s.getBestFromLocal(peer, families)
 		if len(pathList) > 0 {
-			if deferral {
-				pathList = func() []*table.Path {
-					l := make([]*table.Path, 0, len(pathList))
-					for _, p := range pathList {
-						if !p.IsWithdraw {
-							l = append(l, p)
-						}
-					}
-					return l
-				}()
-			}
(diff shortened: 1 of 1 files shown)


fujita:
> // When 'path' is filtered (path == nil), check 'old' has been sent to this peer.
> // If it has, send withdrawal to the peer.
> 
https://github.com/osrg/gobgp/blob/master/pkg/server/server.go#L602-L603

As commented, the intention is sending a withdraw only if the route was sent before. GoBGP doesn't have adj-out, it can't know whether the route was sent or not when sending softresetout. That is, the route could be filtered too with the previous policy. GoBGP doesn't send withdraw rather than leaking information.

chenhaiq:
@fujita this behavior is different with real switch. We need a method to refresh route in peer after policy is updated.

fujita:
Hmm, I heard from an operator that the behavior is same with commercial routers. Btw, the import policy works as you expect. About the export policy, you need hard reset.

Thoro:
But a hard reset disrupts the peering relationship, the idea from a software router would be it's more flexible ...

Other idea:

Any policy updates happen to a policy store that is copied and not applied immediatly, if "ApplyPolicy" or similar is called, the policy store updates the active store, and also invokes softreset (with both old and new policy) then you can figure out the diff

fujita:
I considered such approach but too complicated for me. If other implementations such as bird support such, I guess that we need to think about this.

fujita:
I added the design description on this:
https://github.com/osrg/gobgp/blob/master/docs/sources/policy.md#policy-and-soft-reset

PR #1913, opened 2018-12-15
fix typo in zapi.go: LABLE -> LABEL

irino:
I found a typo (LABLE->LABEL) in my source code zapi.go which was merged in mainstream


fujita:
thanks!

PR #1928, opened 2018-12-29
zebra: Introducing MIN_ZAPIVER and MAX_ZAPIVER. And avoiding double c…

irino:
zebra: Introducing MIN_ZAPIVER and MAX_ZAPIVER. And avoiding double close channel when sequential retry to connect zebra.

- Introducing MIN_ZAPIVER and MAX_ZAPIVER for avoiding editing files except for zapi.go when new ZAPI version will be supported in future.
- Fix a bug avoiding panic by double close for a channel.
- Changing algorithm for sequential retry to connect zebra.

### DIFF
--- a/internal/pkg/zebra/zapi.go
+++ b/internal/pkg/zebra/zapi.go
@@ -37,8 +37,8 @@ const (
 )
 
 const (
-	MIN_ZAPIVER uint8 = 2
-	MAX_ZAPIVER uint8 = 6
+	MinZapiVer uint8 = 2
+	MaxZapiVer uint8 = 6
 )
 
 // Internal Interface Status.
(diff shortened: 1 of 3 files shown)


internal/pkg/zebra/zapi.go
```diff
@@ -35,6 +36,11 @@ const (
 	INTERFACE_NAMSIZ  = 20
 )
 
+const (
+	MIN_ZAPIVER uint8 = 2
+	MAX_ZAPIVER uint8 = 6
+)
```
fujita: Can you use CamelCase instead SnakeCase?

irino: I'm sorry. I will change these variables.

fujita:
squashed and pushed, thanks!

Issue #1939, opened 2019-01-17
Graceful Restart keeps stale routes for dynamic neighbors in case the neighbor never restarts


ffilippopoulos:
I am using GoBGP v2.0.0.

Graceful restart does not work for dynamic neighbors. I have the following configuration:
```
[global.config]
  as = 64513
  router-id = "10.22.22.69"

[[peer-groups]]
  [peer-groups.config]
    peer-group-name = "k8s-bgp-clients"
    peer-as = 64513
  [peer-groups.graceful-restart.config]
    enabled = true
    restart-time = 120
  [peer-groups.route-reflector.config]
    route-reflector-client = true
    route-reflector-cluster-id = "10.22.22.69"
  [peer-groups.route-server.config]
    route-server-client = true
[[dynamic-neighbors]]
  [dynamic-neighbors.config]
    prefix = "10.22.22.0/24"
    peer-group = "k8s-bgp-clients"
```
and a bunch of dynamic peers that connect to this route reflector from `10.22.22.0/24`. When I restart a neighbor I can see that the routes that it was advertising are marked as stale inside the rib, but in case the neighbor doesn't come alive in time the routes are not withdrawn from the rib.

so my rib becomes polluted with all these stale paths that never get removed, etc:
```
gobgp neighbor 10.22.22.45 local
   Network              Next Hop             AS_PATH              Age        Attrs
S*>10.4.0.0/24          10.22.22.30                               02:29:22   [{Origin: i} {LocalPref: 100}]
*  10.4.0.0/24          10.22.22.37                               02:17:00   [{Origin: i} {LocalPref: 100}]
S*>10.4.1.0/24          10.22.22.32                               02:29:21   [{Origin: i} {LocalPref: 100}]
*  10.4.1.0/24          10.22.22.36                               02:17:00   [{Origin: i} {LocalPref: 100}]
S*>10.4.2.0/24          10.22.22.22                               02:29:24   [{Origin: i} {LocalPref: 100}]
S* 10.4.2.0/24          10.22.22.42                               01:28:30   [{Origin: i} {LocalPref: 100}]
S*>10.4.3.0/24          10.22.22.31                               02:29:23   [{Origin: i} {LocalPref: 100}]
*  10.4.3.0/24          10.22.22.38                               02:17:00   [{Origin: i} {LocalPref: 100}]
S*>10.4.4.0/24          10.22.22.35                               02:23:26   [{Origin: i} {LocalPref: 100}]
[...]

PR #1940, opened 2019-01-17
Allow graceful restart for dynamic neighbors

ffilippopoulos:
This is to solve https://github.com/osrg/gobgp/issues/1939

It deletes dynamic neighbors only if graceful restart is not configured, otherwise deletes them when the restart timer expires.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1150,6 +1150,18 @@ func (s *BgpServer) propagateUpdateToNeighbors(source *peer, newPath *table.Path
 	}
 }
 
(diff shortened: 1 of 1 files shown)


fujita:
Very sorry about the delay. Thanks a lot for the report and fix!

ffilippopoulos:
@fujita no worries :) Thank you very much for looking at this. One more question: shall we always be treating master commits as stable? Any estimation on the next release?

fujita:
> shall we always be treating master commits as stable?

Yes, adding tests on this would be very welcome!

> Any estimation on the next release?

https://github.com/osrg/gobgp/blob/master/BUILD.md#releases

PR #1944, opened 2019-01-24
Add experimental BGP-LS feature

amakutunowicz:
This PR adds experimental implementation of BGP-LS. While this is not a production-ready code, it should work correctly for development, testing, or lab purposes.

**What's there?**
- Pretty much all important TLVs are parsed correctly.
- The code was tested with a vendor-supplied implementation of BGP-LS. While no rigorous interoperability tests were run, it seemed to work for all the use cases tried.
- The parsing code was fuzzy tested.
- It was used as a library.

**What's missing?**
- This is an experimental, **NOT** a production-ready code. 
- No extensive/rigorous interoperability testing was performed.
- gobgp CLI support was not tested (use as a library was the only use case).
- Some of the TLVs defined in the standard are not supported (although all the important ones are).
- Documentation/example usage is missing.
- Only incoming mode of BGP-LS was tested (although, not much is probably missing to add support for generating BGP-LS NLRIs/TLVs).

I tried to match protoc/protoc-gen-go versions, so the stray changes to .pb.go files are minimized. As the project uses go modules now, I used _go mod tidy_ to update _go.sum_ file.


fujita:
Great, thanks a lot!
I've pushed this with the change of gobgp.yang consistent with bgp_configs.go

PR #1947, opened 2019-01-26
Question about the use of  a binary release

keyibo:
why after i execute "gobgp", this phrase "context deadline exceeded" will come up?
such as,after i execute the following sentence,
#gobgp vrf add vrf1 rd 10.100:100 rt both 10.100:100 import 10.100:101 export 10.100:102
"context deadline exceeded"comes up


fujita:
Don't create a pull request to ask a question. The message means gobgp can't connect to gobgpd, for example, gobgpd doesn't run.

PR #1954, opened 2019-01-31
Add BGP-LS usage example

amakutunowicz:
This PR adds a simple example showing how to use BGP-LS in the library mode.


fujita:
Nice, thanks!

PR #1969, opened 2019-02-12
[Fix #1967] Fix print statement

### LINKED ISSUE #1967: cli PrintStatement does not print all Conditions
The code at policy.go:480 (printStatement) uses a if-else-structure, but the Condition is not exclusive, therefore it would need to be just if-if-if ..

Therefore the conditions are not displayed correctly on the cli

Edit: actually affects also Actions

Thoro:


fujita:
Thanks! Another 2.X regression in the CLI...

PR #1973, opened 2019-02-12
[Fix #1972] Fix returning of double list size, half nil values

### LINKED ISSUE #1972: v2 ListStatement returns nil values in the first n-entries
make(count) and append is used in the ListStatement, therefore always the first count-entries are nil and passed to the callback

Thoro:


pkg/server/server.go
```diff
@@ -3125,7 +3125,7 @@ func (s *BgpServer) ListStatement(ctx context.Context, r *api.ListStatementReque
 	var l []*api.Statement
 	s.mgmtOperation(func() error {
 		s := s.policy.GetStatement(r.Name)
-		l = make([]*api.Statement, len(s))
+		l = make([]*api.Statement, 0)
```
fujita: l = make([]*api.Statement, 0, len(s))
For consistency with other places and efficiency, can you fix in the above way.

Thoro: sure, updated

fujita:
Thanks!

PR #1977, opened 2019-02-15
reset eor state on graceful-restart peer down

qsqx:
eor state was not reset during peer restarting, and after peer re-established, it will withdraw all stale routes as soon as first update is received regardless of eor


fujita:
Good catch, thanks a lot!

Issue #1938, opened 2019-01-16
Feature Request: a way to see if a route is not exported / imported due to policys


imcom:
eg a R1 sends 10.0.0.0/24 and i have a policy that filteres bogons from that router.

like Quagga or FRR. This feature is quite useful in traffic engineering and troubleshooting , routers like Cisco or Juniper also support similar feature to show hidden routes (filtered) and locally accepted routes etc.

Perhaps a variable on each path that tells me the reason for it being filtered like a reference to what policy blocked it

Please refer to Juniper's [Understanding hidden routes](https://www.juniper.net/documentation/en_US/junos/topics/concept/hidden-routes-understanding.html) for the hidden routes and its purposes and reasons

Thanks in advance

Thoro:
Just for reference, should be possible to make with a few easy changes:

in Path: add a reason of policy rejection (completely internal) and pass that to api
in policy.ApplyPolicy: save the last run policy before a reject was given -> put it into the original path
cli: print the new policy reject field - and possibly add a filter 

That should make it possible to see why a certain path was rejected (for the adj-in table)

For the adj-out table you can get the filtered paths from getBestFromLocal (server.go:getAdjRib) -> that's why adj-out takes a long time, each time it's called all paths are reevaluated

Thoro:
Implemented my proposal in the attached PR, except for the filtered paths, at least you see when a path was Accepted ;)

New field "Policy" in the cli output for RIB shows in the format:

A/import/match_incomplete

[A]ccepted|[R]ejected/Policy Name/Statement Name

mrevilme:
I've been hacking a little on this also.

From what i've concluded until now is;

Either we implement a separat table for rejected incomming routes. 

Or we store the rejected incoming routes in the general table that is already setup. 
But if we do this we need to adapt all the APIs and usage where we apply a policy to the path. Since now if a path is rejected by a policy we return a "nil" instead of a path. 

I dont know where we really should put this.
My initial simple implementation did a separate table for invalid routes but i think its very hacky.

And i think we should save the routes in the table even if they are invalid and do the filtering based information attached to the path by the policys.

1) Path is recieved
2) import policies are applied, if they should be rejected, they are marked with a referense to what policy rejected.
3) if the policy is not rejected update the tables on all neighbours.
4) export policys are applied, if they should be rejected they are marked with a reference to what policy rejected the path. GOTO END
5) if policy is not rejected continue with sending the BGP updates
6) END

[...]

PR #1983, opened 2019-02-19
#1938 Add policy reasons to paths

Thoro:
Just a prove of concept for #1938


internal/pkg/table/path.go
```diff
@@ -144,6 +144,9 @@ type Path struct {
 	// For BGP Nexthop Tracking, this field shows if nexthop is invalidated by IGP.
 	IsNexthopInvalid bool
 	IsWithdraw       bool
+
+	RejectedByPolicy  string
+	AcceptedByPolicy  string
 }
```
fujita: Path object is immutable. Once it's inserted to the table, it can't be changed.

Thoro: I'm aware of the hackyness of the PR ;) - it's just prove of concept that it would work

PR #1987, opened 2019-02-20
[Fix #1986] Restore v1.33 softreset out functionality

### LINKED ISSUE #1986: Regression v1.33 to v2.1.0 - Soft Reset Out does not send path withdrawls
Previously a softresetout sent a withdraw for all filtered paths and send okay paths.

In v.2.1.0 the code to send the filtered paths is missing:

https://github.com/osrg/gobgp/blob/259d9b5264fa22165fdc66843cb00927c6c6fdf2/pkg/server/server.go#L2200
vs.
https://github.com/osrg/gobgp/blob/8e6bd4c7145de5a68c40246418500bf05af9ce39/server/server.go#L1853

Thoro:


fujita:
I added the design description on this:
https://github.com/osrg/gobgp/blob/master/docs/sources/policy.md#policy-and-soft-reset

PR #1991, opened 2019-02-28
Fixing issue #1938

mrevilme:
This is a implementation of issue #1938 

I know there is still the usage of "policy" but i'v asked @fujita for some clarification on Slack about this and thought i'd put this up there for the time being.


fujita:
https://github.com/osrg/gobgp/issues/2033
added the feature to show filtered routes.
Adding the detailed reason fattens api.Path. It hurts the performance of ListPath(), which everyone already complains of.
The client slide can figure out the reason with the policies and routes that gobgpd gives.

Issue #1912, opened 2018-12-05
GoBGP can't connect to FRR API (version 6.0)


joeyqiao12138:
Hello,

I'm using the GoBGP and FRR to set up MPLS-VPN in Centos 7, with the last step, I found PE received routes from CE and import to vpnv4, it was assigned label 0. Then I found a solution was connect to FRR API, but there was an error happened. The FRR's version I installed is 6.0 and zebra is the same version, and the error displayed below, please help!

{"level":"info","msg":"gobgpd started","time":"2018-12-05T22:45:08+08:00"}
{"Topic":"Config","level":"info","msg":"Finished reading the config file","time":"2018-12-05T22:45:08+08:00"}
{"Error":"read unix @-\u003e/var/run/frr/zserv.api: read: connection reset by peer","Topic":"Zebra","level":"error","msg":"failed to read header","time":"2018-12-05T22:45:08+08:00"}
{"Topic":"Zebra","level":"warning","msg":"cannot connect to Zebra with message version 4. going to retry another version...","time":"2018-12-05T22:45:08+08:00"}
{"Error":"read unix @-\u003e/var/run/frr/zserv.api: read: connection reset by peer","Topic":"Zebra","level":"error","msg":"failed to read header","time":"2018-12-05T22:45:08+08:00"}
{"Topic":"Zebra","level":"warning","msg":"cannot connect to Zebra with message version 2. going to retry another version...","time":"2018-12-05T22:45:08+08:00"}
{"Error":"read unix @-\u003e/var/run/frr/zserv.api: read: connection reset by peer","Topic":"Zebra","level":"error","msg":"failed to read header","time":"2018-12-05T22:45:08+08:00"}
{"Topic":"Zebra","level":"warning","msg":"cannot connect to Zebra with message version 3. going to retry another version...","time":"2018-12-05T22:45:08+08:00"}
{"Error":"read unix @-\u003e/var/run/frr/zserv.api: read: connection reset by peer","Topic":"Zebra","level":"error","msg":"failed to read header","time":"2018-12-05T22:45:08+08:00"}
{"Topic":"Zebra","level":"warning","msg":"cannot connect to Zebra with message version 4. going to retry another version...","time":"2018-12-05T22:45:08+08:00"}
{"level":"fatal","msg":"failed to set zebra config: read unix @-\u003e/var/run/frr/zserv.api: read: connection reset by peer","time":"2018-12-05T22:45:08+08:00"}

[...]

Issue #1648, opened 2018-04-18
Dealing with VRFs


amanshaikh75:
It seems to me that the way GoBGP handles VRFs is not correct. Let me illustrate this by describing two work-flows:

**Handling receipt of a BGP update message**
- How GoBGP does it according to my understanding 
       - Check the update for correctness.
       - Add path(s) to Adj-RIB-In of the peer from which the update is received.
       - If the peer belongs to a VRF, import path(s) to the Global RIB by converting IPv4/v6 routes to VPNv4/v6 routes respectively.
       - Apply applicable import policies to the path(s).
               - Run best-path selection in the context of the global RIB.
       - If necessary, install updated routes into kernel forwarding table through Zebra. While doing this, check which VRFs the route can be imported to, and install it in all the VRFs where the route can be imported to.
- Propagate changed routes to all the peers after applying export policies.
- How this should be done in my opinion:
        - Check the update for correctness.
        - Add path(s) to Adj-RIB-In of the peer from which the update is received.
        - Apply applicable import policies to the path(s).
        - Run best-path selection in the context of the VRF in which the peer belongs to.
                - If necessary, install updated routes into VRF’s kernel forwarding table through Zebra.
                - Propagate installed rotes to other peers in the VRF after applying relevant export policies. 
         - If the peer’s VRF is not global VRF, export the route(s) to the Global RIB by converting IPv4/v6 routes to VPNv4/v6 routes respectively.
         - Run best-path selection in the context of the global RIB.
                  - However, Do NOT install VPN routes into kernel’s global forwarding table.
                  - But send updated VPN routes to peers that belong to the global VRF.
         - Check which other VRFs the newly received paths need to be imported into (based on import route-target of the VRF.
         - For each VRF, the route is imported into:
                  - Run the best-path calculation in the context of the VRF.
                          - If necessary, install updated routes into VRF’s kernel forwarding table through Zebra.
          - Propagate installed routes to other peers in the VRF after applying relevant export policies.

-	Addition of a new VRF through configuration
o	How GoBGP handles this:
	If VRF already exists, ignore the request.
	Add a VRF structure to the table-manager.
	For every route-target in the import route-target list of this VRF, create a RouteTargetMemberShipNLRI, send it to appropriate neighbors.
	No VPN routes are imported into the VRF from the global RIB.
o	How this should be done in my opinion:
	Carry out all the above steps.
	Import VPN routes into the VRF.
	Run best-path selection algorithm in the context of the VRF.
[...]

Issue #1611, opened 2018-03-01
Supporting BGP/MPLS VPNs (RFC 4364)


amanshaikh75:
GoBGP supports many ingredients required for supporting BGP/MPLS VPNs (as specified in RFC 4364 - https://tools.ietf.org/search/rfc4364) such VPNv(4|6) NLRI, route targets, route-distinguishers and VRFs.
These ingredients are enough to use GoBGP as a route-reflector (RR) for a VPN service.

However, several features are missing in GoBGP which make it impossibe to use GoBGP as a Provider Edge (PE) router in a VPN service:

- When IP routes are imported from VRFs into the default RIB as VPN routes, the MPLS label assigned is zero. This problem is addressed by https://github.com/osrg/gobgp/pull/1587 which uses Zebra's label manager to get a range of MPLS labels and then assigns one label to every VRF.

- For each VRF label, GoBGP needs to install an MPLS route into the default forwarding table which pops the label from incoming packets and re-directs them to the VRF forwarding table for further processing. In Linux kernel, such route takes the form of "<vrf_label> dev <vrf_name>". Unfortunately, the required support from FRR/Zebra does not exist in the its current stable release (https://github.com/FRRouting/frr/releases/tag/frr-3.0.3). The good news is that FRR has now added support for adding such routes through its Zebra API -- see https://github.com/FRRouting/frr/issues/1759 for more information. The flip side is that this requires modifying GoBGP to support Zebra API (ZAPI) version 5 (in addition to currently supported versions of 3 and 4).

- When VPN routes are received from VPN route-reflector or other PEs, GoBGP correctly imports them into VRFs based on import targets. However, it does not add appropriate route in the VRF's forwarding table so that packets coming in from the CE are encapsulated with two MPLS labels -- inner one for the VRF on the egress PE and the outer one for loopback of the egress PE -- and sent to towards the egress PE. It's unclear to me at this point what is the right way of doing this with FRR/Zebra. I have posted a question on FRR's mailing list which is still awaiting a response. So stay tuned for updates on this subject.

- Finally, I have identified a bug in GoBGP where VPN routes received from route-reflector or other PEs are installed in default forwarding table if no VRF exists that can import these routes. While this is not a show-stopper for correct functioning, it is not correct behavior and can add unnecessarily fill up the default table. I've opened an issue about this: https://github.com/osrg/gobgp/issues/1609.

brotherdust:
[...]

PR #1587, opened 2018-02-09
zclinet: Enable to request MPLS label range to FRRouting

iwaseyusuke:
With FRRouting, MPLS label allocation is maintained on Zebra daemon, so GoBGP need to request the label range allocation to Zebra daemon in order to allocate a label to VRFs with MPLS VPN.

Example of Configuration:

```toml
[zebra.config]
  enabled = true
  url = "unix:/var/run/frr/zserv.api"
  redistribute-route-type-list = ["connect"]
  version = 4
  mpls-label-range-size = 100
```

Note: "version >= 4" is required to use this feature.

PR #1992, opened 2019-03-02
Supporting BGP/MPLS L3VPNs with Frrouting Zebra API 6

irino:
- This commit aims to solve reported problem on issues #1611, #1648 and #1912
 - Partial changes of this commit duplicate with changes on PR #1587 (not merged) and PR #1766 (not merged and already closed)
 - This commit is tested with only FRRouting version 6.0.2 (which uses Zebra API 6)
 - This commit fixes lack of LABEL_MANAGER_CONNECT_ASYNC for ZAPI6.
   (This bug is introduced on commit 2bdb76f2dcf24b891f2b6327a57c31b26463b2dd "Supporting Zebra API version 6 which is used in FRRouting version 6")


internal/pkg/table/path.go
```diff
@@ -133,13 +133,14 @@ type Validation struct {
...
-	aslooped  bool
-	reason    BestPathReason
+	info         *originInfo
+	parent       *Path
+	pathAttrs    []bgp.PathAttributeInterface
+	dels         []bgp.BGPAttrType
+	attrsHash    uint32
+	aslooped     bool
+	reason       BestPathReason
+	receiveVrfId uint32 //VRF in which the path was received.
```
fujita: Don't fatten Path structure. Can you create a map or whatever in zclient.go to store that information?

irino: Thank you for reply and reviewing my source code. 
I introduced receiveVrfId to Path struct referring PR #1766 without careful consideration. 
I sent new source code on this PR for introducing map on zclient instead of deleting receiveVrfID from Path struct.

pkg/server/server.go
```diff
@@ -2055,11 +2081,17 @@ func (s *BgpServer) AddVrf(ctx context.Context, r *api.AddVrfRequest) error {
...
 		}
+
 		if pathList, err := s.globalRib.AddVrf(name, id, rd, im, ex, pi); err != nil {
 			return err
 		} else if len(pathList) > 0 {
 			s.propagateUpdate(nil, pathList)
 		}
+		if vrf, ok := s.globalRib.Vrfs[name]; ok {
+			if s.zclient != nil && s.zclient.mplsLabelRangeSize > 0 {
+				go assignMplsLabel(s, vrf)
```
fujita: why needs to use goroutine here?

irino: The function assignMplsLabel waits to read channel (at line 2045 in servr.go).
https://github.com/osrg/gobgp/pull/1992/files#diff-09856fe9becddf0199651f451409356aR2045

The channel is written when zclient receives body of GET_LABEL_CHUNK message (at line 406 in zclient.go)
https://github.com/osrg/gobgp/pull/1992/files#diff-09856fe9becddf0199651f451409356aR2045

I confirmed the program continue to block following processing when I call assignMplsLabel without goroutine.

fujita:
pushed, let's see how things work. Thanks a lot!

PR #1993, opened 2019-03-05
doc: fix "del" command for removing a neighbor

vincentbernat:
`gobgp neighbor delete` doesn't work. We need to use `gobgp neighbor del`.


fujita:
pushed, thanks!

PR #1995, opened 2019-03-08
cmd/gobgp: use sd_notify protocol to notify the status change to systemd

Koaxiel:
It allows to notify the start-up completion of `gobgpd` to [systemd](https://github.com/systemd) via [sd_notify](https://www.freedesktop.org/software/systemd/man/sd_notify.html).

### DIFF
--- a/cmd/gobgpd/main.go
+++ b/cmd/gobgpd/main.go
@@ -26,6 +26,7 @@ import (
 	"runtime"
 	"syscall"
 
+	"github.com/coreos/go-systemd/daemon"
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/jessevdk/go-flags"
 	"github.com/kr/pretty"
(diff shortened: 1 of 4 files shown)


cmd/gobgpd/main.go
```diff
@@ -205,6 +206,14 @@ func main() {
 	bgpServer := server.NewBgpServer(server.GrpcListenAddress(opts.GrpcHosts), server.GrpcOption(grpcOpts))
 	go bgpServer.Serve()
 
+	if status, err := daemon.SdNotify(false, daemon.SdNotifyReady); !status {
+		if err != nil {
```
fujita: If systemd is not available on a system, we get a warning. I don't think that it's reasonable. Can you add an option to enable this feature?

cmd/gobgpd/main.go
```diff
@@ -26,6 +26,7 @@ import (
 	"runtime"
 	"syscall"
 
+	"github.com/coreos/go-systemd/daemon"
```
fujita: Needs to update the dependency

Koaxiel: Done

cmd/gobgpd/main.go
Koaxiel: I added an option and updated the example of the systemd service

fujita:
Also needs to update files for the module but I'll take care of it. Thanks a lot!

PR #2005, opened 2019-03-17
Deleteing goroutine for assignMplsLabel

irino:
Deleteing goroutine for assignMplsLabel and moveing code relating MPLS  L3VPN (introduced from PR #1992, commit 3a79ad3fdcbb7af0b7911a9c610c6e7af970ad6d) from table_manager.go to zclient.go


fujita:
pushed, thanks! The time to add mpls zebra tests?

irino:
Yes, I should write test codes. I recognize it. Recently FRRouting 7.0 is released. When I will write codes for supporting FRRouting 7.0, I will also write test codes including MPLS test scenario. However, I cannot promise when I will do it because my contribution to GoBGP is private work.

PR #2008, opened 2019-03-22
cmd/gobgp: fix RPKI result in neighbor policy command

nstgt:
This patch fix the output of `gobgp policy` command to show correct `RPKI result:`.

### DIFF
--- a/cmd/gobgp/policy.go
+++ b/cmd/gobgp/policy.go
@@ -511,7 +511,7 @@ func printStatement(indent int, s *api.Statement) {
 		fmt.Printf("%sAsPathLength: %s\n", ind, prettyString(c.AsPathLength))
 	}
 	if c.RpkiResult != -1 {
-		fmt.Printf("%sRPKI result: %s\n", ind, strings.TrimLeft(api.Validation_State(c.RpkiResult).String(), "STATE_"))
+		fmt.Printf("%sRPKI result: %s\n", ind, strings.TrimPrefix(api.Validation_State(c.RpkiResult).String(), "STATE_"))
 	}
 	if c.RouteType != api.Conditions_ROUTE_TYPE_NONE {
 		fmt.Printf("%sRoute Type: %s\n", ind, routeTypePrettyString(c.RouteType))
--- a/cmd/gobgp/policy.go
+++ b/cmd/gobgp/policy.go
@@ -511,18 +511,7 @@ func printStatement(indent int, s *api.Statement) {
 		fmt.Printf("%sAsPathLength: %s\n", ind, prettyString(c.AsPathLength))
 	}
 	if c.RpkiResult != -1 {
-		var result string
-		switch c.RpkiResult {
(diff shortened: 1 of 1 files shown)


cmd/gobgp/policy.go
```diff
@@ -511,18 +511,7 @@ func printStatement(indent int, s *api.Statement) {
...
-			result = "none"
-		case 1:
-			result = "valid"
-		case 2:
-			result = "invalid"
-		case 3:
-			result = "not-found"
-		}
-		fmt.Printf("%sRPKI result: %s\n", ind, result)
+		fmt.Printf("%sRPKI result: %s\n", ind, config.IntToRpkiValidationResultTypeMap[int(c.RpkiResult)])
```
fujita: Can you use api.Validation_State_name instead? api.Condition.RpkiResult should have been api.Validation_State instead of int32.

shusugmt: @fujita `api.Condition.RpkiResult` still seems to be int32
https://github.com/osrg/gobgp/blob/master/api/gobgp.pb.go#L4913

Anyways it's better to cast to `Validation_State` then call `Validation_State.String()` ?

fujita: > api.Condition.RpkiResult still seems to be int32

api.* can't be changed without the major version update. I guess that it's too early.

> Anyways it's better to cast to Validation_State then call Validation_State.String() ?

Your proposal sounds good to me for now.

nstgt: Like this?
```
fmt.Printf("%sRPKI result: %s\n", ind, api.Validation_State(c.RpkiResult))
```

Output:
```
StatementName AS65001-IMPORT_stmt7:
  Conditions:
    RPKI result: STATE_INVALID
  Actions:
     LargeCommunity:  add[64686:1101:13]
     reject
```

fujita: Looks good. If you prefer, removing "STATE_" part also looks good.

shusugmt: I haven't checked through all the code but I guess it is better to keep consistent and stick with the mapping defined in the following place without modifying(in this case, dropping "STATE_" part):
https://github.com/osrg/gobgp/blob/master/api/gobgp.pb.go#L698-L709

nstgt: I think lower case is appropriate... Should I divide line into 2?
```
fmt.Printf("%sRPKI result: %s\n", ind, strings.ToLower(strings.TrimLeft(api.Validation_State(c.RpkiResult).String(), "STATE_")))
```

Output:
```
StatementName AS65001-IMPORT_stmt7:
  Conditions:
    RPKI result: invalid
  Actions:
     LargeCommunity:  add[64686:1101:13]
     reject
```

nstgt:
woops
I'll fix it.

fujita:
squashed and pushed, thanks!

Issue #2003, opened 2019-03-14
Is there a config option to specify BMP sysName and sysDescr?


KumaMonX:
Is there a config option to specify BMP sysName and sysDescr?

According to documents（RFC 7854），BMP Initiation Message must contain sysDescr and sysName Information TLVs， but current release BMP feature didn't send this two mandatory TLVs.

Any plan to fix it?

fujita:
No. You can see the current bmp configuration options:
https://github.com/osrg/gobgp/blob/master/tools/pyang_plugins/gobgp.yang#L560

there are many issues with higher priorities for me right now so a pull request for that is very welcome.

fujita:
https://github.com/osrg/gobgp/commit/80ff3a325cbced352489fa4ea4b570e5f08062ef
now supported.

PR #2009, opened 2019-03-23
Fix issue #2003

JieJhih:


internal/pkg/config/bgp_configs.go
```diff
@@ -1416,7 +1416,9 @@ type BmpServerConfig struct {
 	// gobgp:route-mirroring-enabled's original type is boolean.
 	// Enable feature for mirroring of received BGP messages
 	// mainly for debugging purpose.
-	RouteMirroringEnabled bool `mapstructure:"route-mirroring-enabled" json:"route-mirroring-enabled,omitempty"`
+	RouteMirroringEnabled bool   `mapstructure:"route-mirroring-enabled" json:"route-mirroring-enabled,omitempty"`
+	SysName               string `mapstructure:"SysName" json:"SysName,omitempty"`
+	SysDescr              string `mapstructure:"SysDescr" json:"SysDescr,omitempty"`
 }
```
fujita: bgp_configs.go is automatically generated. you can modify it directly
https://github.com/osrg/gobgp/tree/master/tools/pyang_plugins

internal/pkg/config/default.go
```diff
@@ -350,6 +350,9 @@ func setDefaultConfigValuesWithViper(v *viper.Viper, b *BgpConfigSet) error {
 	}
 
 	for idx, server := range b.BmpServers {
+		if server.Config.SysName == "" || server.Config.SysDescr == "" {
+			return fmt.Errorf("BMP Initiation Message must contain sysDescr and sysName Information TLVs")
+		}
```
fujita: What are appropriate default strings for them? I prefer to set something appropriate if not specified.

internal/pkg/config/bgp_configs.go
fujita: let me know if you like me to take care of changes to bgp_configs.go.

JieJhih: Thanks, I fixed it, and I also fixed the pyang plugins document. Is there have something error please help me fix it out.

internal/pkg/config/default.go
JieJhih: RFC1213
`sysDescr OBJECT-TYPE
              SYNTAX  DisplayString (SIZE (0..255))
              ACCESS  read-only
              STATUS  mandatory
  DESCRIPTION
                      "A textual description of the entity.  This value
                      should include the full name and version
                      identification of the system's hardware type,
                      software operating-system, and networking
                      software.  It is mandatory that this only contain
                      printable ASCII characters."`
`sysName OBJECT-TYPE
              SYNTAX  DisplayString (SIZE (0..255))
 ACCESS  read-write
              STATUS  mandatory
              DESCRIPTION
                      "An administratively-assigned name for this
                      managed node.  By convention, this is the node's
                      fully-qualified domain name."`

JieJhih: I think SysName could have the default value, but I'm not sure SysDescr .

cmd/gobgpd/main.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	"github.com/coreos/go-systemd/daemon"
 	"github.com/golang/protobuf/ptypes/any"
-	"github.com/jessevdk/go-flags"
```
fujita: why?

JieJhih: Oops , I'm sorry, I didn't noticed that. This caused by my vscode plugin.
I fixed it.

fujita:
thanks, squashed and pushed.

PR #2010, opened 2019-03-24
table: fix to avoid setting vacant Large Communities

nstgt:
Currently, GoBGP advertises Large Communities attributes with length of zero when all of them were removed by policy action. Here is a screenshot of a part of packet which was sent from GoBGP.

vacant-large-communities

As described in [RFC7606 Section.4](https://tools.ietf.org/html/rfc7606#section-4), most of path attributes having length of zero shall be considered as syntax error. Actually, the above packet caused BGP peer shutdown due to attribute length error (Notification code 3/5).

This patch fix it by checking length of Large Communities and removing it (if necessary), the same as the way of Path.SetCommunities function.


fujita:
pushed, thanks a lot!

PR #2014, opened 2019-03-24
grpc_server:  Close listener when listen failed

JieJhih:


pkg/server/grpc_server.go
```diff
@@ -62,30 +62,38 @@ func newAPIserver(b *BgpServer, g *grpc.Server, hosts string) *server {
...
+				log.WithFields(log.Fields{
+					"Topic": "grpc",
+					"Key":   host,
+					"Error": err,
+					"Retry": k,
+				}).Warn("listen failed")
+				lis.Close()
+				continue
+			}
+			err = s.grpcServer.Serve(lis)
```
fujita: What's kinda situation could be handled by retrying this function?

JieJhih: I fixed the conditions.
Sometimes listen failed it caused by timeout or the error is temporary , I think it can retry.

fujita:
It doesn't look that Listen() is retried; Accept() is. I don't think that retrying at random times is a good idea. A reasonable way to handle the failure depends on a system.

PR #2017, opened 2019-03-25
Use Itoa method instead sprintf

JieJhih:
Ref. https://gist.github.com/evalphobia/caee1602969a640a4530

### DIFF
--- a/cmd/gobgp/root.go
+++ b/cmd/gobgp/root.go
@@ -17,9 +17,9 @@ package main
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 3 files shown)


JieJhih:
@fujita Some lines like this https://github.com/osrg/gobgp/blob/master/cmd/gobgp/policy.go#L887
If type is string , I think it could use `+`  instead sprintf , and this will improve performance

fujita:
no performance critical path but not a bad thing, I guess. Pushed, thanks.

PR #2024, opened 2019-03-28
fsm: not reject open message with restart flag unset from gr peer

qsqx:


pkg/server/fsm.go
```diff
@@ -1352,10 +1352,7 @@ func (h *fsmHandler) opensent(ctx context.Context) (bgp.FSMState, *fsmStateReaso
 								"Key":   fsm.pConf.State.NeighborAddress,
 								"State": fsm.state.String(),
 							}).Warn("restart flag is not set")
-							// send notification?
-							h.conn.Close()
-							fsm.lock.Unlock()
-							return bgp.BGP_FSM_IDLE, newfsmStateReason(fsmInvalidMsg, nil, nil)
+							// just ignore
```
fujita: Needs to set peer.fsm.pConf.GracefulRestart.State.PeerRestarting to false?

qsqx: we still need peer.fsm.pConf.GracefulRestart.State.PeerRestarting to indicate whether stales routes should be dropped after eor received
```
peer.fsm.lock.RLock()
peerRestarting := peer.fsm.pConf.GracefulRestart.State.PeerRestarting
peer.fsm.lock.RUnlock()
if peerRestarting {
    if peer.recvedAllEOR() {
        peer.stopPeerRestarting()
        pathList := peer.adjRibIn.DropStale(peer.configuredRFlist())
        peer.fsm.lock.RLock()
        log.WithFields(log.Fields{
            "Topic": "Peer",
            "Key":   peer.fsm.pConf.State.NeighborAddress,
        }).Debugf("withdraw %d stale routes", len(pathList))
        peer.fsm.lock.RUnlock()
        s.propagateUpdate(peer, pathList)
    }

    // we don't delay non-route-target NLRIs when peer is restarting
    rtc = false
}
```

PR #2029, opened 2019-03-29
Update image of route server policy model

nstgt:
For corresponding to Single RIB Route Server structure.


fujita:
awesome, thanks a lot!

PR #2030, opened 2019-03-29
Parse IPMSI parameter in cmd

JieJhih:
Ref. https://www.ietf.org/id/draft-ietf-bess-evpn-bum-procedure-updates-05.txt
```

   The per-region I-PMSI route uses an embedded EC in NLRI to identify a
   region.  As long as it uniquely identifies the region and the RBRs
   for the same region uses the same EC it is permitted.  In the case
   where an AS number or area ID is needed, the following can be used:

   o  For a two-octet AS number, a Transitive Two-Octet AS-Specific EC
      of sub-type 0x09 (Source AS), with the Global Administrator sub-
      field set to the AS number and the Local Administrator sub-field
      set to 0.

   o  For a four-octet AS number, a Transitive Four-Octet AS-Specific EC
      of sub-type 0x09 (Source AS), with the Global Administrator sub-
      field set to the AS number and the Local Administrator sub-field
      set to 0.
```

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -10032,8 +10032,8 @@ func ParseExtendedCommunity(subtype ExtendedCommunityAttrSubType, com string) (E
 	}
 }
 
-func ParseRouteTarget(subtype ExtendedCommunityAttrSubType, rt string) (ExtendedCommunityInterface, error) {
-	return ParseExtendedCommunity(subtype, rt)
+func ParseRouteTarget(rt string) (ExtendedCommunityInterface, error) {
+	return ParseExtendedCommunity(EC_SUBTYPE_ROUTE_TARGET, rt)
 }
 
 func SerializeExtendedCommunities(comms []ExtendedCommunityInterface) ([][]byte, error) {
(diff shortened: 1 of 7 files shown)


pkg/packet/bgp/bgp.go
```diff
@@ -10029,8 +10032,8 @@ func ParseExtendedCommunity(subtype ExtendedCommunityAttrSubType, com string) (E
 	}
 }
 
-func ParseRouteTarget(rt string) (ExtendedCommunityInterface, error) {
```
fujita: This function is publicly exported. You can't change it without the major version updated.

JieJhih: Fixed.
Thanks

fujita:
Thanks, now the code looks good. Can you update the doc?
https://github.com/osrg/gobgp/blob/master/docs/sources/evpn.md#cli-syntax

JieJhih:
> Thanks, now the code looks good. Can you update the doc?

Updated.
Thanks

fujita:
squashed and pushed, thanks.

PR #2039, opened 2019-04-02
Fix gofmt

JieJhih:


fujita:
https://github.com/osrg/gobgp/blob/master/.travis.yml#L110-L116

gofmt is included in the CI on pull requests. If it doesn't complain, the format is acceptable.

PR #2040, opened 2019-04-02
Fix spell error

JieJhih:


fujita:
The file is automatically generated, cant be changed directly.
https://github.com/osrg/gobgp/blob/master/tools/pyang_plugins/README.rst

The typos come from the original yang files.

PR #2050, opened 2019-04-07
Watch more state in MonitorPeer

### LINKED ISSUE #2047: Feature Request: don't suppress initial non-established peer state in MonitorPeer
MonitorPeerRequest w/current=true does not return the current state unless it is established. https://github.com/osrg/gobgp/blob/ccdc2e6277020e571da7792c1d9e76cf13e82f44/pkg/server/server.go#L4013-L4023

JieJhih:
Fix #2047 
When watch option ``current=true`` ,  using fsm state reason distinguish established  and non-established peer state

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -4012,13 +4012,18 @@ func (s *BgpServer) watch(opts ...watchOption) (w *watcher) {
 		}
 		if w.opts.initPeerState {
 			for _, peer := range s.neighborMap {
+				m := &fsmMsg{}
 				peer.fsm.lock.RLock()
 				notEstablished := peer.fsm.state != bgp.BGP_FSM_ESTABLISHED
 				peer.fsm.lock.RUnlock()
 				if notEstablished {
-					continue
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -4012,13 +4012,18 @@ func (s *BgpServer) watch(opts ...watchOption) (w *watcher) {
...
-					continue
+					stateReason := newfsmStateReason(fsmNewConnection, nil, nil)
+					m.fsm = peer.fsm
+					m.MsgType = fsmMsgBGPMessage
+					m.MsgSrc = peer.fsm.pConf.State.NeighborAddress
+					m.StateReason = stateReason
 				}
-				w.notify(newWatchEventPeerState(peer, nil))
+				w.notify(newWatchEventPeerState(peer, m))
 			}
```
fujita: We can't change the behavior of exported API.

fujita:
Replace MonitorPeer API with WatchEvent API. This problem doesn't exit anymore.

PR #2051, opened 2019-04-07
Initial grpc health check

JieJhih:
When using grpc server, we need a mechanism to watch its health.

```pkg/server/health.go``` as a health check client, it may through config or cmd `--healthcheck`
 to start watching grpc server healthy.

### CODE REVIEW

File: pkg/server/health.go [hunk 1]
```diff
@@ -73,6 +71,7 @@ func (c *healthClient) Check(ctx context.Context) (bool, error) {
 	return false, err
 }
 
+/* health check client
```
fujita: unused functions should be deleted.

fujita:
Why this feature is necessary?

JieJhih:
In my case, I'm using a project written in Java, it communicates with gobgp through by GRPC. 
I have to make sure the service that through GRPC is still working. If it's not working, then rerun a new gobgp service. 
It can do some error handling through this mechanism when grpc server is not working

fujita:
>I have to make sure the service that through GRPC is still working

If GRPC is not working, then it means gobgpd crashes? If it's what you want to know, then GetBgp API should work?

JieJhih:
Yeah, that is a way. but we have to know the crashed reason.  Distinguish it caused by gobgpd or grpc through this mechanism.And I think when grpc is not working (gobgpd crashed), through this mechanism  that can set by config auto restart is convenient if gobgp support this feature.

Ref. https://github.com/grpc/grpc-go/blob/master/codes/codes.go

fujita:
I don't think that it's possible to get the reason why gobgpd crashed after gobgpd crashed. I don't understand your reference. It's not related at getting the reason of a crash.

JieJhih:
Oops , my bad.
There have some grpc status code. If gobgpd crashed , health check client will get ``Internal Code = 13
`` , other code usually caused by grpc.
I think the feature is not necessary as you say.
Thanks

PR #2052, opened 2019-04-08
Fix route server scenario_test

nstgt:
This patch include following changes.
* Fixing `route_server_policy_test.py` which has not been executed correctly since a commit #0f102be9 was applied
* Removing unnecessary test `InPolicyUpdate2`
* Reusing `InPolicyRejectImplicitWithdraw` as `ImportPolicyRejectImplicitWithdraw`


fujita:
great, thanks a lot!!

PR #2056, opened 2019-04-10
Create makefile for build and unit test

JieJhih:
The makefile could help the beginner to build gobgp through ``make build``.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,27 @@
+EXECUTABLE :=gobgp
+SOURCES ?= $(shell find . -name "*.go" -type f)
+PACKAGES ?= $(shell go list ./... | grep -v /vendor/)
+
+GOFMT ?= gofmt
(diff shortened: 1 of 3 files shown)


JieJhih:
@fujita Do we have ``CONTRIBUTING.md`` , it could mention contributor using ``make test`` to reduce error in CI.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+endif
+
+all: build
+
+build: $(EXECUTABLE)
+
+$(EXECUTABLE): $(SOURCES)
+	go build -v -o bin/$@ ./cmd/gobgp/
+	go build -v -o bin/$@d ./cmd/gobgpd/
+
```
fujita: You always need to enable the module feature.

fujita:
It doesn't look that the Makefile is so useful; it builds only binaries and run unit tests. As you mention, if it supports the all CI tests, I think that surely worth merge but the CI needs VMs and containers on them. I guess that it's not reasonable to assume that everyone needs to set up such environment.

PR #2064, opened 2019-04-26
Consolidate duplicate code blocks both updating peer groups

ecbaldwin:
I can only see two differences: 1) the first block updates
`updatePolicy` and 2) the log statements are a little bit different
(`pg.State` vs `pg.Config` and `%v` vs `%s`).

From the other blocks above it, it seems that `pg.Config` is the right
thing to use in the log statement rather than `pg.State`.


fujita:
Looks good, thanks! No idea how this happened.

PR #2065, opened 2019-05-01
zebra: using LABEL_MANAGER_CONNECT_ASYNC instead of LABEL_MANAGER_CON…

irino:
…NECT

 - This commit changes API from  LABEL_MANAGER_CONNECT to LABEL_MANAGER_CONNECT_ASYNC for connecting zebra label manager to avoid stop processing by synchronous message.
 - This commit updates test for existing API for zebra all version (from 2 to 6) and adds test for some MPLS related APIs for zebra version 5 and 6.


fujita:
squashed and pushed, thanks!

PR #2069, opened 2019-05-07
server: replace github.com/satori/go.uuid

fujita:
satori/go.uuid broke the API without the major version updated. We
could work around this with the module feature but looks like there is
a better option, google/uuid, so let's replace it.


fujita:
pushed

PR #2072, opened 2019-05-08
zebra: fix unit test failure on non linux

fujita:
syscall.AF_INET/AF_INET6 are os/architecture specific. On some non
Linux x86 architectures, failures like the following happen:

--- FAIL: Test_NexthopRegisterBody (0.00s)
        Error Trace:    zapi_test.go:893
        Error:			Not equal: 0x1e (expected)
        			        != 0xa (actual)

        Error Trace:    zapi_test.go:894
        Error:			Not equal: net.IP{0x20, 0x1, 0xd, 0xb8, 0x0, 0x1, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1} (expected)
        			        != net.IP(nil) (actual)

        				Diff:
        				--- Expected
        				+++ Actual
        				@@ -1,2 +1,2 @@
        				-(net.IP) (len=16 cap=16) 2001:db8:1:1::1
        				+(net.IP) <nil>

looks like that frr uses os/os/architecture specific values in the
zapi message.

### DIFF
--- a/internal/pkg/zebra/zapi_test.go
+++ b/internal/pkg/zebra/zapi_test.go
@@ -864,9 +864,9 @@ func Test_NexthopRegisterBody(t *testing.T) {
 
 	// Input binary
 	bufIn := []byte{
-		0x01, 0x00, 0x02, 0x20, // connected(1 byte)=1, afi(2 bytes)=AF_INET, prefix_len(1 byte)=32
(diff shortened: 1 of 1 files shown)


fujita:
@irino if you like to fix the bug in a different way, please create a pr.

fujita:
pushed, thanks.

PR #2075, opened 2019-05-10
Adding a new struct which is used to hold information about the server.

mrevilme:
It now contains 3 fields;
'started_at' which denotes at what time the bgpserver was started
'reconfigured_at' which denotes at what time the server last was modified
'version' which is a struct that contains the version information of the server.


pkg/server/server.go
```diff
@@ -2655,10 +2669,18 @@ func (s *BgpServer) GetBgp(ctx context.Context, r *api.GetBgpRequest) (*api.GetB
...
 				ListenAddresses:  g.Config.LocalAddressList,
 				UseMultiplePaths: g.UseMultiplePaths.Config.Enabled,
+				StartedAt: config.ProtoTimestamp(s.status.startedAt),
+				ReconfiguredAt: config.ProtoTimestamp(s.status.reconfiguredAt),
+				Version: &api.Version{
+					Major: 2,
+					Minor: 4,
+					Patch: 0,
+					Version: "2.4.0",
+				},
```
fujita: I think that embedding static numbers in the code has two problems:
- We have multiple versions that return 2.4.0. We should use git hash or something unique for not released tagged versions.
- We have to update these numbers every time we release a new version.

We need a way to embed a unique version for every commit automatically.

fujita:
How about automatically setting the version number at linking time.
https://github.com/fujita/gobgp/tree/version

fujita:
If you need something new on the top of the version feature, please create a new pr.

PR #2076, opened 2019-05-16
Fix cli crash when showing neighbor stats as list.

christoph-heiss:
Hi,

so basically - if gobgpd is started with TLS support, but `--tls ...` is not passed to the cli, it crashes when listing neighbors with a segfault due to an unhandled error.


fujita:
thanks!

Issue #2079, opened 2019-05-16
AddPath routes have same PathId (same localId).


kishiguro:
When we enable AddPath in Route Reflector and reboot AddPath routes originator router at the same time, sometimes AddPath routes have same PathId=1 (actually same local identifier).
When this happens, Route Reflector client replace AddPath routes with latter one.  So there will be missing routes.  We enable long lived graceful restart and legacy graceful restart.  I'm not sure it affects to this issue or not.
My observation is this is due to missing lock of type Destination member `localIdMap`.  When same prefix routes with different nexthop update comes at the same time, there would be situation of assign same PathId to Path.

fujita:
IIRC, only the main goroutine calls localIdMap methods. Am I missing something?

kishiguro:
if so is there any situation local identifier has same value?  I try to find the root cause but no success...

kishiguro:
Do you think is there any situation where explicit or implicit withdraw override existing local path id in case of AddPath?

```
		if newPath.GetSource().Equal(path.GetSource()) && newPath.GetNlri().PathIdentifier() == path.GetNlri().PathIdentifier() {
			log.WithFields(log.Fields{
				"Topic": "Table",
				"Key":   dest.GetNlri().String(),
				"Path":  path,
			}).Debug("Implicit withdrawal of old path, since we have learned new path from the same peer")

			found = i
			newPath.GetNlri().SetPathLocalIdentifier(path.GetNlri().PathLocalIdentifier())
			break
		}
```

fujita:
Can you explain the details of the setup?

1. you have three peers, rr, rr client, another peer advertising a route to rr?
2. you started rr without addpath enabled, then change the configuration of rr to enable addpath and reboot another peer advertising a route to rr?
3. the duplication of local id happens in rr, right?

kishiguro:
Sorry for late reply.

> you have three peers, rr, rr client, another peer advertising a route to rr?

Yes, we have 3 RRs and many RR clients.  Each RR client connect to 3 RRs at the same time.
RR clients advertise one's own routes to 3 RRs then RR reflect the routes to other RR clients.

> you started rr without addpath enabled, then change the configuration of rr to enable addpath and reboot another peer advertising a route to rr?

Yes, at boot time RR does not have any configuration.  We inject configuration from gRPC API.

> the duplication of local id happens in rr, right?

Yes, duplication of local path id happens only in RR never in client side.

fujita:
[...]

PR #2080, opened 2019-05-16
AddPath routes have same PathId (same localId). #2079.

kishiguro:
Add lock to type Bitmap.

### DIFF
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"net"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/osrg/gobgp/internal/pkg/config"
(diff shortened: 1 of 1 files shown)


internal/pkg/table/path.go
```diff
@@ -69,19 +80,22 @@ func (b *Bitmap) FindandSetZeroBit() (uint, error) {
...
 	b.bitmap = new
+	b.m.Unlock()
 }
 
 func NewBitmap(size int) *Bitmap {
 	b := &Bitmap{}
 	if size != 0 {
 		b.bitmap = make([]uint64, (size+64-1)/64)
 	}
+	b.m = &sync.Mutex{}
```
cstockton: See comment above.

internal/pkg/table/path.go
```diff
@@ -36,21 +37,31 @@ const (
 
 type Bitmap struct {
 	bitmap []uint64
+	m      *sync.Mutex
```
cstockton: This should be the zero value
```suggestion
	m      sync.Mutex
```

internal/pkg/table/path.go
```diff
@@ -69,19 +80,22 @@ func (b *Bitmap) FindandSetZeroBit() (uint, error) {
 }
 
 func (b *Bitmap) Expand() {
+	b.m.Lock()
 	old := b.bitmap
```
cstockton: This appears to be a basic copy operation.
```Go
cpy := make([]uint64, len(b.bitmap)+1)
copy(cpy, b.bitmap)
b.bitmap = cpy
```

cstockton:
What is the rationale of this change?

The primary holder of Bitmap is Destination- which is also not thread safe. Most all Bitmap operations are read operations right? My concern here is this could add a lot of contention for read operations, I think it should be a RWMutex at the very least. However I have a feeling if it's to fix a race condition then others will remain dormant after.

fujita:
I close this because this change surely fixes the reported issue. I'm still waiting for your response on the issue as I commented.

PR #2083, opened 2019-05-17
add version suport

fujita:
Use Makefile and set the version number at linking time.


mrevilme:
I would say this is a very bad way to get this going.

Providing the version for the build at runtime instead in the a proper file seams like a very strange way of thinking.

I would prefer that we code the version INTO the version.go and provide additional data using linker flags. That way we will never have a build that is missing a version number. Have a look at #2084

PR #2084, opened 2019-05-19
Versioning proposal

mrevilme:
As a alternative to https://github.com/osrg/gobgp/pull/2083


BUILD.md
```diff
@@ -15,6 +15,43 @@ Now ready to modify the code and build two binaries, `cmd/gobgp` and `cmd/gobgpd
...
+```internal/pkg/version/version.go``` defines the following variables
+
+```MAJOR``` ```MINOR``` ```PATCH``` these constants are for the Semantic Versioning scheme.
+These will be updated upon release by maintainer.
+
+There is also two more variables that are ment to be changed by ldflags; 
+
+```TAG``` is supposed to be used to denote which branch the build is based upon.
+```SHA``` is supposed to be used to inform about which git sha sum the build is based on.
+
```
fujita: `HASH` is better than `SHA`? No guarantee that git uses SHA for generating a hash value in the future. some feels that `TAG` is redundant? The hash value is unique anyway.

How about let people use strings after the release version as they like. Say, major-minor-path+revision. Whatever you like, branch name, hash value, host name, etc, can be coded into the revision.

mrevilme: Im cool with "HASH" instead of "SHA" thats just sematics.

But regarding the TAG that could be renamed to SUFFIX instead its a part of the SemVer spec.

> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85.

fujita: If we go with terms of the SemVer spec, the following?
``MAJOR``-``MONOR``-``PATCH``-``IDENTIFIERS``
``IDENTIFIERS`` is optional. You can put anything to it.

mrevilme: Yes thats correct thats why we rename TAG into "IDENTIFIERS" or "SUFFIX"

```
if len(TAG) > 0 {
    suffix = fmt.Sprintf("-%s", TAG)
}
```

fujita: but why we need to enforce the specific format (tag-hash)?   the SemVer spec allows various formats. We simply give users ``IDENTIFIERS``?

mrevilme: Ah i think i see what we are talking about i miss read the spec;

2.4.0-IDENTIFIERS+METADATA
2.4.0-alpha1+commit.XXYYZZZZ

So we should rework the proposal to match that?

What im thinking is that we give the builder the flexibility of specifying IDENTIFIERS however they want. 

And we introduce the new "METADATA" variable that can be used to pass on additional build metadata.

And we change SHA to "COMMIT" which is always appended to METADATA parts if "COMMIT" is provided.

```
go build -X github.com/osrg/gobgp/internal/pkg/version.IDENTIFIER=alpha1 -X github.com/osrg/gobgp/internal/pkg/version.COMMIT=XXXYYYZZ
```

Would give a version string like ```2.4.0-alpha1+commit.XXXYYYZZ```

```
go build -X github.com/osrg/gobgp/internal/pkg/version.IDENTIFIER=alpha1 -X github.com/osrg/gobgp/internal/pkg/version.COMMIT=XXXYYYZZ -X github.com/osrg/gobgp/internal/pkg/version.METADATA=date.20190521
```

Would give a version string like ```2.4.0-alpha1+date.20190521.commit.XXXYYYZZ```

What do you think?

fujita: ``2.4.0-IDENTIFIERS+METADATA`` looks fine. I don't see why ``COMMIT`` is necessary. You can decode the commit to ``METADATA``.

mrevilme: Yes but thats just to make COMMIT a more specified argument. So that GIT commits always are repesented with "COMMIT" and not sha, git, version or something that ppl might think up.

mrevilme:
```
Emils-MacBook-Pro:gobgpd emil$ go build -ldflags "-X github.com/osrg/gobgp/internal/pkg/version.COMMIT=`git rev-parse --short HEAD` -X github.com/osrg/gobgp/internal/pkg/version.METADATA="date.`date "+%Y%m%d"`" -X github.com/osrg/gobgp/internal/pkg/version.IDENTIFIERS=alpha1"
Emils-MacBook-Pro:gobgpd emil$ ./gobgpd --version
gobgpd version 2.4.0-alpha1+commit.b461d2f8.date.20190521
```

fujita:
I still think that ``HASH`` is redundant but if you want, it's fine by me. Please fix gofmt and doc format errors.

fujita:
I fixed go fmt and markdown errors and squashed the commits.
https://github.com/osrg/gobgp/pull/2094

fujita:
pushed https://github.com/osrg/gobgp/pull/2094

Issue #2089, opened 2019-06-01
an error in the code


shengslin:
hi, when I read the function pkg/server/zclient.go:newIPRouteBody, I found some mistakes.
In this function, "isWithdraw" is used as a criterion which is an uninitialized return value .
As follows, line 33
```

1. func newIPRouteBody(dst []*table.Path, vrfId uint32, z *zebraClient) (body *zebra.IPRouteBody, isWithdraw bool) {
2. 	version := z.client.Version
3. 	paths := filterOutExternalPath(dst)
4. 	if len(paths) == 0 {
5. 		return nil, false
6. 	}
7. 	path := paths[0]
8. 	l := strings.SplitN(path.GetNlri().String(), "/", 2)
9. 	var prefix net.IP
10. 	var nexthop zebra.Nexthop
11. 	nexthops := make([]zebra.Nexthop, 0, len(paths))
12. 	msgFlags := zebra.MESSAGE_NEXTHOP
13. 	switch path.GetRouteFamily() {
14. 	case bgp.RF_IPv4_UC, bgp.RF_IPv4_VPN:
15. 		if path.GetRouteFamily() == bgp.RF_IPv4_UC {
16. 			prefix = path.GetNlri().(*bgp.IPAddrPrefix).IPAddrPrefixDefault.Prefix.To4()
17. 		} else {
18. 			prefix = path.GetNlri().(*bgp.LabeledVPNIPAddrPrefix).IPAddrPrefixDefault.Prefix.To4()
19. 		}
20. 	case bgp.RF_IPv6_UC, bgp.RF_IPv6_VPN:
21. 		if path.GetRouteFamily() == bgp.RF_IPv6_UC {
22. 			prefix = path.GetNlri().(*bgp.IPv6AddrPrefix).IPAddrPrefixDefault.Prefix.To16()
23. 		} else {
24. 			prefix = path.GetNlri().(*bgp.LabeledVPNIPv6AddrPrefix).IPAddrPrefixDefault.Prefix.To16()
25. 		}
26. 	default:
27. 		return nil, false
28. 	}
29. 	var nhVrfId uint32
30. 	if nhvrfid, ok := z.pathVrfMap[path]; ok {
31. 		// if the path is withdraw, delete path from pathVrfMap after refer the path
32. 		nhVrfId = nhvrfid
33. 		if isWithdraw {   //  an uninitialized return value
34. 			delete(z.pathVrfMap, path)
35. 		}
36. 	} else {
[...]

PR #2090, opened 2019-06-02
fix bug pointed out by issue #2089

irino:


irino:
I passed unit tests on the local environment, however CI timeouts. https://travis-ci.org/osrg/gobgp/jobs/540307053

fujita:
Thanks for the fix! I retried the test and it succeeded. The test sometimes fails. I guess that due to the overloaded systems or something, the timeout happens. Let's see what happens with the longer timeout.

PR #2100, opened 2019-06-12
Fixing a bottleneck in policy handling for community matching

### LINKED ISSUE #2099: BIG bottleneck within community matchin in routepolicys
Converging 30k prefixes over 15minutes and having issues with the api beeing inaccessable. 
While removing all the route-policys for the configuration; almost instant converge.

This is the call chain during the convergence using PPROF.
image

Culprit is located at [policy.go#L1679](https://github.com/osrg/gobgp/blob/master/internal/pkg/table/policy.go#L1679)
```
if x.MatchString(fmt.Sprintf("%d:%d", y>>16, y&0x0000ffff)) {
```

I will submit a PR to fix this.

mrevilme:
Fixes #2099

### DIFF
--- a/internal/pkg/table/policy.go
+++ b/internal/pkg/table/policy.go
@@ -1070,6 +1070,17 @@ func (lhs *regExpSet) Replace(arg DefinedSet) error {
 
 type CommunitySet struct {
 	regExpSet
+	intSet map[*regexp.Regexp]uint32
+}
+
+func (s *CommunitySet) BuildCache() {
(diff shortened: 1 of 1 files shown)


mrevilme:
My colleague had a good point about regex that can be used i missed that part in the documentation i will look into finding a better solution that preserves the Regexp matching capabilities.

fujita:
Please create a new pull request when you find a better way with the current matching capabilities.

Issue #2073, opened 2019-05-08
Question: Is there a reason why ipv4-flowspec is not supported for VRF enslaved neighbor


lechnertech:
I am using the gobgp as a route injector and have each one of my peers in their own VRF.  This works great for announcing ipv4-unicast and ipv6-unicast routes for Source/Dest RTBH.  I want to expand the functionality to support ipv4-flowspec and ipv6-flowspec but receive the error ipv4-flowspec is not supported for VRF enslaved neighbor.  Is there a reason why this is not supported?

Thanks in advance.

fujita:
I guess that nobody is interested in such feature. If you are, please send a pull request.

lechnertech:
I am testing changes and will submit a pull request soon.  Thanks

PR #2101, opened 2019-06-12
Support ipv4-flowspec for VRF enslaved neighbor

lechnertech:
#2073 

Added support for ipv4/ipv6-flowspec announcements/withdraws when a neighbor is configured in a vrf.


fujita:
Looks ok. pushed, thanks!

PR #2103, opened 2019-06-17
fix MPLS L3VPN functionality of zapi

irino:
This PR fixes lookup for pathVrfMap (introducing in commit 276518c8269a2403bf79cdeba7c33b644c238214) of zclient.go. 
Current source code fails lookup because the lookup key is a pointer. This PR uses Path.Equal() in for loop as a lookup method.

### DIFF
--- a/pkg/server/zclient.go
+++ b/pkg/server/zclient.go
@@ -118,9 +118,8 @@ func filterOutExternalPath(paths []*table.Path) []*table.Path {
 
 func addMessageLabelToIPRouteBody(path *table.Path, vrfId uint32, z *zebraClient, msgFlags *zebra.MESSAGE_FLAG, nexthop *zebra.Nexthop) {
 	v := z.client.Version
-	nhVrfId := z.pathVrfMap[path]
 	rf := path.GetRouteFamily()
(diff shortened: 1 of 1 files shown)


fujita:
pushed, thanks!

shengslin:
Thanks, but I think there are still two problems.
1. A BGP neighbor cannot be created from the vrf. The establishment of the neighbor needs to be initiated by the peer.
2. It is not possible to establish a neighbor with the same ip address in a different vrf.
If different users on the same host use vrf for isolation, then different users are likely to configure the same IP address to establish BGP neighbors. Since GoBGP already supports the vrf feature, this should be supported. I tested it with FRR and it works very well.

some log info:
```

{"Key":"1.1.1.1","Topic":"Peer","level":"debug","msg":"state changed","new":"BGP_FSM_ACTIVE","old":"BGP_FSM_IDLE","reason":{"Type":7,"BGPNotification":null,"Data":null},"time":"2019-06-02T12:26:08Z"}
{"Key":"1.1.1.1","Topic":"Peer","level":"debug","msg":"try to connect","time":"2019-06-02T12:26:27Z"}
{"Key":"1.1.1.1","Topic":"Peer","level":"debug","msg":"failed to connect: dial tcp 0.0.0.0:0-\u003e1.1.1.1:179: connect: no route to host","time":"2019-06-02T12:26:27Z"}
```

PR #2105, opened 2019-06-20
Python3 support to gobgp tests

franza73:
This PR removes dependencies on old Fabric version, as it's not
supported by Python3.

The current Fabric versions don't support the colors and indent
used previously, so we found substitute methods from other
libraries and defined these in the library files.

The local function from fabric is now just a wrapper to invoke's
run function. All the files were processed through 2to3 command.

All the tests were executed and we don't see any difference on
the outputs when running Python2 or Python3.

The creation of gobgp container is removed from base.py into
fabfile.py, in order to comply with Fabric2 changes and simplify
dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ _node_js: &_node_js
 
 _python: &_python
   language: python
-  python: "2.7"
+  python: "3.6"
 
 _docker: &_docker
   <<: *_python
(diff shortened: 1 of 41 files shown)


fujita:
Great, I really like to see this in the upstream, thanks a lot!
Errors happened on the CI tests. I would recommend you to set up your travis-ci account and create a pull request for your gobgp repository. After all the CI tests pass, you can create a pull request for the upstream repository.

Issue #2092, opened 2019-06-02
Gobgp does not support VRF very well


shengslin:
```
-----------                             ------------------------
|    CE    |---------------------------|----vrf1             PE |
-----------1.1.1.1/24                  |    1.1.1.2/24    GoBGP |
      FRR                               ------------------------

```
As shown in the figure above，I created an EBGP connection between CE and PE，The interface on PE is located in a VRF, FRR runs on CE and GoBGP runs on PE.
The configuration is as follows:

```
CE:
# do show running-config
Building configuration...
Current configuration:
!
frr version 7.0
frr defaults traditional
hostname vr-cpe2
no ipv6 forwarding
!
router bgp 10002
neighbor 1.1.1.2 remote-as 20001
!
address-family ipv4 unicast
  redistribute connected
exit-address-family
!
line vty
!
end

PE:
#ip -d link
6: vrf1: <NOARP,MASTER,UP,LOWER_UP> mtu 65536 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 32:90:9a:c4:a6:cf brd ff:ff:ff:ff:ff:ff promiscuity 0
    vrf table 100 addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535
16: eth0.100@eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master vrf1 state UP group default qlen 1000
    link/ether 52:54:00:55:18:7e brd ff:ff:ff:ff:ff:ff promiscuity 0
    vlan protocol 802.1Q id 100 <REORDER_HDR>
    vrf_slave table 100 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535
    inet 1.1.1.2/24 scope global eth0.100
       valid_lft forever preferred_lft forever

#cat /proc/sys/net/ipv4/tcp_l3mdev_accept
1

#cat /etc/gobgp.conf
[global.config]
    as=20001
    router-id="2.2.2.2"
 [zebra]
    [zebra.config]
        enabled=true
        url="unix:/var/run/frr/zserv.api"
        redistribute-route-type-list=["connect"]
        version=6

#gobgp global
AS:        20001
Router-ID: 2.2.2.2
Listening Port: 179, Addresses: 0.0.0.0, ::

#gobgp vrf
  Name                 RD                   Import RT            Export RT            ID
[...]

PR #2108, opened 2019-06-23
BGP connection on a speecified interface (including which is associat…

irino:
…ed VRF) using syscall SO_BINDTODEVICE


irino:
This PR enables bgp establishment on VRF interface. This functionality is requested on issue #2092.

fujita:
squashed and pushed. Thanks!

Issue #2095, opened 2019-06-05
Failed Proto Imports - Python gRPC


bgianpetro:
When generating the Python interface as specified in the documentation as follows:

```
python -m grpc_tools.protoc -I./api --python_out=. --grpc_python_out=. api/gobgp.proto api/attribute.proto api/capability.proto
```

I received errors about not being able to find the imports out of the 'gobgp.proto' file, like this:

>gobgp.proto: Import "any/any.proto" was not found or had errors.
>gobgp.proto: Import "empty/empty.proto" was not found or had errors.
>gobgp.proto: Import "timestamp/timestamp.proto" was not found or had errors.

I was able to fix this by modifying the references as follows in the api/*.proto files.

From this:
```
import "any/any.proto";
import "empty/empty.proto";
import "timestamp/timestamp.proto";
```

To this:
```
import "google/protobuf/any.proto";
import "google/protobuf/empty.proto";
import "google/protobuf/timestamp.proto";
```

I don't know if this is an issue with my Python library location (user-based) or if it is just a general Python thing.  If the fix would work for everything I can submit a PR.

fujita:
https://github.com/osrg/gobgp/blob/master/tools/grpc/genproto.sh#L32
You need to specify another include path like the above.

gustavooferreira:
Had exactly the same issue in python and finally managed to solve it.
Was going to open an issue but then realized there was one opened (and closed) already.

Including the path is not enough, at least not for python.

It will generate the files but python code won't work because of the wrong import paths.

This issue should be opened again!

Also, I don't see why the proto files have `any|empty|timestamp/*` when the correct path is `google/protobuf`.

Please reopen this issue.

fujita:
@gmelikov updated the doc. If it doesn't work for you, please update it.

gmelikov:
@gustavooferreira yes, this patch helped me:
```
diff --git a/api/attribute.proto b/api/attribute.proto
index 4cf844f..7bf9452 100644
--- a/api/attribute.proto
+++ b/api/attribute.proto
@@ -21,7 +21,7 @@

 syntax = "proto3";

-import "any/any.proto";
+import "google/protobuf/any.proto";
 import "gobgp.proto";

 package gobgpapi;
diff --git a/api/gobgp.proto b/api/gobgp.proto
index b61b475..52dd9da 100644
--- a/api/gobgp.proto
+++ b/api/gobgp.proto
@@ -21,9 +21,9 @@

[...]

PR #2116, opened 2019-07-16
gRPC docs: include all needed paths

gmelikov:
Closes: #2095.

We need to specify another include path.

Signed-off-by: George Melikov <g.melikov@corp.mail.ru>


fujita:
pushed, thanks.

Issue #2113, opened 2019-07-14
gobgp doesn't support VPNv6 over IPv4 (RFC 4798) ?


DKCopy:
Version: gobgp 2.6.0
I use gobgp connect to Huawei switch (CE12800). when I enable  VPNv6 over IPv4 in Huawei switch,  the gobgpd trigger a panic!
So the gobgp maybe can't support VPNv6 over IPv4 (RFC 4798) ?
 
### Huawei switch bgp config
```
bgp 200
 peer 1.1.1.9 as-number 200
 peer 1.1.1.9 connect-interface 3.3.3.9
 #
 ipv4-family unicast
  peer 1.1.1.9 enable
 #
 ipv4-family vpnv4
  policy vpn-target
  peer 1.1.1.9 enable
 #
 ipv4-family vpn-instance vpna
  import-route direct
 #
 ipv6-family vpnv6
  policy vpn-target
  peer 1.1.1.9 enable
 #
 ipv6-family vpn-instance vpna
  import-route direct
#

```

### gobgp log
```
DEBU[0017] received update                               Key=3.3.3.9 Topic=Peer attributes="[{Origin: ?}  {Med: 0} {LocalPref: 100} {Extcomms: [65040:1]} {MpReach(l3vpn-ipv6-unicast): {Nexthop: 3.3.3.9, NLRIs: [3.3.3.9:11:2001:9:9:9::/64]}}]" nlri="[]" withdrawals="[]"
DEBU[0017] create Destination                            Nlri="3.3.3.9:11:2001:9:9:9::/64" Topic=Table
DEBU[0017] computeKnownBestPath knownPathList: 1         Topic=Table
DEBU[0017] From same AS, ignore.                         Data="{ 3.3.3.9:11:2001:9:9:9::/64 | src: { 3.3.3.9 | as: 200, id: 3.3.3.9 }, nh: 3.3.3.9 }" Key=3.3.3.9 Topic=Peer
DEBU[0017] received update                               Key=3.3.3.9 Topic=Peer attributes="[{MpUnreach(l3vpn-ipv6-unicast): End-of-Rib}]" nlri="[]" withdrawals="[]"
DEBU[0017] EOR received                                  AddressFamily=l3vpn-ipv6-unicast Key=3.3.3.9 Topic=Peer
[...]

PR #2121, opened 2019-07-21
Avoid panic by receiving vpnv6 prefixes with zebra reported in is…

irino:
…sue #2113

- Additionally bug fix which avoid panic caused by displaying vpnv6 prefixes using gobgp vrf rib
- Supporting FRRouting 7.1.x


fujita:
great, thanks!

Issue #2046, opened 2019-04-05
event order across disjointed streaming RPCs


jwhited:
Starting with v2.0 watchers are no longer exported. If we were previously watching peer state changes and update messages on a single channel we now have to invoke 2 RPCs, MonitorPeer and MonitorTable. Using these 2 RPCs, event order is no longer guaranteed and bgpServer event time is a best guess. 

Time information could be included in the streaming responses and may be a good start, but event order still could not be guaranteed/resolved for init* responses (initUpdate, initPeerState, etc). Thoughts?

fujita:
We could have API like WatchEvent to register all kinds of events via a single channel. However, every peer has the own channel to send events (updates, state change, etc) to the main goroutine. Thus, there is no guarantee on the order of events on multiple peers even with such API. It works for you?

PR #2128, opened 2019-07-30
Do not supress initial non-established peer state in MonitorPeer

### LINKED ISSUE #2047: Feature Request: don't suppress initial non-established peer state in MonitorPeer

ecbaldwin:
The issue with the initial update seemed to be that sometimes the
connection has not yet been established. Trying to get the remote and
local connection data caused a segmentation violation because
`fsm.conn` was nil.

Initial testing seems ok but I have not yet run the full suite of tests.

fixes #2047


fujita:
I think that we should not change the MonitorPeer API behavior without the major version updated. Adding a new option to MonitorPeerRequest or something else?

ecbaldwin:
ok, that is fair. I'll update this with an additional to the MonitorPeerRequest which will enable this behavior.

fujita:
Thanks. I've just remembered the following issue then I'm inclining to add a new API to just send all peer states and meet the requirement of the following instead of adding options to the current API.
https://github.com/osrg/gobgp/issues/2046

fujita:
https://github.com/fujita/gobgp/tree/monitor-event

You get all the changes of peer state. If you don't need some, just ignore. The feedback is greatly appreciated.

fujita:
thanks

PR #2132, opened 2019-08-05
Added MPLS-in-UDP support to the CLI parser

peterjhill:
Added CLI option to parser to allow setting MPLSoUDP (mpls-in-udp) as an encap type.


peterjhill:
The encap type is already specified in the bgp.go, but was not added to the parser.

fujita:
pushed, thanks.

PR #2135, opened 2019-08-10
Support connecting zebra of cumulus linux

irino:


irino:
I will reopen this PR after check my code.

PR #2148, opened 2019-08-27
Add BGP Confederation documentation

### LINKED ISSUE #2146: Add BGP Confederation Documentation to GoBGP
Good afternoon,

I would like to open a discussion regarding adding / implementing support for BGP confederation within GoBGP. Based on the following ticket, https://github.com/osrg/gobgp/issues/1428, GoBGP does not currently support confederation. I would like to ask if:

1. This is something that was incorrectly stated (GoBGP does actually support confederation) and a pull request is required to add the supporting documentation so that others may identify how to use this feature / functionality.
2. This is something which is on the roadmap but not yet in a pull request, and does not require discussion / assistance
3. BGP Confederation is not on the roadmap and discussion about it hasn't been completed. At which point I'd ask is the project open to assistance in and/or supporting implementation of BGP confederation.  I'd be willing to try to assist with this, but would need support from the GoBGP team in terms of ensuring that this meets the project's requirements to ensure its maintained moving forward once accepted and implemented. 

## What is BGP Confederation?
In network routing, BGP confederation is a method to use Border Gateway Protocol (BGP) to subdivide a single autonomous system (AS) into multiple internal sub-AS's, yet still advertise as a single AS to external peers. The intent is to reduce iBGP mesh size. This is commonly performed within large Internet Service Providers (ISPs) in part to segment operations by team, region, or other factors.

## Why might a large provider need or want BGP confederation?
A confederated AS is composed of multiple ASs. Each confederated AS alone has iBGP fully meshed and has connections to other ASs inside the confederation. Even though these ASs have eBGP peers to ASs within the confederation, the ASs exchange routing as if they used iBGP. In this way, the confederation preserves next hop, metric, and local preference information. To the outside world, the confederation appears to be a single AS. With this solution, iBGP transit AS problems can be resolved as iBGP requires a full mesh between all BGP routers: large number of TCP sessions and unnecessary duplication of routing traffic. It may also be used to segment a network's responsibility by internal team, region, or other distinct properties.

kkirsche:
Fix #2146

This pull request adds documentation for BGP Confederation, a feature which allows multiple ASNs to be composed to act as though they are a single autonomous system.

Currently, I believe I have the initial configuration piece here. I was wondering if you had a specific program you use or design you prefer for topology diagrams and if you had a suggestion for how to add verification to this.

### DIFF
--- a/docs/sources/bgp-confederation.md
+++ b/docs/sources/bgp-confederation.md
@@ -1,0 +1,37 @@
+# BGP Confederation
+
+This page explains how to configure BGP confederation feature when BGP peers
+are part of a larger mesh representing a single autonomous system (AS).
+
+## Prerequisites
+
+Assume you finished [Getting Started](getting-started.md).
+
+## Contents
+
+- [Configuration](#configuration)
+
+## Configuration
+
+If AS30 is a confederation composed of AS65001 and AS65002, the confederation members must configure
+the following attributes to ensure GoBGP communicates in the correct manner with other member ASNs.
+Each confederated autonomous systems must configure the `[global.confederation.config]` with
+`enabled = true` and `identifier = 30`. The identifier parameter is used to designate what the
+confederation should present as it's ASN non-confederation members. Each member of the confederation
+must also configure `member-as-list` with a list of other ASNs which compose the confederation. For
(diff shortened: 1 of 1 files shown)


fujita:
Pushed, thanks! No specific tool for topology diagrams for GoBGP. Do you have any recommendations?

PR #2150, opened 2019-08-28
Check AfiSafi Graceful Restart configuration in NeedsResendOpenMessage

### LINKED ISSUE #2119: NeedsResendOpenMessage in internal/pkg/config/util.go does not check MpGracefulRestartConfig.Enabled to deteremine if new OPEN message sent
I found this looking for a way to enable graceful restart on-demand, for a planned restart. I only want GR set when explicitly enabled (eg: upgrade event). I posted this finding on the GoBGP slack channel and was confirmed by @fujita that the function should inspect the family GR settings when determining whether to hard reset and create a new session. While uncovering this bug doesn't get me any closer to my desired behavior in the system, I'm bringing this issue over to GH so it can be fixed.. The fix should be fairly straightforward; update the NeedsResendOpenMessage check to include inspecting family-level GR settings.  I might have time in the next week or two to submit a fix.

```
p := &api.Peer{
    GracefulRestart: &api.GracefulRestart{
        LocalRestarting:     false,
        NotificationEnabled: true,
        Enabled:             true,
        RestartTime:         120, 
        DeferralTime:        360, 
    },
           // ... snip ...
    AfiSafis: []*api.AfiSafi{
        {
                        MpGracefulRestart: &api.MpGracefulRestart{
                            Config: &api.MpGracefulRestartConfig{
                                  Enabled: false, // <-- Not checked in function.
                            },
                        },
                        Config: &api.AfiSafiConfig{
                            Family: &api.Family{
                                Afi:  api.Family_AFI_IP,
                                Safi: api.Family_SAFI_UNICAST,
                            },
                            Enabled: true,
                       },
                   },
          },
   }
```
NeedsResendOpenMessage overlooks checking family-level GR settings.

Please let me know if I've overlooked anything with this issue. Thanks

kkirsche:
Fix #2119

If I understood the issue correctly, this adds a check to see whether the AfiSafi slice contains an entry with enabled graceful restart on it to decide whether or not it should perform the open message resend.

This also adds a corresponding unit test to validate that it correctly identifies true and false scenarios. 

Please let me know if I misunderstood or missed any aspect of what this issue required.

### DIFF
--- a/internal/pkg/config/util.go
+++ b/internal/pkg/config/util.go
@@ -221,11 +221,22 @@ func isAfiSafiChanged(x, y []AfiSafi) bool {
 	return false
 }
 
+func isAfiSafiGracefulRestartEnabled(a []AfiSafi) bool {
+	for _, e := range a {
+		if e.MpGracefulRestart.Config.Enabled {
+			return true
+		}
+	}
+
+	return false
+}
+
 func (n *Neighbor) NeedsResendOpenMessage(new *Neighbor) bool {
 	return !n.Config.Equal(&new.Config) ||
 		!n.Transport.Config.Equal(&new.Transport.Config) ||
 		!n.AddPaths.Config.Equal(&new.AddPaths.Config) ||
 		!n.GracefulRestart.Config.Equal(&new.GracefulRestart.Config) ||
+		!isAfiSafiGracefulRestartEnabled(n.AfiSafis) ||
 		isAfiSafiChanged(n.AfiSafis, new.AfiSafis)
 }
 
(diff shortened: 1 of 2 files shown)


fujita:
https://travis-ci.org/osrg/gobgp/jobs/577660662

`go fmt` command returns errors. Please fix the code style. Others might happens due to the CI env. Once you fix the style and push a fix, they might be successful.

fujita:
Only when the gr configuration is changed, the session needs to be reset?
https://github.com/osrg/gobgp/commit/2457948919d2ed295904adfd4de61409ed78c7e2

kkirsche:
I’m not sure I follow, could you elaborate? As I read the comment without additional detail, that doesn’t sound accurate to me, as I would assume the graceful restart configuration itself changing is only indicative of that functionality while this is related to the behavior when that is enabled, not on changes to itself

fujita:
This function is used to if see whether reseting the established session is necessary or not after the configuration of the peer is updated. If the gr configuration is changed, we need to reset the established session to tell the peer about the new configuration. If not, we don't need to tell.

PR #2154, opened 2019-09-01
Fix nexhop overwrite with gobgp mrt inject --no-ipv6 --nexthop x.x.x.x

amitn:
Overwrite only the nexthop attribute and not all attributes

Signed-off-by: Amit Nishry <amit.nishry@gmail.com>


fujita:
pushed, thanks.

PR #2170, opened 2019-09-30
Fix crash on nonexistent vrf delete

mageshgv:


fujita:
Pushed, thanks a lot!

PR #2175, opened 2019-10-04
Maintain unique vrf IDs in watchEventBestPath vrf map

mageshgv:
Fixes missing route propagation to proper vrfs in zapi when we have route import/export from different RTs


fujita:
Pushed, thanks a lot!

PR #2178, opened 2019-10-08
Support vrfs in zapi multipath

### LINKED ISSUE #2177: Feature combination of multipath and vrfs do not work with zebra
When we enable multipath along with vrfs, the routes are sent to zebra with vrf ID as 0. This causes the routes to be not installed in the relevant vrfs. (If the nexthops resolve on the global vrf, zebra might install the route in the wrong table)

mageshgv:
Fixes #2177

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -172,6 +172,9 @@ matrix:
         - TEST=zapi_v3_test.py FROM_IMAGE=osrg/quagga:v1.0
     - <<: *_docker
       env:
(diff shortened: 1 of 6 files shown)


fujita:
Thanks! Better to add the new test to .travis.yml?

fujita:
Looks like the temporary network error. Now CI is successful. Pushed, thanks a lot!

PR #2179, opened 2019-10-10
Do not install route with originator ID as self in adj-in

mageshgv:
Fixes route loop on softreset

### DIFF
--- a/pkg/server/peer.go
+++ b/pkg/server/peer.go
@@ -460,7 +460,7 @@ func (peer *peer) handleUpdate(e *fsmMsg) ([]*table.Path, []bgp.RouteFamily, *bg
 	if len(e.PathList) > 0 {
 		paths := make([]*table.Path, 0, len(e.PathList))
 		eor := []bgp.RouteFamily{}
-		for _, path := range e.PathList {
+		for i, path := range e.PathList {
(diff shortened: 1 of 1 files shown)


mageshgv:
While this is a most simple fix for the issue, we do lose track of the received route count. I am not sure if this is alright, if we want to have the route and not accept it, it needs careful handling in multiple cases (softrest, gr, llgr, policy evaluations, table counters etc). @fujita  what do you think is the right solution.

On a different note, it looks like the counters are already broken without this change if I add a second RR here: https://github.com/osrg/gobgp/blob/master/test/scenario_test/route_reflector_test.py#L232 (received, accepted and advertized count in neighbor and also the adj-out display itself with add path is wrong)

fujita:
we should keep such route in adj-in but not send it to the global rib?

fujita:
I'll clean up a way to filter routes in adj-in before the global.

fujita:
I like that the change of aspath option requires the session reset. Then we don't need to reevaluate paths in adj-in on softreset.
https://github.com/osrg/gobgp/pull/2188

mageshgv:
@fujita sorry for the late response. Thanks for the fix #2188 which looks more generic.  Let me close this pr in favor of yours.

PR #2180, opened 2019-10-11
Remove policy from FSM

ecbaldwin:
The `policy` member of fsm is apparently never used. Clean it up to
avoid confusion. Policy is still set on the server and peers.


fujita:
Thanks! Surely, this should have done when IN policy was removed.

PR #2181, opened 2019-10-11
Fix potential policy deadlock

ecbaldwin:
The first commit fixes the deadlock and changes Lock/Unlock to immediately defer instead of splitting them up. That's where return statements sneak in and cause deadlocks.

The second commit is just an optimization that I notice that isn't really related to the deadlock.

### DIFF
--- a/internal/pkg/table/policy.go
+++ b/internal/pkg/table/policy.go
@@ -3440,18 +3440,24 @@ func (r *RoutingPolicy) reload(c config.RoutingPolicy) error {
 }
 
 func (r *RoutingPolicy) GetDefinedSet(typ DefinedType, name string) (*config.DefinedSets, error) {
-	r.mu.RLock()
+	dl, err := func() (DefinedSetList, error) {
+		r.mu.RLock()
+		defer r.mu.RUnlock()
 
-	set, ok := r.definedSetMap[typ]
-	if !ok {
-		return nil, fmt.Errorf("invalid defined-set type: %d", typ)
-	}
+		set, ok := r.definedSetMap[typ]
+		if !ok {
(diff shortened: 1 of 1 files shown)


internal/pkg/table/policy.go
```diff
@@ -3586,16 +3593,19 @@ func (r *RoutingPolicy) DeleteStatement(st *Statement, all bool) (err error) {
 }
 
 func (r *RoutingPolicy) GetPolicy(name string) []*config.PolicyDefinition {
-	r.mu.RLock()
+	ps := func() Policies {
+		r.mu.RLock()
+		defer r.mu.RUnlock()
```
ecbaldwin: This one is not fixing an existing dead lock. However, I like using the pattern of immediately deferring the unlock after locking. That way, if a future code changes introduces a return statement or something, there is no chance that we can forget to unlock the lock. To maintain the scope of the lock the same as the original, I added this anonymous function.

internal/pkg/table/policy.go
```diff
@@ -3440,18 +3441,24 @@ func (r *RoutingPolicy) reload(c config.RoutingPolicy) error {
 }
 
 func (r *RoutingPolicy) GetDefinedSet(typ DefinedType, name string) (*config.DefinedSets, error) {
-	r.mu.RLock()
+	dl, err := func() (DefinedSetList, error) {
```
ecbaldwin: This anonymous function allowed me to call `defer r.mu.Runlock()` immediately after locking but still keeping the original scope of the lock. In other words, it still unlocks before calling sort.Sort(dl) and finishing the method as it did before.

fujita:
Pushed, thanks a lot! One dead lock bug was fixed and other cleanups look fine.

PR #2183, opened 2019-10-16
Fix adj-out display with add path enabled

mageshgv:


test/lib/gobgp.py
```diff
@@ -275,12 +275,15 @@ def monitor():
...
-    def _get_adj_rib(self, adj_type, peer, prefix='', rf='ipv4'):
+    def _get_adj_rib(self, adj_type, peer, prefix='', rf='ipv4', add_path_enabled=False):
         peer_addr = self.peer_name(peer)
         cmd = 'gobgp neighbor {0} adj-{1} {2} -a {3} -j'.format(peer_addr,
                                                                 adj_type,
                                                                 prefix, rf)
         output = self.local(cmd, capture=True)
+        if add_path_enabled:
+            return self._get_rib(json.loads(output))
+
```
mageshgv: This is the issue with this test function:
`ret = [p[0] for p in json.loads(output).values()]`
It only returns the first path. We should change this, but it involves changing a lot of tests. So I have added a new parameter for this PR.

internal/pkg/table/adj.go
```diff
@@ -89,6 +89,43 @@ func (adj *AdjRib) Update(pathList []*Path) {
 	}
 }
 
+/* The provided pathList is expected to be the real candidate routes after policy evaluation.
+   For routes that are filtered by policy, there could be a mismatch between display
+   and actual rib sent to the peer (if softreset out was not run).
+   Only used to display adj-out because we do not maintain a separate adj-out table
+*/
+func (adj *AdjRib) UpdateAdjRibOut(pathList []*Path) {
```
fujita: adj.Update() does the unnecessary work (handing withdraw, asloop check, etc) so you want to add this function that does only work necessary for adj-out? Or this fixes some bugs in Update() for  adj-out?

mageshgv: The main difference is that adj.Update compares the paths based on PathIdentifier which is correct for adj-Rib-in whereas this function uses PathLocalIdentifier for matching existing paths which should be the behavior for adj-Rib-out.

fujita: Ah, understood. Thanks!

fujita:
pushed, thanks a lot!

internal/pkg/table/adj.go
fujita: > PathLocalIdentifier for matching existing paths

This could actually happen?

PR #2184, opened 2019-10-17
Transition to graceful restart state on hold timer expiry if applicable

### LINKED ISSUE #2174: Question on GR/LLGR
In the current implementation in gobgp, GR/LLGR state transition only happens if TCP session fails (literally remote peer restarting would be the only scenario that would see a tcp layer failure due to tcp reset. We are not setting any timeouts on recv/send, so we will not be hitting any timeout errors on tcp unless hold timer is extremely high). When gobgp runs into hold timer expiry, it clears all the routes from the peer ignoring GR and LLGR.
https://github.com/osrg/gobgp/blob/master/pkg/server/fsm.go#L1807
https://github.com/osrg/gobgp/blob/master/pkg/server/server.go#L1309

While the graceful restart/llgr RFCs do not explicitly mention hold timer expiry, for practical purposes llgr is more useful for the scenarios where the bgp peer is not reachable(not necessarily process/os restart) and we want to work in pure headless mode.

Is the current implementation of ignoring GR on hold timer expiry in gobgp by design or is this a bug ? 

For reference, BIRD transitions to graceful restart state on hold timer expiry, thereby preserving the routes. (Not sure about Cisco/Juniper, but will try to verify this on Cisco if possible).

mageshgv:
Fixes #2174

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -1790,17 +1790,20 @@ func (h *fsmHandler) established(ctx context.Context) (bgp.FSMState, *fsmStateRe
 			// it now.
 			h.outgoing.In() <- err
 			fsm.lock.RLock()
-			if s := fsm.pConf.GracefulRestart.State; s.Enabled &&
-				(s.NotificationEnabled && err.Type == fsmNotificationRecv ||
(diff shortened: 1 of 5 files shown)


fujita:
pushed, thanks a lot!

PR #2188, opened 2019-10-20
avoid installing routes with originator ID to global when softresetin

fujita:


mageshgv:
@fujita thanks for the fix. This patch looks good for the fix.

fujita:
@mageshgv thanks! pushed.

PR #2193, opened 2019-10-21
table: remove unnecessary code in UpdateAdjRibOut()

fujita:
There is no duplication of local id in the rib.


fujita:
@mageshgv what do you think?

fujita:
pushed.

PR #2196, opened 2019-10-22
Split Reset into three methods

ecbaldwin:
The Reset method was difficult to understand. The reason is that it
was called in three different ways and did different things in each
case. It is easier to read when the three different modes are each
their own method.

This came up as I was looking deeper into the threading model around
policies. I think this change makes it easier to understand the code.


fujita:
Looks fine. Thanks! I fixed the staticcheck error and pushed.

PR #2202, opened 2019-10-26
Update python instructions to generate valid grpc interfaces

gmelikov:
Actualize docs, see https://github.com/osrg/gobgp/issues/2095

Signed-off-by: George Melikov <mail@gmelikov.ru>


fujita:
The paths was fixed so let's close this.
https://github.com/osrg/gobgp/pull/2209

PR #2205, opened 2019-10-30
bgp: fix dynamic neighbors with prefix "0.0.0.0/0"

faickermo:
support prefix "0.0.0.0/0".


fujita:
pushed, thanks!

PR #2207, opened 2019-11-02
config: peer groups support add-paths

faickermo:
Support config '[peer-groups.add-paths.config]'.
Neighbor add-paths is set or from peer-groups and saves to neighbor AfiSafis
after reading config file.
After config->api->config, the AddPaths of neighbor and peer-groups is lost.
Neighbor config is set again in func addNeighbor when a peer is added.
The neighbor AfiSafis will be overwritten in func
OverwriteNeighborConfigWithPeerGroup even though it has values.


fujita:
The intention is that (neighbor|peer-groups).add-paths.config is not recommended (should be deprecated in the next major version). addpath in peer-groups.afisafis is preferable.

faickermo:
I see.
close this.

PR #2209, opened 2019-11-06
Use correct import paths for protobuf packages

### LINKED ISSUE #2095: Failed Proto Imports - Python gRPC

ecbaldwin:
The packages `any`, `empty`, and `timestamp` ship with protoc so there
is no need to pull them out of the go package. This simplifies the
generation script and also corrects the import paths for the standard
protobuf types.

Fixes: #2095


tools/grpc/genproto.sh
```diff
@@ -1,9 +1,10 @@
-#!/usr/bin/env bash -x
```
ecbaldwin: I removed the `-x` from here and moved it below because this script wouldn't run for me otherwise. I got this: 

```
$ ./tools/grpc/genproto.sh
/usr/bin/env: ‘bash -x’: No such file or directory
```

It is because the shebang line only allows a single argument after the executable. It interpreted `bash -x` as the argument to `env` and `env` couldn't find an executable by that name.

tools/grpc/genproto.sh
```diff
@@ -26,9 +27,7 @@ for pkg in ${INSTALL_PKGS}; do
 done
 
 GOBGP="${PWD}"
-GOPROTO="$(GO111MODULE=on go list -f '{{ .Dir }}' -m github.com/golang/protobuf)"
 
 echo "generating code"
-protoc -I "${GOPROTO}"/ptypes \
```
ecbaldwin: There is no need to add this import path since these types ship with the protoc compiler under their proper include paths (google/protobuf). So, removing it.

fujita:
I thought that this breaks the compatibility so I need to postpone it until v3. But it seems that it doesn't. Pushed, thanks a lot!

PR #2213, opened 2019-11-08
policy: add nexthop unchanged

faickermo:
Support config set-next-hop = "unchanged" and also cli command.


mrueg:
@fujita can you take a look at this PR if you have time?

fujita:
The config file also needs to support this feature.

faicker:
@fujita Maybe no need to change the config file. It keeps the same with the `set-next-hop = "self"` usage.

fujita:
Oops, `set-next-hop = "unchanged"` in config file works. Merged, thanks a lot!

Issue #1952, opened 2019-01-28
Tests are failing with more recent versions of google/protobuf


vincentbernat:
Hey!

When regenerating `*.pb.go` files with a more recent version, many tests are failing:

```
--- FAIL: Test_AddPathCapability (0.00s)
    capability_test.go:187:
                Error Trace:    capability_test.go:187
                Error:          Not equal:
                                expected: &gobgpapi.AddPathCapability{Tuples:[]*gobgpapi.AddPathCapabilityTuple{(*gobgpapi.AddPathCapabilityTuple)(0xc0002b5f20)}, XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:10}
                                actual  : &gobgpapi.AddPathCapability{Tuples:[]*gobgpapi.AddPathCapabilityTuple{(*gobgpapi.AddPathCapabilityTuple)(0xc0002fa0c0)}, XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}

                                Diff:
                Test:           Test_AddPathCapability
```

This is due to the addition of `XXX_sizecache`. Authors suggest to use `proto.Equal()` instead of `reflect.DeepEqual()`. However, as the test is done through `assert.Equal()`, I am unsure of what the right solution is.

Testify has an issue and a PR to use `go-cmp` instead, but I don't think it would change anything since the structures don't come with an `Equal()` method. We could either have a wrapper around `assert.Equal()` to use `proto.Equal()` if possible first. Or just use a dedicated helper when we compare proto messages. The first approach seems less error-prone.

fujita:
Thanks for the report. What's the version of protobuf?

Due to the following golang bug, I can't update the dependency for the newer protobuf but I'll try after it's fixed. 
https://github.com/golang/go/issues/30833

vincentbernat:
This is with 1.2.0.

vincentbernat:
Hey! Any news on that?

fujita:
I tend to simply update the test code use `proto.Equal()` like
```
@@ -64,7 +65,9 @@ func Test_AsPathAttribute(t *testing.T) {
 
        output := NewAsPathAttributeFromNative(n.(*bgp.PathAttributeAsPath))
        assert.Equal(2, len(output.Segments))
-       assert.Equal(input.Segments, output.Segments)
+       for i := 0; i < len(input.Segments); i++ {
[...]

PR #2216, opened 2019-12-15
tests: fix failing tests when using more recent versions of protobuf

vincentbernat:
When regenerating the `*.pb.go` file, we get many tests failing
because testify is using `reflect.DeepEqual()` under the hood. There
is no option to have a custom comparison function and currently, the
only way around is to use `proto.Equal()` to compare protobuf
messages. The downside of using `proto.Equal()` is that we loose the
expected/actual display in case of a mismatch.

When comparing family, we compare `Api`/`Sapi` in separate asserts
instead of using `proto.Equal()`.

Fix: #1952

If you prefer to keep the more descriptive error messages, I can also maintain the patch myself for Debian use. Hopefully, testify will come with a better solution in the future.

### DIFF
--- a/internal/pkg/apiutil/attribute_test.go
+++ b/internal/pkg/apiutil/attribute_test.go
@@ -19,6 +19,7 @@ import (
 	"net"
 	"testing"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
 	api "github.com/osrg/gobgp/api"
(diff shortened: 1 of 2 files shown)


fujita:
thanks! update go.mod and go.sum too?

vincentbernat:
This would mean also regenerating `*.pb.go`?

fujita:
updating the dependency as follows (and *.pb.go)

https://github.com/fujita/gobgp/commit/9a93ed4cef8a2e73e31bc72202d17715de14b359

But debian doesn't need this change, I guess. I don't have a reason to update grpc so I don't push the above.

PR #2217, opened 2019-12-15
contrib: simplify instructions for systemd

vincentbernat:
First, using setcap is not used when systemd file contains
AmbientCapabilities=CAP_NET_BIND_SERVICE.

Second, by using DynamicUser= directive, we can avoid creating a user
along with the associated directories. systemd does that itself. This
also has the advantage on working with any distribution based on
systemd.

### DIFF
--- a/tools/contrib/centos/README.md
+++ b/tools/contrib/centos/README.md
@@ -1,4 +1,4 @@
-# GoBGP systemd Integration for CentOS
+# GoBGP systemd integration
 
 The following document describes how to manage `gobgp` with `systemd`.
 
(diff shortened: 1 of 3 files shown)


growse:
I was going to raise a separate issue on this, but I've just noticed `DynamicUser=yes` pop up in Debian Unstable on `v2.11.0-1` (this might be a debian-specific change? I can't find reference in this codebase). The problem with this is now `gobgpd` doesn't have permissions to talk to the local zebra `zerv.api` socket, so isn't able to update the local routing table.

vincentbernat:
Yes, this is a Debian-specific modification (also, the paths are not the same as we use `/etc/gobgpd.conf` in Debian, not `/etc/gobgp/gobgpd.conf`).

I didn't think about the Zebra part when doing that. I don't know if there is an easy solution with systemd for this purpose. So, maybe it's better to keep a local user instead of a dynamic one. I'll update the Debian package accordingly on the next update.

growse:
Thanks.

For the moment I've overridden the unit file and am running gobgpd as `frr`. The zebra socket seems to only be read/writable by its owner, so I'll need to read the manual a bit more to figure out how to make it group rw.

fujita:
should I merge this or drop?

vincentbernat:
Let's close it and keep it as a reference. I'll open another one to remove the `setcap` which is not needed.

PR #2218, opened 2019-12-18
contrib: simplify instructions for systemd

vincentbernat:
Using setcap is not used when systemd file contains
AmbientCapabilities=CAP_NET_BIND_SERVICE.


fujita:
thanks!

Issue #2223, opened 2020-01-04
RPKI validation status of show rib command is unstable


taruta811:
When RPKI validation is enabled, the display of RPKI validation status changes every time `gobgp global rib` is executed

## Actual behavior

```
root@R1:~# gobgp rpki server
Session                 State  Uptime     #IPv4/IPv6 records
192.168.100.11:8282     Up     00:03:45   2/0

root@R1:~# gobgp rpki table
Network            Maxlen AS         Server
192.168.0.1/32     32     65010      192.168.100.11:8282
192.168.100.0/24   24     65000      192.168.100.11:8282

root@R1:~# for i in `seq 3`; do echo "------" ;gobgp global rib; done
------
   Network              Next Hop             AS_PATH              Age        Attrs
V*>192.168.0.1/32       192.168.10.1         65000                00:04:13   [{Origin: i} {Med: 1}]
N*>192.168.0.2/32       0.0.0.0                                   00:04:18   [{Origin: i} {Med: 1}]
N*>192.168.10.0/30      0.0.0.0                                   00:04:18   [{Origin: i} {Med: 1}]
I* 192.168.10.0/30      192.168.10.1         65000                00:04:13   [{Origin: i} {Med: 1}]
N*>192.168.100.0/24     192.168.10.1         65000                00:04:13   [{Origin: i} {Med: 1}]
------
   Network              Next Hop             AS_PATH              Age        Attrs
I*>192.168.0.1/32       192.168.10.1         65000                00:04:13   [{Origin: i} {Med: 1}]
N*>192.168.0.2/32       0.0.0.0                                   00:04:18   [{Origin: i} {Med: 1}]
N*>192.168.10.0/30      0.0.0.0                                   00:04:18   [{Origin: i} {Med: 1}]
N* 192.168.10.0/30      192.168.10.1         65000                00:04:13   [{Origin: i} {Med: 1}]
V*>192.168.100.0/24     192.168.10.1         65000                00:04:13   [{Origin: i} {Med: 1}]
------
   Network              Next Hop             AS_PATH              Age        Attrs
N*>192.168.0.1/32       192.168.10.1         65000                00:04:13   [{Origin: i} {Med: 1}]
V*>192.168.0.2/32       0.0.0.0                                   00:04:18   [{Origin: i} {Med: 1}]
N*>192.168.10.0/30      0.0.0.0                                   00:04:18   [{Origin: i} {Med: 1}]
[...]

PR #2224, opened 2020-01-04
gobgpd: use map for validation table

taruta811:
`Table.destinations` is a map, so the order changes.
This commit allows getting validate information from the path by using map for validate table

Fix: https://github.com/osrg/gobgp/issues/2223

Signed-off-by: Toshiki Tsuchiya <taruta0811@gmail.com>


fujita:
great, thanks a lot!

PR #2227, opened 2020-01-10
Parse BMP Stats Report messages with unknown Stat Types

lechnertech:
While parsing Open BMP messages found some devices that send unknown stat types.  Added code to parse them correctly instead of failing.  Per RFC

A BMP implementation MUST ignore unrecognized stat types on receipt, and likewise MUST ignore unexpected data in the Stat Data field.

Included an example packet in bmp_test.go


fujita:
thanks!

PR #2237, opened 2020-02-11
local CLUSTER_ID check in CLUSTER_LIST.

kishiguro:
local CLUSTER_ID check in CLUSTER_LIST should be performed to non route reflector client as well.
otherwise multi cluster configuration may have routing loop.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -505,6 +505,9 @@ func filterpath(peer *peer, path, old *table.Path) *table.Path {
 			if info.RouteReflectorClient {
 				ignore = false
 			}
+			if peer.isRouteReflectorClient() {
+				ignore = false
+			}
 			// RFC4456 8. Avoiding Routing Information Loops
 			// If the local CLUSTER_ID is found in the CLUSTER_LIST,
 			// the advertisement received SHOULD be ignored.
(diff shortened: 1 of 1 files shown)


fujita:
>local CLUSTER_ID check in CLUSTER_LIST should be performed to non route reflector client as well.

Hmm, with the following description of rfc4456, I thought that only RR uses this attribute: 

>When an RR reflects a route, it MUST prepend the local CLUSTER_ID to
   the CLUSTER_LIST.  If the CLUSTER_LIST is empty, it MUST create a new
   one.  Using this attribute an RR can identify if the routing
   information has looped back to the same cluster due to
   misconfiguration. 

Looks like that bird does the same:  
https://github.com/CZ-NIC/bird/blob/master/proto/bgp/attrs.c#L1376

fujita:
This is now handled on the receive side. Commit 544c4bea ("server: reject local cluster loops on import") checks the CLUSTER_LIST of a received path against the cluster IDs of all route-reflector-client neighbors, and rejects the path before best path selection. RFC 4456 Section 8 says "the advertisement received SHOULD be ignored", so
that is the right place for it.

Closing. Please reopen if you still see a loop with the current code.

Issue #2235, opened 2020-02-05
get table info from vrf


pvorontsovd:
is it possible get routes count from vrf?

from cli i can do `gobgp global rib summary` of similar grpc method `GetTable`
but this works only on global table
i want get similar from vrf

pvorontsovd:
i try do this https://github.com/pvorontsovd/gobgp/commit/30fb2e019c3287ec9a6c13b5f885b007dab76a80

fujita:
can you create a pull request? thanks!

PR #2238, opened 2020-02-18
add support GetTable method for vrfs #2235

pvorontsovd:


pvorontsovd:
tests fails because in docker container disable ipv6 network, but in test it used

fujita:
travis-ci updated the version of docker it broke the ci. I fixed it and pushed these commits. Thanks!

PR #2249, opened 2020-03-09
fixed style: unnecessary type cast after type switch and some minor issues

serejkus:
The main thing is eliminating double type casts (one in type switch case, another in type switch body).


fujita:
thanks!

PR #2251, opened 2020-03-14
optimized allocations in packet/bgp

serejkus:
* used stack-based buf for serializing numbers
* added capacity to buffer allocations
* replaced `fmt.Errorf("string without any %fmt directive")` with `errors.New()`
* replaced `fmt.Sprint(a, b)` with `a + b`


fujita:
pushed, thanks!

PR #2252, opened 2020-03-15
optimized allocations in packet/bgp validations

serejkus:
* optimized comparision of LargeCommunities: it compared strings, now it compares uint32 values. I hadn't replaced O(n^2) comparision due to unlikeliness of a significant length of this list.
* got rid of possible double type cast in `ValidateUpdateMsg()`
* used preallocation in `BitmaskFlagOp.String()`
* fixed misspell


fujita:
thanks!

PR #2253, opened 2020-03-15
optimized packet/mrt allocations, mostly by reuse of error values

serejkus:
* stopped creating error value with `fmt.Errorf`on each `DecodeFromBytes()` operation.
* used preallocation in `RibEntry.Serialize()`. It can be optimised even more with allocating a full buffer at the function start, but it comes at cost of a more complex code, so I decided to drop it for the moment
* replaced some unnecessary `fmt.Errorf()` with `errors.New()`


fujita:
thanks!

PR #2268, opened 2020-04-25
Refactoring Zebra and supporting frr7.3

irino:
- Change const values to camel case from snake case to make zero golint's warnigns
- Introduce convert functions (toEach, toCommon) between the newest version to older versions
- Merge code about decode nexthop to reduce lines of code
- Add frr7.3 statemet in allowable software for config
- Rename and Update generated file by stringer


fujita:
great, thanks a lot!

Issue #2265, opened 2020-04-23
bgpattrtype_string.go out of date


dnikulin:
`BGPAttrType` itself has had several commits since the file was auto-generated, but none of those commits re-generated the corresponding stringer code. This means several types do not have corresponding names, even though they appear in public BGP sessions.

fujita:
I'm not sure what you mean. BGPAttrType is defined in bgp.go, which isn't auto-generated.

dnikulin:
That's right, but `bgpattrtype_string.go` (in the issue title) is out of date with respect to bgp.go
None of the last several changes to `BGPAttrType` have included corresponding changes to `bgpattrtype_string.go` (generated by stringer)

fujita:
Ah, understood. Can you create a pull request for updating?

dnikulin:
I am very happy to do that, but even though it's trivial, it will require undertaking a legal process with my employer. I'll try to get that started and get back to you.

PR #2297, opened 2020-06-17
Add //go:generate stringer directives for public enum types.

dnikulin:
All such directives can be invoked by running
$ go generate

Stringer itself can be obtained by running
$ go get golang.org/x/tools/cmd/stringer

See https://github.com/osrg/gobgp/issues/2265


fujita:
pushed, thanks!

PR #2298, opened 2020-06-17
Regenerate stringer code for public enum types.

dnikulin:
This adds support for several values which were added to enums without
corresponding changes to the stringer generated code.

New versions of stringer, including the one used for this change, add
compile-time checks for whether any existing enum values have changed.
Adding new values does not violate this check.

Resolves https://github.com/osrg/gobgp/issues/2265


fujita:
pushed, thanks!

PR #2303, opened 2020-07-12
implement LsAddrPrefix marshal json function

### LINKED ISSUE #2293: this function return nil ,correct or make mistake ?
type LsAddrPrefix struct {
	PrefixDefault
	Type   LsNLRIType
	Length uint16
	NLRI   LsNLRIInterface
}

func (l *LsAddrPrefix) MarshalJSON() ([]byte, error) {
	return nil, nil
}
>>>>why this function returns nil  ? should it be unusefull  ?

JieJhih:
implement LsAddrPrefix marshalJson() function
Fix #2293


fujita:
pushed, thanks.

PR #2305, opened 2020-07-26
zebra: refactoring and fix bug and add processing messageBackupNextho…

irino:
…p (added in frr7.4)

- Move function for checking softwareName from internal/pkg/config/default.go to internal/pkg/zebra/zapi.go.
- Delete support for frr7.1.
- Fix bug by inverse processing MessageMTU and messageTag, and add messageTableId in decodeFromBytes and serialize of IPRouteBody.
- Add processing messageBackupNexthop (which is added in frr7.4) in decodeFromBytes and serialize of IPRouteBody.
- Update document.


fujita:
Great, thanks!

PR #2306, opened 2020-07-28
api: add ls identifier and protocol ID attributes

vma:
The LS identifier and Protocol ID are missing currently from the grpc API.
This patch adds them to `LsNodeNLRI` message:
- `ls_protocol_id` is the named LS protocol (`ISIS-L1`, `ISIS-L2`, `OSPFv2`, ...)
- `identifier` is the LS node ID


fujita:
These information are missing in LsLinkNLRI, LsPrefixV4NLRI, and LsPrefixV6NLRI?

vma:
Hi, you are correct: in bgp.go, all theses types are derived from base type `LsNLRI` which contains `LsProtocolID` and `Identifier` attributes but the corresponding api types does not have this entry. For example,

- in bgp.go, `LsLinkNLRI` is defined as
```
type LsLinkNLRI struct {
	LsNLRI
	LocalNodeDesc  LsTLVInterface
	RemoteNodeDesc LsTLVInterface
	LinkDesc       []LsTLVInterface
}
```
- and in attribute.proto we have:
```
message LsLinkNLRI {
  LsNodeDescriptor local_node = 1;
  LsNodeDescriptor remote_node = 2;
  LsLinkDescriptor link_descriptor = 3;
}
```

So would you agree if I define `LsNLRI` type in the proto file and add it as first entry of `LsNodeNLRI`, `LsLinkNLRI`, `LsPrefixV4NLRI`, and `LsPrefixV6NLRI`?

vma:
Here is a new patch, I have added the LS fields to LsAddrPrefix so that it is distributed to all derived types (LsNodeNLRI, LsLinkNLRI, LsPrefixV4NLRI, and LsPrefixV6NLRI)

api/attribute.proto
```diff
@@ -309,16 +309,29 @@ enum LsNLRIType {
...
+  uint32 length = 2;
+  LsProtocolID protocol_id = 3;
+  uint64 identifier = 4;
   // One of:
   // - LsNodeNLRI
   // - LsLinkNLRI
   // - LsPrefixV4NLRI
   // - LsPrefixV6NLRI
-  google.protobuf.Any nlri = 2;
+  google.protobuf.Any nlri = 5;
```
fujita, line 334: This breaks the compatibility?

vma: Correct, I will update the PR to keep the same numbers for existing fields

vma: done in last commit 1366db0

fujita:
Sorry about the delay. Pushed.

PR #2309, opened 2020-08-04
best path selection - compare neigh addr

### LINKED ISSUE #2308: non-deterministic path selection w/equal router ID
path selection implementation appears to be incomplete and stop at router ID comparison:

https://github.com/osrg/gobgp/blob/v2.17.0/internal/pkg/table/destination.go#L483

`compareByRouterID()` will return `nil, nil`. In the case where router IDs are equal this leads to non-deterministic path selection. 

This can occur in a topology where there are 2 GoBGP speakers, with 2 eBGP sessions built between them over independent links.

[compareByRouterID()](https://github.com/osrg/gobgp/blob/v2.17.0/internal/pkg/table/destination.go#L883
) references [RFC5004](https://tools.ietf.org/html/rfc5004#section-3), which instructs:

```
In addition, the algorithm SHOULD NOT be applied when both paths are
from peers with an identical BGP identifier (i.e., there exist
parallel BGP sessions between two BGP speakers).  As the peering
addresses for the parallel sessions are typically allocated by one AS
(possibly with route selection considerations), the algorithm (if
applied) could impact the existing routing setup.  Furthermore, by
not applying the algorithm, the allocation of peering addresses would
remain as a simple and effective tool in influencing route selection
when parallel BGP sessions exist.
```

I believe neighbor address comparison needs to be implemented to satisfy [RFC4271-section9.1.2.2](https://tools.ietf.org/html/rfc4271#section-9.1.2.2)

jwhited:
This PR adds a `compareByNeighborAddress()` path selection function that implements RFC 4271 9.1.2.2. g

This is needed in a topology where there are 2 parallel eBGP sessions between the same two devices, otherwise path selection is non-deterministic

fixes #2308

### DIFF
--- a/internal/pkg/table/destination.go
+++ b/internal/pkg/table/destination.go
@@ -50,6 +50,7 @@ const (
 	BPR_ROUTER_ID
 	BPR_OLDER
 	BPR_NON_LLGR_STALE
+	BPR_NEIGH_ADDR
 )
 
 var BestPathReasonStringMap = map[BestPathReason]string{
@@ -68,6 +69,7 @@ var BestPathReasonStringMap = map[BestPathReason]string{
 	BPR_ROUTER_ID:          "Router ID",
 	BPR_OLDER:              "Older",
 	BPR_NON_LLGR_STALE:     "no LLGR Stale",
+	BPR_NEIGH_ADDR:         "Neighbor Address",
 }
 
 func (r *BestPathReason) String() string {
@@ -479,6 +481,10 @@ func (dst *Destination) sort() BestPathReason {
 				}).Error("Could not get best path by comparing router ID")
 			}
 			reason = BPR_ROUTER_ID
+		}
+		if better == nil {
+			better = compareByNeighborAddress(path1, path2)
+			reason = BPR_NEIGH_ADDR
 		}
 		if better == nil {
 			reason = BPR_UNKNOWN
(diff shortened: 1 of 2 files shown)


internal/pkg/table/destination.go
```diff
@@ -918,6 +924,25 @@ func compareByRouterID(path1, path2 *Path) (*Path, error) {
...
+func compareByNeighborAddress(path1, path2 *Path) *Path {
+	// Select the route received from the peer with the lowest peer address as
+	// per RFC 4271 9.1.2.2. g
+	log.WithFields(log.Fields{
+		"Topic": "Table",
+	}).Debug("enter compareByNeighborAddress")
+
+	for i := 0; i < 16; i = i + 4 {
+		a := binary.BigEndian.Uint32(path1.GetSource().Address[i : i+4])
+		b := binary.BigEndian.Uint32(path2.GetSource().Address[i : i+4])
```
fujita, line 936: path1.GetSource().Address is nil if the path is injected locally.

jwhited: thanks for taking a look, updated

internal/pkg/table/destination.go
```diff
@@ -918,6 +924,34 @@ func compareByRouterID(path1, path2 *Path) (*Path, error) {
...
+	p1 := path1.GetSource().Address
+	if p1 == nil {
+		return path1
+	}
+	p2 := path2.GetSource().Address
+	if p2 == nil {
+		return path2
+	}
+
+	for i := 0; i < 16; i = i + 4 {
```
fujita, line 943: This function should handle ipv6 too?
how about simply using bytes.Compare()?

jwhited: updated w/bytes.Compare()

fujita:
pushed, thanks a lot!

PR #2314, opened 2020-08-31
chore: general py setting for vscode development; suppress py3 syntax warnings in py test files

imcom:
added `.env` file for vscode to find test/lib module
replace `is not` exp in *_test.py files to suppress python syntax warning with py3


fujita:
thanks

PR #2316, opened 2020-09-16
remove incomingCh when deleting dynamic peer

mrchrisb:
When a dynamic neighbor/peer is deleted (i.e. BGP session goes down), its incoming channel is closed, but never removed from the server's list. This results in excessive iterations of the loop in BgpServer.Serve func, as it constantly selects on the closed channel. This in turn causes **very** high CPU utilization, mainly from the repeated creation and garbage collection of the slice being passed to reflect.Select.

This PR fixes the issue by removing the incoming channel from the server's list when the dynamic neighbor is deleted (which also mirrors the behaviour of when a non-dynamic peer is deleted)


fujita:
Great, thanks a lot!

PR #2320, opened 2020-10-18
fix #2289: improve regex to give expected results

### LINKED ISSUE #2280: BGP unnumbered and rfc2545 nexthop field compliance
Hello. 
When using bgp unnumbered and advertising some prefix, the bgp updates are sent with the MP_REACH_NLRI. The next hop attribute is sent only with one ipv6 address, in this case a link-local address. This is not rfc2545 compliant, where it says it should be a global ipv6 address, plus the link-local. I'm having some issues with network devices, because of this.

```
" A BGP speaker shall advertise to its peer in the Network Address of
   Next Hop field the global IPv6 address of the next hop, potentially
   followed by the link-local IPv6 address of the next hop.

   The value of the Length of Next Hop Network Address field on a
   MP_REACH_NLRI attribute shall be set to 16, when only a global
   address is present, or 32 if a link-local address is also included in
   the Next Hop field." 
```
Current bgp update:

image

Expected bgp update with link-local + global ipv6 set by policy:

image

By default, I can see that gobgp doesn't send the bgp update with two next-hops. 
I've been able to change some lines in the code to make it work with my use case, but not sure the best approach to fix this globally. 

Any thoughts on this?

### LINKED ISSUE #2289: Inconsistent regex result for parseCommunityRegexp
Hi,

I have just noticed that parseCommunityRegexp returns inconsistent results if the community itself is parseable.

````
parseCommunityRegexp("6830:24370$")  => ^6830:247380$$
parseCommunityRegexp("^6830:24370$")  => ^^6830:247380$$
````

I believe the correct result should be:

```
parseCommunityRegexp("6830:24370$")  => 6830:24738$
parseCommunityRegexp("^6830:24370$")  => ^6830:247380$
```

This is caused by the matching regex repexpCommunity (regexp.MustCompile(`(\d+.)*\d+:\d+`)) which only matches on the numbers, but then the full arg is used:

```
        if repexpCommunity.MatchString(arg) {
		return regexp.Compile(fmt.Sprintf("^%s$", arg))
	}
```

I propose to either change the regex to:

```
regexp.MustCompile(`^(\d+.)*\d+:\d+$`)
```

or

```
regexp.MustCompile(`((\d+.)*\d+:\d+)`)
```

and use the substring match regex function.

Thoro:


fujita:
Thanks! pushed with a style fix.

PR #2335, opened 2020-11-03
Use MonitorTableRequest.Name to filter by peer

alistairking:
It seems that `GetTableRequest.Name` is used to retrieve only the table for the given peer, but while the `MonitorTableRequest.Name` field exists, from what I can tell it is unused.

This uses the `MonitorTableRequest.Name` field to filter returned Paths based on provided peer address (similar to
GetTable). If ADJ_IN is monitored with Current is set, then this filtering is done in the appropriate section of the `watch` method, which should significantly improve performance for users trying to monitor a single peer's table.


fujita:
Can you fix the unittest?
https://travis-ci.org/github/osrg/gobgp/jobs/741192471

alistairking:
My apologies -- missed that one!
Fixed.

fujita:
Pushed, thanks! Just out of curiosity, for what you use GoBGP?

alistairking:
Thanks.
We're using GoBGP over here at Kentik to build a BGP route collection and distribution system. GoBGP let's us do this in a nice software-defined way -- thanks for the great project!

fujita:
Thanks! Always nice to hear that GoBGP is used in production.

PR #2349, opened 2020-11-11
fix wrong vrf neighbors display

srt180:
I found this bug when I run the command gobgp vrf <vrf name> neighbor, it just show all neighbor without filtering vrf.


fujita:
thanks!

PR #2359, opened 2020-11-29
Support of FRR7.5

irino:
This commit works with FRR7.5 when "version = 6" is specified in zebra.config
- Changes Zebra internal API commands to FRR7.5 which is incompatible with older version.
- Changes SendHello() to include AFI
- Changes SendRouteIDAdd() to include AFI
- Changes decodeFromBytes/serialize of IPRouteBody and NexthopUpdateBody, because size of MESSAGE_FALG is changed.


fujita:
great, thanks a lot!

PR #2362, opened 2020-11-30
Add missing marshal of Prefix SID path attribute

sbezverk:

Currently when API call is issued to List for L3VPN prefix with Prefix SID, prefix sid info has not been returned along with other path attributes. This PR add marshalling Prefix SID into the reply to List call.


fujita:
pushed, thanks.

Issue #2115, opened 2019-07-16
Policy prefix-set is not take effect in vpnv4 prefix


grandhuhu:

grandhuhu:
I saw the code in policy.go line 1454 and function is PrefixCondition. In 1466 line:
 `	case bgp.RF_IPv4_UC:
		masklen = path.GetNlri().(*bgp.IPAddrPrefix).Length
		key = keyf(path.GetNlri().(*bgp.IPAddrPrefix).Prefix, int(masklen))
	case bgp.RF_IPv6_UC:
		masklen = path.GetNlri().(*bgp.IPv6AddrPrefix).Length
		key = keyf(path.GetNlri().(*bgp.IPv6AddrPrefix).Prefix, int(masklen))`
Is it not support for vnpv4 prefix?

fujita:
You are right. What policy representations are used for vpnv4 prefix in other bgp implementations?

PR #2365, opened 2020-12-16
Enable IPv6 and IPv4 labeled prefixes to match against prefix set in policy

tahirazim:
This change is required because currently GoBGP does not apply prefix filters to IPv6-labeled and IPv4-labeled prefixes. Instead, it simply blocks such prefixes from reaching the global RIB if a prefix set is specified in the policy. This issue has been raised for GoBGP before for IPv4 VPN prefixes (https://github.com/osrg/gobgp/issues/2115), although this change doesn't address that particular issue.

This change solves the problem by fixing `nlriToIPNet` to handle IPv4 and IPv6 labeled prefixes correctly, adding unit tests for it in the process. It also modifies the `Evaluate` function for `PrefixCondition` to initially return false based only on different AFIs rather than the AFI-SAFI pair.


fujita:
Looks good, thanks!

PR #2367, opened 2020-12-20
fixed possible crashes on parsing of bgp messages

serejkus:
There were:
* panics and an infinite loop in parsing `OptionParameterCapability`
* panic in parsing `PathAttributeAigp`
* panic in parsing `TLV`
* panic in parsing `TunnelEncapSubTLVSRSegmentList`

It is https://github.com/dvyukov/go-fuzz who have found these panics. I've stored the minimal examples under `testdata` dir and not in the test itself because possible inputs are lengthy and it'll just litter the `bgp_test.go` file.


fujita:
looks good, thanks!

PR #2368, opened 2020-12-28
api: corrects typo in validation reason enum 

vma:
minor typo correction: s/reasot/reason/


fujita:
Unfortunately, we can't break the API compatibility.

fujita:
We can break the compatibility for now so pushed this, thanks.

PR #2369, opened 2020-12-29
defines LsOspfRouteType enum for LsPrefixDescriptor

vma:
Allows better serialization as this field is automatically removed from
prefix descriptor when empty/unknown instead of getting
`LsOspfRouteType(0)` default value from Stringer.


fujita:
We can't break the API compatibility.

fujita:
We can break the compatibility for now so pushed this, thanks.

PR #2370, opened 2020-12-29
adds `monitor global ls` command to gobgp

vma:
This command continuously monitors and prints as json the BGP-LS updates received by the
daemon.


fujita:
`monitor global rib -a ls -j`, using 'a' and 'j' options is clean, I think.

PR #2372, opened 2021-01-07
block next hop attribute for ipv6

zhbngchen:


fujita:
RFC says it should be blocked?

zhbngchen:
The Multi-Protocol Reach NLRI (Type 14) attribute has next_hop in it
already.

Under the same circumstances, we tested Quagga bgpd which does not have
stand-alone next_hop and it works.

fujita:
Unless RFC says so, I don't think that it needs to be blocked.

fujita:
or other implementations do this?

PR #2375, opened 2021-01-12
Fix IsEBGPPeer check to use LocalAs

alistairking:
Previously the LocalAs setting for the neighbor was ignored when
determining if the session is eBGP or not.


fujita:
thanks!

PR #2379, opened 2021-01-20
Support UDP destination port sub-TLV 

mikma:
This PR adds support for UDP destination port sub-TLV as specified in draft-ietf-idr-tunnel-encaps (which has been approved as a proposed standard).


fujita:
pushed, thanks!

PR #2380, opened 2021-01-22
server: add option to bind listener to a (VRF) device

milanlenco:
This pull request allows to bind the listener of GoBGP server to a given [VRF](https://www.kernel.org/doc/Documentation/networking/vrf.txt) device. This allows to run multiple instances of GoBGP, potentially listening on the same IP:port address but in the scope of different VRFs.
For the client side of GoBGP this has been already done as mentioned in https://github.com/osrg/gobgp/issues/2092, but the server/listener side was missing.

Signed-off-by: Milan Lenco <milan.lenco@pantheon.tech>


pkg/server/server.go
```diff
@@ -89,11 +106,11 @@ func newTCPListener(address string, port uint32, ch chan *net.TCPConn) (*tcpList
 				}).Warn("Failed to AcceptTCP")
 				return err
 			}
-			ch <- conn
+			ch <- conn.(*net.TCPConn)
```
sbezverk, line 109: Is it safe to convert `conn` interface to net.TCPConn without validating, it will cause panic otherwise.

milanlenco: Thank you for the review - I have changed it to the safe cast. While a type casting fail should be unreachable (network type `tcp4` or `tcp6` is given to Listen), it is certainly better to stay on the safe side as you suggested.

The PR should be ready, hence it is not marked as draft. However if there is anything missing (for example with respect to the added field into the proto file), please let me know.

fujita:
sorry about the delay, pushed. thanks guys.

PR #2390, opened 2021-02-21
Add support for tunnel egress endpoint sub-TLV

mikma:
This PR adds support for tunnel egress endpoint sub-TLV as specified in draft-ietf-idr-tunnel-encaps-22.


pkg/packet/bgp/bgp.go
```diff
@@ -11623,6 +11624,98 @@ func NewTunnelEncapSubTLVColor(color uint32) *TunnelEncapSubTLVColor {
...
+
+func (t *TunnelEncapSubTLVEgressEndpoint) DecodeFromBytes(data []byte) error {
+	value, err := t.TunnelEncapSubTLV.DecodeFromBytes(data)
+	if err != nil {
+		return err
+	}
+	if t.Length != 6 && t.Length != 10 && t.Length != 22 {
+		return NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST, nil, "Not all TunnelEncapSubTLVEgressEndpoint bytes available")
+	}
+	address_family := binary.BigEndian.Uint16(value[4:6])
```
sbezverk, line 11640: What about to use `switch` by address family and in each case make length check? I think it will be more readable.

pkg/packet/bgp/bgp.go
```diff
@@ -11623,6 +11624,98 @@ func NewTunnelEncapSubTLVColor(color uint32) *TunnelEncapSubTLVColor {
...
+	if address_family == 0 && t.Length != 6 || address_family == AFI_IP && t.Length != 10 || address_family == AFI_IP6 && t.Length != 22 {
+		return NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST, nil, "Not all TunnelEncapSubTLVEgressEndpoint bytes available")
+	}
+	if address_family == 0 {
```
sbezverk, line 11644: what about:
```
t.Address = nil
if address_family != 0 {
t.Address = blah
}
```
In this case you do not need if/else and it will be easier to read.

pkg/packet/bgp/bgp.go
mikma: I changed it into a `switch`, and also added named constants to make it more readable.

pkg/packet/bgp/bgp.go
mikma: I removed the else branch from this if statement and from two other similar if statements. I also was able to remove a switch statement with duplicated code.

fujita:
ready for merging?

mikma:
> ready for merging?

I think I have addressed the things @sbezverk mentioned.

fujita:
pushed, thanks.

PR #2394, opened 2021-02-24
Fix LsTLVIgpRouterID.DecodeFromBytes

shaitan:
According to https://tools.ietf.org/html/rfc7752#section-3.2.1.4, IGPRouterID could be 4-octet for OSPFv2 or OSPFv3 non-pseudonode.


fujita:
thanks!

PR #2397, opened 2021-02-25
fixing an issue when gobgp propagates paths from MP peer to non-MP peer

imcom:
…P-capable peer

whereby MP-capable peer uses MP_REACH_NLRI to send updates to gobgp
gobgp accepts it but without setting NEXT_HOP pathattr
then gobgp propagates the paths to the non-MP-capable peer with update.NLRI set but without NEXT_HOP
the non-MP-capable peer will treat the update as withdraw message cause routes cannot propagate properly

### DIFF
--- a/internal/pkg/table/message.go
+++ b/internal/pkg/table/message.go
@@ -442,6 +442,12 @@ func (p *packerV4) pack(options ...*bgp.MarshallingOption) []*bgp.BGPMessage {
 			paths := c.paths
 
 			attrs := paths[0].GetPathAttrs()
+			// we can apply a fix here when gobgp receives from MP peer
+			// and propagtes to non-MP peer
+			// we should make sure that we next-hop exists in pathattrs
+			// while we build the update message
+			// we do not want to modify the `path` though
+			attrs = append(attrs, bgp.NewPathAttributeNextHop(paths[0].GetNexthop().String()))
 			attrsLen := 0
 			for _, a := range attrs {
 				attrsLen += a.Len()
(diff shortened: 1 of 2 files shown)


fujita:
This is about [Advertising IPv4 Network Layer Reachability Information with an IPv6 Next Hop](https://tools.ietf.org/html/rfc5549), right? If two gobgp routers doesn't negotiate this capability, a router should not advertise such route to another router? If they negotiate the capability, a router should advertise a route with MP_REACH_NLRI to another?

imcom:
Yea, it really depends on when and where MP_REACH should be used. The situation we are facing is that there are 3 routers in total. Router-A peers with Gobgp-RS, both agree on IPv4 NLRI with IPv6 next-hop, however, Router-A uses MP_REACH even with IPv4 UC + IPv4 Next-Hop. Gobgp-RS accepts the MP_REACH route and further it needs to propagate the routes to Gobgp-N. In our setup, Gobgp-RS and Gobgp-N do not agree on MP_REACH, Gobgp-N does not support the capability (v4 on v6 next-hop). In this scenario, Gobgp-RS sends the route learnt from Router-A in an UPDATE message where NLRI is populated, but NEXT_HOP path attr is missing, because when propagating the route, `gobgp` does not check if certain attr is present, but rather copy the attrs from origin path directly, hence the problem. Gobgp-N will see an UPDATE that contains NLRI, MP_REACH but there is no NEXT_HOP in path attr and then reject the route.

imcom:
Perhaps this is more visible

fujita:
>Router-A uses MP_REACH even with IPv4 UC + IPv4 Next-Hop.

Well well well. Merged, thanks a lot!

Issue #2396, opened 2021-02-24
Libc incompatibility on CentOS 8


dclaisse:
When using pre-compiled binary, on CentOS 8 (glibc 2.28-145.el8), gobgpd (latest version 2.24) crashes due to a SIGSEGV.
I've narrowed down the issue to DNS resolution, where the resolver is using cgo instead of plain go resolution, triggering a segmentation fault due to newer libc being used.
Indeed, if I set environment variable GODEBUG=netdns=go, I don't reproduce the issue anymore.
Also, if I build gobgpd on a newer system (Ubuntu 20.04.1), I don't reproduce the issue.

Would you consider bumping build environment to ensure compatibility with newer libc?

Please find attached a stack trace of the crash:
[gobgp_stack_trace.txt](https://github.com/osrg/gobgp/files/6036847/gobgp_stack_trace.txt)

Kind regards,

Damien

PR #2398, opened 2021-02-25
goreleaser: improve portability

dclaisse:
- Use tags=netgo to ensure net package won't use cgo
- Rebuild stdlib (go build -a) to really disable cgo as it was
  tried in https://github.com/osrg/gobgp/pull/2300
- Remove static ldflags as it would statically compile cgo


dclaisse:
This should fix https://github.com/osrg/gobgp/issues/2396

fujita:
thanks!

PR #2405, opened 2021-03-13
tools: don't use a shebang for bash completion

vincentbernat:
They are not expected to be executed. This seems a silly path, but
some people bother filing bug reports about it, so if you are OK with
it, it could be "fixed".


fujita:
pushed, thanks.

PR #2407, opened 2021-03-16
Add support for the gRPC server to listen on a unix domain socket

### LINKED ISSUE #2230: Supporting Unix sockets for gRPC api?
Have you thought about allowing Unix sockets for the gRPC api?  We use are playing with running in multiple docker containers and it would make managing the multiple containers easier, at least for us.  It's doesn't look terribly hard so if you want to discuss the correct way to design, I could probably do the work.

bbassingthwaite:
Fixes #2230


fujita:
thanks, pushed!

PR #2416, opened 2021-05-10
Fixed dead lock bug caused by hold-time expiry

zhouzijiang:
when gobgp and its peer encounter hold-time expiry at the same time, gobgp will be in a dead lock state，which leads to a false bgp established state but the tcp connection is closed. The details are as follows:

1. gobgp encounters hold-time expiry, so it creates BGP_ERROR_HOLD_TIMER_EXPIRED notification message and pushes the message to outgoing channel. And if GracefulRestart is not enabled, so it just returns, but blocked on 'defer wg.Wait()', waiting for sendMessageloop and recvMessageloop goroutines to exit. (source code is in function established of file pkg/server/fsm.go)

2. gobgp receives the BGP_ERROR_HOLD_TIMER_EXPIRED notification message which is sent by its bgp peer. so it pushes a fsmNotificationRecv message to stateReasonCh channel and return nil. And the code go into function recvMessageWithError again, but this time function readAll will cause an error and pushes a fsmReadFailed message to stateReasonCh channel. Finnally, the recvMessageloop goroutine exits, and stateReasonCh channel now has two messages to be processed. (source code is in function recvMessageloop and recvMessageWithError of file pkg/server/fsm.go)

3. sendMessageloop goroutine reads from outgoing channel and calls the send function. In the send function , it will write message to stateReasonCh channel directly. But now, stateReasonCh channel is full(its capacity is 2). So sendMessageloop goroutine is blocked on sending message to stateReasonCh channel. But stateReasonCh's consumer established goroutine is blocked on 'defer wg.Wait()', waiting for sendMessageloop goroutine to exit. So the dead lock occues

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -1569,6 +1569,14 @@ func (h *fsmHandler) openconfirm(ctx context.Context) (bgp.FSMState, *fsmStateRe
 }
 
 func (h *fsmHandler) sendMessageloop(ctx context.Context, wg *sync.WaitGroup) error {
+	sendToStateReasonCh := func(typ fsmStateReasonType, notif *bgp.BGPMessage) {
+		// probably doesn't happen but be cautious
+		select {
+		case h.stateReasonCh <- *newfsmStateReason(typ, notif, nil):
+		default:
+		}
+	}
+
 	defer wg.Done()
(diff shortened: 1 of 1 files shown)


fujita:
thanks a lot!

Issue #2422, opened 2021-05-23
Export/Import policy overrides due MP_REACH_NLRI is used


imcom:
The issue happens when a peer using MP_REACH_NLRI with IPv4 prefix (NLRI) + IPv4 Nexthop or IPv4 prefix (NLRI) + IPv6 Nexthop.

This issue happens because when GoBGP applies modification policies like MED, Community, AS Path etc.. it does not touch path attribute MP_REACH_NLRI of the Path being modified. So if a peer sends an update that using MP_REACH_NLRI contains 3 NLRIs (1.1.1.0/24, 1.1.2.0/24 and 1.1.3.0/24), 1) GoBGP installs the routes as three Paths in local rib and each of the Path has the origin path attribute, so the path of 1.1.1.0/24 has the MP_REACH_NLRI attributes as origin [1.1.1.0/24, 1.1.2.0/24 and 1.1.3.0/24], the same goes to the other two Paths.

Then if there is export policy we implement to match 1.1.1.0/24 and change the route community to some other values. When GoBGP advertises the 3 paths, we end up in an overriding situation like below.

1. - Path (1.1.1.0/24) has community A:A with MP_REACH_NLRI (1.1.1.0/24, 1.1.2.0/24, 1.1.3.0/24)
2. - Path (1.1.2.0/24) has community B:B with MP_REACH_NLRI (1.1.1.0/24, 1.1.2.0/24, 1.1.3.0/24)
3. - Path (1.1.3.0/24) has community B:B with MP_REACH_NLRI (1.1.1.0/24, 1.1.2.0/24, 1.1.3.0/24)

This will confuse the downstream peers as the (1) update tells the all 3 prefixes should have community A:A whereas the update of (2)&(3) because of path attrs hash, they are merged as one update, tells all 3 prefixes should have community B:B.

I have a fix in mind like what I did with the missing next-hop attribute that I can amend the MP_REACH_NLRI right before GoBGP sends the update out. Path attributes hash is fine, just need to alter the MP_REACH_NLRI itself.

It goes like:

```go
loop(attrsLen, paths, func(nlris []*bgp.IPAddrPrefix) {
				msgs = append(msgs, bgp.NewBGPUpdateMessage(nil, attrs, nlris))
				// till the cb is called
				// we know for sure what NLRIs we have in this message
				// so we rectify the path attribute here
				rectified_attrs := make([]bgp.PathAttributeInterface, len(attrs))
				for _, a := range attrs {
[...]

PR #2423, opened 2021-05-23
bugfix: fix the issue of overriding bgp attrs due to mp_reach_nlri is unchanged per policy

imcom:
This addresses the issue https://github.com/osrg/gobgp/issues/2422

### DIFF
--- a/internal/pkg/table/message.go
+++ b/internal/pkg/table/message.go
@@ -474,7 +474,7 @@ func (p *packerV4) pack(options ...*bgp.MarshallingOption) []*bgp.BGPMessage {
 						// here comes the overriding
 						// So we need align the MP_REACH to the actual NLRI we want to advertise
 						// Since this is merge situation, we need to use all the nlris
-						mp_nlris := make([]bgp.AddrPrefixInterface, len(nlris))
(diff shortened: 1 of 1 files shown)


imcom:
oh, I will fix the tests asap

fujita:
using packerMP instead of packerV4 is simpler? I guess that we really need to clean up the code though.

imcom:
Indeed, but it does not cover all scenarios. First of all, we could not know for sure when to use packerMP if both NLRI and next-hop are IPv4 UC, the difference maker is policy but whether an impactful policy is applied is now known when choosing a packer. Secondly, if GoBGP sits in between two peers where one peer supports MP capability and the other one does not, it would not work if using packerMP. Feel free to correct me if I got it wrong. Thanks in advance.

imcom:
There are certain cases that have to use MP:
1) NLRI is not IPv4_UC
2) NLRI is IPv4_UC but next hop is IPv6

However, in our production environment, we do see just a few routers announce IPv4 + IPv4 via MP_REACH_NLRI

imcom:
This static check does not look like related to this pull request , could be golang tool upgrade ?

imcom:
better remove MP_REACH_NLRI directly, will open another pull request

PR #2425, opened 2021-05-27
Remove MP_REACH_NLRI if the attr exists for IPv4_UC with IPv4 nexthop

imcom:
refactor: nlri/nexthop requires MP will get MP explicitly already. So we can safely remove the MP_REACH_NLRI from ipv4_uc with ipv4 nexthop

However, I am wondering is it safe to discard the MP_REACH_NLRI on the receiving side directly. Since MP_REACH_NLRI is not counted for hash calculation


fujita:
thanks!

PR #2426, opened 2021-06-03
zebra: close the incoming channel

worxli:
In general the writer should close the channel, thus we can close the channel here directly.
If zebra has disconnected the `closeChannel` method will in most cases not even close the channel
because no more messages can be received on it.


fujita:
looks fine. thanks, pushed.

PR #2428, opened 2021-06-13
Add support for tunnel encap in gobgp policy

guangtong-cao:
gobgp now supports adding global rib with encap tunnel type, but tunnel encapsulation is not supported in policy configuration, this pull requests is add add support of tunnel encapsulation in gobgp policy configuration by extending extended community with encap, also add geneve encap in CLI.


fujita:
Thanks, can you add the description to the doc?
https://github.com/osrg/gobgp/blob/master/docs/sources/policy.md

guangtong-cao:
I have updated policy.md by adding encap tunnel in extended communities set section.

fujita:
Thanks! I fixed the lint warning and pushed.

PR #2431, opened 2021-06-23
SR Policy add support SegmentTypeB

zvfvrv:
In the current implementation it is possible to inject SR Policy with SegmentTypeA (mpls only). This PR wants to add the support of the SRv6BindingSID and the SegmentTypeB in order to allow the injection of SR policy for SR over IPv6, as described in the draft [Advertising Segment Routing Policies in BGP](https://datatracker.ietf.org/doc/html/draft-ietf-idr-segment-routing-te-policy-13)


internal/pkg/apiutil/attribute.go
```diff
@@ -1848,7 +1871,31 @@ func MarshalSRSegments(segs []bgp.TunnelEncapSubTLVInterface) []*any.Any {
...
 				},
 			}
-			// TODO (sbezverk) Add Type B Segment when SRv6 Binding SID gets finalized.
+		case *bgp.SegmentTypeB:
+			flags := &api.SegmentFlags{
+				VFlag: s.Flags&0x80 == 0x80,
+				AFlag: s.Flags&0x40 == 0x40,
+				SFlag: s.Flags&0x20 == 0x20,
+				BFlag: s.Flags&0x10 == 0x10,
+			}
```
sbezverk, line 1880: Could you change `if else` to just a single `if` 
```
r = &api.SegmentTypeB{
					Flags: flags,
					Sid:   s.SID,
				}
if s.SRv6EBS != nil {
// Assign EndpointBehaviorStructure
}
```

pkg/packet/bgp/sr_policy.go
```diff
@@ -676,19 +676,86 @@ func (s *SegmentTypeA) MarshalJSON() ([]byte, error) {
...
-	Behavior api.SRv6Behavior
+	Behavior SRBehavior
 	BlockLen uint8
 	NodeLen  uint8
 	FuncLen  uint8
 	ArgLen   uint8
 }
 
+func (s *SRv6EndpointBehaviorStructure) DecodeFromBytes(data []byte) error {
+	behavior := binary.BigEndian.Uint16(data[0:2])
```
sbezverk, line 730: if length of data is less than 2 you will get panic. Please validate data length before doing any manipulations

internal/pkg/apiutil/attribute.go
zvfvrv: done, thanks

pkg/packet/bgp/sr_policy.go
zvfvrv: done, thanks

fujita:
let me know when this becomes ready.

sbezverk (approved):
looks good to me

fujita:
pushed, thanks.

PR #2444, opened 2021-08-03
Add support for the gRPC client to connect on a unix domain socket

bbassingthwaite:
This is a follow up item from #2407


fujita:
pushed, thanks!

PR #2449, opened 2021-08-11
Update the hold timer when a BGP Update message is received

peterjhill:
According to RFC4271,
       The calculated value indicates the maximum number of
         seconds that may elapse between the receipt of successive
         KEEPALIVE and/or UPDATE messages from the sender.

This change will reset the hold timer in the update FSM case.
(note, I am super new with Go. I'm not quite sure how to craft a good test case for this.)


fujita:
I thought that resetting hold timer for update messages is handled but can't find. Hmm. Pushed, thanks.

Issue #2399, opened 2021-03-04
Parsing error of updates from BIRD with LLGR


kzemek:
Hi! I'm having GoBGP bail out on parsing updates (sends back `UPDATE Message Error`: `Malformed Attribute List`) when Long Lived Graceful Restart is enabled on BIRD side.

```
WARN[0052] Session will be reset due to malformed BGP message  Key=10.1.0.1 State=BGP_FSM_ESTABLISHED Topic=Peer error="network bit length is too long"
```

The issue looks like an off-by-one-byte error. Looking at the debug output, it's parsing routes like: `22.123.124.125/29`, while BIRD is in fact advertising `123.124.125.0/22 PathId 29` . I dumped the TCP packets and confirmed that at least Wireshark parses the updates correctly (`123.124.125.0/22`).

The issue is reproducible, e.g. with config for BIRD 2.0.7:

```
ipv4 table test_routes;

protocol static static_test_routes {
  ipv4 { table test_routes; };
  route 10.123.123.0/24 unreachable;
  route 123.123.0.0/22 unreachable;
  route 123.120.0.0/19 unreachable;
}

protocol bgp load_balancers {
    local 10.1.0.1 as 64123;
    neighbor range 10.0.0.0/8 internal;
    graceful restart on;
    long lived graceful restart on;

    ipv4 {
        table test_routes;
        add paths tx;
        next hop keep;
        import none;
        export all;
    };
}
```

And GoBGP:

```
[global.config]
  as = 64123
  router-id = "192.168.255.1"
  port = -1

[[neighbors]]
  [neighbors.config]
    neighbor-address = "10.1.0.1"
    peer-as = 64123
  [neighbors.timers.config]
    hold-time = 3600
    keepalive-interval = 60
  [neighbors.transport.config]
    local-address = "10.1.0.2"
  [neighbors.add-paths.config]
    receive = true
  [neighbors.graceful-restart.config]
    enabled = true
    notification-enabled = true
    long-lived-enabled = true
```

fujita:
Can you attach the pcap file?

kzemek:
[gobgp-bird.zip](https://github.com/osrg/gobgp/files/6337535/gobgp-bird.zip)

kzemek:
## The configs I used

### bird.conf
```
router id 127.0.0.2;

ipv4 table test_routes;

protocol static static_test_routes {
  ipv4 { table test_routes; };
  route 10.123.123.0/24 unreachable;
[...]

PR #2451, opened 2021-08-12
Fix extended-nexthop & add-path capability parsing.

kzemek:
The capabilities' DecodeFromBytes() was eating as much data as given,
while these functions would normally be called with all the data
remaining to parse - e.g. including following capabilities.

Fixes osrg/gobgp#2399


fujita:
Looks good, thanks!

Issue #2452, opened 2021-08-13
frr7.4 is missing from summary-of-combination-of-version-and-software Table


gshirazi:
Missing frr7.4 from table after https://github.com/osrg/gobgp/commit/b55c32ce65c9b6e0c9f4620e6acb27e0e49f238d#r54831858

It should be clarified that 7.3 and 7.4 are the same in the table.
Can someone verify that this statement is correct? Or is something different needed for 7.4?

gshirazi:
from @irino : I haven't tested with frr7.4. I assume frr7.4 doesn't work correctly regardless 'software-name' configuration.

PR #2453, opened 2021-08-13
Add frr7.4 to the summary-of-combination table

gshirazi:
Fix for https://github.com/osrg/gobgp/issues/2452
Can someone verify that this is correct? Or is something different needed for 7.4?

### DIFF
--- a/docs/sources/zebra.md
+++ b/docs/sources/zebra.md
@@ -83,7 +83,7 @@ You need to enable the zebra feature in the Global configuration as follows.
 |5      |cumulus      |Cumulus Linux VX 3.7.7   |                     |
 |5      |frr4         |FRRouting 4.0.x          |                     |
 |6      |             |FRRouting 7.5.x          |                     |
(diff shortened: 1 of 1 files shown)


irino:
I'm a maintainer of functionality of zebra integration. This fix for the documentation is incorrect. If frr7.4 is specified as 'software-name', the 'software-name' configuration is ignored (which is same as unspecified 'software-name') because the 'software-name' configuration is checked in IsAllowableSoftwareName function in zapi.go.
BTW, why do you need frr7.4 instead of frr7.5?

gshirazi:
Hi @irino thanks for the fast reply.
I need 7.4 and cannot upgrade to 7.5 due to dependency issues.
Would setting `frr7.3` as `software-name` work for 7.4?

irino:
I haven't tested with frr7.4. I assume frr7.4 doesn't work correctly regardless 'software-name' configuration.

PR #2454, opened 2021-08-13
zapi: For IPv4, set offset=12 for NetworkBytes

gshirazi:


gshirazi:
I see, the caller is supposed to do `.To4()`

PR #2456, opened 2021-08-19
.github/release: Use goreleaser action

mrueg:
https://github.com/goreleaser/goreleaser-action#workflow

This simplifies the workflow, allows to define the golang version it is
built with.


fujita:
Thanks.

PR #2464, opened 2021-09-08
internal/pkg/table: remove debug logging from Destination.Calculate

mdlayher:
Signed-off-by: Matt Layher <mlayher@fastly.com>

While using gobgp and handling a large number of route updates from multiple BGP peers, we noticed a huge amount of allocations related to the debug logging for Destination.Calculate. Removing these non-actionable logs (and empty functions) cut memory allocations tremendously.

autopilot-routed-alloc-space

I suspect some of these may have been added during the development process, but ultimately they don't seem useful compared to unit tests which exercise this sort of behavior. Removing these logs in the hot path tremendously reduced the amount of memory allocations for our service.


fujita:
Looks great, thanks a lot! As you suspect, this has been used during development.
Happy to know that Fastly uses GoBGP in production!

Issue #2466, opened 2021-09-15
bgp bandwidth community -- EC_SUBTYPE_LINK_BANDWIDTH


gpl:
Hello,

I'm trying to use EC_SUBTYPE_LINK_BANDWIDTH ( https://datatracker.ietf.org/doc/html/draft-ietf-idr-link-bandwidth-06) with gobgp, but it looks like the only extended communities supported are RT, SoO, and encap.

Could this be added?  Thank you!

ton31337:
I can work on this.

fujita:
surely, go ahead.

fujita:
@gpl is this what you wanted?

gpl:
Hello,

We'll test this patch out next week and see if it works, thank you!

gpl:
Hello,

Apologies for the late reply.  It looks like that patch is working, thank you!

fujita:
please create a new issue if the current code doesn't work as expected.

gpl:
apologies, we have had time to look at this and have been busy testing it.  however there is a typo in the name of 
LinkBandiwdthExtended

should be
LinkBandwidthExtended

I will make a new PR shortly to correct this, if that works

fujita:
thanks!

PR #2467, opened 2021-09-20
BGP link-bandwidth extended community support

ton31337:
Closes https://github.com/osrg/gobgp/issues/2466


ton31337:
@fujita could you please take a look here?

api/attribute.proto
```diff
@@ -400,6 +400,13 @@ message FourOctetAsSpecificExtended {
   uint32 local_admin = 4;
 }
 
+message LinkBandiwdthExtended {
+  bool is_transitive = 1;
+  uint32 sub_type = 2;
```
fujita, line 405: How `is_transitive` and `sub_type` is used?

ton31337:
Basically ￼not used at the moment. I'll drop this better. Initial idea was to use the same way as FRR (it can be send transitive), but it against RFC.

fujita:
Understood. Let's drop them for now. Better to avoid adding unused stuff to a proto file.

api/attribute.proto
ton31337: Fixed.

fujita:
pushed, thanks a lot!

PR #2468, opened 2021-10-08
Return MpGracefulRestart.State in ListPeer() api call

floatingstatic:
I have a use-case to check the value of `EndOfRibReceived` in `ListPeer()` api calls. It looks like the code was only returning `MpGracefulRestart.Config`. This adds `MpGracefulRestart.State` (which includes `EndOfRibRceived`).


fujita:
Thanks!

Issue #2474, opened 2021-10-28
Invalid SR Segment label unmarshalling 


GeorgyKA:
SR segment label is 32-bit wide number containing TTL (8 bit), BottomOfStack Flag (1 bit), TrafficClass (2 bit) and the label value (20 bits). 
The marshaling routine encodes a label as is including all containing fields whereas unmarshaling function treats its input as single label value to be shifted left for 12 bits. Such makes impossible to set SR segment TTL, S and TC fields trough API.

PR #2475, opened 2021-10-28
Fix SR segment label unmarshaling

GeorgyKA:
Closes https://github.com/osrg/gobgp/issues/2474


fujita:
@sbezverk Any comment?

fujita:
pushed, let's see if someone would complain.

PR #2482, opened 2021-11-09
AsPathPrependAction support add replace del

chaoliang-wei:
Signed-off-by: ChaoLiang <chaoliang.wei@nexusguard.com>


chaoliang-wei:
AsPathPrependAction support add replace del

PR #2484, opened 2021-11-10
When the length is greater than 255 the wrong asnumber will be appended here

chaoliang-wei:


chaoliang-wei:
Yes When the total is greater than 255 
newAsList := append(asns[:int(repeat)], asList...)
will overwrite asns with the contents of asList starting from repeat；
This PR should be cancelled because it's not quite right for me to change it this way, so if you think this needs to be changed I'll resubmit a PR

chaoliang-wei:
var idx = int(repeat)
			if int(repeat)+len(asList) > 255 {
				idx = int(repeat) - (255 - len(asList))
			}
			newAsList := append(asns[idx:], asList...)
			asPath.Value[0] = bgp.NewAs4PathParam(segType, newAsList)
			asns = asns[:idx]

chaoliang-wei:
Also we want to add del replace to AsPathPrependAction; do you have any comments or suggestions on your side?

chaoliang-wei:
Sorry your previous write up was fine I made a mistake.
repeat is uint8 does not exceed 255, so it will never be overwritten

fujita:
What bug does this fix?

fujita:
Looks like that this doesn't

PR #2487, opened 2021-11-16
Add SIGINT as shutdown signal

grongor:
...so that you can stop the GoBGP server gracefully via `CTRL+C` when started directly from CLI


fujita:
pushed, thanks.

Issue #2488, opened 2021-11-22
Local routes with origin AS = LocalAS filtered between global rib and neighbor adj-out


floatingstatic:
I have gobgp configured as an IBGP neighbor with another BGP speaker (running bird). When I issue an `AddPathRequest` to inject a route into the gobgp global rib the route is correctly installed. However when I attempt to share this injected route with the IBGP neighbor with a route filter applied it appears to be dropping the route but letting others pass through without issue. Upon further investigation this only happens when the origin asn of the injected route is the same as the LocalAS configured for the neighbor.

The problematic section of code appears to be here: 
https://github.com/osrg/gobgp/blob/b931ccd0566e2852a0bf1c25b77488319d1882d0/pkg/server/server.go#L580-L582

I am not using the route-server feature but as this is an IBGP session the `isASLoop()` function is returning true, as a result the `filterpath()` function is returning nil for my injected route. I feel this is probably not correct behavior as the route I have injected is locally sourced (not learned from another IBGP neighbor so we are not violating anything related to split horizon). Also the lines in this function that do check things related to IBGP are mainly concerned with routes for which `IsLocal()` does not return true: https://github.com/osrg/gobgp/blob/b931ccd0566e2852a0bf1c25b77488319d1882d0/pkg/server/server.go#L518 
To me it seems like this function may be neglecting to handle the case of locally sourced routes.

So to summarize:

- I inject a route `192.0.2.0/24`, with origin ASN 64512 with `AddPathRequest`
- Route is learned in global rib
- Route filter (very permissive) does not let the route pass to a neighbor with `LocalASN` of 64512

floatingstatic:
Fixed in #2489 , thanks for merging @fujita

PR #2489, opened 2021-11-22
Bypass as-path loop filtering on locally injected routes

floatingstatic:
Addresses issue raised in: https://github.com/osrg/gobgp/issues/2488


fujita:
looks like makes sense. Pushed, thanks!

PR #2490, opened 2021-11-23
fsm: make variables more readable

lx1036:
make variables more readable, use stateReason instead of err.


fujita:
this makes the core readable? Not sure but I prefer shorter name like `reason` if `err` is really confusing.

fujita:
Please rebase this against the latest master if you think that it's worth.

Issue #2485, opened 2021-11-14
Please consider updating to new protobuf library


gibmat:
I'd like to request that you consider updating this program's use of protobuf to utilize the newer version of the library (github.com/protocolbuffers/protobuf-go). My selfish reason for wanting this is that it will allow LXD, which depends on this code, to then eliminate its use of the old protobuf library, which would in turn simplify my packaging of LXD for Debian.

A quick glance through this code shows use of protobuf's `ptypes` library, which was deprecated within the older version of the library, but has what appears to be [easy updates](https://github.com/golang/protobuf/pull/1217/files).

mikma:
Which protoc version do you want to use? I tried protoc 3.14.0 from Ubuntu 20.04LTS. Some change were needed, but not the ones you mentioned it seems.
```
Update to protoc 3.14.0

Add go_packages options, required in future protoc versions.
Require grpc v1.32.0, as specified in api/gobgp_grpc.pb.go.
Include api.UnimplementedGobgpApiServer in server struct as required
for forward compatibility.
```
https://github.com/mikma/gobgp/commit/c113a6bf8cee440fc8ee8794bb5f1eb66986fd4e

gibmat:
@mikma, I'm not sure where your comment went, although I did get an email notification, so I'll quote it here:

> Which protoc version do you want to use? I tried protoc 3.14.0 from Ubuntu 20.04LTS. Some change were needed, but not the ones you mentioned it seems.
> ```
> Update to protoc 3.14.0
> 
> Add go_packages options, required in future protoc versions.
> Require grpc v1.32.0, as specified in api/gobgp_grpc.pb.go.
> Include api.UnimplementedGobgpApiServer in server struct as required
> for forward compatibility.
> ```
> https://github.com/mikma/gobgp/commit/c113a6bf8cee440fc8ee8794bb5f1eb66986fd4e

I'm not too concerned about the specific version of protoc that's used; Debian has v3.12.4 in unstable. The big changes needed are updating the imported version of the protobuf library itself. I'm not super familiar with Go's protobuf implementations, but basically all imports of `github.com/golang/protobuf` need to be changed/adjusted to `google.golang.org/protobuf`, and `github.com/golang/protobuf` removed from `go.mod`.

Does that make sense?

fujita:
This breaks the backward compatibility? So needs to update the major version?

gibmat:
[...]

PR #2492, opened 2021-12-05
Remove bgp package dependency on api and protobuf packages

grongor:
THIS IS A BREAKING CHANGE.

I couldn't figure out a way to do this without BC break...if you think of something or want the things done differently, let me know and I'll update the PR.

For context, I tried this here https://github.com/grongor/tmp-dummy-package . Check the difference between "before" and "after":

before

after


fujita:
Thanks. The api directory is intended to have only photo files and the generated files from them. So api/prefix_sid.go is placed in internal/pkg/apiutil?
Also seems that people want the public APIs for the conversion between pkg/packet and api so probably needs to make internal/pkg/apiutil public, I guess.

grongor:
Yeah, I thought that you wouldn't like me moving it to the `api` package :) Yes, I agree, `internal/pkg/apiutil` was my first choice, but as it wasn't public I decided to go with `api`.

I can't agree more with making `apiutil` public - the functions there are very useful; once I worked with GoBGP API I was quite confused about why is this package internal.

So, I adjusted the first commit to put the thing into `apiutil`, and then added another commit to make `apiutil` public.

fujita:
I plan to push https://github.com/osrg/gobgp/issues/2485 this week. This conflicts with the pull request so can you work on this after it?

> I worked with GoBGP API I was quite confused about why is this package internal.

I preferred to use the protocol buffer structures and APIs rather than making two public APIs (the pb stuff and pkg/packet). But it's too late to fix my mistake to create public pkg/packet, I guess.

grongor:
Are the #2485 changes ready? If so, I can rebase this PR onto your v3 branch.

And well, since you are planning on releasing a new major version, maybe consider this...making `pkg/packet` internal would be a huge mistake in my opinion. Just now I'm writing a custom BGP server (I need just a small subset of functionality, basically I'm just gathering all routes and pushing them to Kafka), and I'm using `pkg/package/bgp` to do the heavy lifting of parsing the data from routers. But what would really make sense to me is to split the "GoBGP" part from the `generic BGP` part -> splitting the BGP client/server logic from the raw data parsing logic into two separate repositories. So people like me who need to parse raw BGP data into some usable structs would only use the `parser` package, which would ideally have basically zero dependencies, and the rest would keep using this package, which would internally depend on the parser. This would also mean that the `parser` part would be "more stable" (have fewer versions), as it would change only when the BGP protocol itself changes.

fujita:
> Are the #2485 changes ready? If so, I can rebase this PR onto your v3 branch.

Passed the tests so after confirming that this is what people want, I'll push
https://github.com/fujita/gobgp/pull/133
> And well, since you are planning on releasing a new major version, maybe consider this...making pkg/packet internal would be a huge mistake in my opinion.

Don't worry, I don't do such drastic changes (I understand it's too late). What I'm planing are

- Allowing passing logger instance to  BgpServer.
- Merge MonitorPeer and MonitorTable APIs to get both kinds of events in order.
- Minor cleanups on proto files.

grongor:
Ok, I'll rebase on that branch later today :)

If you were aboard with the split of "basic BGP stuff" and the server itself, I'd gladly do it :) I think it would improve the experience for people using "just the parsing logic", and wouldn't touch people using the whole GoBGP server/client in any way :) And I don't think that we would need to do much renaming/moving around, so the BC break would basically be only changing the import path. For a new major version, this seems reasonable.

fujita:
Sorry about lots of conflicts. Can you rebase?

grongor:
Yeah sorry, will do shortly, got some unexpected work... Sorry

fujita:
No need to rush. I think that it takes 2 months or so until releasing v3.0.0. Thanks!

grongor:
@fujita Okay, done - hopefully correctly. Just a reminder, I saw some old imports paths (without v3 suffix) in docs, tests, ... so don't forget them before releasing :)

```
$ find . -type f | xargs grep 'github.com/osrg/gobgp/[^v3]'
./tools/pyang_plugins/bgpyang2golang.py:    print('"github.com/osrg/gobgp/pkg/packet/bgp"', file=fd)
./tools/contrib/centos/README.md:cd /tmp/gobgp && curl -s -L -O https://github.com/osrg/gobgp/releases/download/v1.31/gobgp_1.31_linux_amd64.tar.gz
./tools/completion/README.md:    If you run 'go get github.com/osrg/gobgp/gobgp', gobgp command is installed
./tools/completion/README.md:    % source $GOPATH/src/github.com/osrg/gobgp/tools/completion/gobgp-completion.bash
./tools/completion/README.md:GOBGP_COMP=$GOPATH/src/github.com/osrg/gobgp/tools/completion/zsh
./README.md:Try [a binary release](https://github.com/osrg/gobgp/releases/latest).
./README.md:[LICENSE](https://github.com/osrg/gobgp/blob/master/LICENSE) for the full
./.github/workflows/ci.yml:          if [ "$(go env GOARCH)" = "amd64" ]; then go test -race github.com/osrg/gobgp/pkg/packet/bgp -run ^Test_RaceCondition$; else echo 'skip'; fi
./pkg/server/fsm.go:			// Reference: https://github.com/osrg/gobgp/issues/2174
./pkg/server/server.go:					// https://github.com/osrg/gobgp/issues/1777
./test/scenario_test/README.md:    $ cd $GOPATH/src/github.com/osrg/gobgp/test/scenario_test
./test/scenario_test/README.md:    $ cd $GOPATH/src/github.com/osrg/gobgp/test/scenario_test
./test/scenario_test/graceful_restart_test.py:        # https://github.com/osrg/gobgp/pull/1707
./test/scenario_test/rtc_test.py:    # https://github.com/osrg/gobgp/issues/1682
./test/scenario_test/rtc_test.py:        # https://github.com/osrg/gobgp/issues/1630
./test/scenario_test/rtc_test.py:        # https://github.com/osrg/gobgp/issues/1688#issuecomment-391456615
./test/scenario_test/rtc_test.py:        # https://github.com/osrg/gobgp/pull/1703
./test/scenario_test/rtc_test.py:        # https://github.com/osrg/gobgp/issues/1688
./docs/sources/grpc-client.md:[`tools/grpc/python/add_path.py`](https://github.com/osrg/gobgp/blob/master/tools/grpc/python/add_path.py)
./docs/sources/grpc-client.md:Use [`tools/grpc/cpp/Makefile`](https://github.com/osrg/gobgp/blob/master/tools/grpc/cpp/Makefile).
./docs/sources/grpc-client.md:The above to generate the server and client interface and the binary to add a route by using `AddPath` API, ['tools/grpc/cpp/add_path.cc'](https://github.com/osrg/gobgp/blob/master/tools/grpc/cpp/add_path.cc).
./BUILD.md:	"-X github.com/osrg/gobgp/internal/pkg/version.COMMIT=`git rev-parse --short HEAD` \
./BUILD.md:	 -X github.com/osrg/gobgp/internal/pkg/version.METADATA="date.`date "+%Y%m%d"`" \
./BUILD.md:	 -X github.com/osrg/gobgp/internal/pkg/version.IDENTIFIER=alpha"
```

fujita:
pushed, thanks. Not sure about `apiutil` package name but I don't have better options.

thanks for heads-up about v3 suffix. looks like most of them are fine (refer to the real path rather than virtual path including v3 suffix) but I'll check out again.

PR #2501, opened 2021-12-10
server: fix output of gobgp rpki server

a16:
gobgp rpki server command hasn't been working correctly since #a40bc53.
This code will fix it.


fujita:
thanks!

PR #2513, opened 2021-12-21
Ensure EndOfRibReceived is always cleared on Peer Down

floatingstatic:
I observed that the `EndOfRibReceived` state is "stuck" set to true except when a peer is restarting due to GR. We should always ensure that `EndOfRibReceived` is cleared even when this is not the case otherwise we cannot tell when a peer has sent all of its initial route updates when re-establishing.


fujita:
Thanks!

PR #2515, opened 2021-12-28
chore: introduce buf for code generation

imcom:
I would like to introduce the usage of `buf` with gobgp project. The tool can help to maintain and generate code from .proto files

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -17,6 +17,8 @@ Before sending pull request, please make sure that your changes have passed both
 
 ## Changing the gRPC API
 
+### Origin go-tools approach
+
 If you change the gRPC API, generate `api/*.pb.go` in the following way:
 
 ```bash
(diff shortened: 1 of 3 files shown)


fujita:
Can force the specific version of protoc?

imcom:
> Can force the specific version of protoc?

I believe the unknown version comes from this [replacement](https://docs.buf.build/generate/high-performance-protoc-replacement)

It has its own complier, the only difference I see now is the `unknown` tag in generated code. Anything else should I pay attention to ?

imcom:
more information as follows:
https://docs.buf.build/build/internal-compiler

fujita:
Do you propose replacing protoc with this tool?

imcom:
> Do you propose replacing protoc with this tool?

The `buf` project is getting more and more attention from the community and does provide more tooling stuff than `protoc`, not quite a replacement for `gobgpd` but at least it is an option when people try to build and develop `gobgp`.

I do see the tool's lint reports many concerns we do not want/have to adopt at the moment. `protoc` and `buf` can coexist with no harm.

fujita:
`buf` might generate code slightly different from `protoc`, right? Then if someone submit a pull request including the code generated by `buf`, it's supposed to be rejected. We don't need pointless changes of `*.pb.go`, right?

imcom:
> `buf` might generate code slightly different from `protoc`, right? Then if someone submit a pull request including the code generated by `buf`, it's supposed to be rejected. We don't need pointless changes of `*.pb.go`, right?

I got your point, although `buf` claims to be just as same as `protoc`. The additional risk is true, as you said. I believe we can keep this pull request open for a prolonged evaluation/observation of `buf` project and continue our discussion from there. Sounds good?

fujita:
sure, let this pr open.

imcom:
let's close this and keep `protoc` for consistency

PR #2516, opened 2021-12-29
chore: added missing dep

imcom:
found a missing dep `go-cmp` in `go.mod`

the package is being used in `server_test.go`


fujita:
thanks!

PR #2535, opened 2022-03-19
Expand first use of BGP acronym in Readme.

fragmede:
Linked to out of context, it's not immediately clear what this project is about. Expanding the first usage of the BGP acronym heaps users understand the project.


fujita:
Why people who don't know "BGP" is interested in GoBGP?

PR #2536, opened 2022-03-21
pkg/packet/bgp: fix PathAttributeUnknown output format

themiron:
Formatting PathAttributeUnknown binary value as a string breaks text
output and may cause data truncation while line-by-line parsing.

Signed-off-by: Vladislav Grishenko <themiron@yandex-team.ru>


fujita:
make sense, thanks!

PR #2539, opened 2022-03-23
documentation : BGP-SR via python grpc example

cpmarvin:
Provide a documentation example for BGP-SR using python grpc.


fujita:
I've fixed the format issues and pushed. Thanks!

cpmarvin:
thanks!
looks like the script went to tools/grpc/sr_policy.py instead of tools/grpc/**python**/sr_policy.py. will do another update.

PR #2540, opened 2022-03-23
optimised ExtCommunityCondition.Evaluate() by caching community string

serejkus:
It saves a lot of resources when there is a lot of communities to match against,
say hundreds of RTs. The same approach is used in some other conditions,
ex. AsPathCondition.


fujita:
thanks!

Issue #2541, opened 2022-03-25
AddPath IPv6 injected routes are not honoring PathIdentifier 


floatingstatic:
I observed that injecting multiple paths for a prefix with `AddPath` that the `Identifier` passed with the paths are not being honored. A quick way to reproduce this:

```
$ gobgp global rib add -a ipv6 2001:db8::/64 nexthop fd00::1 origin igp aspath 2914 local-pref 100 large-community 65104:1:0 identifier 1
$ gobgp global rib -a ipv6
   Network              Next Hop             AS_PATH              Age        Attrs
*> 2001:db8::/64        fd00::1              2914                 00:00:05   [{Origin: i} {LocalPref: 100} {LargeCommunity: [ 65104:1:0]}]
$ gobgp global rib add -a ipv6 2001:db8::/64 nexthop fd00::2 origin igp aspath 174,7922 local-pref 100 large-community 65104:1:0 identifier 2
$ gobgp global rib -a ipv6
   Network              Next Hop             AS_PATH              Age        Attrs
*> 2001:db8::/64        fd00::2              174 7922             00:00:02   [{Origin: i} {LocalPref: 100} {LargeCommunity: [ 65104:1:0]}]
```

I expect to see both paths in the global rib. Here is an example with IPv4 where this is working:

```
$ gobgp global rib add 203.0.113.0/24 nexthop 192.0.2.2 origin igp aspath 1299 local-pref 100 large-community 65104:1:0 identifier 1
$ gobgp global rib add 203.0.113.0/24 nexthop 192.0.2.1 origin igp aspath 174,7922 local-pref 100 large-community 65104:1:0 identifier 2
$ gobgp global rib
   Network              Next Hop             AS_PATH              Age        Attrs
*> 203.0.113.0/24       192.0.2.2            1299                 00:00:10   [{Origin: i} {LocalPref: 100} {LargeCommunity: [ 65104:1:0]}]
*  203.0.113.0/24       192.0.2.1            174 7922             00:00:03   [{Origin: i} {LocalPref: 100} {LargeCommunity: [ 65104:1:0]}]
```

I believe what is happening is that we initially honor and set the `PathIdentifier` from the `Identifier` passed in the API here:

https://github.com/osrg/gobgp/blob/459bcfedf44e571dbb93e5c697b904e9a5348d1c/pkg/server/grpc_server.go#L306

[...]

PR #2542, opened 2022-03-25
Fix AddPath IPv6 injected routes not honoring PathIdentifier

floatingstatic:
Fixes the problem described in https://github.com/osrg/gobgp/issues/2541

I also added some simple test cases to make ensure the scenario described in this issue are covered in `AddPath` and `DeletePath` tests.


fujita:
Pushed, thanks a lot!
IIRC, there is an use-case that sets bgp.PathAttributeMpReachNLRI with empty Path.Nlri. Let's see if someone will complain.

pavel-odintsov:
Hey! I'm here to complain https://github.com/osrg/gobgp/issues/2673 ;)

PR #2543, opened 2022-04-01
Add the possibility to choose the local port in the TCP connection to a peer.

hardeker:
It allows to have more control on the connection established to peers by GoBGP. 

In my use case, I have an access list mechanism for incoming packets. Knowing the port in advance allows to accept the packets corresponding to a given BGP peering session. If I cannot set the port before initiating the connection, I would have to learn it dynamically, but that means dropping the first return packets from the remote peer (SYN+ACK), which is pretty ugly and unpredictable.

### DIFF
--- a/internal/pkg/config/bgp_configs.go
+++ b/internal/pkg/config/bgp_configs.go
@@ -1141,7 +1141,7 @@ type ZebraState struct {
 	MplsLabelRangeSize uint32 `mapstructure:"mpls-label-range-size" json:"mpls-label-range-size,omitempty"`
 	// original -> gobgp:software-name
 	// Configure zebra software name.
-	// frr4, cumulus, frr6, frr7, frr7.2, and frr7.3 can be used.
+	// frr4, cumulus, frr6, frr7, frr7.2 and frr7.3 can be used.
 	SoftwareName string `mapstructure:"software-name" json:"software-name,omitempty"`
 }
 
@@ -1171,7 +1171,7 @@ type ZebraConfig struct {
 	MplsLabelRangeSize uint32 `mapstructure:"mpls-label-range-size" json:"mpls-label-range-size,omitempty"`
 	// original -> gobgp:software-name
 	// Configure zebra software name.
-	// frr4, cumulus, frr6, frr7, frr7.2, and frr7.3 can be used.
+	// frr4, cumulus, frr6, frr7, frr7.2 and frr7.3 can be used.
 	SoftwareName string `mapstructure:"software-name" json:"software-name,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


fujita:
You can't edit bg_configs.go file as the head of the file says. Instead you need to modify yang files.
https://github.com/osrg/gobgp/tree/master/tools/pyang_plugins

hardeker:
I missed that, sorry. Updated the PR accordingly.

hardeker:
Hi @fujita, it seems to me ci/lint failed because the golang version in use switched from go 1.17.8 to go 1.18.
The use of `strings.Title` (cmd/gobgp/neighbor.go:1084) became deprecated in 1.18, which is independent from this PR.

fujita:
> Hi @fujita, it seems to me ci/lint failed because the golang version in use switched from go 1.17.8 to go 1.18.
> The use of `strings.Title` (cmd/gobgp/neighbor.go:1084) became deprecated in 1.18, which is independent from this PR.

fixed.

internal/pkg/config/bgp_configs.go
```diff
@@ -1141,7 +1141,7 @@ type ZebraState struct {
 	MplsLabelRangeSize uint32 `mapstructure:"mpls-label-range-size" json:"mpls-label-range-size,omitempty"`
 	// original -> gobgp:software-name
 	// Configure zebra software name.
-	// frr4, cumulus, frr6, frr7, frr7.2, and frr7.3 can be used.
+	// frr4, cumulus, frr6, frr7, frr7.2 and frr7.3 can be used.
```
fujita, line 1144: Directly modified?

internal/pkg/config/bgp_configs.go
```diff
@@ -1171,7 +1171,7 @@ type ZebraConfig struct {
...
```
fujita, line 1174: ditto.

internal/pkg/config/bgp_configs.go
hardeker: Nope, it appeared with the pyang auto-generation

internal/pkg/config/bgp_configs.go
hardeker: Same

internal/pkg/config/bgp_configs.go
hardeker: It seems to me there was some manual change of `bgp_configs.go` in b55c32ce65c9b6e0c9f4620e6acb27e0e49f238d
At least, I see that in this commit the comma was present in the file `bgp_configs.go` but not in `tools/pyang_plugins/gobgp.yang`

fujita:
pushed, thanks.

PR #2546, opened 2022-04-29
Bmp AdjRIBOut

mayuresh82:
This PR adds support for parsing BMP AdjRIBOut route monitoring messages per https://datatracker.ietf.org/doc/html/draft-ietf-grow-bmp-adj-rib-out-03
It also fixup a couple of things:
- create an IP from data slice by using copy(). Using net.IP() directly can have unexpected results . ( this probably also needs to change in the bgp packet parsing also ).
- for the scanner split func, only parse the header instead of full message since we only need header info to decide message boundaries.

### DIFF
--- a/pkg/packet/bmp/bmp.go
+++ b/pkg/packet/bmp/bmp.go
@@ -51,8 +51,14 @@ const (
 	BMP_PEER_FLAG_IPV6        = 1 << 7
 	BMP_PEER_FLAG_POST_POLICY = 1 << 6
 	BMP_PEER_FLAG_TWO_AS      = 1 << 5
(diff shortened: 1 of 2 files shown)


fujita:
> - create an IP from data slice by using copy(). Using net.IP() directly can have unexpected results
What do you mean?

mayuresh82:
I noticed that peerAddress in the BmpPeerHeader was sometimes mangled ( included garbage data ) when trying to pass it around in my code. Using the copy() function fixed that . I will try to repro this as soon as my BMP test bed is up again.

fujita:
pushed, thanks.

PR #2547, opened 2022-05-03
Listener Queue: reduce length

tobymitico:
This PR reduces the Listener channel queue Length.
this poses a problem causing gobgp to completely trip up on a heavy loaded system with many (hundreds) of peers. eventually it will just stop accepting connections all together. by reducing the queue we stop the kernel from opening sockets and forwarding it to the listeners as it won't be able to process it anyway. by the time it gets to process the openened tcp socket the client will already have killed it due to keepalive timers and initiated a bunch of other connections in the queue.
reducing this queue fixes this and allows clients to eventuall establish even on busy queues/systems

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -2239,7 +2239,7 @@ func (s *BgpServer) StartBgp(ctx context.Context, r *api.StartBgpRequest) error 
 		}
 
 		if c.Config.Port > 0 {
-			acceptCh := make(chan *net.TCPConn, 4096)
+			acceptCh := make(chan *net.TCPConn, 32)
 			for _, addr := range c.Config.LocalAddressList {
 				l, err := newTCPListener(s.logger, addr, uint32(c.Config.Port), g.BindToDevice, acceptCh)
 				if err != nil {


fujita:
> eventually it will just stop accepting connections all together.

https://github.com/osrg/gobgp/blob/master/pkg/server/server.go#L123
the queue is full, so the goroutine will be blocked; the goroutine don't call accept().

>  by reducing the queue we stop the kernel from opening sockets and forwarding it to the listeners

What this means? Can you give the details?

tobymitico:
if the system is busy (and has many clients) the goroutine will be blocked but only after it has already 4000 items in the queue to still process... so as it tries to catch up, the items in the queue are already expired from the client side - the client will disconnect and re-connect again, adding itself back at the end of the queue, wasting cycles and never actually being able to connect at the same time. 

I'm still not sure why eventually gobgp will completely stop processing clients completely. and at that point the kernel will just fill up the connection queue to whatever its limit is (net.core.somaxconn) and no matter what limit you set in net.core.somaxconn it will just fill up. I am not 100% sure why this is happening but reducing the channel queue length fixes both problems: 
1. the client never times out, 
2. 2 connection queue doesn't fill up anymore


see attachment log of tcpdump (ips randomized). this is one tcp connection of a client. you can see the BGP open message (timestamp 16:17:34.767491), which the kernel immediately ACK. but because gobgp is blocked it doesn't process it and is stuck in the channel
and 9 seconds later the keepalive expired message again from the client (timestamp: 16:17:43.767561)

in these 9 seconds gobgp never replies with a BGP open message due to load while still trying to catch up other connectiones ahead in the queue
if the channel queue is shorter. the kernel will no longer even accept the tcp connection and the client will just retry using SYN, (1. it will do so faster without blocking for the keepalive timeout and 2. if there IS a slot opening in the queue, gobgp will actually be fast enough to reply with a BGP open message and connection can establish.

the problem does not occur when connections are already established, the system is able to keep the connections alive

[test.log](https://github.com/osrg/gobgp/files/8682046/test.log)

fujita:
>see attachment log of tcpdump (ips randomized). this is one tcp connection of a client. you can see the BGP open message (timestamp 16:17:34.767491), which the kernel immediately ACK. but because gobgp is blocked it doesn't process it and is stuck in the channel

I don't think that the main gobgp goroutine is blocked. It should never be blocked (if it's blocked, it's a bug). But if the main goroutine is too busy, likely the channel for accepted sockets might not be handled.

> if the channel queue is shorter. the kernel will no longer even accept the tcp connection

Yeah, the goroutine for accept() is blocked when the channel is full so the kernel doesn't send an ACK.

Seems that reducing the length doesn't have any ill effect.  Pushed, thanks!

PR #2548, opened 2022-05-04
Support FRRouting 8.2 and introduce zebra.Software structure.

irino:
Support FRRouting 8.2 and introduce zebra.Software structure to identify FRRouting version.


tools/pyang_plugins/gobgp.yang
```diff
@@ -1267,7 +1267,7 @@ module gobgp {
       type string;
       description
         "Configure zebra software name.
-        frr4, cumulus, frr6, frr7, frr7.2 and frr7.3 can be used.";
+        frr4, cumulus, frr6, frr7, frr7.2, frr7.3, frr7.4, frr7.5, frr8, frr8.1 can be used.";
```
fujita, line 1270: Probably, you need "and" befefore frr8.1. Placing a comma before "and" or not is your choice:
https://en.wikipedia.org/wiki/Serial_comma

btw, I think that the following sentence is more accurate if you have to set one here:
_set to either frr4, cumulus, frr6, frr7, frr7.2, frr7.3, frr7.4, frr7.5, frr8, or frr8.1_

fujita:
Great, thanks a lot!
I have one comment but it's trivial and this change is really nice so I've pushed this.
If you want to fix the comment, please create a new pull request.

tools/pyang_plugins/gobgp.yang
irino: Yes, I agree that "or" should be used instead of "and".
Moreover, in this PR, this variable is not constrainted and it is divided to software name and version in internal Software structure as a result of changing behavior of zapi.go. I think changing comments for tracking releases of FRRouting does not make sense, I want to change this comment fundamentally in next PR.

PR #2549, opened 2022-05-05
Modify comment for sofiware-name in gobgp.yang.

irino:
This PR contains modification of sofiware-name and addition of route-family typedef (for avoid pyang error) in gobgp.yang.
This PR does not contain bgp_congis.go because pyang generates new bgp_configs.go which has big diffrence from current one.


fujita:
This isn't necessary? I didn't have time to set up a working env for pyang so haven't tried this yet.

PR #2554, opened 2022-06-30
fix(cmd/gobgp): parse listen-port as u16, not i16

Tuetuopay:
A port is in the range 0..64k, not -32k..32k. Hence unsigned.


fujita:
Ci tests failed due to an unrelated style issue so rebased and pushed, thanks.

Issue #2555, opened 2022-07-01
Panic calling table.GetLongerPrefixDestinations with invalid cidr


floatingstatic:
When sending an invalid cidr string with gobgp we cause gobgpd to panic:
```
$ gobgp global rib 203.0.113.0/24
Network not in table
$ gobgp -p 50700 global rib 203.0.113.0 longer-prefixes
rpc error: code = Unavailable desc = error reading from server: EOF
```

Trace:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xaba49e]
goroutine 173 [running]:
github.com/osrg/gobgp/v3/internal/pkg/table.(*Table).GetLongerPrefixDestinations(0xc000992520, {0xc2c4d6c280, 0xe})
 /go/src/github.com/<REDACTED>/vendor/github.com/osrg/gobgp/v3/internal/pkg/table/table.go:215 +0xde
github.com/osrg/gobgp/v3/internal/pkg/table.(*Table).Select(0xc000992520, {0xc028f09e20, 0x1, 0xc?})
 /go/src/github.com/<REDACTED>/vendor/github.com/osrg/gobgp/v3/internal/pkg/table/table.go:388 +0x62b
github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).getRib.func1()
 /go/src/github.com/<REDACTED>/vendor/github.com/osrg/gobgp/v3/pkg/server/server.go:2538 +0x1cd
github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).handleMGMTOp(0xf85900?, 0xc315c16d68)
 /go/src/github.com/<REDACTED>/vendor/github.com/osrg/gobgp/v3/pkg/server/server.go:273 +0x89
github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).Serve(0xc0004daf00)
 /go/src/github.com/<REDACTED>/vendor/github.com/osrg/gobgp/v3/pkg/server/server.go:460 +0x5da
created by github.com/<REDACTED>/internal/routed.(*Server).ServeBGP
 /go/src/github.com/<REDACTED>/internal/routed/bgp.go:58 +0xbc
```

PR #2556, opened 2022-07-01
Fix panic in table.GetLongerPrefixDestinations() with invalid cidr

floatingstatic:
Fixes issue outlined in https://github.com/osrg/gobgp/issues/2555


floatingstatic:
Unrelated to the original change I address a linter error that was picked up by CI checks in this repo:

```
internal/pkg/table/path.go#L73
S1001: should use copy() instead of a loop (gosimple)
```

fujita:
thanks!

PR #2559, opened 2022-07-21
ci: use go 1.18

higebu:
go 1.19 will be released in August 2022 and go 1.17 will reach end of
life in August 2022.
Also update go version in go.mod with `go mod tidy -go=1.18`.


fujita:
Pushed, thanks! I'll fix the CI failure unrelated with this pr.

PR #2564, opened 2022-08-27
packet/bmp: enable parsing/serializing add-path-enabled BGP updates

vincentbernat:
`ParseBody()` and `Serialize()` need to be aware if we Add-Path is
enabled to be able to parse/serialize BGP updates correctly.

This solves only half of the problem: server/bmp also needs to be
updated.

### DIFF
--- a/pkg/packet/bmp/bmp.go
+++ b/pkg/packet/bmp/bmp.go
@@ -173,19 +173,19 @@ func NewBMPRouteMonitoring(p BMPPeerHeader, update *bgp.BGPMessage) *BMPMessage 
 	}
 }
 
-func (body *BMPRouteMonitoring) ParseBody(msg *BMPMessage, data []byte) error {
+func (body *BMPRouteMonitoring) ParseBody(msg *BMPMessage, data []byte, options ...*bgp.MarshallingOption) error {
 	var err error
 
-	body.BGPUpdate, err = bgp.ParseBGPMessage(data)
+	body.BGPUpdate, err = bgp.ParseBGPMessage(data, options...)
 
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


vincentbernat:
The exposed API is not usable. The Add-Path capability is tied to a peer. We cannot just pass a set of options to `ParseBMPMessage()`. It could be:
- a map with peers as keys
- a function to tell the options to use when provided with the BMP header
I think the second option is better?

fujita:
Adding bgp.MarshallingOption looks reasonable but we can't break APIs without major version updated so needs to add new functions like ParseBodyWithOption, I guess.

vincentbernat:
Even if the additional arguments are provided through variadic arguments?

fujita:
Oops, adding variadic arguments is fine because the existing code doesn't need modification, I guess.

vincentbernat:
I have added a `ParseBMPMessageWithOptions` taking a function providing options to use from the peer header. No other changes.
The build failure is unrelated.

fujita:
How about simply adding `options ...*bgp.MarshallingOption` to ParseBMPMessage()? I prefer less exported functions.

vincentbernat:
Because you need to know which peer it is to know if you need to enable Add-Path or not.

fujita:
Can the caller executes optionsFunc() and passes the option to ParseBMPMessage()?

vincentbernat:
It cannot. The Add-Path is per-BGP peer. For one BMP peer, you have several BGP peers (in case you use Adj-RIB-in). You need to parse the open messages in the peer up notifications to know which BGP peer should send Add-Path NLRI. And only for these peers, you need to pass the appropriate option.

fujita:
Understood. Pushed, thanks a lot!

PR #2565, opened 2022-08-29
ci: use ubuntu-20.04

vincentbernat:
ubuntu-18.04 is deprecated. Moreover, the shipped versions for pip
does not know how to handle Python's cryptography correctly. Let's
check if it works with ubuntu-20.04.


fujita:
thanks!

PR #2569, opened 2022-09-01
mup: fixes for BGP-MUP

higebu:
- pkg/packet/bgp,api: add SRv6 behavior for SRv6 MUP
- mup: fixes for BGP-MUP
    - Route Target support for MUP routes
    - Only Direct Segment Discovery route needs MUP Extended Community
    - TEID value of 0 is invalid

Fixed usage: https://gist.github.com/higebu/90c0504545978e0cf12cc97b89cb5c81

Known issues:
- Interwork Segment Discovery route and Direct Segment Discovery route update doesn't have a prefix SID attribute. Because gobgp doesn't support SRv6 locator settings for now.
    - Workaround: use AddPath API.


fujita:
Thanks!

PR #2572, opened 2022-09-06
bmp: correctly encode path identifier in NLRI when needed

vincentbernat:
The FSM has the right decoding options. First, we change the
`marshallingOptions` field to a slice instead of a single option for
consistency with the remaining of the code (`Serialize()` and
`Parse*()` accepts several options). This is not really needed as we
only initialize a single option or none, but I suppose this may matter
in the future.

Then, we pass the decoding options to the `watchUpdateEvent` struct.
From my understanding, this is the only one needing it. When
transmitting the local RIB, we don't need to encode paths with
AddPath. And when mirroring routes, the routes are already encoded.

I am however unable to correctly pass the decoding options in the
post-policy case. I don't see an obvious way to access the FSM in this
case.

### DIFF
--- a/pkg/server/bmp.go
+++ b/pkg/server/bmp.go
@@ -154,8 +154,8 @@ func (b *bmpClient) loop() {
 				tickerCh = t.C
 			}
 
-			write := func(msg *bmp.BMPMessage) error {
-				buf, _ := msg.Serialize()
+			write := func(msg *bmp.BMPMessage, options ...*bgp.MarshallingOption) error {
+				buf, _ := msg.Serialize(options...)
 				_, err := conn.Write(buf)
 				if err != nil {
 					b.s.logger.Warn("failed to write to bmp server",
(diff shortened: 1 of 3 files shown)


fujita:
Thanks for the patch, and sorry for the long silence.

The bug is real, but it is now fixed in master by a different set of
changes. This PR was written against v3 and no longer applies.

Two details differ from the patch. Adj-RIB-In monitoring follows the
receive direction of the add-path capability, not the send direction
the encoder uses. And the identifier to report is the one the peer
sent, not the one gobgp assigns for its own re-advertisement (RFC 7911
section 2). The post-policy case you could not reach is handled: the
negotiated state is on the watch event, in Neighbor.

Route mirroring must not re-encode at all. RFC 7854 section 4.7 asks
for "verbatim duplication of messages as received", so gobgp now
forwards the octets it received.

  18d807d3 server/bmp: build route monitoring messages in a separate function
  7aca34a1 server/bmp: mirror the message as it was received
  4ba57249 packet/bmp: parse the mirrored message with the session options
  b05a9897 server/bmp: encode the path identifier in route monitoring
  3f856cda server/bmp: key the route monitoring cache by path identifier

Closing. Thanks again.

PR #2577, opened 2022-09-20
bgp: turn RouteFamily into uint32

vincentbernat:
1. On 64-bit platform, it takes less space to store an uint32 because an int is equivalent to int64. An uint32 is enough as AFI is 16-bit and SAFI is 8-bit.

2. On 32-bit platform, int is int32 and it is not big enough to encode all route families, notably the ones with AFI > 2**15. At runtime, this is not a problem, however, at compile time, we get an error.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -8273,7 +8273,7 @@ func RouteFamilyToAfiSafi(rf RouteFamily) (uint16, uint8) {
 	return uint16(int(rf) >> 16), uint8(int(rf) & 0xff)
 }
 
-type RouteFamily int
+type RouteFamily uint32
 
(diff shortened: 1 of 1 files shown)


fujita:
makes sense but is it possible that this breaks the existing code?

vincentbernat:
I am unsure. You cannot store to an int without casting. I am unable to find a case where it would break.

fujita:
I prefer keep this open until the next major release because I'm not completely sure that this doesn't break anything.

vincentbernat:
On 2022-09-23 12:00, FUJITA Tomonori wrote:
> I prefer keep this open until the next major release because I'm not 
> completely sure that this doesn't break anything.

No problem!

fujita:
Merged, thanks!

PR #2579, opened 2022-09-28
mup: handle Type 1 ST Route Prefix as Prefix instead of single address

higebu:
See https://www.ietf.org/archive/id/draft-mpmz-bess-mup-safi-01.html#name-bgp-type-1-session-transfor for details.


api/attribute.proto
```diff
@@ -391,7 +391,7 @@ message MUPType1SessionTransformedRoute {
   // - RouteDistinguisherIPAddress
   // - RouteDistinguisherFourOctetASN
   google.protobuf.Any rd = 1;
-  uint32 prefix_length = 2;
+  uint32 prefix_length = 2 [deprecated = true];
```
fujita, line 394: No code uses this field?

higebu: Yes.

higebu: There are 3 options to remove existing field.

1. Remove the field.
2. [Add deprecated field option](https://developers.google.com/protocol-buffers/docs/proto3#options).
3. [Reserve the field number](https://developers.google.com/protocol-buffers/docs/proto3#reserved).

Which one is better for the gobgp API?

fujita:
`deprecated` looks fine. Pushed, thanks a lot!

PR #2586, opened 2022-10-14
Refactor Prefix-SID and support SRv6 L2 Service TLV

higebu:


api/attribute.proto
```diff
@@ -841,10 +841,10 @@ message UnknownAttribute {
...
 }
 
-// https://tools.ietf.org/html/draft-dawra-bess-srv6-services-02#section-2.1.1
+// https://www.rfc-editor.org/rfc/rfc9252.html#section-3.2.1
 message SRv6StructureSubSubTLV {
-  uint32 local_block_length = 1;
-  uint32 local_node_length = 2;
+  uint32 locator_block_length = 1;
+  uint32 locator_node_length = 2;
   uint32 function_length = 3;
```
fujita, line 848: rename?

higebu: Thank you for your review. I'll add locator_* as new field and mark old fields as deprecated.

fujita: This feature is still based on RFC drafts. Breaking is fine by me.

fujita:
I'll push this pull request as is if it's ok with you.

api/attribute.proto
higebu: If gobgp prefers RFC-compliant to API backward compatibility, I think we can rename these fields. Because this PR is based on RFC9252.

higebu:
@fujita It's ok for me. Please merge this PR. Thank you.

fujita:
pushed, thanks.

PR #2592, opened 2022-10-26
grpc: fix goroutine leak in `WatchEvent`

Tuetuopay:
The context used there was a background context, which was not inherited from the stream. Thus when the client ends the stream (e.g. ^C `gobgp monitor global rib`), the context never completed, preventing the stop of most goroutines responsible for forwarding events. The only case where it completes is when an event gets generated, the transmit channel was closed, ending the chain of goroutines.

Fix by inheriting the context from the stream, which completes when the stream ends, properly cleaning up all resources.

### Steps to reproduce

1. Start gobgp
2. Observe the goroutine count using the pprof http server
3. Run and stop multiple `gobgp monitor global rib` and observe the goroutine count go up

Using the pprof interfaces also confirms that all the leaky goroutines are just parked and not doing anything.


fujita:
Great, thanks a lot!

PR #2595, opened 2022-10-28
mup: fix decode/serialize prefix in Interwork Segment Discovery Route

higebu:
This PR fixes decoding and serializing the Prefix in Interwork Segment Dicsovery Route. The Prefix should be encoded as RFC 4760 5. NLRI Encoding: https://www.rfc-editor.org/rfc/rfc4760.html#section-5

>    b) Prefix:
> 
>       The Prefix field contains an address prefix followed by enough
>       trailing bits to make the end of the field fall on an octet
>       boundary.  Note that the value of trailing bits is irrelevant.


fujita:
thanks!

PR #2597, opened 2022-11-01
pkg/packet/bgp: initial support for BGP VPLS

### LINKED ISSUE #2590: Support for VPLS route family
Hey!

Currently, route family VPLS (AFI L2VPN/SAFI VPLS) is not supported. A BGP session with the family enabled will trigger an error during parsing due to the inability to instantiate a prefix.

I can try to implement it as it seems simple enough (it is not parametrized). Is there a recent commit adding a route family that I could use?

vincentbernat:
This includes BGP VPLS NLRI and extended community. There is no CLI/API support yet. Also, VPLS AFI/SAFI is shared with RFC 6074 which says:

> Both BGP-AD and VPLS-BGP [RFC4761] use the same AFI/SAFI.  In order
> for both BGP-AD and VPLS-BGP to co-exist, the NLRI length must be
> used as a demultiplexer.
>
> The BGP-AD NLRI has an NLRI length of 12 bytes, containing only an
> 8-byte RD and a 4-byte VSI-ID.  VPLS-BGP [RFC4761] uses a 17-byte
> NLRI length.  Therefore, implementations of BGP-AD must ignore NLRI
> that are greater than 12 bytes.

This is not handled in this PR.

The generic transitive experimental extended community types are shared with FlowSpec. The parsing is extended to handle VPLS as well, but not other L2 encapsulations.

Fix #2590

Please note, I don't know much about all that. Don't assume anything I wrote is right. I am relying on the 4 written tests to ensure some correctness. The encoded NLRI and extended communities are extracted from a vMX setup.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -184,6 +184,55 @@ const (
 	EC_SUBTYPE_UUID_BASED_RT ExtendedCommunityAttrSubType = 0x11
 )
 
+// RFC6624
+type Layer2EncapsulationType uint8
+
+const (
+	LAYER2ENCAPSULATION_TYPE_FRAMERELAY      Layer2EncapsulationType = 1
+	LAYER2ENCAPSULATION_TYPE_ATM_AAL5        Layer2EncapsulationType = 2
+	LAYER2ENCAPSULATION_TYPE_ATM_TRANSPARENT Layer2EncapsulationType = 3
+	LAYER2ENCAPSULATION_TYPE_ETHERNET_VLAN   Layer2EncapsulationType = 4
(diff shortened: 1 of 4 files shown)


fujita:
Great, thanks a lot!
Looks like needs to add `vpls` to https://github.com/osrg/gobgp/blob/master/tools/spell-check/dictionary.txt

vincentbernat:
Done!

fujita:
pushed, thanks! looking forward to follow-up pull requests.

vincentbernat:
> pushed, thanks! looking forward to follow-up pull requests.

Oh, sorry if I didn't say so before. I am only interested in the parsing part, so I didn't plan to do more. If you prefer to not have incomplete stuff, feel free to revert.

fujita:
Understood. No problem. Someone might implement in the future.

Issue #2593, opened 2022-10-26
Request to export package `internal/pkg/config`


wenovus:
Since toml/yaml configuration is part of the public API, I'm wondering if `config.BgpConfigSet` (defined at internal/pkg/config/bgp_configs.go) can safely be made public too since they're mapped 1:1? It would enable programmatic access of declarative configuration by allowing functions like `InitialConfig`/`UpdateConfig` ([ref](https://pkg.go.dev/github.com/osrg/gobgp/v3@v3.7.0/pkg/config)) to be accessible to external callers. Otherwise users only have config file access to the declarative configuration logic via gobgpd, or take the untasteful alternative of having to fork and modify a large chunk of the repository.

Doing this likely also requires exporting `internal/pkg/table` since `InitialConfig`/`UpdateConfig` refers to this as well.

Thanks,
Wen

fujita:
Why can't you use InitialConfig/UpdateConfig()?

> Doing this likely also requires exporting internal/pkg/table since InitialConfig/UpdateConfig refers to this as well.

internal/pkg/table includes many internal stuff that shouldn't be public.

wenovus:
> Why can't you use InitialConfig/UpdateConfig()?
> 
Because [config](https://pkg.go.dev/github.com/osrg/gobgp/v3@v3.7.0/internal/pkg/config).[BgpConfigSet](https://pkg.go.dev/github.com/osrg/gobgp/v3@v3.7.0/internal/pkg/config#BgpConfigSet) is in an internal package, so you can't create it programmatically. The only way is to create a config file and call `ReadConfigFile` on it like how gobgpd does it.

fujita:
I see. config.BgpConfigSet can be exported but I don't think that we can export internal/pkg/table.

wenovus:
> internal/pkg/table includes many internal stuff that shouldn't be public.
If policy.go can be moved to its own package that's exported that would also work. Do you think this makes for correct packaging?

The benefit of doing this would be to allow for writing a custom config handler for policies similar to `InitialConfig`/`UpdateConfig` where they refer to the definitions in policy.go.

fujita:
policy.go needs the implementation details, which should not be public. Needs to split policy.go; config and implementation, I guess.

wenovus:
Would you be open to a PR that exports `internal/pkg/config` and the public parts in policy.go which InitialConfig/UpdateConfig() use? We haven't finalized our design yet but this might influence our design decision. Thanks.

[...]

PR #2599, opened 2022-11-04
Make generated BGP configuration code public

wenovus:
Move internal/pkg/config -> pkg/bgpconfig

Please advise if there are internal stuff here and I can try to figure out if I can move them back to internal.

Background: https://github.com/osrg/gobgp/issues/2593


fujita:
Why simply moving to `pkg/config`?

wenovus:
> Why simply moving to `pkg/config`?

Two reasons:
One is to avoid taking away "config" as a variable name.

https://go.dev/blog/package-names
```
Don’t steal good names from the user. Avoid giving a package a name that is commonly used in client code. For example, the buffered I/O package is called bufio, not buf, since buf is a good variable name for a buffer.
```

Another reason is "config" is a [generic name](https://pkg.go.dev/search?q=config), so for example if I'm writing a virtual router with configuration from different protocols "config" doesn't really tell me which one I'm working with, making the code harder to read.

wenovus:
> I don't think `config` is a good variable name.
> 
> > Another reason is "config" is a [generic name](https://pkg.go.dev/search?q=config), so for example if I'm writing a virtual router with configuration from different protocols "config" doesn't really tell me which one I'm working with, making the code harder to read.
> 
> You can't import a package as a different name?
> 
> I don't think using `bgp` prefix for gobgp packages is a good idea. Just verbose.

I see, that's reasonable. However, when I merge these packages I run into a dependency cycle:
package github.com/osrg/gobgp/v3/cmd/gobgp                                                                                                         
        imports github.com/osrg/gobgp/v3/internal/pkg/table                                                                                        
        imports github.com/osrg/gobgp/v3/pkg/config                                                                                                
        imports github.com/osrg/gobgp/v3/internal/pkg/table: import cycle not allowed                                                              

This is because "table" now imports types from "config" (aka. "bgconfig"), which in turn imports from the table package.
```
table/destination.go
31:var SelectionOptions config.RouteSelectionOptionsConfig
32:var UseMultiplePaths config.UseMultiplePathsConfig
121:func NewPeerInfo(g *config.Global, p *config.Neighbor) *PeerInfo {
```

When I moved policy.go out into its own package to avoid "config" importing "table", I now see
package github.com/osrg/gobgp/v3/internal/pkg/table                                                                                                                                                                                                                                                   
        imports github.com/osrg/gobgp/v3/pkg/config                                                                                                
        imports github.com/osrg/gobgp/v3/pkg/server                                                                                                
        imports github.com/osrg/gobgp/v3/internal/pkg/table: import cycle not allowed                                                              

This is because the original "config" refers to server.BgpServer, and "server" in turn imports from the internal "table" package other than policy.go

```
server/rpki.go     
62:     table     *table.ROATable                                       
66:func newROAManager(table *table.ROATable, logger log.Logger) *roaManager {
101:    m.table.DeleteAll(host)
```

What do you think is an appropriate solution?

wenovus:
Hi, a couple ideas here to solve the dependency issue:
1. The current `config` package is not about GoBGP's configuration schema, but rather about applying the configuration on an existing GoBGP server. So, we can justify calling the `internal/pkg/config` package `pkg/schema` instead of merging it with the existing `pkg/config` to make it clear that this is about the configuration schema rather than configuration operations. In this way we also avoid the dependency cycle.
2. We can directly break the dependency cycle by creating an interface for `server.BgpServer`'s methods and have `pkg/config` call its functions instead of referring to the `server.BgpServer` struct directly.

PR #2601, opened 2022-11-18
Add enable_only_binary flag to grpc ListPath method

bayrinat:
Adding enable_only_binary allows using only binary representation of nlri and attributes on the ListPath call. For clients who uses only binary representation it helps to significantly reduce resource consumption by refusing unnecessary conversion.

It is vital while processing a large number of paths, e.g. full-view.

This change doesn't break backward compatibility.


bayrinat:
@fujita hello, could you plz merge this tiny change?

fujita:
Looks fine. Pushed, thanks!

Issue #2591, opened 2022-10-21
gRPC - ListPathRequest wrong type


cpmarvin:
i'm using the script here (https://github.com/osrg/gobgp/pull/2539/commits/ca69adce8bcc7d0eeb0e946d036c9b341635950f#diff-026dfc5dec62f6ff2c0bd28d262c042601b3c732984574bf95f63e03b33e28ed):  to create an SR-TE policy.  I can see the attributes are programmed with the correct url/encapsulation 

```
[type_url: "type.googleapis.com/apipb.NextHopAttribute"
value: "\n\r10.192.180.17"
, type_url: "type.googleapis.com/apipb.OriginAttribute"
, type_url: "type.googleapis.com/apipb.ExtendedCommunitiesAttribute"
value: "\nJ\n5type.googleapis.com/apipb.IPv4AddressSpecificExtended\022\021\020\002\032\r10.192.180.18"
, type_url: "type.googleapis.com/apipb.TunnelEncapAttribute"
value: "\n\220\003\010\017\022=\n7type.googleapis.com/apipb.TunnelEncapSubTLVSRPreference\022\002\020\013\022m\n7type.googleapis.com/**apipb.TunnelEncapSubTLVSRBindingSID**\0222\n0\n&type.googleapis.com/apipb.SRBindingSID\022\006\032\004\000\004\223\344\022\335\001\n8type.googleapis.com/apipb.TunnelEncapSubTLVSRSegmentList\022\240\001\n\002\020\014\0222\n&type.googleapis.com/apipb.SegmentTypeA\022\010\n\000\020\200\340\320\206\003\0222\n&type.googleapis.com/apipb.SegmentTypeA\022\010\n\000\020\200\200\320\206\003\0222\n&type.googleapis.com/apipb.SegmentTypeA\022\010\n\000\020\200\300\320\206\003"
]

```

When listing the paths with below code i see the type beeing set to any for one of the TLV field.

```
family = gobgp_pb2.Family(
        afi=gobgp_pb2.Family.AFI_IP, safi=gobgp_pb2.Family.SAFI_SR_POLICY
    )

path_list = stub.ListPath(
    gobgp_pb2.ListPathRequest(
        family=family
    )
)
paths = list(path_list)

ipdb> paths[0].destination.paths[0].pattrs[2]
type_url: "type.googleapis.com/apipb.TunnelEncapAttribute"
[...]

PR #2602, opened 2022-11-20
#2591 Correct value of bsid in grpc list path

cpmarvin:
This allows ListhPath request for a SR Policy to correctly display the type and decode the value inside the TLVs


fujita:
Removed gobgp binary in this pull request and pushed. Thanks!

PR #2608, opened 2022-12-13
Watch EOR grpc method

bayrinat:
Add new EOR type for WatchEvent method.
It provides opportunity of getting 'readiness' of the current peer.


bayrinat:
@fujita what do you think?

bayrinat:
@fujita gentle ping.

bayrinat:
> Sorry about the delay. why you don't use TableEvent?

Thanks for replying.
Could you please explain the reasons? Is the TableEvent is a decent type? 
 
I mean EOR-method signals that all eor messages from peers are received. It corresponds with table and peers as well. 
If I add it to TableEvent or to PeerEvent, it will be confusing a little bit.

fujita:
I thought that simply EoR messages can be sent via TableEvent.Paths. But you prefer to know when a peer gets EoR messages in all families? If so, the proposal looks like reasonable.

bayrinat:
Actually, I like your idea to watch each eor message via Paths. It will be even more useful, will be come soon with implementation.

bayrinat:
I've tried to make it at least minimalistic as possible.

bayrinat:
@fujita What do you think about this PR?

api/gobgp.proto
```diff
@@ -125,7 +125,7 @@ message WatchEventRequest {
 
   message Table {
     message Filter {
-      enum Type { BEST = 0; ADJIN = 1; POST_POLICY = 2; }
+      enum Type { BEST = 0; ADJIN = 1; POST_POLICY = 2; EOR = 3; }
```
fujita, line 128: Is this necessary? ADJIN is supposed to send all messages include EoR?

bayrinat: > Is this necessary? ADJIN is supposed to send all messages include EoR?

Adding a new filter provides opportunity of getting eor message itself without handling adj-in tables that can be huge.
I thought about adding it to Best actually, but it will probably not the right place (tell me if I'm wrong).

fujita: I see. Is this backward compatible?

bayrinat: It looks so. If you don't add a filter, you won't get an eor message.

fujita: The old binary (using Watch gRPC API) works with the gobgpd binary including this change?

mai2412:
I would really like this feature as well (unless there was another way of watching EOR). Would you recommend opening a new pr or is it possible to merge this?

fujita:
If someone confirms that this change doesn't break the backward compatibility of the gRPC API, then I'll merge.

mai2412:
Tested that:
$ gobgp monitor global rib / adj-in / neighbor
works with the gobgpd built from this change.

mai2412:
@fujita  Ahh I see that the POST_POLICY filter has to be [used](https://github.com/osrg/gobgp/blob/master/pkg/server/server.go#L4635), I did not see this before and only tried testing with BEST and ADJIN. It would be ideal if there's a filter for getting just the EOR message without the other updates. I think the original proposal of this PR that introduces a new message type EOR might also work well. What do you think?

Do correct me if i said anything wrong, thank you for looking into this.

bayrinat:
At last I rebased the code with some minor changes.

- This change still doesn't break the backward compatibility and works stable.

bayrinat:
@fujita gentle ping

fujita:
Pushed, thanks.

PR #2610, opened 2022-12-25
Add Windows Support

GSGBen:
Both `gobgpd` and `gobgpd` now build for and run on Windows. This is based on https://github.com/osrg/gobgp/pull/2250/ with less required changes. Tested and confirmed it can peer successfully and receive routes. It still builds on Linux (doesn't change syscalls for non-windows files).


fujita:
Someone actually tries to use GoBGP on Windows? I'm not sure but seems this doesn't break anything so I pushed it. Thanks!

PR #2611, opened 2022-12-27
BGP-LS extensions for Egress Peer Engineering of Segment Routing (RFC9086)

kanaya516:
Currently, GoBGP cannot advertise Egress Peer Engineering (EPE) SIDs.
This pull request is an extension of BGP-LS for EPE as defined in [RFC9086](https://datatracker.ietf.org/doc/rfc9086/).
This pull request modifies BGP-LS to send and receive BGP Peer Node/Adjacency/Set SID of SR-MPLS SID for EPE.
Interoperability was verified.
The following devices were able to receive and send SR-MPLS SIDs for EPE via BGP-LS with GoBGP.
- Juniper cRPD 22.2R1.9
- Cisco XRd 7.8.1


fujita:
Thanks a lot!

PR #2613, opened 2023-01-04
evpn type2 use in graceful restart crash

liangrunqiang:
Gobgp will crash when use evpn type2 sync in graceful restart and peer(crash or out by kill -9/reboot)


fujita:
thanks!

PR #2615, opened 2023-01-07
Natively modify host route table

GSGBen:
This pull request provides experimental functionality to inject learned routes into the routing table of the host running GoBGP. i.e. it provides a subset of something like Zebra's functionality without requiring the external dependency of Zebra.

The reference implementation provided is for Windows, but it's structured using the existing `<filename>_<platform>.go` pattern so it can be extended to other platforms. The implementation and flow mirrors the Zebra integration for consistency.

Config is under a new top-level `experimental` level to emphasize to users this isn't mainline functionality. Documentation heavily emphasizes this too.

Documentation is provided on usage, architecture/flow, and how to extend, in `docs/sources/modify-host-fib.md`.

I've tested this on Windows 10 with adds, updates, deletes, cleanup-on-shutdown (not sure if the existing Zebra integration does this but I wanted it), and a reasonably sized BGP topology - all is working. I haven't added any tests to CI/CD as GitHub Actions doesn't work well (or at all) with Windows containers. If in the future someone adds Linux support, it would be easy to add tests for that using the existing test structure you have because each container will have a route table you can check. 

For Windows route table support I've added a dependency on wireguard-go, which exposes Windows route table APIs in a clean way. Wireguard-go is a well known and trusted project.


fujita:
I like the idea of playing with the routing table in operating systems directly but I'm not sure about the needs for running GoBGP in Windows with this feature. It's very difficult to maintain a feature that few people need.

GSGBen:
That's a good point.

In the documentation I've mentioned **This is not supported by the core GoBGP team at all. Use at your own risk.** Would it work if under that I added

**Additionally, the Windows implementation is not maintained by the core GoBGP team. Issues regarding the Windows implementation, and all modify-host-fib functionality will be closed.**

Then you wont have any additional maintenance burden.

sc68cal:
I would like to see more support for GoBGP doing native route table implementation, since having to split configuration between GoBGP and Zebra was [somewhat annoying](https://gist.github.com/sc68cal/7a4e8c59de053245dd6e766116817b9f). It would be great to be able to configure all of it in a single place, instead of juggling route maps in two different components.

olljanat:
> I like the idea of playing with the routing table in operating systems directly but I'm not sure about the needs for running GoBGP in Windows with this feature. It's very difficult to maintain a feature that few people need.

@fujita is there some reason why it have not been implemented but instead of GoBGP is relying for Zebra?

As far I see, it is actually quite simple to create and remove local routes with https://github.com/vishvananda/netlink ( `RouteAdd` and `RouteDel` functions).

fujita:
> > I like the idea of playing with the routing table in operating systems directly but I'm not sure about the needs for running GoBGP in Windows with this feature. It's very difficult to maintain a feature that few people need.
> 
> @fujita is there some reason why it have not been implemented but instead of GoBGP is relying for Zebra?
> 
> As far I see, it is actually quite simple to create and remove local routes with https://github.com/vishvananda/netlink ( `RouteAdd` and `RouteDel` functions).

What about non-Linux OSes?

adamdunstan:
I have created a fork to import and export routes using netlink. I did it as a component of the kubernetes Load Balancer called PureLB.  In addition their is a CRD controller for configuring gobgp.  Its at https://github.com/purelb/gobgp-netlink.  

Adding netlink also allowed me to fix IPv6 link local peers.

scratchmex:
> What about non-Linux OSes?

can't we have a feature only for linux? let the user enable if needed, see https://github.com/purelb/gobgp-netlink approach

> It's very difficult to maintain a feature that few people need.

I think its very useful since one of the most common use cases is to use a linux machine as a router

adamdunstan:
I'm doing an update to gobgp-netlink now.  Merging in the key bug fixes and a couple of the new features since I created the friendly forked version.     

As I mentioned, there is also some changes  that uses netlink to get IPv6 link local addresses dynamic peering.  When I took a look at the main, the problem still exists.  

Its worth noting that the windows reason is not really valid, I dont think Zebra runs on windows (the other method for doing route redistribution that adds a lot of complexity).  I left zebra in my fork but if I find time, I will probably remove it.

As this stage it would be difficult, not impossible, to merge gobgp-netlink back to orsg/gobgp.

PR #2617, opened 2023-01-08
BGP-Confederation implementation for BGP-LS extensions

kanaya516:
We implemented BGP-LS extensions for Egress Peer Engineering of Segment Routing in #2611.
However #2611 does not support Member-ASN TLV of BGP Confederation mentioned in [RFC9086](https://tools.ietf.org/html/rfc9086#section-4.3).

This pull request modifies the BGP-LS to send and receive Member-ASN via BGP-LS.


kanaya516:
We added tests for BGP Confederation.

fujita:
Thanks!

Issue #2616, opened 2023-01-08
BGP-LS EPE Implementation


watal:
Hi. We are working on BGP-LS Egress Peer Engineering (BGP-LS EPE) extensions. (e.g., https://github.com/osrg/gobgp/pull/2611)
There are two BGP-LS extensions for EPE: **SR-MPLS** mentioned in RFC9086 and **SRv6** in draft-ietf-bgpls-srv6-ext.
We will implement these two approaches. 

To share our BGP-LS EPE Roadmap, we opened issues as follows. We are looking forward to heading from the community members to share other tasks related to EPE or contribute to our activities.

Thank you.

## SR-MPLS 
Reference： [RFC9086](https://datatracker.ietf.org/doc/rfc9086/)
- [x] C-Plane implementation (#2611)
    - Note: We have not implemented the BGP Confederation option (RFC9086 section 4.2/4.3).
- [ ] Essential BGP-LS data exchange functions for gRPC
    - Current GoBGP does not support the function to exchange data using gRPC for BGP-LS.
- [ ] The functions to add and remove the BGP Peering Segments to BGP rib according to the BGP peer state.
    - We will implement PeerNode SID to steer to a specific peer node.
    - We will only implement the function to declare the EPE SID statically from the config. (Not automatically generate the EPE SID)
- [ ] FIB manipulation
    - We have not yet organized the necessary implementation required.

## SRv6
Reference: [draft-ietf-bgpls-srv6-ext](https://datatracker.ietf.org/doc/draft-ietf-idr-bgpls-srv6-ext/)
- [ ] C-plane implementation
- [ ] Essential BGP-LS data exchange functions for gRPC
- [ ] The functions to issue and remove the BGP Peering Segments according to the BGP peer state.
- [ ] FIB manipulation
    - This has yet to be considered either.

## Others
- [ ] Documentation
    - We will create documents for EPE use cases using BGP-LS.
- [ ] Tutorial
    - We will summarize examples of simple network topology necessary for EPE using our BGP-LS implementation.
    - This tutorial may be published to external references like [Tutorial: Using GoBGP as an IXP connecting router](https://www.slideshare.net/shusugimoto1986/tutorial-using-gobgp-as-an-ixp-connecting-router).

fujita:
Great, sounds exciting!

> - [ ] Essential BGP-LS data exchange functions for gRPC
>    - Current GoBGP does not support the function to exchange data using gRPC for BGP-LS.

Finally!
[...]

PR #2619, opened 2023-01-10
Improving the processing of SID flag of BGP-LS EPE.

yas-nyan:
This PR is a part of #2616 
Modify BGP-LS EPE related structures, `bgp.LsAttributeBgpPeerSegment`. 
The flags and SID length can now be correctly processed via gRPC. This feature is required to insert a BGP-EPE route via gRPC.

Also, BGP Peering Segments can now be displayed correctly from the CLI in an easy-to-understand manner.

ref: https://datatracker.ietf.org/doc/html/rfc9086

- `root@gobgp:/# gobgp global rib -a ls`

```
*> NLRI { LINK { LOCAL_NODE: 1.1.1.1 REMOTE_NODE: 2.2.2.2 LINK: 10.0.0.1->10.0.0.2} }                  172.100.100.101      65001                00:18:54   [{Origin: i} {LsAttributes: {Peer Node SID: 1000002} }]
```

- `root@gobgp:/# gobgp global rib -a ls`

```json
{
    "NLRI { LINK { LOCAL_NODE: 1.1.1.1 REMOTE_NODE: 2.2.2.2 LINK: 10.0.0.1-\u003e10.0.0.2} }": [
        {
            "nlri": {
                "type": 2,
                "length": 81,
                "nlri": "NLRI { LINK { LOCAL_NODE: 1.1.1.1 REMOTE_NODE: 2.2.2.2 LINK: 10.0.0.1-\u003e10.0.0.2} }"
            },
            "age": 1673351046,
            "best": true,
            "attrs": [
                {
                    "type": 1,
                    "value": 0
                },
                {
                    "type": 2,
                    "as_paths": [
                        {
                            "segment_type": 2,
                            "num": 1,
                            "asns": [
                                65001
                            ]
                        }
                    ]
                },
                {
                    "type": 29,
                    "flags": 128,
                    "node": {},
                    "link": {},
                    "prefix": {},
                    "bgp_peer_segment": {
                        "bgp_peer_node_sid": {
                            "flags": {
                                "value": true,
                                "local": true,
                                "backup": false,
                                "persistent": true
                            },
                            "weight": 0,
                            "sid": 1000002
                        }
                    }
                },
                {
                    "type": 14,
                    "nexthop": "172.100.100.101",
                    "afi": 16388,
                    "safi": 71,
                    "value": [
                        {
                            "type": 2,
                            "length": 81,
                            "nlri": "NLRI { LINK { LOCAL_NODE: 1.1.1.1 REMOTE_NODE: 2.2.2.2 LINK: 10.0.0.1-\u003e10.0.0.2} }"
                        }
                    ]
                }
            ],
            "stale": false,
            "source-id": "1.1.1.1",
            "neighbor-ip": "172.100.100.101"
        },
    ]
}
```


api/attribute.proto
```diff
@@ -830,10 +830,23 @@ message LsAttributePrefix {
...
+  uint32 sid = 3;
+}
+
 message LsAttributeBgpPeerSegment {
-	uint32 bgp_peer_node_sid = 1;
-	uint32 bgp_peer_adjacency_sid = 2;
-	uint32 bgp_peer_set_sid = 3;
+  LsBgpPeerSegmentSID bgp_peer_node_sid = 1;
+  LsBgpPeerSegmentSID bgp_peer_adjacency_sid = 2;
+  LsBgpPeerSegmentSID bgp_peer_set_sid = 3;
```
fujita, line 849: This breaks the backward compatibility? LsAttributeBgpPeerSegment was introduced in v3.10, right?

fujita: I just like to understand your intension. I don't ask for workaround. Probably breaking this without the major number updated is fine.

yas-nyan: Thanks for the review. I'm sorry for not considering compatibility.

As you pointed out, in this PR,  we have changed the gRPC structure to support the flags parameter, which has just been implemented.
This change will enable, for example, the BGP-LS EPE route generation that we plan to do in the future. But on the other hand, compatibility will be affected if existing gRPC clients use this structure.

However, this change has no effect on non-BGP-LS applications. A problem occurs only when using this structure via gRPC. 
Since the gobgp command can now retrieve BGP-LS from this PR, the backward compatibility from the gobgp command in v3.10.0  is not broken.

Thus, in our opinion, it does not break any gRPC operations in existing releases.

If you have any comments, please let us know your opinion.

// FYI, the `message LsAttributePrefix.sr_prefix_sid` should also require the same modification.

fujita: Sure. Unlikely, this gRPC API change bothers anyone. Pushed, thanks!

> // FYI, the message LsAttributePrefix.sr_prefix_sid should also require the same modification.

like this?

https://github.com/fujita/gobgp/commit/b662da1f77e98cd08accf1946f68a9bdd0c59a71

yas-nyan: Thank you for your quick response!

> like this?
Sorry, not the same change, but it's a similar kind of change.
The TLVs of each SID in BgpPeeringSegment and the TLV of Prefix-SID are similar in structure, but the meanings of the flags are completely different. In other words, the dedicated SID structure for the Prefix-SID must be prepared.

According to [RFC9085](https://datatracker.ietf.org/doc/html/rfc9085#section-2.3.1), the meaning of flags is different for OSPFv2, OSPFv3, and IS-IS. I think it will be a little more difficult to handle...

fujita: Better to change the API soon but sounds considerable amount of work...

PR #2621, opened 2023-01-28
packet/bgp,mup: fix decode Interwork Segment Discovery route

higebu:
Current  code broke received data because it adds bytes to received data.


fujita:
Thanks!

PR #2623, opened 2023-01-30
Concat instead of fmt.Sprintf

bayrinat:
Using fmt.Sprintf adds additional time for getting string representation of each prefix and each attribute.

This PR replaces fmt.Sprintf with concatenation on some structs.

Synthetic benchmark shows x2 speed up:

```
type MockPrefix struct {
	Prefix net.IP
	Labels []uint32
	RD     string
}

var TestPrefix = &MockPrefix{
	Prefix: net.ParseIP("10.11.12.13"),
	Labels: []uint32{1234},
	RD:     "65534:1",
}

func Fmt(m *MockPrefix) string {
	masklen := 32 - uint8(8*(len(m.Labels)+len(m.RD))) // useless test-only mask
	return fmt.Sprintf("%s/%d", m.Prefix, masklen)
}

func Concat(m *MockPrefix) string {
	masklen := 32 - uint8(8*(len(m.Labels)+len(m.RD))) // useless test-only mask
	return m.Prefix.String() + "/" + strconv.FormatUint(uint64(masklen), 10)
}


func BenchmarkFmt(b *testing.B) {
	for i := 0; i < b.N; i++ {
		Fmt(TestPrefix)
	}
}

func BenchmarkConcat(b *testing.B) {
	for i := 0; i < b.N; i++ {
		Concat(TestPrefix)
	}
} 
```

```
BenchmarkFmt
BenchmarkFmt-12       	 3429468	       386.2 ns/op
BenchmarkConcat
BenchmarkConcat-12    	 7327934	       185.9 ns/op
```

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -2123,7 +2123,7 @@ func (l *LabeledIPAddrPrefix) String() string {
 	if isIPv4MappedIPv6(l.Prefix) {
 		prefix = "::ffff:" + prefix
 	}
-	masklen := int(l.Length)-l.Labels.Len()*8
+	masklen := int(l.Length) - l.Labels.Len()*8
 	return prefix + "/" + strconv.FormatUint(uint64(masklen), 10)
 }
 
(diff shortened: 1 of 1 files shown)


fujita:
Looks like good numbers. Pushed, thanks!

PR #2625, opened 2023-02-11
packet/bgp,mup: fix prefix format in Type 1 ST route

higebu:
https://www.ietf.org/archive/id/draft-mpmz-bess-mup-safi-01.html#section-3.1.3
The Prefix field must be encoded as described in RFC4271 4.3. UPDATE Message Format.
https://www.rfc-editor.org/rfc/rfc4271.html#section-4.3

The attached image is fixed UPDATE message.
image


fujita:
thanks!

PR #2627, opened 2023-02-21
pyang_plugins: add MUP SAFI to YANG definition

higebu:
I forgot to use the YANG definition to generate bgp_configs.go.


fujita:
thanks

PR #2628, opened 2023-02-21
Update x/net to 0.7.0

dawn-minion:
Updates x/net to the latest available version to address SAST warning for CVE-2022-41723: https://github.com/golang/go/issues/57855. `go mod tidy` additionally autoupdated `x/text` to match versions.


fujita:
thanks!

Issue #1507, opened 2017-11-22
prometheus support


fujita:
Is anyone interested in?

CoreTex:
Yes, but more interesting would be Zabbix.

benerudolph:
Maybe there is a common way to handle both monitoring via prometheus and other monitoring systems in a generic way?

7310510:
These metrics must be helpful for us.
- routes # of adj-in, accepted, and local RIB for route server mode
- per peer bgp packets counter
- buffering status

chippyleet:
Prometheus support would be greatly appreciated.

imcom:
+1 for Prometheus

tonidas:
+1 for Prometheus!

xiaopeng163:
+1 for Prometheus!

greenpau:
> Is anyone interested in?

@fujita, I am interested 😃 There are at least two ways to do that.

1. A separate project for `gobgp_exporter`, which queries `gobgp` API.
2. A separate go routine in `gobgp`

Any preference?

fujita:
Sorry about the delay, I've just started to investigate implementation options.

@greenpau why can't we pick the best of option 1 and 2? Having a separate project (exporter) forces users to manage another daemon. We can't run a go routine to execute the gRPC API and handle http requests from prometheus?

I've just create a branch to only send the received number in each peer:
https://github.com/fujita/gobgp/commit/090cca2967b3251d4656b06e13864de4c1f8ac4b

greenpau:
@fujita, it depends on the circumstances.

1. Sync vs Async polling...imagine you want to monitor a routing table contining 1000s routes. Do you want the polling to be sync or async?
2. Containerizing daemons. Once you containerized gobgp with prom built-in, the different go routine compete for the same cpu, because 1 cpu for a container.
3. Relid configs. How do you reload prom config?
4. etc.

Prom exporters are designed on per app basis. Compare node_exporter and backbox_exporter and you will get the gist.

fujita:
> 1. Sync vs Async polling.

I don't understand. I don't see how this is related with the exporter and exposing models.

> 2. Containerizing daemons.

1 cpu for a container? Why? You can configure what you like.

> 3. Relid configs. How do you reload prom config?

I had quick look at the official exporters, mysql, consul. Looks like they don't support the config reloading.

greenpau:
@fujita , see below.

> 1 cpu for a container? Why? You can configure what you like.

[...]

PR #2630, opened 2023-03-01
basic prometheus metrics

kikht:
This patch exposes some basic gobgp metrics in prometheus format. It reuses existing http endpoint that is used for pprof. For now only very basic metrics are implemented:
1. Per-neighbor bgp message counters. 
2. Number of recieved/accepted routes for each neighbor and each route family. 
3. GRPC api counters. 

Issue: #1507


fujita:
Why not using the gRPC API like this?
https://github.com/greenpau/gobgp_exporter

kikht:
> Why not using the gRPC API like this? https://github.com/greenpau/gobgp_exporter

Mostly for the grpc api call counters. While it is possible to collect them and extend api to expose them, it is much easier to do it directly in prometheus format. Also having metrics in gobgpd is slightly more convinient from deployment perspective - we have only single self-contained daemon binary.

kikht:
@fujita gentle ping. Is this approach feasible or we should pass all necessary metrics through gobgp_exporter?

pavel-odintsov:
Really nice patch, we added native support for Prometheus while ago and customers love it.

cmd/gobgpd/main.go
```diff
@@ -88,7 +91,14 @@ func main() {
 		runtime.GOMAXPROCS(opts.CPUs)
 	}
 
+	metricRegistry := prometheus.NewRegistry()
+
 	if !opts.PProfDisable {
```
mrueg, line 96: Should this line be changed? Otherwise if you disable pprof, you will also disable metrics, right?

fujita:
Monitoring software comes and goes. So I prefer the feature to be implemented in a way that it would be replaced with a different software easily. If people prefer native prometheus, it's fine by me, however, I don't want the prometheus support to play with GoBGP internal directly. Instead, I prefer it to play with the gRPC APIs.

kikht:
Sure, I can take less invasive approach and use only public api to get necessary information. Please, see the latest commits. Now all metrics stuff is moved to another package and uses public ListPeer method like grpc server does. This way we only need to add initialization stuff to gobgpd main and do not otherwise mess with its internals.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,136 @@
+package metrics
```
fujita, line 1: `internal/pkg/metrics` is more appropriate place?

fujita:
> Now all metrics stuff is moved to another package and uses public ListPeer method like grpc server does. This way we only need to add initialization stuff to gobgpd main and do not otherwise mess with its internals.

Thanks, looks good. But I have only one comment, why this is placed at `/pkg/metrics/` instead of `/internal/pkg/metrics/`?

fujita:
Thanks, pushed.
I wasn't sure about supporting prometheus natively last time but let's see.

kochkarev:
@fujita @kikht Hi! In my opinion it would be convenient to have this package in `/pkg/metrics/` instead of `/internal/pkg/metrics/`. GoBGP server can be imported and used while metrics package has to be copied :(

PR #2632, opened 2023-03-07
server: use SetsockoptTCPMD5Sig from golang.org/x/sys/unix

tklauser:
Use the TCPMD5Sig type and the corresponding SetsockoptTCPMD5Sig func added upstream in golang.org/x/sys v0.6.0


fujita:
Finally. Let's see how it works. Thanks!

Issue #2066, opened 2019-05-01
is there auth for grpc api calls? 


blackgold:
I am running a gobgp server and add a path using grcp api calls. There is there a way to add authentication?

robloxrob:
It looks like you can use TLS mutual auth between the client and the server
```
Usage:
  gobgpd [OPTIONS]

Application Options:
  -f, --config-file=      specifying a config file
  -t, --config-type=      specifying config type (toml, yaml, json) (default:
                          toml)
  -l, --log-level=        specifying log level
  -p, --log-plain         use plain format for logging (json by default)
  -s, --syslog=           use syslogd
      --syslog-facility=  specify syslog facility
      --disable-stdlog    disable standard logging
      --cpus=             specify the number of CPUs to be used
      --api-hosts=        specify the hosts that gobgpd listens on (default:
                          :50051)
  -r, --graceful-restart  flag restart-state in graceful-restart capability
  -d, --dry-run           check configuration
      --pprof-host=       specify the host that gobgpd listens on for pprof
                          (default: localhost:6060)
      --pprof-disable     disable pprof profiling
      --sdnotify          use sd_notify protocol
      --tls               enable TLS authentication for gRPC API
      --tls-cert-file=    The TLS cert file
      --tls-key-file=     The TLS key file
      --version           show version number

Help Options:
  -h, --help              Show this help message
```

Client:
```Usage:
  gobgp [flags]
  gobgp [command]

Available Commands:
  bmp
  global
  help        Help about any command
  monitor
  mrt
  neighbor
  policy
  rpki
  vrf

Flags:
      --bash-cmpl-file string   bash cmpl filename (default "gobgp-completion.bash")
  -d, --debug                   use debug
  -c, --gen-cmpl                generate completion file
  -h, --help                    help for gobgp
  -u, --host string             host (default "127.0.0.1")
  -j, --json                    use json format to output format
  -p, --port int                port (default 50051)
  -r, --pprof-port int          pprof port
  -q, --quiet                   use quiet
      --tls                     connection uses TLS if true, else plain TCP
      --tls-ca-file string      The file containing the CA root cert file```

fujita:
If it's not enough, please send a pull request.

jtkristoff:
This can remain closed, but in case anyone else comes looking, mutual TLS authentication is not currently supported.  Here is a hint at what would need to be done: https://github.com/grpc/grpc-go/issues/403

PR #2633, opened 2023-03-07
Implement TLS client authentication

icedream:
This PR addresses https://github.com/osrg/gobgp/issues/2066#issuecomment-640789765 by implementing authentication of API clients using TLS on top of the existing server authentication, effectively turning authentication mutual.

Implemented with this PR are:

- New command line flags `tls-client-cert-file` and `tls-client-key-file` for the `gobgp` binary. These flags are both respectively paths and in combination enable TLS client authentication against the gRPC API.
- New command line flag `tls-client-ca-file` for the `gobgpd` binary. If this path is set, it is loaded as a CA certificate against which clients are required to authenticate.

This would allow for gobgp's API to be used in a safe way from external clients. Use cases could include managing routes using a separately implemented software solution running on cloud infrastructure or other isolated resources where tunneling via a VPN often is technically not a workable solution.

### DIFF
--- a/cmd/gobgpd/main.go
+++ b/cmd/gobgpd/main.go
@@ -148,7 +148,7 @@ func main() {
 		// server cert/key
 		cert, err := tls.LoadX509KeyPair(opts.TLSCertFile, opts.TLSKeyFile)
 		if err != nil {
-			logger.Fatalf("Failed to load server certificate/keypair: %v", err)
+			logger.Fatalf("Failed to load server certificate/key pair: %v", err)
 		}
 		tlsConfig := &tls.Config{Certificates: []tls.Certificate{cert}}
 
--- a/cmd/gobgpd/main.go
+++ b/cmd/gobgpd/main.go
@@ -17,6 +17,8 @@
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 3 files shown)


icedream:
lintdoc complained about `keypair` in the certificate/key pair load error message, fixed it up by writing `key pair` instead.

Originally that was `Failed to generate credentials` which I thought was a bit confusing since nothing was actually getting generated. If there are better suggestions what to place as an error message there, let me know!

cmd/gobgpd/main.go
```diff
@@ -142,10 +145,27 @@ func main() {
 	maxSize := 256 << 20
 	grpcOpts := []grpc.ServerOption{grpc.MaxRecvMsgSize(maxSize), grpc.MaxSendMsgSize(maxSize)}
 	if opts.TLS {
-		creds, err := credentials.NewServerTLSFromFile(opts.TLSCertFile, opts.TLSKeyFile)
+		// server cert/key
+		cert, err := tls.LoadX509KeyPair(opts.TLSCertFile, opts.TLSKeyFile)
```
fujita, line 149: why `tls.LoadX509KeyPair` is better than `credentials.NewServerTLSFromFile`?

icedream: `credentials.NewServerTLSFromFile` does not give us access to the `*tls.Config` that is created as part the call for which we later possibly want to set the client CA certificate. This is at least the easiest way to code it in my knowledge.

`tls.LoadX509KeyPair` and then creating a `*tls.Config` is exactly the same as [`credentials.NewServerTLSFromFile` is coded to work](https://github.com/grpc/grpc-go/blob/f2fbb0e07ebf3dd46aa641ee89c9f17e8083eaf6/credentials/tls.go#L187-L192).

fujita:
Thanks!

PR #2635, opened 2023-03-08
server: disable tcpkeepalive

ChenYahui2019:
The default value for tcpkeepalive in the golang net package is 15 seconds, while the default value for BGP keepalive is 30 seconds. If the network is not good, the TCP connection is disconnected before BGP can send keepalive messages.

Therefore, BGP keepalive is effectively invalid.

The solution is that since we have BGP keepalive, we don't need to open tcp keepalive.

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -555,6 +555,7 @@ func (h *fsmHandler) connectLoop(ctx context.Context, wg *sync.WaitGroup) {
 			d := net.Dialer{
 				LocalAddr: laddr,
 				Timeout:   time.Duration(tick-1) * time.Second,
+				Keepalive: -1,
 				Control: func(network, address string, c syscall.RawConn) error {
 					return dialerControl(fsm.logger, network, address, c, ttl, ttlMin, password, bindInterface)
 				},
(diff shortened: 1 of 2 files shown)


fujita:
looks like frr supports tcp keepliave. bird doesn't use. Not sure what is appropriate. I tend to keep this alone because nobody hasn't complained about this until now.

ChenYahui2019:
> looks like frr supports tcp keepliave. bird doesn't use. Not sure what is appropriate. I tend to keep this alone because nobody hasn't complained about this until now.

Unfortunately, we've had this problem for over a year.

The network of our cloud IAAS is complex. When there is heavy traffic, the BGP connection may be disconnected.

We tried to adjust the `BGP keepalive` interval, but this didn't work.

Finally, we locate the root cause of the problem is the `TCP keepalive` problem. This is OK to verify in our environment.

fujita:
https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html

ChenYahui2019:
By default, the socket created by OS does not enable tcpkeepalive. 

By default, golang's net package enables the socket tcpkeepalive and sets the keepalive period to 15 seconds. Please have a look at source code of link below:
https://github.com/golang/go/blob/71c84d4b4149bebc2abcc495ef744e1a010a18e7/src/net/dial.go#L17

Therefore, modifying the tcpkeepalive configuration of the OS does not affect this bug.

fujita:
Rebased and merged, thanks.

PR #2638, opened 2023-03-22
mup: fix TEID processing in Type 1/2 ST Route

higebu:
# mup: fix TEID processing in Type 1/2 ST Route
Fix TEID processing in Type 2 ST Route:
- TEID can be 0-32 bits value as described in I-D.

Accept hex and IPv4 format for TEID in CLI and show TEID as IPv4 format in CLI.
- TEID is generally represented in hexadecimal format.
- IPv4 format is useful because Type 2 ST Route can be used in the longest match algorithm as described in I-D.

# test(rtr): remove deprecated function rand.Seed()
See https://pkg.go.dev/math/rand@go1.20#Seed


api/attribute.proto
```diff
@@ -394,7 +394,7 @@ message MUPType1SessionTransformedRoute {
   google.protobuf.Any rd = 1;
   uint32 prefix_length = 2 [deprecated = true];
   string prefix = 3;
-  uint32 teid = 4;
+  bytes teid = 4;
```
fujita, line 397: Is this change necessary? I prefer to not break the compatibility. Even if unlikely someone would complain about the API breakage due to this change.

higebu: Thank you for your review.
I understand the compatibility is important. I will revert this change and push unchanged code.

higebu: I reverted the API change.

fujita:
thanks a lot!

PR #2640, opened 2023-03-26
cli,mup: fix prefix and route type handling

higebu:
This PR enables to use `<prefix>` as not only route type but also route key.


fujita:
Thanks!

PR #2644, opened 2023-03-31
API: Preserve UUID in path table

dawn-minion:
Changes the AddPath API to add the UUID generated to the path object. Previously this UUID was only usable for deleting routes, and would be dropped when inserted into the table. That meant then if the same route was passed to a table event watcher, the UUID was set to nil, despite being made by the API.

With this change, the UUID is preserved and will be passed to callers of ListPath, or when passed to an event watcher. To implement this an additional UUID field has been added onto the table path object, and the AddPath, api2Path functions modified to preserve them. The Marshaller was also updated to return this value to end users as well.

Additionally the DeletePath call has been simplified when deleting paths by UUID. Since each path now stores the UUID they are checked directly rather than using a lookup map. This also means that the `uuidMap` has been removed.

### DIFF
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -24,6 +24,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/osrg/gobgp/v3/internal/pkg/config"
 	"github.com/osrg/gobgp/v3/pkg/log"
 	"github.com/osrg/gobgp/v3/pkg/packet/bgp"
@@ -141,6 +143,9 @@ type Path struct {
 	// For BGP Nexthop Tracking, this field shows if nexthop is invalidated by IGP.
 	IsNexthopInvalid bool
 	IsWithdraw       bool
+
+	// Only exists for paths added via AddPath API
+	Uuid *uuid.UUID
 }
 
 var localSource = &PeerInfo{}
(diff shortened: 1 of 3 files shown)


fujita:
The reason UUID isn't stored in Path struct is that we don't want to avoid increasing the size of Path struct.

dawn-minion:
Oh i see. Could I ask why though? 16 bytes increase seems fairly negligible, and with the removal of uuidMap it would end up being an equal amount of total memory usage.

fujita:
There are use-cases that GoBGP receives millions of routes from peers so 16 bytes in Path struct matters. On the other hand, I never heard that someone inserts millions local routes.

dawn-minion:
Thank you for the explanation. Wouldn’t the memory usage be the same though? 16 bytes in uuidMap today, or 16 bytes in Path struct. This change only moves it from one place to another, it’s not 16 extra bytes. Sorry if I’m misunderstanding of course :)

fujita:
> Wouldn’t the memory usage be the same though? 16 bytes in uuidMap today, or 16 bytes in Path struct.

I don't think so. Suppose that gobgpd establishes with a peer having millions routes. Then gobgpd has millions of Path objects but uuidMap is empty.

dawn-minion:
Wouldn’t it be the same here though? UUID is only set if the path is added via the gRPC AddPath, otherwise the pointer is nil. So if 1mil routes were received from a peer all uuids are nil. 

I chose a pointer type in the internal Path so the size change is only the size of a pointer. Or is the pointer size the concern here?

PR #2646, opened 2023-04-18
Use standard library context package

tklauser:
Use Go's standard library `context` package instead of the `golang.org/x/net/context` package. Package `context has been available in the standard library since Go 1.7 and `x/net/context.Context` is merely an alias for the standard library type.


fujita:
Thanks!

PR #2647, opened 2023-04-24
Updating support of L2VPN VPLS address family (AFI 25, SAFI 65)

MeneTelk0:
This is an update and extension of L2VPN VPLS address family support made in PR #2597

What's new:
1) Extension of API - adding new VPLSNLRI type and VPLSExtended extended-communities attribute. With marshaling/unmarshaling and tests
2) Update of VPLS NLRI parsing - changing type of Label Block Base from MPLSLabelStack to uint32 type. Generally speaking, Label Block Base in VPLS NLRI is not a MPLS label. This is a base, from which labels are constructed. So usage of MPLSLabelStack is incorrect as it is wrongly interpreted as stack of MPLS labels, which requires "Bottom of Stack" bit. This bit isn't present in Label Block Base and led to errors and wrong parsing before this update
3) Simple console handling of l2vpn-vpls address family. Now it's possible to read l2vpn-vpls from BGP table using gobgp.

What's NOT covered - adding of L2VPN VPLS routes via console


fujita:
Thanks!

PR #2651, opened 2023-05-03
Add MRT-specific MP_REACH_NLRI encoding

### LINKED ISSUE #2649: MRT decoding fails due to unsupported MP_REACH_NLRI format
When trying to parse some public MRT dumps, e.g. https://data.ris.ripe.net/rrc01/2023.04/bview.20230417.0000.gz, some RIB entries fail to parse with the following error `error: unknown route family. AFI: 8224, SAFI: 1`.

Specifically, this failure happens in the [MP_REACH_NLRI](https://github.com/osrg/gobgp/blob/master/pkg/packet/bgp/bgp.go#L10940) decoding. I believe this is because gobgp doesn't support the abbreviated form of this attribute described in https://www.rfc-editor.org/rfc/rfc6396.html:
> There is one exception to the encoding of BGP attributes for the BGP
   MP_REACH_NLRI attribute (BGP Type Code 14) [[RFC4760](https://www.rfc-editor.org/rfc/rfc4760)].  Since the AFI,
   SAFI, and NLRI information is already encoded in the RIB Entry Header
   or RIB_GENERIC Entry Header, only the Next Hop Address Length and
   Next Hop Address fields are included.  The Reserved field is omitted.
   The attribute length is also adjusted to reflect only the length of
   the Next Hop Address Length and Next Hop Address fields.

These MRT dumps are correctly decoded by bgpdump, which does support this abbreviated form (https://github.com/RIPE-NCC/bgpdump/blob/master/bgpdump_lib.c#L1499).

hfreeb:
This form, as defined in [RFC 6396 4.3.4](https://www.rfc-editor.org/rfc/rfc6396.html#section-4.3.4), is found in many public MRT dumps.

Closes #2649.

### Testing

This has been tested against public MRT dumps, e.g. https://data.ris.ripe.net/rrc18/2023.05/bview.20230503.0800.gz (CATNIX).

When parsing the dump without this patch, many messages error with `Failed to parse body: unknown route family. AFI: 4128, SAFI: 1`. With this patch all messages are successfully parsed.

Basic code used to test this: https://gist.github.com/hfreeb/ded114f29d81a50f854e187ef1a00063.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -32,8 +32,21 @@ import (
 )
 
 type MarshallingOption struct {
-	AddPath    map[RouteFamily]BGPAddPathMode
-	Attributes map[BGPAttrType]bool
+	AddPath        map[RouteFamily]BGPAddPathMode
+	Attributes     map[BGPAttrType]bool
+	ImplicitPrefix AddrPrefixInterface
+}
+
+// GetImplicitPrefix gets the implicit prefix associated with decoding/serialisation. This is used for
+// the MRT representation of MP_REACH_NLRI (see RFC 6396 4.3.4).
+func GetImplicitPrefix(options []*MarshallingOption) AddrPrefixInterface {
+	for _, opt := range options {
+		if opt != nil && opt.ImplicitPrefix != nil {
+			return opt.ImplicitPrefix
+		}
+	}
+
+	return nil
 }
 
 func IsAddPathEnabled(decode bool, f RouteFamily, options []*MarshallingOption) bool {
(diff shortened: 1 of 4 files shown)


fujita:
Thanks a lot!

`(e *RibEntry) DecodeFromBytes` is a public API so we can't break the API (adding an argument) without a major version 
updated. Other ways to handle this bug? `DecodeFromBytes` could serialize the prefix, concat, and call `(p *PathAttributeMpReachNLRI) DecodeFromBytes`?

fujita:
Thanks a lot!
Pushed with a hack to keep the compatibility.

PR #2658, opened 2023-05-22
Allow EbgpMultihop change in UpdatePeer

rastislavs:
This PR allows changing `EbgpMultihop` configuration via `UpdatePeer` RPC.

Before this change the `EbgpMultihop` changes are ignored in `UpdatePeer` and old configuration remains in place.

As TTL is configured via socket options at connect, we need to reconnect if it is updated.


fujita:
thanks!

PR #2659, opened 2023-05-22
Fix defaulting for UpdatePeer + make IdleHoldTimeAfterReset actually configurable

rastislavs:
This PR contains 2 somehow connected fixes:
 - makes `IdleHoldTimeAfterReset` actually configurable via `AddPeer` / `UpdatePeer` RPC (before this change, this value was ignored),
 - ensures that neighbor defaulting (`SetDefaultNeighborConfigValues`) is applied during  `UpdatePeer` processing even when `PeerGroup` is not configured for the neighbor.

These two issues together make it impossible to perform non-disruptive `UpdatePeer` RPC without changing the timer configuration - as the `IdleHoldTimeAfterReset` is defaulted to `30s` on `AddPeer`, but not defaulted on `UpdatePeer` when `PeerGroup` is not configured, and cannot be configured ATM.


fujita:
thanks a lot!

PR #2660, opened 2023-05-25
graceful-restart: expose peer graceful restart state information via API

harsimran-pabla:
Exposing PeerRestartTime and PeerRestarting info in config.NewPeerFromConfigStruct.


fujita:
thanks!

PR #2662, opened 2023-05-30
Dedicated RIB option

GeorgyKirichenko:
The option allows goBGP to distribute routes according to neighbor identifier. There is a dedicated routes table for each peer with this option enabled. Also PeerRib parameter is added to AddPath and DeletePath routines to point target router table out.


fujita:
Thanks for the patch, and sorry for the delay.

I am going to close this.

gobgp can already send different routes to different neighbors. Add the route to the global RIB and use an export policy on the neighbor to decide which neighbor gets it. No new table is needed.

A dedicated RIB is also almost the same as the route server client table. Both use the neighbor address as the table ID. The only difference is how the table is selected. gobgp would then have three RIB models, and every code path that looks up a table would have to handle all three. That cost is too high.

If an export policy does not work for your use case, please open an issue and describe the use case. I want to understand the problem before adding another RIB.

PR #2663, opened 2023-06-05
Fix setting of TTL on a nil connection in updateNeighbor

rastislavs:
This fixes a bug introduced by my previous PR: https://github.com/osrg/gobgp/pull/2661

Avoids attempt to set TTL on a nil connection in `updateNeighbor`.

This can happen e.g. if `UpdatePeer` comes very quickly after `AddPeer` and the connection is not yet created.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -3354,7 +3354,7 @@ func (s *BgpServer) updateNeighbor(c *config.Neighbor) (needsSoftResetIn bool, e
 		peer.fsm.pConf.TtlSecurity.Config = c.TtlSecurity.Config
 		setTTL = true
 	}
-	if setTTL {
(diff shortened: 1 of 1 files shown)


rastislavs:
Thanks for merging @fujita !

We would need this fix in this Cilium PR https://github.com/cilium/cilium/pull/25708 - would it be potentially possible to crate a patch release (e.g. `v3.15.1`) with it, or do we need to wait till the next minor version release?

fujita:
We've not made a patch release. Can you work around on your side until the next minor version?

rastislavs:
> We've not made a patch release. Can you work around on your side until the next minor version?

understood, we will stay on an unreleased version in the meantime

PR #2665, opened 2023-06-06
cmd/gobgp: Fix evpn parsing

costasd:
This PR:
1. Adds tests for all EVPN cli parsers using the full examples under `docs/source/evpn.md` for each.
2. Fixes a parsing issue for evpn a-d, not parsing `esi-label` with optional arguments correctly.
3. Fixes a parsing issue for evpn macadv parsing breaking when `default-gateway` is set


costasd:
hmm, seems like a transient failure between stopping the server and writing to it - I can't reproduce it locally. 

Looking at the code, it is definitely unrelated to this PR's changes (happens in pkg/bgp/server_test.go, which doesn't use anything out of `cmd/gobgp`).

fujita:
Thanks. Surely, unrelated to this pr.

Issue #796, opened 2016-03-24
How to bind API to localhost?


pavel-odintsov:
Hell, folks!

Bu default gobgp API listen on all interfaces (including external) and it's not expected behavior :)

Actually we have option to change default port but we haven't any way to specify custom host:
```bash
 /opt/dps/libraries/gobgp_1_4_0_git/gobgpd --help 2>&1|grep grpc
  -g, --grpc-port=        grpc port (default: 50051)
```

Finally, I have two suggestions:
1) Let's bind to 127.0.0.1 by default
2) Offer configurable option to specify custom host for listening 

Thanks!

pavel-odintsov:
Hi, folks!

Do you have some updates? :) That's real security issue for production use and I would be very glad if you could fix it.

fujita:
why not iptables? If it doesn't work for you, I guess that we could extend '-g' option or something.

pavel-odintsov:
Hello!

Will be fine to have option to tune bind interface. On servers with bunch
of interfaces I want to have multiple GoBGP instances for different
tasks (for example, first gobgp on 192.168.155.1:179, second on
10.10.10.12:179 etc). So we could not handle this configuration with
iptables.

Also iptables configuration adds another level of complexity and very error
prone for new users of gobgp. And open api interface offer very dangerous
level of acces for external malicious user.

 So I will be very glad to avoid it.



On Saturday, 9 April 2016, FUJITA Tomonori <notifications@github.com> wrote:

> why not iptables? If it doesn't work for you, I guess that we could extend
> '-g' option or something.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/osrg/gobgp/issues/796#issuecomment-207854321>
>


-- 
Sincerely yours, Pavel Odintsov

fujita:
https://github.com/osrg/gobgp/pull/828

Works for you?

pavel-odintsov:
Yep, looks awesome! Waiting in master! :)

On Sun, Apr 10, 2016 at 8:53 AM, FUJITA Tomonori <notifications@github.com>
wrote:

> #828 <https://github.com/osrg/gobgp/pull/828>
>
> Works for you?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/osrg/gobgp/issues/796#issuecomment-207988870>
>



-- 
Sincerely yours, Pavel Odintsov

[...]

PR #2669, opened 2023-06-30
Proposal to bind gRPC API to localhost by default

pavel-odintsov:
Hello!

I just pulled latest gRPC 3.15.0 and noticed that it binds gRPC to 0.0.0.0 / :: by default and exposes API to wild world of the Internet:
```
tcp6       0      0 :::50051                :::*                    LISTEN      3234326/./gobgpd    
```

As you can see I did not use any arguments or options for it. I'm pretty sure that majority of new deployments starts similar way and they will be vulnerable to attacks and will be insecure by default.

We discussed this issue while ago and it was partially addressed here by providing command line argument to specify API bind host (--api-hosts="::1"): https://github.com/osrg/gobgp/issues/796

In this PR I changed default wildcard bind to :: which listens on all available interfaces to more specific IPv6 localhost ::1 which makes GoBGP secure by default. 

Some may argue that not all servers have IPv6 connectivity but IPv6 localhost is available on all modern distributions. 

Thank you!

### DIFF
--- a/cmd/gobgp/root.go
+++ b/cmd/gobgp/root.go
@@ -86,7 +86,7 @@ func newRootCmd() *cobra.Command {
 		},
 	}
 
-	rootCmd.PersistentFlags().StringVarP(&globalOpts.Host, "host", "u", "127.0.0.1", "host")
+	rootCmd.PersistentFlags().StringVarP(&globalOpts.Host, "host", "u", "::1", "host")
 	rootCmd.PersistentFlags().IntVarP(&globalOpts.Port, "port", "p", 50051, "port")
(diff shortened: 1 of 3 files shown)


fujita:
This is a big change since GoBGP accepts any from day one. But I guess that the more secure default configuration is better. So I could merge this with the major version updated.
I guess that you need to fix some tests since they depend on this behavior.

robbat2:
Query: how to correctly specify binding to both ::1 and 127.0.0.1?

Systems where IPv6 is disabled via `net.ipv6.conf.all.disable_ipv6=1` won't have ::1

PR #2672, opened 2023-07-13
Enhance AsPath attribute String() format

rastislavs:
Enhances AsPath attribute `String()` format to be consistent with others, e.g.

`String()` of a Path attributes before the fix:
```
[{Origin: i} 65000 65001 {Nexthop: 172.0.0.1}]
```

after the fix:
```
[{Origin: i} {AsPath: 65000 65001} {Nexthop: 172.0.0.1}]
```


fujita:
Thanks!

PR #2674, opened 2023-07-15
Updated GitHub CI to Ubuntu 22.04

pavel-odintsov:
Hello!

I started playing with test suite on my Ubuntu 22.04 machine and noticed some issues caused by modern version of Python. So I decided to migrate your CI to latest Ubuntu. 

Looks like nose library does not like Python 3.10 and I added workaround to fix it: https://stackoverflow.com/questions/69515086/error-attributeerror-collections-has-no-attribute-callable-using-beautifu


fujita:
Thanks a lot!

PR #2679, opened 2023-07-31
Fix SRv6 prefix SID unmarshalling

rastislavs:
During SRv6 prefix SID unmarshalling, wrong Path Attribute length was applied (`4`B instead of `3`B). This was causing a 1 byte difference in PathAttribute length after marshalling & unmarshalling.

Screenshot from the `TestFullCyclePrefixSID` unit test shows the difference in PathAttribute Length (`38` vs `37`) before the fix (now asserted in the unit test): 

Screenshot from 2023-07-31 10-26-14


fujita:
pushed, thanks.

PR #2683, opened 2023-08-05
Fix panic on malformed MRT injection

rodrigopv:
This PR fixes a comparison that was not being triggered when a RIB had a not valid peerIndex, causing the client to panic:

`panic: runtime error: index out of range [1] with length 1`

The same malformed MRT file would throw the following error with BGPDump as a comparison:
`Assertion failed: (e->peer_index < entry->dump->table_dump_v2_peer_index_table->peer_count), function process_mrtd_table_dump_v2_ipv6_unicast, file bgpdump_lib.c, line 687`

With this patch gobgp now shows the following error before exiting (as expected):
`invalid peer index: 1 (PEER_INDEX_TABLE has only 1 peers)`

Another test, when the peer index table is present, but contains 0 peers, still works as intended:
`invalid peer index: 0 (PEER_INDEX_TABLE has only 0 peers)`

### DIFF
--- a/cmd/gobgp/mrt.go
+++ b/cmd/gobgp/mrt.go
@@ -111,7 +111,7 @@ func injectMrt() error {
 				paths := make([]*api.Path, 0, len(rib.Entries))
 
 				for _, e := range rib.Entries {
-					if len(peers) < int(e.PeerIndex) {
+					if len(peers) <= int(e.PeerIndex) {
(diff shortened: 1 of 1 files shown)


rodrigopv:
Just a heads up – it looks like the CI reports are indicating failures on both PRs, but they don't seem related to my changes, would it be possible to re-run the tests to confirm if they consistently fail?

`go test -count=1 -race -timeout 240s ./...` report no errors locally, and unit386 is also passing 🤔

Thanks!

fujita:
thanks!

PR #2685, opened 2023-08-08
Use filters for watch pre/post-update watch events

bayrinat:
This change extends WatchEvent API(backward compatible) by adding peer_address and peer_group fields.
It will help to reduce processing time by filtering out PRE- and POST- paths on server side.

It also adds filtering mechanism to watcher for using filter fields not only for the initial state, but for the all paths.


bayrinat:
@fujita hello, gentle ping. Is there any blocker to merge this filtering? :)

fujita:
Pushed, sorry about the delay.

PR #2703, opened 2023-09-06
Update configuration documentation for overriding peer-as checking during FSM

aredmon85:
Proposing update to the configuration documentation on bypassing autonomous system checking during FSM.


fujita:
thanks

PR #2707, opened 2023-09-19
[Fuzzing] fuzzing support for oss-fuzz

0x34d:
Adding  [go-fuzz](https://go.dev/security/fuzz)  fuzz testing support.


fujita:
Can you fix the lint error of the documentation?

fujita:
thanks

PR #2709, opened 2023-09-19
Allow testing multiple GoBGP instances on localhost.

wenovus:
Currently GoBGP does not accept UPDATE messages with nexthops pointing to a loopback address. This disallows multiple GoBGP instances from running at the same time on 127.0.0.0/8.

This PR proposes removing this constraint when the RouterID of the current GoBGP instance itself resides within the testing subnet of 127.0.0.0/8.


fujita:
> nexthops pointing to a loopback address

routerId?

wenovus:
> > nexthops pointing to a loopback address
> 
> routerId?

The code just checks loopback for now:
https://github.com/osrg/gobgp/blob/8bfcc66bcd3fd17c6b1ed9ebefe73d787cf161bc/pkg/packet/bgp/validate.go#L172

Are you suggesting simply to change the check for `p.Value.IsLoopback()` to checking whether it equals the RouterID? So essentially a less-restrictive check?

Or do you mean I have a bug here that I need to check whether next-hop == RouterID as well, even in this test environment?

pkg/server/fsm.go
```diff
@@ -1074,7 +1075,16 @@ func (h *fsmHandler) recvMessageWithError() (*fsmMsg, error) {
...
 				rfMap := h.fsm.rfMap
 				h.fsm.lock.RUnlock()
-				ok, err := bgp.ValidateUpdateMsg(body, rfMap, isEBGP, isConfed)
+
+				// Allow updates from loopback addresses if the GoBGP instance
+				// itself is assigned to 127.0.0.0/8, since this can happen when
+				// testing, where multiple GoBGP instances might be created within
+				// 127.0.0.0/8.
+				var allowLoopback bool
+				if routerIDAddr, err := netip.ParseAddr(h.fsm.gConf.Config.RouterId); err == nil && routerIDAddr.Is4() {
```
fujita, line 1084: What do you try to do here?

pkg/server/fsm.go
```diff
@@ -1074,7 +1075,16 @@ func (h *fsmHandler) recvMessageWithError() (*fsmMsg, error) {
 				h.fsm.lock.RLock()
 				rfMap := h.fsm.rfMap
 				h.fsm.lock.RUnlock()
-				ok, err := bgp.ValidateUpdateMsg(body, rfMap, isEBGP, isConfed)
+
+				// Allow updates from loopback addresses if the GoBGP instance
+				// itself is assigned to 127.0.0.0/8, since this can happen when
```
fujita, line 1080: What `assigned` mean?

wenovus: "assigned" means that the GoBGP server is expected set the next-hop attribute to one of these addresses, which generally should not happen.

The use case I have is I'm starting multiple GoBGP servers on the same host on different loopback addresses for lightweight testing. In order for this to work, I need the UPDATE messages to not be dropped when the next-hop is set to 127.0.0.0/8. I set this up by configuring Neighbor.Transport.Config.LocalAddress to the specific IP address within 127.0.0.0/8 for each test GoBGP instance.

In order to distinguish this test scenario from production scenarios, I thought the right way would be to check whether the RouterID falls within 127/8.

I wasn't able to figure out how GoBGP exactly sets the nexthop attribute, which should never be 127/8 in production use cases. Do you think there is a better way to check for this to make GoBGP more testable?

pkg/server/fsm.go
wenovus: see comment above

pkg/server/fsm.go
fujita: You can set RouterID to anything so I don't think that it's the right way to check if nexthops points to a loopback address or not.
Why not using `fsm.peerInfo.Address` or `fsm.peerInfo.LocalAddress`?

The following code updates the next hop attribute:
https://github.com/osrg/gobgp/blob/146b2b8f598d5f5c1c674fa451a2ccf0781220bf/internal/pkg/table/path.go#L197

fujita:
I think that Checking Address or LocalAddress is enough but I pushed anyway. Thanks!

PR #2712, opened 2023-09-24
server: improve ListPath's memory consumption with batched send

themiron:
With a lot of paths (hundreds of thousands) gobgp may oom or stuck in swapping.
Allow to specify max batch size via grpc and use 16K by default without preallocation on the first run, so it should not affect perfomance/allocations with small ammount of paths and safely improve it with any number greater 16K.

Supersedes https://github.com/osrg/gobgp/pull/2605 as less invasive, fixed error order, more memory friendly and production used.


themiron:
@fujita hello, kindly please take a look

api/gobgp.proto
```diff
@@ -265,6 +265,8 @@ message ListPathRequest {
   // enable_only_binary == true means that only nlri_binary and pattrs_binary
   // will be used instead of nlri and pattrs for each Path in ListPathResponse.
   bool enable_only_binary = 9;
+  // max ammount of paths to be allocated, default is 16K
+  int64 batch_size = 10;
```
fujita, line 269: Why not unsigned type like uint64?

themiron: just to avoid sign bit while conversion to int on x64, at the other hand with uint64 error return will be unnecessary

pkg/server/grpc_server.go
```diff
@@ -42,6 +42,8 @@ import (
 	"github.com/osrg/gobgp/v3/pkg/packet/bgp"
 )
 
+const defaultListPathBatchSize = 16384
```
fujita, line 45: Why 16k? By default, we should not change the behavior. u64 max?

themiron: Before 21093fbc8739d0a8e4f84ec6e52c98dad0894612 default was 1 and causes quite slow perfomance. After upgrade gobgp to v3 with n*100K of paths we begin to experience ooms, so it was empirically found 16K by default has still good performance (and allocation one by one is kept for small path ammounts) and have no dramatic memory impact as u64 max which means there'll be no limit by default.
What do you think about to treat here original default as 1 and 16K as feasible (and noticeable) improvement with optional possibility to specify u64 max if one knows what he's doing and has the free memory for such kind of load?

fujita: The default was 1 three years ago. We set the default to no limit for the last three years. I don't think that we need to change the default behavior.

themiron: Got it, that you for the suggestion. Please refer change to MaxInt64 by default, it still too high to be used as unlimited and allows to avoid casting len(l) to uint64

fujita:
thanks

themiron:
thank you for the merge

PR #2714, opened 2023-10-03
Fix data race in newWatchEventPeer

dawn-minion:
This method was attempting to read from peer.fsm before acquiring a read lock, leading to a data race as this struct is written by a different goroutine in parallel. Commit moves the call to RLock before the first read from the struct. 

Issue was flagged when the project was built with Golang's data race detector.


fujita:
How this is related with?
https://github.com/osrg/gobgp/pull/2715

We need both or either?

dawn-minion:
All 3 are separate data races, and all 3 are needed - I opened them as 3 pull requests in case changes were needed.

fujita:
thanks

PR #2715, opened 2023-10-03
Clone OriginInfo in Path Clone

dawn-minion:
OriginInfo assigned to Path objects is a pointer to the originating peer's info, which is populated by the FSM goroutine handling the peer's connection. This value is typically protected by the FSM lock, but in the event a path watch is running, it gets converted to a protobuf object in toPathAPI in a separate goroutine. This method reads from this same structure without any data synchronisation, leading to a data race.

To ensure this cannot happen, this commit adds a line to clone the field before it is passed to the event watcher, ensuring the data race cannot occur.

This issue was found when the project was built using Golang's data race detector.

### DIFF
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -328,11 +328,14 @@ func (path *Path) IsIBGP() bool {
 
 // create new PathAttributes
 func (path *Path) Clone(isWithdraw bool) *Path {
+	newOriginInfo := *path.info
+
 	return &Path{
 		parent:           path,
 		IsWithdraw:       isWithdraw,
 		IsNexthopInvalid: path.IsNexthopInvalid,
 		attrsHash:        path.attrsHash,
+		info:             &newOriginInfo,
 	}
 }
 


dawn-minion:
I wasn't quite sure what you folks would prefer here to resolve this, and this seemed the least invasive as it's a simple copy. Let me know if you want it changed of course.

fujita:
Clone() is used everywhere and this increases memory footprint. Can we copy peer's info only when Clone() is used for watcher?

fujita:
The race you found is gone. peer.peerInfo is now an atomic.Pointer[table.PeerInfo], and the PeerInfo is built fresh by NewPeerInfo() when the peer reaches ESTABLISHED. Nothing writes to a PeerInfo field in place any more, so the watcher goroutine reading path.GetSource() is safe.

Note that this patch would not have fixed it either. Clone() sets parent, and OriginInfo() returns root().info, so the info field set on a clone is never read. The copy was dead.

Looking into this did turn up a real race on originInfo.stale, which AdjRib.StaleAll writes through root() while the watcher goroutines read it with no lock. That field is now an atomic.Bool.

Closing.

PR #2716, opened 2023-10-03
Move SelectionOptions and UseMultiplePath to server-local vars

dawn-minion:
In destination.go there were two global variables used for controlling path and multi-path support for BGP servers, which were populated and modified by gRPC/public API calls. In scenarios where multiple BGP servers are running using the public API it would mean that the last server created would override the options for all running servers. It also lead to data races (Confirmed by using Go's data race detector) that could crash Go's runtime as the value would be overridden unexpectedly.

To resolve this, both options have been moved into the TableManager itself which takes a pointer option to both structs populated via the StartBgp API call. These are then passed to the internal methods as new function parameters, ensuring that every server has a unique copy of these structs.

All tests were updated to conform to the new API.

### DIFF
--- a/internal/pkg/table/destination.go
+++ b/internal/pkg/table/destination.go
@@ -28,9 +28,6 @@ import (
 	"github.com/osrg/gobgp/v3/pkg/packet/bgp"
 )
 
-var SelectionOptions oc.RouteSelectionOptionsConfig
-var UseMultiplePaths oc.UseMultiplePathsConfig
-
 type BestPathReason uint8
 
 const (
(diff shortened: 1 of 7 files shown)


dawn-minion:
This one was quite a refactor - I moved everything as function parameters where originally it read the global vars. None of the public API signatures are changed, so I am hoping it's still acceptable. Let me know if anything should be changed, thanks!

fujita:
Thanks for the patch, and sorry for the long delay.

The bug is real. The globals break any process that creates more than one BgpServer, and they also leak between tests in the table package.

The tree has changed a lot since 2023, so this patch no longer applies. I reimplemented the same fix on current master. TableManager now holds both option structs by value, set once in NewTableManager, and pkg/server reads them through the table manager.

Closing in favor of https://github.com/osrg/gobgp/commit/c7b863412693e443d75a73fc770feeb30771eeb8. Thanks for the report.

PR #2722, opened 2023-10-16
feat(config): autoreload on config changes

SkalaNetworks:
This merge request adds a new flag to gobgpd to enable the detection of changes in the configuration file, triggering an auto reload.

This is especially useful in containerized environments, like Kubernetes, where pushing a new configuration through ConfigMaps will overwrite the content of the configuration file in the container, but not trigger a reload because GoBGP doesn't know about the changes.

The new flag is '-a' or '--config-auto-reload'. I'm simply using Viper's ability to detect new configurations and emulating a SIGHUP signal.


mbanatt:
unsubscribe

fujita:
Why sending a signal is difficult for reloading?

SkalaNetworks:
@fujita not every deployment technology can send signals by itself on a config change, Kubernetes could theoretically send a signal through an HTTP webhook to gobgp (if it were to support it) to tell it to reload. We change the configuration of our bgp speakers quite often through GitOps, which means changing the config in the git results in a smooth automatic deployment with no manual operation.

For standard deployments (not through k8s), the alternative would be a script detecting the changes and pushing a signal to gobgp. With this PR, gobgp takes care of that by itself.

PR #2724, opened 2023-10-17
Fix systemd service reload command

ym:
gobgpd -r` is not the command to reload gobgpd, the correct way is to send a HUP signal.

https://github.com/osrg/gobgp/blob/419c50dfac578daa4d11256904d0dc182f1a9b22/cmd/gobgpd/main.go#L239


fujita:
thanks

PR #2727, opened 2023-10-24
Export zebra package and some symbols

wenovus:
Use case is to allow a different Go process/thread to exchange routes with GoBGP via Zebra API. For example, a custom Go-based RIB manager.


fujita:
You are actually implementing a custom RIB manager that can replace zebra?

wenovus:
> You are actually implementing a custom RIB manager that can replace zebra?

Currently we're only doing this in a limited way in order to integrate GoBGP as the BGP speaker for OpenConfig's reference implementation: https://github.com/openconfig/lemming/blob/main/sysrib/zapi.go, sysrib is the custom RIB manager for lemming.

fujita:
Understood. If the CIs succeed, I'll merge this.

pkg/packet/bgp/internal/generate.go
```diff
@@ -51,7 +51,7 @@ func generateSRBehaviour() {
 	buf.Write(constants)
 
 	if err := os.WriteFile(srBehaviourTargetFile, buf.Bytes(), 0o664); err != nil {
-		doPanic("failed to write file %s: %w", )
+		doPanic("failed to write file %s: %w", srBehaviourTargetFile, err)
```
fujita, line 54: What's this change? The generated code has a bug?

wenovus: Oh I just wanted to re-trigger CI, and this looked like a typo in existing code so I fixed it:
https://github.com/osrg/gobgp/blob/3ad3adb76aed1acbd949992fa52d8bef08f41124/pkg/packet/bgp/internal/generate.go#L57-L59

I can revert if you like.

fujita: Looks like this is generated code so we should not touch.

wenovus: ok reverted.

fujita:
@irino any comments?

irino:
@fujita OK, no proble,

Issue #2589, opened 2022-10-17
gobgpd with auth-password and dynamic neighbors


mikejoh:
Hi,

i have the following setup:
- A `kind` (Kubernetes in Docker) cluster running `cilium` with the [`--enable-bgp-control-plane`](https://docs.cilium.io/en/latest/network/bgp-control-plane/#cilium-bgp-control-plane) flag set to `true`.
- `gobgpd` running `v3.7.0` in a Docker container attached to the same Docker network as the `kind` cluster are running in.

The BGP implementation in `cilium` that are feature toggled using the flag above are using `gobgp` in code. The current implementation does not allow us to set an authentication password when configuring BGP peering in `cilium`.

What i'm trying to wrap my head around is why this `gobgpd` configuration in YAML:
```
global:
  config:
    as: 65130
    router-id: 172.18.0.6
    port: 179
    local-address-list:
      - "172.18.0.6"

peer-groups:
  transport:
    config:
      local-address: "172.18.0.6"
  config:
    peer-group-name: cilium-nodes
    peer-as: 65131
    local-as: 65130
    auth-password: "password"

dynamic-neighbors:
  config:
    prefix: 172.18.0.0/16
    peer-group: cilium-nodes
```
don't yield an error stating that the `cilium` peers doesn't use authentication and still let the dynamic neighbors connect and establish connections?

Have i missed something in how this is should work? Or maybe i have overlooked something vital in how BGP works?

I'm using [this](https://github.com/mikejoh/k8s-tech-demos/tree/main/cilium-bgp-demo) part of a repository created to test this.

Thanks!

mikejoh:
> We would also like to use md5 and dynamic neighbors. Based on https://github.com/osrg/gobgp/issues/2182 it seems that it is currently not supported. The information however seems not correct (anymore). Since Linux kernel 4.14+ TCP MD5 should be possible with prefixes (reference: https://github.com/FRRouting/frr/pull/4084) 

@simplysoft thanks for mentioning that issue, that discussion didn't really give any answers on this topic unfortunately. Interesting to read up on how FRR implemented this though!

fujita:
If I remember correctly, auth-password in peerGroup isn't implemented.

fujita:
@mikejoh Supporting md5 for dynamic neighbors needs the feature that @simplysoft mentioned.

mikejoh:
@fujita Thanks a bunch for the info!

jdewald:
[...]

PR #2739, opened 2023-11-07
Implementation of Dynamic Neighbor MD5 (Issue #2589)

jdewald:
Adds usage of the "prefix-based" TCP MD5 for dynamic neighbors. Non-dynamic neighbors will continue to use non-prefix based, which makes it more compatible with running on older kernels, as only 4.14+ includes the necessary support.

This change also includes tests of dynamic peers in general.

In the tests I tried to follow some of the naming/addressing that was in place already but might not have stuck with that as well as I could have.

Intended as fix for Issue #2589


fujita:
Pushed, thanks!

PR #2741, opened 2023-11-10
Add replace option to AddDefinedSet rpc

floatingstatic:
I have a use-case to replace a defined set (neighbor set) in-place. It seems `AddDefinedSet` currently does an append if one adds a set with the same name as something that already exists as seen here:
https://github.com/osrg/gobgp/blob/02fa6129afc715acdc9a0efa5ca83332aaaa6237/internal/pkg/table/policy.go#L3501-L3504

It seems slightly wasteful to add an entirely new RPC method (ie. `ReplaceDefinedSet` or `UpdateDefinedSet`). Instead I propose adding an optional `replace` parameter to `AddDefinedSetRequest` messages. If set to true we do an in-place replacement instead of appending. Given its a bool that defaults to false this should be backwards compatible with how most people are calling this RPC today if not explicitly set.


fujita:
The original design was handling policy stuff as immutable but we already have some exceptions. So I've merged this, thanks.

PR #2742, opened 2023-11-14
[Fuzzing] add cifuzz for PR fuzzing

0x34d:
Add [Cifuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/)  support for Pull request fuzzing.


0x34d:
update is on the way https://github.com/google/oss-fuzz/pull/11220

PR #2745, opened 2023-11-20
.github: Add a dependabot config

mrueg:
Adds a [Dependabot](https://docs.github.com/en/code-security/dependabot) configuration that checks for golang and github actions dependency updates


mrueg:
It will create a PR if there's a new version of a golang dependency released or there's a new version of a used github action published.

fujita:
I would like to update a package only when it has a security issue or such.

PR #2750, opened 2023-11-30
evpn: fix quadratic evpn mac-mobility handling

Tuetuopay:
## Description

This patch adds a special case in the destination hashmap for EVPN Type-2 routes, to index them by MAC address. This allows for direct access to the destination struct, instead of iterating over all destination and all paths. Such an access is performed for every EVPN Type-2 route received for EVPN MAC mobility handling.

In effect, this replaces an iteration over all known paths by a quick lookup to the MAC, leaving only an iteration to multiple paths to the same MAC (e.g. multihoming or through multiple VNIs).

The practical effect is a reasonable convergence time for large EVPN instances.

- before: 6m 7s
- after: 11s

The comparison was performed on a Xeon Silver 4209T, and an EVPN instance comprising of 13k EVPN type-2 routes. The time is measured by comparing the timestamp of the first and the last routes logged by the cli's monitor mode.

Given the extreme difference, no further work was done for a more accurate measurment.

## Discussion

~~From the short testing done, I have seen no adverse effect conflating routes with the same MAC address to the same `Destination` struct. However, this is the point I want to attract attention to as I am not familiar enough with the GoBGP internals.~~

After more testing, such a conflating resulted in incorrect `ListPath` behavior and most importantly route propagation, as only the best path between all of them to the same MAC were OK. It was then an issue as a MAC present in multiple VNIs led to only a single path.

The implementation was changed to use RD+MAC as the destination key. For the MAC mobility usecase, an index was added mapping a MAC address to the set of RDs that announced this MAC, allowing to build the set of destination keys for listing the paths related to a MAC.

Thanks!

### DIFF
--- a/internal/pkg/table/table.go
+++ b/internal/pkg/table/table.go
@@ -57,6 +57,10 @@ type Table struct {
 	routeFamily  bgp.RouteFamily
 	destinations map[string]*Destination
 	logger       log.Logger
+	// index of route distinguishers with paths to a specific MAC
+	// this is a map[MAC address]map[RD]struct{}
+	// this holds a map for a set of RD.
+	macIndex map[string]map[string]struct{}
 }
 
 func NewTable(logger log.Logger, rf bgp.RouteFamily, dsts ...*Destination) *Table {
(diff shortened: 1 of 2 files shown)


Tuetuopay:
Hi! The force push was to rebase atop the current `master` branch. However, I see that CI is failing, but it does not look to by my fault.

fujita:
I fixed the lint errors.

Tuetuopay:
Thanks, rebased. Oh, and noticed I forgot to remove the draft status.

fujita:
thanks

PR #2751, opened 2023-12-01
Make sure the TCP connection is closed when a peer is deleted

hardeker:
It could be that a peer gets deleted and added back while it was transitionning from active to open confirm. In that case, the TCP connection from the old version of the peer could still be up. This is a problem if the server is a listener only as the remote peer would consider the old TCP connection as being valid and it won't be able to connect until the TCP connection is eventually cleaned by the Golang GC.


fujita:
Thanks!

Issue #2574, opened 2022-09-17
BGP EVPN MPLS: unknown evpn subtype: 4


bv-oc:
Hi,

Trying to setup Gobgp as a route-reflector for a BGP EVPN MPLS setup, however the BGP UPDATE message from my network devices are being rejected by Gobgp with the following error:

{"Error":"unknown evpn subtype: 4","Key":"3.3.3.3","State":"BGP_FSM_ESTABLISHED","Topic":"Peer","level":"warning","msg":"the received Update message was treated as withdraw","time":"2022-09-17T15:48:35Z"}

I'm attaching a screenshot of the content of the BGP UPDATE message from my device.

![Screenshot from 2022-09-17 18-00-37](https://user-images.githubusercontent.com/113788419/190865828-56f2b921-7960-409e-8d3f-826f0b9555f0.png)

Looking at this doc: https://www.iana.org/assignments/bgp-extended-communities/bgp-extended-communities.xhtml#evpn
which says that the EVPN subtype 4, for an extended community when the value of the first octet (the "Type" field) is 0x06, should be EVPN Layer 2 Attributes, shouldn't Gobgp accept the UPDATE message ?

Thanks

PR #2752, opened 2023-12-05
packet: Add initial support for evpn subtype 4

Mrflatt:
Fixes: https://github.com/osrg/gobgp/issues/2574

Very initial support for [evpn subtype 4](https://www.rfc-editor.org/rfc/rfc8214.html#page-8), works when gobgp is used as route-reflector.
Currently we have tested these changes in our lab and now running in production prototype.


fujita:
Pushed, thanks!

PR #2753, opened 2023-12-15
mup: add Source Address to Type 1 ST Route

higebu:
This patch adds Source Address to Type 1 ST Route as defined in draft-mpmz-bess-mup-safi-03.
See https://datatracker.ietf.org/doc/html/draft-mpmz-bess-mup-safi-03#section-3.1.3


fujita:
Thanks!

Issue #2746, opened 2023-11-23
Why does bmp not support MultiPathList?


sephiroth81:
case *watchEventBestPath:
						info := &table.PeerInfo{
							Address: net.ParseIP("0.0.0.0").To4(),
							AS:      b.s.bgpConfig.Global.Config.As,
							ID:      net.ParseIP(b.s.bgpConfig.Global.Config.RouterId).To4(),
						}
						for _, p := range msg.PathList {
							u := table.CreateUpdateMsgFromPaths([]*table.Path{p})[0]
							if payload, err := u.Serialize(); err != nil {
								return false
							} else if err = write(bmpPeerRoute(bmp.BMP_PEER_TYPE_LOCAL_RIB, false, 0, true, info, p.GetTimestamp().Unix(), payload)); err != nil {
								return false
							}
						}
This one only takes the data from the PathList, but doesn't care about the MultiPathList. So multipath only receives the bmp message of the first rib, can I ask why it is designed this way?

sephiroth81:
Could you give me some advise, thank you.  @fujita

fujita:
multiple paths are converted into multiple BMP messages?

sephiroth81:
Yes.  The question is if only take the PathList's path, only the first path will have the bmp message, so the route information get from the bmp message won't be the same as the real one.  But if open multiple paths also sends bmp messages, I'm not sure if it will carry different pathID.

sephiroth81:
@fujita  I try to convert multiple paths to BMP messages, then I do get a bmp message for each route changes. But the new question is updates from multiple paths will only send the full paths,  couldn't know  which path is withdraw.  If I want to use BMP to track path changes, how can i do? please give me some advises, thank you.

fujita:
I'm not sure what you want to do. When gobgpd receives a BGP update message including multiple paths, each path separately is stored in a routing table. gobgpd doesn't store which paths were received in one message.

sephiroth81:
@fujita   Could you give some advises？ Or I have some misconceptions about the use of BMP?

PR #2758, opened 2024-01-15
multi path process

chaoliang-wei:
multi path process

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,5 +1,6 @@
 # Development Guide
 
+
 ## Building the development environment
 
 You need a working [Go environment](https://golang.org/doc/install) (1.16 or newer).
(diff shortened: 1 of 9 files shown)


imcom:
could you describe why would there be the addSet and delSet for multi-path? What's the intended use case? 
cc @fujita

imcom:
Hi @fujita  seeking your advice here. This PR seems to have breaking changes. However, I do like to know what's the best way to handle the Multi Path watch / chagnes
Thanks in advance

fujita:
There are some pull requests that break the APIs. We can merge them with the major version bumped.
I need the detailed description on this, not just "multi path process".

imcom:
Hi @fujita  I found an issue related to this PR. I assume this PR is trying to address the discussion here https://github.com/osrg/gobgp/issues/2746
Would you please take another look so we can get this PR moving forward
Discussion is more than welcomed. We would like to settle on the implementation and use case

fujita:
This breaks the public API (watch API), right? So we need to update the major version?

imcom:
@fujita Yes fujita, this breaks the `BMP` watcher hence a major version change is required.
Besides, do you think the tests should cover the `MPath` return values and probably there shall be specific test cases to demonstrate the use cases with `BMP`

PR #2760, opened 2024-01-22
Ability to kill the Serve method

Thekthekthek:
Currently, the bgp server is running endlessly. It's goten shut down when the main program stops. This PR have as a purpose to supply an optional channel in order to stop gracefully the Serve method whenever it's needed.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -381,8 +381,12 @@ func (s *BgpServer) passConnToPeer(conn *net.TCPConn) {
 
 const firstPeerCaseIndex = 3
 
-func (s *BgpServer) Serve() {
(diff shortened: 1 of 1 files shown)


fujita:
Already supported:
https://github.com/osrg/gobgp/blob/815683fa5e0d1e210e436bfa589588d9d2c19015/api/gobgp.proto#L36

saser:
Also note that the `BgpServer` type has a `Stop` method that will call the `StopBgp` API method for you: https://github.com/osrg/gobgp/blob/815683fa5e0d1e210e436bfa589588d9d2c19015/pkg/server/server.go#L219-L225. Just calling `Stop()` has been sufficient when I've used it.

Issue #2524, opened 2022-01-13
Graceful restart cause a slow initialization for a gobgp server


hhyasdf:
Hello,

Something wrong happened when I was using gobgpd to test the BGP GR procedure.

It seems like a slow initialization (about 300 seconds) happens if the gobgpd is configured to support ipv4 and ipv6 with `neighbors.afi-safis.mp-graceful-restart.config` while the remote bgp peer supports ipv4 only.

By the words in RFC 4724:
> Once the session between the Restarting Speaker and the Receiving Speaker is re-established, the Restarting Speaker will receive and process BGP messages from its peers. However, it MUST defer route selection for an address family until it either (a) receives the End-of-RIB marker from all its peers (excluding the ones with the “Restart State” bit set in the received capability and excluding the ones that do not advertise the graceful restart capability) or (b) the Selection_Deferral_Timer referred to below has expired. It is noted that prior to route selection, the speaker has no routes to advertise to its peers and no routes to update the forwarding state.

I think maybe gobgpd is just waiting for a defer time during the low initialization, because of the missing of ipv6 EoR message from remote peer. Is there any way to avoid this?

PR #2761, opened 2024-01-25
Only wait for GR End of RIB for the received address family

YutaroHayakawa:
Currently, graceful restart waits for the EoR message for all address families "enabled" for the peer, but it should only wait for "received" address families (the address families the peer is capable of handling).

Fixes: #2524


fujita:
Great, thanks!

PR #2768, opened 2024-02-16
Use net.IP.Equal in peer.filterPathFromSourcePeer

bepayne:
- Compares IPs using net.IP.Equal instead of using a string comparison in order to avoid unnecessary allocations.

- Adds peer.routerID to access IP, called by peer.RouterID.


fujita:
Pushed, thanks!

PR #2769, opened 2024-02-16
Adjust size of packing slice before allocating

bepayne:
- Changes table.(*packerV4).pack.func1 (split) to adjust the max parameter before using it to allocate slice size.  Previously the full max size was allocated then max was (possibly) truncated before further use.


fujita:
pushed, thanks!

PR #2772, opened 2024-02-27
logs: Guard frequently called logs behind if check

dawn-minion:
Moves a number of heavily called debug lines behind if checks to avoid needless allocations of Fields objects and stringification of fields. For cases where the server is not set to "debug" log level, these fields were allocated on the heap and then immediately discarded - as well a number of these were stringifying state / NLRIs regardless of log level.

In servers with significant amounts of routes and BGP peers, this lead to a large amount of wasted allocations - in our case looking at Go's memory profiler, 25% of all allocations were from these lines alone.


fujita:
Thanks for the profiling result, pushed. How do you use GoBGP?

PR #2778, opened 2024-03-04
Revert turning on onlyBinary flag

### LINKED ISSUE #2777: Breaking change in WatchEvent API introduced in v3.24.0
The `v3.24.0` introduced a breaking change in WatchEvent API by the commit https://github.com/osrg/gobgp/commit/aff055b44d0f036da26f5d82275d15e3cb23fee0

As of this commit, `Path` in generated events no longer has `Nlri` and `Pattrs` attributes set (they are always `nil`), and instead `NlriBinary` and `PattrsBinary` attributes are set.

Not sure what was the motivation behind [this change](https://github.com/osrg/gobgp/commit/aff055b44d0f036da26f5d82275d15e3cb23fee0#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996L4146-L4170), but as it changes the behavior of the API, it is breaking for the API consumers.

bayrinat:
Fix #2777

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -4179,11 +4179,11 @@ func (s *BgpServer) WatchEvent(ctx context.Context, r *api.WatchEventRequest, fn
 							l = append(l, p...)
 						}
 						for _, p := range l {
-							pl = append(pl, toPathApi(p, nil, true, false, false))
+							pl = append(pl, toPathApi(p, nil, false, false, false))
 						}
 					} else {
 						for _, p := range msg.PathList {
-							pl = append(pl, toPathApi(p, nil, true, false, false))
+							pl = append(pl, toPathApi(p, nil, false, false, false))
 						}
 					}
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -4179,11 +4179,11 @@ func (s *BgpServer) WatchEvent(ctx context.Context, r *api.WatchEventRequest, fn
 							l = append(l, p...)
 						}
 						for _, p := range l {
-							pl = append(pl, toPathApi(p, nil, true, false, false))
+							pl = append(pl, toPathApi(p, nil, false, false, false))
```
rastislavs, line 4182: should the same change be applied on the line 4164 (`watchEventUpdate` case), or is there any reason for keeping it there?

fujita:
@bayrinat Please create a new pull request for EoR event feature.

PR #2781, opened 2024-03-11
allow to pass add/del route -a vpnv46 identifier

sknat:
This patch adds an 'identifier %d' option to the route add / del CLI.
This is helpful for testing add-paths capabilities in vpv46 contexts.


fujita:
pushed, thanks!

PR #2783, opened 2024-03-15
prePolicyFilterpath: respect max routes config

maxime-peim:


maxime-peim:
Hi @fujita 
I was not able to run zapi test locally, will it be possible to run the CI please?

fujita:
I've just run the CI. btw, you can run the CI by forking the gobgp repository and creating a pull request for your repository.

maxime-peim:
Thanks!
This is what I have done, but the CI does not seem to be triggered when force-pushing

maxime-peim:
Sorry, it should be good now. However, I didn't find a way to re-run the CI from my side apart from closing and reopening the PR...

api/gobgp.proto
```diff
@@ -535,15 +535,16 @@ message Path {
...
-  bool stale = 13;
-  bool is_from_external = 14;
-  string neighbor_ip = 15;
-  bytes uuid = 16; // only paths installed by AddPath API have this
-  bool is_nexthop_invalid = 17;
-  uint32 identifier = 18;
-  uint32 local_identifier = 19;
-  bytes nlri_binary = 20;
-  repeated bytes pattrs_binary = 21;
+  bool send_max_filtered = 13;
```
bayrinat, line 538: This is an API breaking change

maxime-peim: Should I put the new field at the end of the structure?

maxime-peim: done

maxime-peim:
Hi @fujita @bayrinat,
Do you have comments to do on that PR, please?

maxime-peim:
Anyone?

fujita:
Thanks a lot for working the feature!
So each peer needs to keep filtered paths in memory. looks like it could consume lots of memory in some configurations. Any other way to implement that feature?

maxime-peim:
> Thanks a lot for working the feature! So each peer needs to keep filtered paths in memory. looks like it could consume lots of memory in some configurations. Any other way to implement that feature?

Hi Fujita,
Each peer keeps a key to the path that is stored in the local RIB, it does not have a copy of the path.
Maybe we could only store N filtered paths if the limit is set to N. So, we could replace all the advertised paths are withdrawn at once.

fujita:
Thanks a lot!

Issue #2765, opened 2024-02-02
How to use `ListPath` to get OpenConfig BGP RIB information?


wenovus:
**TLDR**: The API behaviour of `ListPath` is hard to understand. After experimentation I can't get the correct information to populate OpenConfig BGP RIB schema. Specifically, `ListPath` may be incorrectly populating path attributes for post-policy ADJ_IN.

**Ask**: Requesting guidance on how to use or improve this API to get pre/post-policy ADJ_IN/ADJ_OUT RIB information.

## Background
I'm using `ListPath` with `ADJ_IN` and `ADJ_OUT` to get the BGP RIB for the following types of RIBs:
* ADJ_IN routes before applying import policies. (adj-rib-in-pre)
* ADJ_IN routes after applying import policies. (adj-rib-in-post)
* ADJ_OUT routes before applying export policies. (adj-rib-out-pre)
* ADJ_OUT routes after applying export policies. (adj-rib-out-post)

The use case is to populate the BGP RIBs defined in [OpenConfig](https://openconfig.net/projects/models/paths/index.html).

## Problem

At first, the `EnableFiltered` parameter seems to do what I want: it applies the policies, and marks each route as either "filtered", (or rejected).

### Case/Attempt 1
So I did this:
* ADJ_IN routes before applying import policies. (adj-rib-in-pre)
  * Call `ListPath` with `EnableFiltered=false`, and use all routes
* ADJ_IN routes after applying import policies. (adj-rib-in-post)
  * Call `ListPath` with `EnableFiltered=true`, and use routes where `route.Filtered=false`
* ADJ_OUT routes before applying export policies. (adj-rib-out-pre)
  * Call `ListPath` with `EnableFiltered=false`, and use all routes
* ADJ_OUT routes after applying export policies. (adj-rib-out-post)
  * Call `ListPath` with `EnableFiltered=true`, and use routes where `route.Filtered=false`

#### Observations
* ACCEPT/REJECT actions were wrong for adj-rib-out-pre (i.e. EXPORT policy was being applied).
* community path attributes were wrong for all but adj-rib-in-pre.
* community path attributes seems to be reversed between adj-rib-out-post and adj-rib-out-pre -- they seem to have what each other's should have.

### Case/Attempt 2
So now I tried flipping `EnableFiltered` for adj-rib-out-*:
* ADJ_IN routes before applying import policies. (adj-rib-in-pre)
  * Call `ListPath` with `EnableFiltered=false`, and use all routes.
[...]

PR #2784, opened 2024-03-15
Fix `ListPath` for `ADJ_IN` and `EnableFiltered=true` and add tests.

wenovus:
Currently, even though `ApplyPolicy` is called for determining the accepted routes after apply policy, the new route with attribute modifications is not returned. This is problematic for gRPC API users.

Validation is added within `TestListPathEnableFiltered` for all four cases that were described in https://github.com/osrg/gobgp/issues/2765. This PR makes the behaviour conform to "Case/Attempt 2" described in the issue.


fujita:
I thought about a new parameter to avoid breaking the API but I guess that we call this issue a bug so pushed. Thanks a lot!

PR #2787, opened 2024-03-22
Add BatchSize parameter to WatchEvent

Tuetuopay:
When dealing with large BGP instances, the path list can be so large it makes a response message larger than the gRPC max message size.

This adds a BatchSize field to the request to send at most BatchSize paths in a single WatchEventResponse message, greatly limiting the size of each stream item.

For reference, this was hit on a production BGP-EVPN instances with over 80k paths.

    # gobgp mon --current global rib -a evpn
    rpc error: code = ResourceExhausted desc = grpc: received message larger than max (140281387 vs. 4194304)

Setting a batch size allows us to get the current full table without resorting to ListPaths calls, with all the synchronization issues that ensues. This is pretty important because the limit can be reached quite quickly: we observed single paths well over 15kB on the wire, and even a batch size of 4096 leads to oversized messages (already a far cry from the 80k of the default implementation).


fenollp (approved):

Tuetuopay:
Rebased.

fujita:
Pushed, thanks.

PR #2794, opened 2024-04-10
graceful restart: send initial paths list to all neighbors

hardeker:
Before this PR, when graceful restart was configured for a neighbor and when the restart flag was set for the restarting speaker, if the neighbor was not advertising the GR capability, the initial paths list was never sent by the restarting speaker to its neighbor

This is a problem when the server is configured with graceful restart for all its peers without knowing if the peer supports it. If some of the peers don't support it, they may never receive the routes from the restarting speeaker, leading to an inconsistent routing state.


hardeker:
hey @fujita, any comment on this PR?

fujita:
seems that makes sense. Hmm, what was the intention not to send initial paths to a peer that without GR cap advertised?

hardeker:
> seems that makes sense. Hmm, what was the intention not to send initial paths to a peer that without GR cap advertised?
I think the code was deisgned with the idea that the restart flag should not be set when talking to a peer that does not support the GR capability.

However, the issue is that you don't know the GR caps of the peer until you've received the BGP open message from the peer and you need to defined the restart flag when configuring the peer (before you've received the Open message).

If a goBGP server restarts and does not know which peers support the GR-cap or not, it should always set the "restart" flag for all peers. In the current implementation, if that is the case, only GR-capable peers will receive the initial RIB

hardeker:
Thinking about it, I could rather change the code from:
 ```
if !p.isGracefulRestartEnabled() {
	continue
}
```
to something like
```
if !p.isGracefulRestartEnabled() && !p.isLocalRestarting() {
	continue
}
```
This way, goBGP would still not send the routes to non-GR-capable for whcih the "restart" flag is not set

PR #2799, opened 2024-04-23
fix data race when getting capabilities from neighbor config

hardeker:
capabilitiesFromConfig was always taken under the read lock. However, when graceful restart is enabled for some families, this function writes to the neighbor config which creates a data race.


fujita:
Thanks!

PR #2803, opened 2024-04-30
graceful restart: send initial paths list to all neighbors

hardeker:
Before this PR, when graceful restart was configured for a neighbor and when the restart flag was set by the restarting speaker, if the neighbor was not advertising the GR capability, the initial paths list was never sent by the restarting speaker to its neighbor

This is a problem when the server is configured with graceful restart for all its peers without knowing if the peer supports it. If some of the peers don't support it, they may never receive the routes from the restarting speaker, leading to an inconsistent routing state.

@fujita, I closed previous PR #2794 and re-created this new one with some modification based on the discussion there


fujita:
Thanks a lot!

PR #2806, opened 2024-05-06
metrics: add metrics for BGP peer state and number of advertised routes

anitgandhi:


anitgandhi:
@fujita 👋 hello! We'd really love for this to be in the next release, if possible. When you have a moment, can you please approve the workflow run, and review the changes? Thanks in advance!

fujita:
How prometheus can handle the backward compatibility?

anitgandhi:
I don't think I'm modifying any of the existing metrics, so it shouldn't cause any backwards compatibility issues, I believe.

fujita:
pushed, thanks.

PR #2811, opened 2024-05-30
fix send-max limit in case of route update

maxime-peim:
- move send-max check post filtering
- don't increment route count on update and update test
- fix EOR `LabeledVPNIPAddr` nil RD


fujita:
ready for merge?

maxime-peim:
Yes :)

fujita:
Pushed, thanks!

PR #2812, opened 2024-06-04
fix flops value was not copied to api struct

kacpekwasny:
Flops count was only in the internal state, methods that allowed access to Peer values from the internal state omitted this value during creating the copy of the `state.Flops`.

One of the ways to observe this issue was by using command: `gobgp neighbor <IP address>`, and the Flops count was always zero.


fujita:
thanks!

PR #2815, opened 2024-06-12
Add ability to inject gzip/bzip2 compressed MRT files

vista-:
This small addition to the MRT command makes it easier to directly inject gzip and bzip2 compressed MRT files. RIPE RIS provides them in the former, RouteViews provides them in the latter format.

A quick benchmark shows that injecting a gzipped and bzip2ed MRT file (from RIPE route collector rrc18) takes almost the same time as injecting the uncompressed file:

```
$ time ./gobgp mrt inject global ~/latest-bview.gz

real	0m52.668s
user	0m51.392s
sys	0m5.315s

# Restart gobgpd

$ time ./gobgp mrt inject global ~/latest-bview.bz2

real	0m53.537s
user	0m48.198s
sys	0m3.960s

# Restart gobgpd

$ time ./gobgp mrt inject global ~/latest-bview

real	0m52.175s
user	0m52.177s
sys	0m8.418s
```

### DIFF
--- a/cmd/gobgp/mrt.go
+++ b/cmd/gobgp/mrt.go
@@ -16,6 +16,7 @@
 package main
 
 import (
+	"compress/bzip2"
 	"compress/gzip"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


fujita:
Looks like you need to update dictionary.txt

fujita:
pushed, thanks!

PR #2817, opened 2024-06-14
Support policy predicate `set next-hop peer-address`

wenovus:
Cisco support:
https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/routing/command/reference/rr40crs1book_chapter8.html#wp710515721:~:text=Sets%20the%20next%2Dhop%20to%20the%20IP%20address%20of%20the%20remote%20Border%20Gateway%20Protocol%20%28BGP%29%20peer


fujita:
Nice, thanks a lot!

PR #2824, opened 2024-07-25
ci: build with latest stable Go version

anitgandhi:
Previously, we were using `go-version-file: 'go.mod'`, which refers to the `go: 1.20` directive in `go.mod`. That means that `go1.20.14` was being used as the toolchain version. That's very out-of-date and no longer maintained. We have run into sporadic runtime fatal errors, and I believe that using a newer Go version will help with this.

I could specify `'1.22'` instead, if that's preferred. I chose `stable` just because when future major Go versions are released, they will be picked up automatically on the next build-and-release GitHub Actions run.

The version used to build a specific `gobgpd` binary can be found with `go version -m ./gobgpd`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
-          go-version-file: 'go.mod'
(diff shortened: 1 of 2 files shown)


fujita:
Thanks!

PR #2825, opened 2024-07-29
send-max: fix flag when path is not filtered anymore

hardeker:
when a path that was previously filtered becomes advertised because another path is withdrawn, the sendMaxFiltered flag needs to be unset. Otherwise, this path would never be deleted from the peer unless the peer is removed.


fujita:
Looks correct.
@maxime-peim any comments?

maxime-peim:
Indeed, that is correct, we need to unset the filtered boolean for a path ready to be advertised.

fujita:
Thanks guys!

PR #2843, opened 2024-10-07
Re-allow unspecified IPv6 local transport address

rastislavs:
Fixes regression introduced by 832b2d99c7de2a5461e3028ed56aec1c8b59ad14 which breaks peerings with unspecified IPv6 local transport address.

Fixes: https://github.com/osrg/gobgp/issues/2844


fujita:
Thanks a lot!

Issue #2857, opened 2024-11-18
Exposing internal prometheus metrics


mwarzech21:
What are the reasons for keeping the Prometheus metrics inside the internal package?
Would it be possible, to move them out from internal folder, in order to give access for gathering those metrics , while using only some specific gobgp parts and features?

kacpekwasny:
+1

fujita:
It means that it's more difficult to keep the compatibility because the internal might be changed in every release?

SkalaNetworks:
Usually you try not to change prometheus metrics (at least their name and structure) in a way that isn't backward compatible. Even for internal packages, this would is the rule.

The reason is simple, modifying the name or deleting labels on a metric breaks panels and Prometheus rules.
There's probably nothing wrong with this package going public.

mwarzech21:
Yes, thats what I was thinking. And unfortunately now, when those metrics are in the internal pkg, I cannot use them in my project.

Heres the PR for this issue I made a while ago:
https://github.com/osrg/gobgp/pull/2858

piotrsuchy:
This would be useful for me as well. Looking forward to get it merged!

PR #2858, opened 2024-11-20
Moved metrics from internal to public pkg

mwarzech21:
As said in issue #2857, there is currently no option to use those metrics, while using only some part of functionalities, since there is not any option to import the /internal package with metrics. 
All I did in this PR, was to moved those folders and update its import in the main.go. The metrics were presented as expected.

Issue:
https://github.com/osrg/gobgp/issues/2857


mwarzech21:
Hi @fujita , since this change would satisfy many users, as show in this PR or in the issue comments: https://github.com/osrg/gobgp/issues/2857 
Are we going to move forward with it and merge this metrics relocation?

fujita:
Merged, let's see what will happen.

PR #2859, opened 2024-11-20
grpc: fix ListPath batch_size parameter

Tuetuopay:
With the append being gated by the length check, for each batch sent, we were dropped the current path. Change to always append first, then check the length.

The effect on small batch sizes was dramatic, and best shown with a size of 1: exactly half of the routes were missing.

GoBGP with 500 routes:

    $ ./gobgp global rib -a evpn | wc -l
    501
    $ ./gobgp global rib -a evpn -b 1 | wc -l
    251
    $ ./gobgp global rib -a evpn -b 2 | wc -l
    335
    $ ./gobgp global rib -a evpn -b 500 | wc -l
    501

Note: the above example uses 500 routes, but the issue is present even with two routes.

This commit also adds a --batch-size to gobgp rib commands, to be able to use the tweak from the cli. It is very useful with a large number of routes to avoid GoBGP gobbling up gigabytes of memory.

### DIFF
--- a/cmd/gobgp/common.go
+++ b/cmd/gobgp/common.go
@@ -100,6 +100,7 @@ const (
 
 var subOpts struct {
 	AddressFamily string `short:"a" long:"address-family" description:"specifying an address family"`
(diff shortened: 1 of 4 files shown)


fujita:
Pushed, thanks!

PR #2863, opened 2024-12-25
Do not preserve MPReachNLRI attribute the way we receive it from wire 

myaut:
Right now `table.ProcessMessage()` rebuilds attribute list, but preserves
MPReachNLRI attribute as is. This causes trouble if GoBGP has a neighbor
that merges multiple NLRIs having same nexthop in a single BGP update
message and if GoBGP has a watcher. Thus watcher receives up to n^2
nlris (n paths each containing n NLRIs) causing enormous memory
consumption (since serialization/deserialization removes deduplication).

Since having all NLRIs in each path is useless in most cases, the
behavior is changed so only relevant NLRI is present in BGP path, but
for backward compatibility old behavior can be reenabled using config
option:
```yaml
    bgp-update-processing:
      config:
        preserve-nlris: true
```

In our tests memory consumption of such watcher had reduced from 19 Gb
to 2Gb (with around 300K routes in RIB).

### DIFF
--- a/internal/pkg/table/table_manager.go
+++ b/internal/pkg/table/table_manager.go
@@ -88,10 +88,7 @@ func ProcessMessage(m *bgp.BGPMessage, peerInfo *PeerInfo, timestamp time.Time) 
 		pathList = append(pathList, p)
 	}
 	if reach != nil {
-		reachAttrs := make([]bgp.PathAttributeInterface, len(attrs)+1)
-		copy(reachAttrs, attrs)
-		// we sort attributes when creating a bgp message from paths
-		reachAttrs[len(reachAttrs)-1] = reach
+		nexthop := reach.Nexthop.String()
 
 		for _, nlri := range reach.Value {
 			// when build path from reach
@@ -99,6 +96,11 @@ func ProcessMessage(m *bgp.BGPMessage, peerInfo *PeerInfo, timestamp time.Time) 
 			// this happens when a MP peer send update to gobgp
 			// However nlri is always populated because how we build the path
 			// path.info{nlri: nlri}
+			// Compute a new attribute array for each path with one NLRI to make serialization
+			// of path attrs faster
+			nlriAttr := bgp.NewPathAttributeMpReachNLRI(nexthop, []bgp.AddrPrefixInterface{nlri})
+			reachAttrs := makeAttributeList(attrs, nlriAttr)
+
 			p := NewPath(peerInfo, nlri, false, reachAttrs, timestamp, false)
 			p.SetHash(hash)
 			pathList = append(pathList, p)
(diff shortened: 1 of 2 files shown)


pkg/config/oc/bgp_configs.go
```diff
@@ -3175,9 +3215,6 @@ type NeighborConfig struct {
 	// original -> bgp:peer-group
 	// The peer-group with which this neighbor is associated.
 	PeerGroup string `mapstructure:"peer-group" json:"peer-group,omitempty"`
-	// original -> gobgp:send-software-version
-	// gobgp:send-software-version's original type is boolean.
-	SendSoftwareVersion bool `mapstructure:"send-software-version" json:"send-software-version,omitempty"`
```
myaut, line 3180: Honestly, I do not know why this didn't surface up earlier. 
`neighbor-interface` is defined at line 919 https://github.com/osrg/gobgp/blob/master/tools/pyang_plugins/gobgp.yang#L919
`send-software-version` at 1010: https://github.com/osrg/gobgp/blob/master/tools/pyang_plugins/gobgp.yang#L1010
It is logical that it goes after it. 

Changing order of field broke test, though, so I fixed `overrideConfig()` in a separate commit

fujita:
I can't recall why but we preserve that attribute for bmp/mrt?

myaut:
Per my understanding:
- Route Monitoring is affected, but it doesn't guarantee that it produces exact state nevertheless:
> Route monitoring messages are state-compressed.
https://datatracker.ietf.org/doc/html/rfc7854#section-4.6

- Route Mirroring is not affected as it uses raw `bgp.BGPMessage` which I do not modify - I only touch `table.Path`

Makefile
```diff
@@ -0,0 +1,50 @@
+SHELL := /bin/bash -o pipefail
```
fujita, line 1: What's this Makefile?

fujita:
I think that we can drop the config option because this doesn't break the gRPC API.

Makefile
myaut: Ughh, it's from our internal build system, it shouldn't be here :( 
Anyway, I've rewritten commit and removed config option

pkg/config/oc/default.go
```diff
@@ -516,13 +516,12 @@ func OverwriteNeighborConfigWithPeerGroup(c *Neighbor, pg *PeerGroup) error {
 
 func overwriteConfig(c, pg interface{}, tagPrefix string, v *viper.Viper) {
 	nValue := reflect.Indirect(reflect.ValueOf(c))
-	nType := reflect.Indirect(nValue).Type()
 	pgValue := reflect.Indirect(reflect.ValueOf(pg))
```
fujita, line 519: The changes to this file are related with this pull request?

myaut: Since I removed `bgp_configs.go` change, no, they're no longer are. I put it into separate PR - https://github.com/osrg/gobgp/pull/2873
Thanks.

fujita:
Pushed, thanks.

PR #2865, opened 2025-01-02
feature: add sentry integration to gobgpd

imcom:
We would like to use Sentry with `gobgpd` to capture unexpected crashes
We've seen occasionally `gobgpd` crashed with corrupted memory and leaving basically no trace to investigate

With Sentry integration, we hope to there is a chance to capture more information from the crash


fujita:
Thanks!

PR #2871, opened 2025-01-17
test: put scenario test dockerfile inside the source tree

dvalter:
Avoid using directories outside of the Gobgp source tree on the docker context. Otherwise docker uses a parent directory for its context, which causes slow builds and permission issues

### DIFF
--- a/test/lib/fabfile.py
+++ b/test/lib/fabfile.py
@@ -16,15 +16,10 @@ def make_gobgp_ctn(ctx, tag='gobgp',
 
     c = CmdBuffer()
     c << 'FROM {0}'.format(from_image)
-    c << 'COPY gobgp/gobgpd /go/bin/gobgpd'
-    c << 'COPY gobgp/gobgp /go/bin/gobgp'
+    c << 'COPY gobgpd /go/bin/gobgpd'
+    c << 'COPY gobgp /go/bin/gobgp'
 
-    rindex = local_gobgp_path.rindex('gobgp')
-    if rindex < 0:
-        raise Exception('{0} seems not gobgp dir'.format(local_gobgp_path))
-
-    workdir = local_gobgp_path[:rindex]
-    os.chdir(workdir)
+    os.chdir(local_gobgp_path)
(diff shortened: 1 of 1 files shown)


test/lib/fabfile.py
```diff
@@ -16,15 +16,10 @@ def make_gobgp_ctn(ctx, tag='gobgp',
 
     c = CmdBuffer()
     c << 'FROM {0}'.format(from_image)
-    c << 'COPY gobgp/gobgpd /go/bin/gobgpd'
-    c << 'COPY gobgp/gobgp /go/bin/gobgp'
+    c << 'COPY gobgpd /go/bin/gobgpd'
+    c << 'COPY gobgp /go/bin/gobgp'
```
fujita, line 21: Why the above change is necessary?

dvalter: The current dockerfile puts quite unnecessary restrictions on the folder structure:
1. Docker uses the whole `/clone/path/for/gobgp/..` for it's context. If this directory contains a lot of files (say Linux kernel git repo), docker build may take extremely long time
1. Gobgp sources *must* be in `gobgp` folder, something like `gobgp-custom` won't work due to hardcoded paths in the dockerfile

In this PR I try to fix both of there issues

fujita: The current code avoids to create Dockerfile inside gobgp directory. I think that when this code was implemented, it's a recommended way (for the performance, I can't recall). It's not the case anymore?

dvalter: In terms of docker build performance, you want to reduce docker [build context](https://docs.docker.com/build/concepts/context/) as much as possible. Putting it in the parent directory achieves the opposite

For me the following experiment shows docker reading a neighboring directory to gobgp source
```bash
mkdir ../not-a-gobgp-dir
cd ..
inotifywait -r -m  not-a-gobgp-dir &
cd -
fab -r ./test/lib make-gobgp-ctn
```

fujita:
Thanks, pushed.

PR #2873, opened 2025-02-03
Fix overwriteConfig if field order mismatches

myaut:
As mentioned in https://github.com/osrg/gobgp/pull/2863#discussion_r1897329474 - automatically generating bgp_configs.go moves SendSoftwareVersion to the bottom (expected), but that requires fix to `overwriteConfig` so peerGroup fields would be applied correctly if their order doesn't patch order in neighbor config.


fujita:
gobgp.yang should be fixed instead?

myaut:
My guess was that someone eventually will add a field to `gobgp.yang` that break the order again, so it's better to make code ready for it.

fujita:
Right, pushed. Thanks!

PR #2876, opened 2025-02-09
support for injecting the prefixes in MRT file only from a specified AS number

tonusoo:
I created a small patch that adds support for injecting prefixes from a MRT file only for a specified AS number. For example, suppose one needs to inject only the prefixes received from `AS 6667` by the [RIPE RIS](https://www.ripe.net/analyse/internet-measurements/routing-information-service-ris/) [RRC07 (Route Collector 7)](https://www.ris.ripe.net/peerlist/all.shtml). With the proposed patch, this could be accomplished with `gobgp mrt inject global --peer-asn 6667 latest-bview-rrc07.gz` where `latest-bview-rrc07.gz` is the compressed MRT file.


fujita:
pushed, thanks.

PR #2877, opened 2025-02-10
Show the peer-group name in the output of the 'gobgp neighbor' cli command

taitov:
Example:
```
$ gobgp neighbor
Peer          Group     AS      Up/Down State       |#Received  Accepted
127.0.0.1 reflector 123456 16d 11:56:29 Establ      |   195075    195075
127.0.0.2 reflector 123456 16d 11:55:46 Establ      |   195075    195075
127.0.0.3 reflector 123456 16d 11:56:04 Establ      |   195075    195075
127.0.3.1    group2 123456 16d 11:53:09 Establ      |     5166      5166
```

### DIFF
--- a/cmd/gobgp/neighbor.go
+++ b/cmd/gobgp/neighbor.go
@@ -162,6 +162,7 @@ func showNeighbors(vrf string) error {
 		return nil
 	}
 	maxaddrlen := 0
+	maxgrouplen := len("Group")
 	maxaslen := 2
 	maxtimelen := len("Up/Down")
 	timedelta := []string{}
@@ -189,6 +190,9 @@ func showNeighbors(vrf string) error {
 		} else if j := len(n.State.NeighborAddress); j > maxaddrlen {
 			maxaddrlen = j
 		}
+		if l := len(n.Conf.PeerGroup); l > maxgrouplen {
+			maxgrouplen = l
+		}
 		if l := len(getRemoteASN(n)); l > maxaslen {
 			maxaslen = l
 		}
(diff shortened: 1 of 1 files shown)


fujita:
How about the output of peers not in a peer group?

taitov:
> How about the output of peers not in a peer group?

```
$ ./gobgp neig
Peer           Group     AS      Up/Down State       |#Received  Accepted
127.0.2.1     group1  12345 18d 23:31:55 Establ      |       66        66
127.0.3.1     group2  12345 18d 23:31:53 Establ      |        1         1
127.0.0.1            212345        never Active      |        0         0
127.0.0.2  reflector 212345        never Active      |        0         0
127.0.0.3  reflector 212345        never Active      |        0         0
```
and without any groups:
```
$ ./gobgp neig
Peer      Group     AS Up/Down State       |#Received  Accepted
127.0.0.1       212345   never Active      |        0         0
127.0.0.2       212345   never Active      |        0         0
127.0.0.3       212345   never Active      |        0         0
```

fujita:
Assuming that PeerGroup is often unused, the output doesn’t look good. How about adding an option to enable showing this? If other BGP CLIs (like Cisco, Juniper, etc) works in this way by default, it should be ok though.

fujita:
I am closing this. The group is worth showing, but not in the summary table.

FRR has no group column in `show bgp summary`. Even the wide form only adds LocalAS (BGP_SHOW_SUMMARY_HEADER_ALL_WIDE in bgpd/bgp_vty.h). It prints "Member of peer-group X for session parameters" in `show bgp neighbors` instead, and it has a separate `show bgp peer-group` command. Cisco and Juniper split it the same way. In all of them the summary is one line per session, and the configuration goes in the per-neighbor output.

So I took a different approach and committed it. It prints the group in `gobgp neighbor <address>`, which did not show it at all. `gobgp neighbor` is left as it is. For scripts, `gobgp neighbor -j` already has conf.peer_group.

Thanks for the patch, and sorry for the long wait.

PR #2878, opened 2025-02-10
fsm: Speed-up initial connection time

rastislavs:
In the existing implementation, the initial connection from GoBGP side was not attempted until the timer with a random value from the interval `[5, 10)` seconds expired (derived from hardcoded `minConnectRetryInterval`, not configurable by any way).

This change makes the initial connection attempt immediate, and lowers the `minConnectRetryInterval` to `1s`. This helps to reduce convergence times, which is helpful especially in the datacenter deployment scenarios.

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	minConnectRetryInterval = 5
+	minConnectRetryInterval = 1
 )
 
 type fsmStateReasonType uint8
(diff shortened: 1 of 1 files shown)


fujita:
Supposed to be closed?

rastislavs:
Yes, closing this one in favor of https://github.com/osrg/gobgp/pull/2885

Could not make the tests pass after removing the sleep before the initial connect completely, not sure if it was a test issue or I missed something else. For now https://github.com/osrg/gobgp/pull/2885 is a good first step, we may look into removing the initial sleep later.

Issue #2866, opened 2025-01-08
Use of Any vs OneOf in API protobuff file


crankynetman:
Hello! 

When using the API protobuf and gRPC, I've noticed that the protobuf definition file makes heavy use of the `google.protobuf.Any` type and then there are comments in the protobuf file saying which type you need to pack to actually send across the wire. To me, this seems like a perfect use case for the `Oneof` message type, but full disclosure, I'm still very inexperienced with protobuf and auto-generated clients. Is this intended behavior and  a conscious choice, and if so, is there any information on why this path was chosen? One of the side-effects of using the `Any` type seems to be that you can't merge or copy objects into an existing object, at least with python generated code. For example, when creating a `Path` message to send in an `AddPathRequest` for the `AddPath` RPC, you have to generate the objects and pack them, then assemble the final `Path` message. If you don't pack and just use the object itself, you'll get an error. For example:

```python
path_object = gobgp_pb2.Path(
    nlri=attribute_pb2.IPAddressPrefix(
                prefix_len=32,
                prefix="10.0.0.6",
            ),
)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: Parameter to MergeFrom() must be instance of same class: expected google.protobuf.Any got apipb.IPAddressPrefix.
```

Instead, as shown in the demo documentation, you have to create an empty `Any` object and then pack that with the appropriate GoBGP object, like so:

```python
nlri = Any()
nlri.Pack(
    attribute_pb2.IPAddressPrefix(
                prefix_len=32,
                prefix="10.0.0.6",
            )
)
path_object = gobgp_pb2.Path(nlri=nlri)
```

This may not seem like a big deal (and it's not a huge issue) but it seems to ignore a lot of the reasons for using protobufs (specifically around type safety) and it makes it a bit of a pain to modify an object at any point when you're trying to construct payloads. 

To me, it seems like converting the API protobuf from the current state of: 

```protobuf
message Path {
  // One of the following defined in "api/attribute.proto":
  // - IPAddressPrefix
  // - LabeledIPAddressPrefix
  // - EncapsulationNLRI
  // - VPLSNLRI
  // - EVPNEthernetAutoDiscoveryRoute
  // - EVPNMACIPAdvertisementRoute
  // - EVPNInclusiveMulticastEthernetTagRoute
  // - EVPNEthernetSegmentRoute
  // - EVPNIPPrefixRoute
  // - EVPNIPMSIRoute
[...]

PR #2880, opened 2025-02-18
Rewrite the gRPC API to ditch google.protobuf.Any in favor of protobuf native oneof

Tuetuopay:
Hi,

This pull request introduces a major change to the GoBGP API: transitioning from `google.protobuf.Any` to protobuf's native `oneof` feature.

## Why

The BGP protocol uses a lot of list of various types, that reflects in the gRPC API. For example, path attributes can be any type of a specific subset. Currently, the API uses runtime dynamic types in the form of `Any`, allowing to hold any protobuf type. Which values are accepted is documented by a comment above the field, indicating which types are legal.

There are, however, a few issues with this:

- the API is not easily discoverable, as this is akin to a `void*` in C, and one has to rely on the comment for this
- those manual lists have to be maintained, and can fall out of sync between the actual implementation and the comment. In fact, while performing this rewrite, I found several such occurrences while adapting the various Marshal and Unmarshal helpers for Attribute, Nlri, etc.
- debugging tools cannot leverage Any, and easily drop down to the raw proto bytes contained by the Any (e.g. Wireshark)
- similarly, most API tools will stumble upon this (e.g. grpcurl)
- api producers and consumers must wrangle with the Any type, and manually (un)marshal to the correct type, instead of relying on protobuf's typing, especially in generated code.

In a nutshell, the main motivation is type safety and discoverability.

In addition to those usability aspects, a performance aspect is present as well, since it requires dynamic typing and parsing of the actual type URL. The GC will also be happier, as it means less allocations and garbage (it was, in fact, looking at pprof that initially triggered this initiative).

There also is an improvement on the wire usage, as all variants of a oneof share the same memory once encoded. Since that is always a sub-message for gobgp, it's always a pointer, so all those oneofs only eat up a pointer on the wire. A net win.

## How

The API structure did not change. All `Any`s were taken one by one, and a corresponding `oneof` message was written.

- oneofs specific to a single message type were nested near their usage (e.g. `TunnelEncapTLV`'s `tlvs` has its oneof message nested since it's only used in this message)
- oneofs that are used in multiple places have their own message (e.g. `NLRI`, `RouteDistinguisher`)
- oneofs that are large and significant enough have also their own non-nested message (e.g. `ExtendedCommunity` that does even have its own `.proto` file)

Once those were identified, the various `Marshal`/`Unmarshal` helpers were used to infer the correct set of accepted values in the oneof (many documentation comments were out of sync, especially for stuff used at several places).

Due to include cycles, the big `attribute.proto` file had to be broken down in smaller files, like `extcom.proto`, `common.proto`, `nlri.proto`, etc.

## Result

In the end, the API is much more self-describing, avoids wrangling with Any, and (subjectively) better overall. A simple look at some of the main oneofs ought to be enough: NLRI, Attribute, ExtendedCommunity just to name those three.

For the performance aspect, I measured notable improvements on the gRPC API. The tests were performed using EVPN type-2 routes and a small Rust program for route injection.

| Load                           | Anyful  | Anyless | Improvement             |
|--------------------------------|---------|---------|-------------------------|
| Insertion 100k, unary backend  | 4352/s  | 4459/s  | +2.5% paths/s           |
| Insertion 100k, stream backend | 49619/s | 59925/s | +21%  paths/s           |
| Insertion 1M, stream backend   | 43989/s | 52693/s | +20%  paths/s           |
|                                | 44.63s  | 36.67s  | -18%  cpu time (user)   |
|                                | 3.35s   | 2.79s   | -17%  cpu time (system) |
| Listing 1M, batch 128          | 40.82s  | 33.91s  | -17%  wall time     |
| Memory after insertion         | 2.74GB  | 2.83GB  | -3%   memory            |

Note: "unary backend" means calling the "AddPath" gRPC endpoint, while "stream backend" means using a single "AddPathStream" gRPC call and streaming all routes. I include both for completeness: most users will use the unary backend and won't see a meaningful improvement, while heavy users (like myself) will use the stream backend and will see meaningful improvements.

CPU time was measured by sampling `/proc/pid/stat` before and after applying the workload. It is included as an indicator of how much less work gobgp does, and that the increased performance does not come at a CPU cost. We do less work.

The tests were performed on a Scaleway POP2-4C-16G instance running Ubuntu 24.04.

## Downsides

A few downsides are present with this approach:

- it is not possible to have a different set of variants depending on where the type is used. previously, there could be some variants (e.g. of nlri) illegal in some place but legal in any other. while I personally don't see where this could be useful, and think it would be bad api design given the goal of this api, it is up to the maintainer to weigh in on this.
  (it could be documented via comments that a variant is illegal in some situation, which is IMHO better, as to explicitly document exceptions)
- some languages may see _increased_ memory usage for deserialized messages. AFAICT only Rust is affected: it has tagged enums, and a oneof is represented as a rust enum. Those will be sized to fit their largest variants, which means that any other variant pays the price. (e.g. with this new oneofs, the Rust generated Attribute is more than 500 bytes).
  IMHO this is a language / tooling issue, as a simple patch in protoc-gen-prost would allow to box enum variants, and reduce the enum size to a pointer.
- the other class of languages _not_ having tagged enums resort to wrapping structs with interfaces (e.g. Go), which can make the code a bit more verbose to write (see the attribute utils tests). I think this is a small price to pay for an actually discoverable and safe API.
- this is a breaking API change

However, I think the benefits far outweigh the drawbacks.

## Other words

I don't expect this pull request to be merged straight away, or as-is. It definitely could, however. But I don't come to get this merged ASAP, and I'm open to discussion. This is a big changed, and we already discussed on slack about this. I hope this clear most of the doubts that could cloud this initiative.

On a side note, I did not fully nuke Anys from the proto. Deep down SRv6, there is a pair of Anys left, which I didn't touch, because I'm far from being knowledgeable enough to touch this (though I'll read up on it and end up doing it, it's a tiny leaf struct anyways).

Thank you for considering this :)


Tuetuopay:
Update: I did the SRv6 TLVs, so now the API truly is Any-free.

fujita:
Does anyone have any opinions on this?

checorone:
Hi!

+1 on this, currently I'm facing problems of huge memory allocations, OOM and GC pressure while listing paths from gobgp.
Guess the whole problem could be solved by not using anypb marshalling logic.
image

@Tuetuopay Thank you for your work

Tuetuopay:
@checorone thanks! I indeed know this flamegraph all too well :grin:

Just a warning, this is not a silver bullet, but it will indeed help. Another thing that may interest you, is the `batch_size` parameter in the ListPath RPC, that will avoid blowing up the GoBGP server memory. I fixed its implementation recently, and using it in prod relieved my memory :D

fenollp:
Note: this fixes https://github.com/osrg/gobgp/issues/2866

@checorone Nice TBs! What's your $DAYJOB?

Tuetuopay:
@checorone I just remembered, but the `batch_size` parameter in the `ListPaths` RPC will help you _a log_ regarding memory pressure. By default, GoBGP will copy all known paths to an in-memory slice, then send all of them. This setting will limit the slice size, and flush it to the gRPC stream when reaching the batch size. Basically, it will cap memory used. Since we've deployed this setting in prod, we have way way less memory issues.

Issue #2448, opened 2021-08-10
Gobgp support GracefulRestart without delete route path


forsakening:
Hi all:
I use Gobgpd as route reflector, and enable GR. But when peer down, routes received from peer were all delete.
Followd Log:
 gobgpd[225828]: time="2021-08-10T20:53:56+08:00" level=warning msg="sent notification" Code=4 Data="[]" Key=10.35.98.21 State=BGP_FSM_ESTABLISHED Subcode=0 Topic=Peer
 gobgpd[225828]: time="2021-08-10T20:53:56+08:00" level=info msg="peer graceful restart" Key=10.35.98.21 State=BGP_FSM_ESTABLISHED Topic=Peer
 gobgpd[225828]: time="2021-08-10T20:53:56+08:00" level=info msg="Peer Down" Key=10.35.98.21 Reason=graceful-restart State=BGP_FSM_ESTABLISHED Topic=Peer
 gobgpd[225828]: time="2021-08-10T20:53:56+08:00" level=debug msg="state changed" Key=10.35.98.21 Topic=Peer new=BGP_FSM_IDLE old=BGP_FSM_ESTABLISHED reason=graceful-restart
 gobgpd[225828]: time="2021-08-10T20:53:56+08:00" level=debug msg="Removing withdrawals" Key="[type:Prefix][rd:10.35.98.21:1000][etag:0][prefix:11.1.1.2/32]" Topic=Table    -----  This path is deleted which not as expect.

===== when GR timeout, about 5min later, logs as :
gobgpd[225828]: time="2021-08-10T20:58:56+08:00" level=warning msg="graceful restart timer expired" Key=10.35.98.21 State=BGP_FSM_ACTIVE Topic=Peer

so, I wondered if Gobgpd not support this mode ?

My gobgpd conf:
global:
 config:
  as: 60001
  router-id: 10.35.98.21
 graceful-restart:
    config:
      enabled: true
      restart-time: 300
neighbors:
 - config:
    neighbor-address: 10.50.130.199
    peer-as: 60001
   graceful-restart:
      config:
        enabled: true
        restart-time: 300
   afi-safis:
    - config:
        afi-safi-name: l2vpn-evpn
    - config:
        afi-safi-name: l3vpn-ipv4-unicast
    - config:
        afi-safi-name: l3vpn-ipv4-flowspec
 - config:
    neighbor-address: 10.50.130.100
    peer-as: 60001
   graceful-restart:
      config:
        enabled: true
[...]

PR #2882, opened 2025-02-19
fix: prevent AS loop

maxime-peim:
Revert https://github.com/osrg/gobgp/issues/2488
We should not allow a path with an AS loop to be advertised.
iBGP peers should not add locally originated routes with there own AS in their path.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -691,7 +691,7 @@ func filterpath(peer *peer, path, old *table.Path) *table.Path {
 		return nil
 	}
 
-	if !peer.isRouteServerClient() && isASLoop(peer, path) && !path.IsLocal() {
+	if !peer.isRouteServerClient() && isASLoop(peer, path) {
 		return nil
 	}
 	return path


fujita:
I agree that this is the expected behavior. Hmm, I can't recall why I merged #2488

fujita:
I thought that It would be better to allow whatever a user wants to do (even it's a mistake from the BGP perspective). No?

maxime-peim:
Well, I don't think this is a good idea to break from BGP behavior to allow mistakes made by the user.

fujita:
@floatingstatic Opinions?

floatingstatic:
@fujita This will break an [application I built around gobgp](https://www.fastly.com/blog/turning-a-fast-network-into-a-smart-network-with-autopilot) where I need to advertise some routes with an origin ASN that matches the ASN of an iBGP peer in some but not all cases. Can I suggest we make a configuration option to enable the behavior I need instead of reverting? We can make this disabled by default which should suffice for @maxime-peim 's needs. I'm happy to do the work for this and can open a PR later today to implement this if that works for you.

floatingstatic:
@maxime-peim Sorry for confusion due to my wording. For my use case the paths that are filtered without this check are routes i originate so I used the term "origin asn" but the as-path loop logic doesn't actually care where the asn is in the path generally. To your question about why I can't use an export policy is because the as-path loop check happens before any policy is applied (see `prePolicyFilterpath()` function in the `BgpServer` type).

I have a branch I'm testing right now that just adds a per-neighbor/peer option to allow me to get the behavior I need for my traffic engineering application to function and by default it would revert to behavior prior to #2448 so that should make all of us happy :)

mpeim:
The check `isASLoop` that you were trying to avoid is only comparing AS from the AS path attribute.
Concerning the export policy, indeed the `isASLoop` check is perform in `prePolicyFilterpath` which is called before applying export policies (in `BgpServer.filterpath`). So you could have an export policy that add your ASN in the AS path without adding it beforehand in the path.

floatingstatic:
@maxime-peim yep got it. i understand how the as-path check works. regarding the export policy thing, without explaining in great detail how my application works, no this is not possible as we are mirroring routes that already have as-path set.
PR is coming. I don't feel we need to debate this further.

mpeim:
@floatingstatic perfect thanks!

floatingstatic:
@fujita I'd like to propose we go with this instead of this revert PR:

https://github.com/osrg/gobgp/pull/2883

Open to comments about the config option name or what struct you feel this should go under. I tested pulling my fork into my route injector application and setting this configuration option allows me to get what I want and not setting it (default) reverts to behavior @maxime-peim is expecting.

PR #2883, opened 2025-02-24
Add neighbor config option to bypass as-path loop detection on local (static) routes

floatingstatic:
Addresses issue being raised under #2882 without breaking behavior for users that wish to bypass as-path loop detection when advertising static routes to ibgp peers with the peer asn in the path.


pkg/server/server.go
```diff
@@ -691,8 +691,15 @@ func filterpath(peer *peer, path, old *table.Path) *table.Path {
...
-		return nil
+	if !peer.isRouteServerClient() && isASLoop(peer, path) {
+		// Do not filter local (static) routes with as-path loop
+		// if configured to bypass these checks in the peer
+		// as-path options config.
+		if path.IsLocal() && !peer.allowAsPathLoopLocal() {
+			return nil
+		} else if !path.IsLocal() {
+			return nil
+		}
```
mpeim, line 702: Could we compute this before checking for the AS loop? this would prevent an unecessary check on the AS path attributes which could be of unknown length.

mpeim:
One small comments, but overall thanks!

pkg/server/server.go
floatingstatic: I'm not sure that would work well. If you are suggesting an early exit or something if the path is local and we allow this option we may end up bypassing additional conditions that may be added in the future.
Personally I feel its cleaner to keep this in the section of code where we check for the loop, especially given this option is only relevant to loop checking. I defer to @fujita for input on this.

fujita: I don't have a strong preference, but checking in the order of BGP's general rules followed by special cases is easy to understand and likely won't have a significant impact on performance, so the current code is fine as it is by me.

fujita:
Thanks!

PR #2885, opened 2025-02-27
fsm: Lower minimal ConnectRetry interval to 1s

rastislavs:
Lower the `minConnectRetryInterval from` `5s` to `1s`.

As this interval also defines the wait time before the initial connection attempt, lowering it helps in situations where fast re(connection) times are required.

The change also adds more variation for the random jitter applied to the ConnectRetry timer.

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -37,7 +37,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


fujita:
It makes sense, given that the use cases for GOBGP within the data center are increasing.
Pushed, thanks!

PR #2887, opened 2025-02-28
Update gobgp.yang for AllowAsPathLoopLocal config option

fujita:
Update gobgp.yang for AllowAsPathLoopLocal config option and generate bgp_configs.go.

Seems that the commit 16ea4f9c592a ("Add neighbor config option to bypass as-path loop detection on local (static) routes") updates bgp_configs.go by hand.


fujita:
@floatingstatic

PR #2889, opened 2025-02-28
Update genproto.sh protoc and protoc-gen-go versions 

floatingstatic:
In #2883 we regenerated protobufs using latest stable protoc (`25.6`). This change updates `tools/grpc/genproto.sh` to reflect this. In addition we update to the latest `protoc-gen-go` version `v1.36.5`.

Updating `protoc-gen-go-grpc` requires refactoring some gRPC related code so holding of on changing the pinned version for that at this point.


fujita:
Thanks!

PR #2890, opened 2025-03-05
pkg/server: Do not count EOR paths in ListPeer advertised routes counter

Matus-p:
EOR paths are present in path list when graceful restart is enabled, they should not be counter in advertised routes count


fujita:
Can we modify the interface of getBetFromLocal to avoid looping through the list again?

fujita:
Fixed by https://github.com/osrg/gobgp/commit/78dbfea2cb46200ec4bc6645093d6c701cb8e22b

Matus-p:
> Fixed by [78dbfea](https://github.com/osrg/gobgp/commit/78dbfea2cb46200ec4bc6645093d6c701cb8e22b)

thank you very much

PR #2891, opened 2025-03-05
pkg/metrics: Fix local router_id label in bgp_peer_local_asn metric

Matus-p:
use routerId from bgp server instead of transport local address


fujita:
Does anyone have any opinions? I'm not familiar with the prometheus support.

Matus-p:
using p.Transport.GetLocalAddress() is wrong, because for example for ipv6 you will get 128-bit address, while router ID is only 32bit.

fujita:
pushed, thanks.

Issue #2897, opened 2025-03-17
Error occurs when executing the example command of Opaque


shell720:
In [Using GoBGP](https://github.com/osrg/gobgp/blob/master/docs/sources/cli-command-syntax.md), there is a CLI command example below.
`gobgp global rib add -a opaque key hello value world`

However, when I tried this, the following error occurred.
`root@n1:/gobgp/cmd/gobgp# ./gobgp global rib add -a opaque key hello value world` 
`rpc error: code = Unknown desc = failed to unmarshal nlri: proto: invalid empty type URL`

PR #2898, opened 2025-03-17
Fix: Opaque NLRI

shell720:
Fix: Opaque NLRI
I cannot execute the below example command.
`gobgp global rib add -a opaque key hello value world`
`rpc error: code = Unknown desc = failed to unmarshal nlri: proto: invalid empty type URL`

Then, I fixed this. 
As a result, the result of the rib after executing the command is as follows.

#### local bgp daemon
```
root@n1:/gobgp/cmd/gobgp# ./gobgp global rib add -a opaque key hello value world
root@n1:/gobgp/cmd/gobgp# ./gobgp global rib -a opaque
*> hello                0.0.0.0                                   00:00:19   [{Origin: ?}]

```
#### peer bgp daemon
```
root@n2:/gobgp/cmd/gobgp# ./gobgp global rib -a opaque
*> hello                172.20.20.2          65000                00:01:10   [{Origin: ?}]
```
closes: #2897

### DIFF
--- a/pkg/apiutil/attribute.go
+++ b/pkg/apiutil/attribute.go
@@ -1345,6 +1345,11 @@ func MarshalNLRI(value bgp.AddrPrefixInterface) (*apb.Any, error) {
 			Rd:    rd,
 			Rules: rules,
 		}
+	case *bgp.OpaqueNLRI:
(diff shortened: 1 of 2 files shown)


fujita:
Thanks

PR #2899, opened 2025-03-20
Fix invalid nexthop advertisement with nexthop-unchanged

YutaroHayakawa:
When we apply nexthop-unchanged action to the locally-originated route with unspecified nexthop (0.0.0.0 or ::), we advertise the path with unspecified nexthop. This may ends up with the session shutdown with Invalid Update Message Notification (only tested with FRR 10.2.1, following is a capture of the path with`bgp allow-martian-nexthop` which forcefully allows such an invalid nexthop).

```

BGP routing table entry for 10.244.0.0/24, version 5
Paths: (2 available, best #2, table default)
  Advertised to non peer-group peers:
  fc00:c111::2 fc00:c111::5
  65000
    0.0.0.0 (inaccessible, import-check enabled) from fc00:c111::5 (172.19.0.5)
      Origin IGP, invalid, external
      Last update: Tue Mar 18 16:41:54 2025

```

Fix this issue by setting table.PolicyOptions.OldNextHop to the local peering address when the path is locally-originated and has an unspecified nexthop.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -781,8 +781,17 @@ func (s *BgpServer) prePolicyFilterpath(peer *peer, path, old *table.Path) (*tab
 
 	peer.fsm.lock.RLock()
 	options := &table.PolicyOptions{
-		Info:       peer.fsm.peerInfo,
-		OldNextHop: path.GetNexthop(),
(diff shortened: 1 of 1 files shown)


fujita:
Thanks!

PR #2900, opened 2025-03-23
rtc: implement default route target and cli support

themiron:
* add support for cli add/del operations
* add support for default route target via api/cli
* fix panic while deduplication of default route target
* fix rt membership validation, partial prefixes are still not supported, just 0/32/96-bit for consistency
* add tests


fujita:
Thanks!

PR #2905, opened 2025-04-14
multi path process

chaoliang-wei:


fujita:
Thanks for the patch, and sorry for the long delay. Your later version is already in master. 

This was fixed in master by commit c57da629 ("fix multipath when use bmp or
monitor"), merged as #3448. It takes the same approach, computing the
difference between the old and the new multipath set, with some changes:

- The difference is computed in Update.GetMultiBestPathDiff(), so
  GetChanges() keeps its signature.
- Paths are matched by source and path ID instead of by next hop. A matched
  pair is also compared by attributes, so an attribute change is reported as
  an update.
- A withdrawn path is a clone, so the path in the RIB is not modified.
- The result is passed in watchEventBestPath as UpdatePathList and
  WithdrawPathList. It is a flat list, so one event covers all destinations,
  and both BMP and the monitor API use it.

So I'm closing this. Thanks again.

PR #2940, opened 2025-05-19
server: fix LLGR memory and timer leak

dvalter:
Prevent multiple LLGR reentrances on idle peers causing community memory leak and multiple LLGR timers running.


pkg/config/oc/bgp_configs.go
```diff
@@ -4285,6 +4285,9 @@ type GracefulRestartState struct {
 	// original -> gobgp:long-lived-enabled
 	// gobgp:long-lived-enabled's original type is boolean.
 	LongLivedEnabled bool `mapstructure:"long-lived-enabled" json:"long-lived-enabled,omitempty"`
+	// original -> gobgp:long-lived-running
+	// gobgp:long-lived-running's original type is boolean.
+	LongLivedRunning bool `mapstructure:"long-lived-running" json:"long-lived-running,omitempty"`
 }
```
fujita, line 4292: Don't edit this file directly:
https://github.com/osrg/gobgp/blob/master/tools/pyang_plugins/README.md

OpenConfing is about configurations from config files. I don't think we need to configure `LongLivedRunning` from a config file. So you add `LongLivedRunning` to a different place.

fujita:
Pushed, thanks.

PR #2941, opened 2025-05-19
fix covert withdraw of an updated rtc path

bor1go:
When an rtc-path is updated with a value that belongs to the same Destination, but is not equal (!path.Equal(newPath)), the system sends it to prePolicyFilterPath. There, the new path is dropped and the old one is marked as withdrawn:
`path = old.Clone(true)`
However, the destination and table logic works correctly, and we can see the RTC path in the global rib and in the peer adj-out. Still, there are no responses from the peer, because the RTC path was withdrawn.

Both the new and old paths appear in the `implicitWithdraw` and `explicitWithdraw` functions, using the "source and path identifier equality" criteria, and are then processed in `prePolicyFilterPath `with the `Equal()` criteria ("source and attributes equality"). 

I believe this behavior is an artifact of the infinite reflection prevention logic. There is now a scenario test for infinite reflection.

As part of this PR, I have:
- Removed the line shown above,
- Added a test for covert withdraw,
- Added a test for VRF deletion (the comment for this line is: "We assume the VRF with the specific RT is deleted.").

### DIFF
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -1033,6 +1033,21 @@ func (path *Path) GetLocalPref() (uint32, error) {
 	return lp, nil
 }
 
+func (lhs *Path) EqualByNlri(rhs *Path) bool {
+	if rhs == nil {
+		return false
+	}
+	if lhs == rhs {
+		return true
+	}
+
+	if !lhs.GetSource().Equal(rhs.GetSource()) {
+		return false
+	}
+
+	return lhs.GetNlri().PathIdentifier() == rhs.GetNlri().PathIdentifier()
+}
+
 func (lhs *Path) Equal(rhs *Path) bool {
 	if rhs == nil {
 		return false
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -722,8 +722,10 @@ func (s *BgpServer) prePolicyFilterpath(peer *peer, path, old *table.Path) (*tab
 		}
 
 		if old != nil && old.IsLocal() {
-			// We assumes VRF with the specific RT is deleted.
-			path = old.Clone(true)
+			// If path == nil it will be set to old.Clone(true) in the
+			// func (s *BgpServer) filterpath(peer *peer, path, old *table.Path).
+			// Otherwise this is the local path changing without rt change. We
+			// need to update path or do nothing if path == old.
 		} else if peer.isRouteReflectorClient() {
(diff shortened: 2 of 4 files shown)


bor1go:
@fujita could you please take a look at it, if it is possible?
We discovered this issue when updating the rtc path: the response paths disappeared.

It's weird that the zebra test failed here. It doesn't seem to be related to this PR (RTC only).

internal/pkg/table/path.go
```diff
@@ -1033,6 +1033,21 @@ func (path *Path) GetLocalPref() (uint32, error) {
...
+	}
+
+	if !lhs.GetSource().Equal(rhs.GetSource()) {
+		return false
+	}
+
+	return lhs.GetNlri().PathIdentifier() == rhs.GetNlri().PathIdentifier()
+}
+
 func (lhs *Path) Equal(rhs *Path) bool {
```
fujita, line 1051: This is just refactoring?

pkg/server/server.go
```diff
@@ -722,8 +722,10 @@ func (s *BgpServer) prePolicyFilterpath(peer *peer, path, old *table.Path) (*tab
 		}
 
 		if old != nil && old.IsLocal() {
-			// We assumes VRF with the specific RT is deleted.
-			path = old.Clone(true)
+			// If path == nil it will be set to old.Clone(true) in the
+			// func (s *BgpServer) filterpath(peer *peer, path, old *table.Path).
+			// Otherwise this is the local path changing without rt change. We
+			// need to update path or do nothing if path == old.
 		} else if peer.isRouteReflectorClient() {
```
fujita, line 729: This code is for a local path.
> When an rtc-path is updated with a value that belongs to the same Destination, but is not equal (!path.Equal(newPath)), the system sends it to prePolicyFilterPath.

Your above explanation is about local paths?

internal/pkg/table/path.go
bor1go: Yes, it is. I just tried to show out the difference between this code and Path.Equal(). But I can revert if it was bad idea.

pkg/server/server.go
bor1go: Yes, it is about local paths.

fujita: A local path is updated with another local path belonging to the same Destination? In what situations does this occur? Can you update the description?

fujita:
Pushed, thanks.

PR #2960, opened 2025-05-27
packet/bgp: fix security issue on DecodeFromBytes()

nplanelcisco:
DecodeFromBytes() return value are not tested


fujita:
Pushed, thanks!

PR #2961, opened 2025-05-27
fix: MRT extended timestamp

maxime-peim:
As per https://datatracker.ietf.org/doc/html/rfc6396#section-3


fujita:
Pushed, thanks!

PR #2962, opened 2025-05-28
api: introduce new WatchEvent path(s) library mode API call

nplanelcisco:
```
WatchEvent(ctx, api.request, WatchEventMessageCallbacks{
	OnPathUpdate func([]*apiutil.Path, time.Time)
	OnBestPath   func([]*apiutil.Path, time.Time)
	OnPathEor    func([]*apiutil.Path, time.Time)
	OnPeerUpdate func(*apiutil.WatchEventMessage_PeerEvent, time.Time)
})
```

This will expose the simple and light `apiutil.Path` aim to
get access events without a big overhead, just a translation
from internal table.Path to flat apiutil.Path type.Introducing a new API call, that would give us access to the Events without exposing the watch() method.

Overall, this will give us access to fast BGP info, by removing the protobuf serialization (and object size) overhead.
Please note that GRPC similar API name, relay on bgpServer.API

### DIFF
--- a/pkg/apiutil/util.go
+++ b/pkg/apiutil/util.go
@@ -60,18 +60,18 @@ type Path struct {
 }
 
 type PeerConf struct {
-	PeerAsn           uint32
-	LocalAsn          uint32
+	PeerASN           uint32
+	LocalASN          uint32
 	NeighborAddress   net.IP
 	NeighborInterface string
 }
 type PeerState struct {
-	PeerAsn         uint32
-	LocalAsn        uint32
+	PeerASN         uint32
+	LocalASN        uint32
 	NeighborAddress net.IP
 	SessionState    bgp.FSMState
 	AdminState      api.PeerState_AdminState
-	RouterId        net.IP
+	RouterID        net.IP
 	RemoteCap       []bgp.ParameterCapabilityInterface
 }
 type Transport struct {
--- a/pkg/apiutil/util.go
+++ b/pkg/apiutil/util.go
@@ -38,22 +38,25 @@ const (
 // WatchEventMessages API type
 type WatchEventMessage_PeerEvent struct {
 	Type PeerEventType
-	Peer *Peer
+	Peer Peer
 }
 
 // used by server.WatchEventMessages API
 type Path struct {
-	Nlri  bgp.AddrPrefixInterface      `json:"nlri"`
-	Age   int64                        `json:"age"`
-	Best  bool                         `json:"best"`
-	Attrs []bgp.PathAttributeInterface `json:"attrs"`
-	Stale bool                         `json:"stale"`
-	// true if the path has been filtered out due to max path count reached
-	SendMaxFiltered bool   `json:"send-max-filtered,omitempty"`
-	Withdrawal      bool   `json:"withdrawal,omitempty"`
-	SourceASN       uint32 `json:"source-asn,omitempty"`
-	SourceID        net.IP `json:"source-id,omitempty"`
-	NeighborIP      net.IP `json:"neighbor-ip,omitempty"`
+	Nlri       bgp.AddrPrefixInterface      `json:"nlri"`
+	Age        int64                        `json:"age"`
+	Best       bool                         `json:"best"`
+	Attrs      []bgp.PathAttributeInterface `json:"attrs"`
+	Stale      bool                         `json:"stale"`
+	Withdrawal bool                         `json:"withdrawal,omitempty"`
+	SourceASN  uint32                       `json:"source-asn,omitempty"`
+	SourceID   net.IP                       `json:"source-id,omitempty"`
+	NeighborIP net.IP                       `json:"neighbor-ip,omitempty"`
+	// true if the path has been filtered out due to max path count reached (used by ListPath API)
+	SendMaxFiltered    bool `json:"send-max-filtered,omitempty"`
+	IsFromExternal     bool `json:"is-from-external,omitempty"`
+	NoImplicitWithdraw bool `json:"no-implicit-withdraw,omitempty"`
+	IsNexthopInvalid   bool `json:"is-nexthop-invalid,omitempty"`
 }
 
 type PeerConf struct {
--- a/pkg/apiutil/util.go
+++ b/pkg/apiutil/util.go
@@ -43,15 +43,15 @@ type WatchEventMessage_PeerEvent struct {
 
 // used by server.WatchEventMessages API
 type Path struct {
-	Nlri       bgp.AddrPrefixInterface      `json:"nlri"`
-	Age        int64                        `json:"age"`
-	Best       bool                         `json:"best"`
-	Attrs      []bgp.PathAttributeInterface `json:"attrs"`
-	Stale      bool                         `json:"stale"`
-	Withdrawal bool                         `json:"withdrawal,omitempty"`
-	SourceASN  uint32                       `json:"source-asn,omitempty"`
-	SourceID   net.IP                       `json:"source-id,omitempty"`
-	NeighborIP net.IP                       `json:"neighbor-ip,omitempty"`
+	Nlri        bgp.AddrPrefixInterface      `json:"nlri"`
+	Age         int64                        `json:"age"`
+	Best        bool                         `json:"best"`
+	Attrs       []bgp.PathAttributeInterface `json:"attrs"`
+	Stale       bool                         `json:"stale"`
+	Withdrawal  bool                         `json:"withdrawal,omitempty"`
+	PeerASN     uint32                       `json:"peer-asn,omitempty"`
+	PeerID      net.IP                       `json:"peer-id,omitempty"`
+	PeerAddress net.IP                       `json:"peer-address,omitempty"`
 	// true if the path has been filtered out due to max path count reached (used by ListPath API)
 	SendMaxFiltered    bool `json:"send-max-filtered,omitempty"`
 	IsFromExternal     bool `json:"is-from-external,omitempty"`
@@ -107,8 +107,8 @@ func NewDestination(dst *api.Destination) *Destination {
 			Stale:           p.Stale,
 			SendMaxFiltered: p.SendMaxFiltered,
 			Withdrawal:      p.IsWithdraw,
-			SourceID:        net.ParseIP(p.SourceId),
-			NeighborIP:      net.ParseIP(p.NeighborIp),
+			PeerID:          net.ParseIP(p.SourceId),
+			PeerAddress:     net.ParseIP(p.NeighborIp),
 		})
 	}
 	return &Destination{Paths: l}
--- a/pkg/apiutil/util.go
+++ b/pkg/apiutil/util.go
@@ -26,7 +26,22 @@ import (
 	tspb "google.golang.org/protobuf/types/known/timestamppb"
 )
 
-// workaround. This for the json format compatibility. Once we update senario tests, we can remove this.
+type PeerEventType uint32
+
+const (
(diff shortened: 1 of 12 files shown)


fujita:
Why we need this?

nplanelcisco:
Hi @fujita,

As we discussed offline, we would like to improve the performance of gobgp as a library, without relaying on serialization/deserialization of objects (apiutil, packet/bgp)

I introduce this new API call, that would give us access to the Events without exposing the watch() method.
Overall, this will give us access to the complete BGP message contents including NLRI, path attributes, and other BGP-specific information and remove the serialization overhead.

internal/pkg/table/path.go
```diff
@@ -368,6 +368,62 @@ func (path *Path) root() *Path {
 	return p
 }
 
+// path.pathattrs and path.info.nlri are not cloned as there are interfaces, so the caller should not modify those by using Serialize() helper.
```
kerart: nitpick: "they are"

internal/pkg/table/path.go
```diff
@@ -368,6 +368,62 @@ func (path *Path) root() *Path {
...
+func CloneTablePathTree(paths []*Path) []*Path {
+	oldnewPath := make(map[*Path]*Path, len(paths))
+	oldnewOriginInfo := make(map[*originInfo]*originInfo)
+	oldnewSource := make(map[*PeerInfo]*PeerInfo)
```
kerart: why "oldnew"?

internal/pkg/table/path.go
```diff
@@ -368,6 +368,62 @@ func (path *Path) root() *Path {
...
+			}
+			oldnewPath[pp].info = oldnewOriginInfo[pp.info]
+		}
+	}
+	// update the parent pointers
+	for _, p := range paths {
+		for pp := p; pp != nil; pp = pp.parent {
+			new, found := oldnewPath[pp]
+			if !found {
+				panic("no new path found")
```
kerart: do we really want to panic in that function? shouldn't we simply log an error or ignore as this cloning is only for watchers?

internal/pkg/table/path.go
nplanelcisco: it's like convert old path to new path

internal/pkg/table/path.go
nplanelcisco: it's a bit of paranoia, there no chance we would hit this point with the current code
but if someone change something later, this will guarantee this will not fail silently

pkg/server/server.go
```diff
@@ -47,6 +47,8 @@ import (
 	"github.com/osrg/gobgp/v4/pkg/zebra"
 )
 
+type Path = table.Path
```
joeahn-sbg-cisco: Can we also expose the `PeerInfo` struct and the helper functions in `internal/pkg/table/path.go` like `NewPath`,  `setPathAttr` , and `getPathAttr`? It would be useful in unit testing.

fujita: I don't think that using internal structures like `Path` is a good idea. It includes data used only internally. We need to define a new Path structure for API.

internal/pkg/table/path.go
kerart: Understood. Would it make sense to log an error instead to avoid killing the whole process if that ever happens?

nplanelcisco: may I can return nil in this case, it's a bit less aggressive, up to the caller to log an error. WDYT ?

kerart: Sounds reasonable to me.

pkg/server/server.go
nplanelcisco: @fujita we try here to use gobgp as a library and avoiding coping data (if possible), using an API with different object would imply copy and/or serialization of data.
On the same topic/logic why not exposing `Path` to other golang package ? (renaming paths `internal` to `common`)

fujita: > @fujita we try here to use gobgp as a library and avoiding coping data (if possible), using an API with different object would imply copy and/or serialization of data. On the same topic/logic why not exposing `Path` to other golang package ? (renaming paths `internal` to `common`)

If we don't copy object like Path, the API gives a reference to the object? Then a user of the API can modify Path object and break GoBGP?

nplanelcisco: I did implement a deep path/origin object copy `CloneTablePathTree` aim to solve that, so the user/caller can do what ever he want without interfere with gobgp instance.

If the user would like to access directly to the object (not the default behavior) he can as an option `WatchDontCloneMessage` would pass the Paths pointers, so it's zerocopy, but the responsibility is on the user/caller side.

pkg/apiutil/util.go
```diff
@@ -36,10 +51,42 @@ type Path struct {
...
 	Withdrawal      bool   `json:"withdrawal,omitempty"`
+	SourceASN       uint32 `json:"source-asn,omitempty"`
 	SourceID        net.IP `json:"source-id,omitempty"`
 	NeighborIP      net.IP `json:"neighbor-ip,omitempty"`
 }
 
+type PeerConf struct {
+	PeerAsn           uint32
+	LocalAsn          uint32
+	NeighborAddress   net.IP
```
fujita, line 65: Can you use netip instead? I want to replace all net.IP usages.

pkg/apiutil/util.go
```diff
@@ -36,10 +51,42 @@ type Path struct {
...
+}
+
+type Peer struct {
+	Conf      PeerConf
+	State     PeerState
+	Transport Transport
+}
+
+func (peer *Peer) SessionState() api.PeerState_SessionState {
+	return api.PeerState_SessionState(int(peer.State.SessionState) + 1)
```
fujita: I want to avoid this kinda of conversion. If either of them changes, this code can easily break.
While direct conversion would be preferable from an efficiency standpoint, is there any way to avoid this fragility?

nplanelcisco: I agree on the fragility, but this come from the initial implementation https://github.com/osrg/gobgp/blob/master/pkg/server/server.go#L4473

but technically I can remove this code as it's only used for by test and compatibility with the current/protobuf api

pkg/apiutil/util.go
nplanelcisco: Shall will do that in a separate PR, moving to `netip.Addr` and `netip.Prefix` types (I can follow up on that)

pkg/server/server.go
```diff
@@ -4367,6 +4363,158 @@ func (s *BgpServer) ResetRpki(ctx context.Context, r *api.ResetRpkiRequest) erro
...
+	}
+	return p
+}
+
+type WatchEventMessageCallbacks struct {
+	OnPathUpdate func([]*apiutil.Path, time.Time)
+	OnBestPath   func([]*apiutil.Path, time.Time)
+	OnPathEor    func([]*apiutil.Path, time.Time)
+	OnPeerUpdate func(*apiutil.WatchEventMessage_PeerEvent, time.Time)
+}
```
fujita, line 4410: Does this native API correspond one-to-one with the gRPC API?

pkg/server/server.go
```diff
@@ -4367,6 +4363,158 @@ func (s *BgpServer) ResetRpki(ctx context.Context, r *api.ResetRpkiRequest) erro
...
+
+func (s *BgpServer) WatchEventMessages(ctx context.Context, r *api.WatchEventRequest, callbacks WatchEventMessageCallbacks) error {
+	if r == nil {
```
fujita: I'm not opposing the idea — just to confirm, is the native API intended to use the gRPC request types?

pkg/server/server.go
```diff
@@ -4367,6 +4363,158 @@ func (s *BgpServer) ResetRpki(ctx context.Context, r *api.ResetRpkiRequest) erro
...
+				return status.Errorf(codes.InvalidArgument, "unknown filter type %s", filter.Type)
+			}
+		}
+	}
+	if len(opts) == 0 {
+		return fmt.Errorf("no events to watch")
+	}
+	w := s.watch(opts...)
+
+	go func() {
```
fujita, line 4418: I think that we have the similar code for gRPC API. Is there a way to share something, like by passing a closure that handles data conversion as an argument?

pkg/server/server.go
nplanelcisco: @joeahn-sbg-cisco @fujita please review the new API

pkg/server/server.go
nplanelcisco: on latest version, native API have now his proper API, GRPC API doesn't change and call/relay on the server.API

pkg/server/server.go
nplanelcisco: it's very similar, server.API can replace current grpc.API

fujita:
Is it possible to start by adding just the Watch API? The current diff is quite large, which makes it hard to review.

nplanelcisco:
> Is it possible to start by adding just the Watch API? The current diff is quite large, which makes it hard to review.

sure I can split the PR

nplanelcisco:
> > Is it possible to start by adding just the Watch API? The current diff is quite large, which makes it hard to review.
> 
> sure I can split the PR

@fujita done

fujita:
Please fix the embedded CI failure.

pkg/server/grpc_server.go
```diff
@@ -312,9 +295,108 @@ func (s *server) ListPath(r *api.ListPathRequest, stream api.GoBgpService_ListPa
 	return send()
 }
 
+func (s *server) watchEvent(ctx context.Context, r *api.WatchEventRequest, fn func(*api.WatchEventResponse, time.Time)) error {
+	if r == nil {
+		return fmt.Errorf("nil request")
+	}
```
fujita, line 301: `watchEvent()` should return gRPC errors?
https://pkg.go.dev/google.golang.org/grpc/codes

pkg/apiutil/util.go
```diff
@@ -26,18 +26,64 @@ import (
...
-	SourceID        net.IP `json:"source-id,omitempty"`
-	NeighborIP      net.IP `json:"neighbor-ip,omitempty"`
+	Nlri       bgp.AddrPrefixInterface      `json:"nlri"`
+	Age        int64                        `json:"age"`
+	Best       bool                         `json:"best"`
+	Attrs      []bgp.PathAttributeInterface `json:"attrs"`
+	Stale      bool                         `json:"stale"`
+	Withdrawal bool                         `json:"withdrawal,omitempty"`
+	SourceASN  uint32                       `json:"source-asn,omitempty"`
+	SourceID   net.IP                       `json:"source-id,omitempty"`
```
fujita: I know that you simply use the same member name but it should be RouterID or something?

pkg/apiutil/util.go
```diff
@@ -26,18 +26,64 @@ import (
...
+	NeighborAddress   net.IP
+	NeighborInterface string
+}
+type PeerState struct {
+	PeerAsn         uint32
+	LocalAsn        uint32
+	NeighborAddress net.IP
+	SessionState    bgp.FSMState
+	AdminState      api.PeerState_AdminState
+	RouterId        net.IP
```
fujita: `RouterID`?

pkg/server/grpc_server.go
nplanelcisco: done

pkg/apiutil/util.go
nplanelcisco: I changed to Peer{ASN,ID,Adress}

fujita:
Congratulations! Just a reminder. Don’t forget to replace the ip package with the netip package. I plan to update all exported APIs to use netip before v4.

PR #2971, opened 2025-06-02
path: GetMed() optimize error path

nplanelcisco:
Error path could his quite a lot by talbe.getMultiBestPath()->path.Compare()->GetMed() fmt.Errorf() would do an allocation/free doing pressure on the GC

Introduce ErrNoMedPathAttr aim to fix this issue


fujita:
Make sense, thanks!

PR #2972, opened 2025-06-02
path: optimize by caching bgp.COMMUNITY_LLGR_STALE attribute

nplanelcisco:
Currently a lookup of all the attributes is done to know if the COMMUNITY_LLGR_STALE is set.
This PR will cache it in the path.cache.IsLLGRStale


nplanelcisco:
drop it in favor of https://github.com/osrg/gobgp/pull/2975

PR #2974, opened 2025-06-04
pkg/packet: fuzzing fixup DecodeFromBytes

nplanelcisco:
golang fuzzing toolchain pass on all deserialization DecodeFromBytes() objects pkg/packet/{bgp,rtr,mrt,bmp}


fujita:
Great, thanks a lot!

PR #2975, opened 2025-06-04
path: fast path attributes O(1) and hash optimization

nplanelcisco:
The general goal of this PR is optimize access to path attributes and path compare, aim to have fast table.destination.computeKnownBestPath()

This PR introduce 
* O(1) access to path attributes
* fast path compare (no need to sort attributes anymore)
  * attribute hash
  * path attributes hash
* GetTransversalPathAttrs() will solve the path attributes if path is cloned

Running the recent unit test benchmark form @joeahn-sbg-cisco merged in gobgp
This PR we can see 10% improvement.

```
go test -benchmem -run=^$ -bench ^BenchmarkCalculate$ github.com/osrg/gobgp/v4/internal/pkg/table

Before : BenchmarkCalculate-12          519141          2225 ns/op         376 B/op          14 allocs/op
After  : BenchmarkCalculate-12          602040          2013 ns/op         376 B/op          14 allocs/op   
```

### DIFF
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -23,7 +23,7 @@ import (
 	"math"
 	"net"
 	"slices"
-	"sort"
+	"sync"
 	"time"
 
 	"github.com/osrg/gobgp/v4/pkg/config/oc"
@@ -72,10 +72,7 @@ func (b *Bitmap) FindandSetZeroBit() (uint, error) {
 }
 
 func (b *Bitmap) Expand() {
-	old := b.bitmap
-	new := make([]uint64, len(old)+1)
-	copy(new, old)
-	b.bitmap = new
+	b.bitmap = append(b.bitmap, uint64(0))
 }
 
 func NewBitmap(size int) *Bitmap {
@@ -132,11 +129,46 @@ type Validation struct {
 	UnmatchedLength []*ROA
 }
 
+// it's faster than using a map for the path attributes.
+// Golang will waist lots of cpu to calculate the 8 bits hash key.
+// PathAttributes is a sorted slice of bgp.PathAttributeInterface that we can access in O(logN) time.
+type PathAttributes []bgp.PathAttributeInterface
+
+var searchAttribute = sync.Pool{
+	New: func() interface{} {
+		return &bgp.PathAttributeUnknown{}
+	},
+}
+
+func (pa PathAttributes) Get(a bgp.BGPAttrType) bgp.PathAttributeInterface {
+	sattr := searchAttribute.Get().(*bgp.PathAttributeUnknown)
+	defer searchAttribute.Put(sattr)
+	sattr.Type = a
+	n, found := slices.BinarySearchFunc(pa, bgp.PathAttributeInterface(sattr), func(a, b bgp.PathAttributeInterface) int {
+		return int(a.GetType()) - int(b.GetType())
+	})
+	if !found {
+		return nil
+	}
+	return pa[n]
+}
+
+func (pa *PathAttributes) Set(a bgp.PathAttributeInterface) {
+	n, found := slices.BinarySearchFunc(*pa, a, func(a, b bgp.PathAttributeInterface) int {
+		return int(a.GetType()) - int(b.GetType())
+	})
+	if found {
+		(*pa)[n] = a
+	} else {
+		*pa = slices.Insert(*pa, n, a)
+	}
+}
+
 type Path struct {
 	info      *originInfo
 	parent    *Path
-	pathAttrs []bgp.PathAttributeInterface
-	dels      []bgp.BGPAttrType
+	pathAttrs PathAttributes
+	dels      map[bgp.BGPAttrType]struct{}
 	attrsHash uint32
 	rejected  bool
 	// doesn't exist in the adj
@@ -181,7 +213,7 @@ func NewPath(source *PeerInfo, nlri bgp.AddrPrefixInterface, isWithdraw bool, pa
 		return nil
 	}
 
-	return &Path{
+	p := &Path{
 		info: &originInfo{
 			nlri:               nlri,
 			source:             source,
@@ -189,8 +221,12 @@ func NewPath(source *PeerInfo, nlri bgp.AddrPrefixInterface, isWithdraw bool, pa
 			noImplicitWithdraw: noImplicitWithdraw,
 		},
 		IsWithdraw: isWithdraw,
-		pathAttrs:  pattrs,
+		dels:       make(map[bgp.BGPAttrType]struct{}),
 	}
+	for _, a := range pattrs {
+		p.setPathAttr(a)
+	}
+	return p
 }
 
 func NewEOR(family bgp.Family) *Path {
@@ -200,6 +236,7 @@ func NewEOR(family bgp.Family) *Path {
 			nlri: nlri,
 			eor:  true,
 		},
+		dels: make(map[bgp.BGPAttrType]struct{}),
 	}
 }
 
@@ -360,6 +397,7 @@ func (path *Path) Clone(isWithdraw bool) *Path {
 		IsWithdraw:       isWithdraw,
 		IsNexthopInvalid: path.IsNexthopInvalid,
 		attrsHash:        path.attrsHash,
+		dels:             make(map[bgp.BGPAttrType]struct{}),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


fujita:
I think that using map for attributes would make sense.
However, I don't think that it's a good idea to add a hash value to PathAttribute.
- packet package is supposed to just decode/encode. a hash value isn't included in a packet format. Note that I want to remove PathIdentifier* methods in AddrPrefixInterface for the same reason.
- Path attributes are usually updated, so computing the hash during parsing has limited benefit. Calculating a hash after updating attributes should be ok (Bird does the same, IIRC)

nplanelcisco:
Thank you @fujita for your responses

> I think that using map for attributes would make sense. However, I don't think that it's a good idea to add a hash value to PathAttribute.
> 
> * packet package is supposed to just decode/encode. a hash value isn't included in a packet format. Note that I want to remove PathIdentifier* methods in AddrPrefixInterface for the same reason.

Yes it will never be serialized, that's why object are for, we can add metadata that are helpful (other idea see [1])

> * Path attributes are usually updated, so computing the hash during parsing has limited benefit. Calculating a hash after updating attributes should be ok (Bird does the same, IIRC)

I hash computing is done during parsing because we have already all the bytes for, so why not doing `hash(bytes)` on `DecodeFromBytes()` , it's quite cheap .... in the worst case I can leave it to the default hash (0), and then it will be calculated if needed.


[1] In another track/idea, it would be nice to expose Path and Attibutes outside the `internal` package, I think moving these to `common`, think way gobgp could be used a library and from outside packages, the general idea would be we would like to do a new product that is a extenssion of gobgp

fujita:
Structures used internally should be separated from those exposed as part of the API.

nplanelcisco:
> Structures used internally should be separated from those exposed as part of the API.

I follow-up the discussion to this PR/thread https://github.com/osrg/gobgp/pull/2962#discussion_r2144648182

internal/pkg/table/path.go
```diff
@@ -501,61 +515,53 @@ func (a PathAttrs) Less(i, j int) bool {
...
+		for t := range p.dels {
+			delete(modified, t)
+		}
+	}
+	return modified
+}
+
+func (path *Path) GetPathAttrs() []bgp.PathAttributeInterface {
+	modified := path.GetTransversalPathAttrs()
+	list := make([]bgp.PathAttributeInterface, len(modified))
```
joeahn-sbg-cisco: For this, can we allocate just the capacity `make([]bgp.PathAttributeInterface, 0, len(modified))` and `append` the `a` elements to it so that when the caller of the function iterates over the list of PathAttributeInterface it doesn't actually have to check for nil elements?

nplanelcisco: I did some changes and more tests, the returned list need to be in order because it's expected like so, and removed the nil from the list.
I did much more tests and there where not so obvious regression (breaking LLGR CI).

joeahn-sbg-cisco: > the returned list need to be in order because it's expected like so, and removed the nil from the list.

I don't fully understand. 

To clarify what I mean is when we allocate the `list` to be of type `bgp.PathAttributeInterface` and length equal to `modified`, the initial value of each element in `list` is `nil`. We overwrite the elements `list[i] = a` conditionally if the `modified` element exists, so if somehow `i != len(modified)` by the end of this for loop, then there will be `nil` elements still in the list. 

By doing `make([]bgp.PathAttributeInterface, 0, len(modified))`, my suggestion is to allocate the capacity but not the length for the list, and do an `list = append(list, a)` so that when the callers of this function iterate over the returned list like `for a := range list`, it doesn't need to check for non-overwritten `nil` elements

Example of the difference between the two: https://go.dev/play/p/rITIQ63fK_a

nplanelcisco: I see your point, but this condition will never occur, as i is incremented if modified[t] exist, and the relation is that i will be always in count(t) as t will iterate all the possible known types of modified. I can add a fence at the end of the loop like `if i != len(modified) { panic("something is wrong, we should not hit this") }

I do this way because it faster (create an array and setup the elements) and there are no risk for i outside of bounds
```
pkg: github.com/osrg/gobgp/v4/internal/pkg/table
cpu: 13th Gen Intel(R) Core(TM) i7-1365U
BenchmarkA/PathAttributeInterfaceList-12              2886636           420.6 ns/op           0 B/op           0 allocs/op
BenchmarkA/PathAttributeInterfaceListAppend-12        1081370          1161 ns/op        2155 B/op           0 allocs/op

func BenchmarkA(b *testing.B) {
    modified := make(map[bgp.BGPAttrType]bgp.PathAttributeInterface)
    for _, t := range bgp.BGPAttrTypeIterator {
        modified[t] = nil
    }

    list := make([]bgp.PathAttributeInterface, len(modified))
    b.Run("PathAttributeInterfaceList", func(b *testing.B) {
        b.ResetTimer()
        for i := 0; i < b.N; i++ {
            i := 0
            for _, t := range bgp.BGPAttrTypeIterator {
                if a, ok := modified[t]; ok {
                    list[i] = a
                    i++
                }
            }
        }
    })

    list = make([]bgp.PathAttributeInterface, 0, len(modified))
    b.Run("PathAttributeInterfaceListAppend", func(b *testing.B) {
        b.ResetTimer()
        for i := 0; i < b.N; i++ {
            for _, t := range bgp.BGPAttrTypeIterator {
                if a, ok := modified[t]; ok {
                    list = append(list, a)
                }
            }
        }
    })
}
```

nplanelcisco:
@fujita Could you take a look/re-review this PR, as it's in a good shape and ready to merge.
Some others part of the code (like bgp update message) will be addressed later.

aspath UT seems to be facky on the CI (I'm not allowed to restart it), but can't be reproduce it locally

internal/pkg/table/path.go
```diff
@@ -132,12 +128,18 @@ type Validation struct {
 	UnmatchedLength []*ROA
 }
 
+type PathAttributes [bgp.BGP_ATTR_TYPE_MAX]bgp.PathAttributeInterface
```
nplanelcisco: @fujita I did double check, and you right everything was fine here as we don't test path with unknown attributes, I added an unit test. Thanks

internal/pkg/table/path.go
```diff
@@ -132,12 +128,20 @@ type Validation struct {
 	UnmatchedLength []*ROA
 }
 
+// BGPAttrType(256) here as it's faster than using a map for the path attributes.
+// Golang will waist lots of cpu to calculate the 8 bits hash key.
+type PathAttributes [256]bgp.PathAttributeInterface
+
```
fujita, line 136: Add 4096 bytes to Path? Considering the possibility of having multiple full-route peers, I don't think this is a good idea.
If using hash with a map consumes too much CPU, then we would go with list (like BIRD). Or we might use a structure that keeps all kinds of attributes (like FRR).

pkg/packet/bgp/bgp.go
```diff
@@ -10570,14 +10574,36 @@ type PathAttributeInterface interface {
...
+		if attr == nil || slices.ContainsFunc(exclude, func(e BGPAttrType) bool { return e == BGPAttrType(t) }) {
+			continue
+		}
+		h = fnv1a.AddUint64(h, GetPathAttributeHash(attr))
+	}
+	return h
+}
+
 type PathAttribute struct {
+	hash   uint64 // hash for the whole payload, excluding the header Flags, Type, Length
```
fujita: Please don't add hash here. packet/bgp for a library for marshaling unmarshaling.

nplanelcisco: I'm not sure to followup here, `PathAttribute` is an object and we can add metadata, that will help our processing time quite a lot. We don't want to calculate the hash every time we need to compare PathAttribute, PathAttribute are not mutable so hash will be calculated once if needed (Serialization or Compare)

Why do you want strict object fields for serialization, this field will not be used by Serialization/de-Serialization, it's just metadata hook to this object.

internal/pkg/table/path.go
nplanelcisco: it's more >2048 ish (256*8) bytes, but I agree I will move to something another solution with less memory footprint

fujita: > it's more >2048 ish (256*8) bytes, but I agree I will move to something another solution with less memory footprint

On my env, this uses 4096 bytes. I think that the interface is a pointer with type information.

pkg/packet/bgp/bgp.go
fujita: - For code that uses this library solely for packet serialization/de-serialization, the hash value is unnecessary.
- When storing millions of Path objects, we should be cautious about increasing the size of Path attributes.
- Basic path attributes are small values like u8 (for example, the origin attribute). Storing a u64 hash alongside a u8 attribute doesn’t make sense at all.
- The AS Path attribute is probably the attribute where a hash would be most effective, but in most use cases, the AS Path is modified before transmission. So computing the hash during deserialization is wasteful.

The hash of path attributes is used during transmission to group NLRIs, and the hash should be computed over the entire set of attributes, not each attribute individually. In my opinion, the real issue with the current code is that it's serializing just to compute a hash. It might be better to store the attributes in their serialized form—as a flat memory layout—like BIRD does.

internal/pkg/table/path.go
fujita: I think we shouldn't try to optimize access to all types of path attributes. What really matters for BGP performance are the attributes that are accessed repeatedly—specifically, those used in best path selection. That’s likely just a small subset (probably attributes with type codes under 10).

For example, we could keep using a list in the current implementation, and store the indexes of the frequently accessed attributes in the Path struct.

nplanelcisco:
dropping this PR as https://github.com/osrg/gobgp/pull/3026 as been merged and path object size must kept small

PR #2976, opened 2025-06-06
fix: migrate to modern testing interface

emmanuel-ferdman:
## PR Summary
This small maintenance PR is migrates from `unittest.assertEquals` to `unittest.assertEqual` which is deprecated from Python2.7:
```python
DeprecationWarning: Please use assertEqual instead.
```
It also fixes a few typos along the way.


fujita:
Thanks!

PR #2980, opened 2025-06-11
fuzzing: 2nd pass findings

nplanelcisco:
The big fixup are
* zebra protocol
* OpaqueNLRI
* ParseRTR


fujita:
Pushed with a lint fix. Thanks!

PR #2983, opened 2025-06-16
path: fixup destination implicitWithdraw()

nplanelcisco:
destination.Calculate() should not replace a path that has been marked "no implicit withdraw"
Added some unit tests.


fujita:
Pushed, thanks!

btw, `noImplicitWithdraw` is used?

nplanelcisco:
> Pushed, thanks!
> 
> btw, `noImplicitWithdraw` is used?

AFIK not on our side, but I read this code and seen it was wrong, so here the PR :)

PR #2984, opened 2025-06-16
table: faster key without allocation

nplanelcisco:
Note: There are no allocation for IPAddrPrefix IPv6AddrPrefix types

    ```
    Before:
    BenchmarkTableKey/TableKey_known_types-12                6205518               177.9 ns/op            96 B/op          6 allocs/op
    BenchmarkTableKey/TableKey_with_unknown_type-12          1224622              1132 ns/op             416 B/op         17 allocs/op
    
    After:
    BenchmarkTableKey/TableKey_known_types-12               13046750                79.58 ns/op           16 B/op          2 allocs/op
    BenchmarkTableKey/TableKey_with_unknown_type-12          1000000              1014 ns/op             336 B/op         13 allocs/op
    ```


fujita:
Since hash collisions are possible, I don't think it's a good idea to use hash values directly as map keys in a BGP implementation. As in the BIRD implementation, after generating a hash value, it's necessary to compare the original data if a matching key already exists.

I think that it might be better to focus on only IPAddrPrefix and IPv6AddrPrefix about this improvement. We can get a unique key (u64) from IPAddrPrefix easily.

fujita:
One thing I'd like to mention in relation to this change is that I'm considering using the netip package.  IPAddrPrefix and IPv6AddrPrefix could be implemented by using netip.Prefix.

nplanelcisco:
> One thing I'd like to mention in relation to this change is that I'm considering using the netip package. IPAddrPrefix and IPv6AddrPrefix could be implemented by using netip.Prefix.

I think would be nice, as we will not hit https://pkg.go.dev/vuln/GO-2024-2887 as we depend on go 1.23.0

nplanelcisco:
> Since hash collisions are possible, I don't think it's a good idea to use hash values directly as map keys in a BGP implementation. As in the BIRD implementation, after generating a hash value, it's necessary to compare the original data if a matching key already exists.
> 
> I think that it might be better to focus on only IPAddrPrefix and IPv6AddrPrefix about this improvement. We can get a unique key (u64) from IPAddrPrefix easily.

I think if we hit a collision on 64bits hash there are really something wrong with the hash algorithm, as we don't have billion of billions of entries. I tried here to make golang map hash life a bit easier.

for the IP{,v6}AddrPrefix types, I think would make sense to introduce `PrefixHash() uint64` in the `AddrPrefixInterface` interface ? WDYT ?

fujita:
> > Since hash collisions are possible, I don't think it's a good idea to use hash values directly as map keys in a BGP implementation. As in the BIRD implementation, after generating a hash value, it's necessary to compare the original data if a matching key already exists.
> > I think that it might be better to focus on only IPAddrPrefix and IPv6AddrPrefix about this improvement. We can get a unique key (u64) from IPAddrPrefix easily.
> 
> I think if we hit a collision on 64bits hash there are really something wrong with the hash algorithm, as we don't have billion of billions of entries. I tried here to make golang map hash life a bit easier.

BGP developers are paranoia.

YutaroHayakawa:
I was wondering if changing the table data structure to trie helps reducing the memory consumption and solves the concern around the hash collision. Recently Cilium introduced the [pure-go trie implementation](https://pkg.go.dev/github.com/cilium/cilium@v1.17.4/pkg/container/bitlpm) which we use for network policy engine. It also contains `CIDRTrie` which takes `netip.Prefix` as a key. 

Unfortunately, I cannot try this soonish, but if you're interested in changing the table's implementation, it might be worth checking out.

nplanelcisco:
> I was wondering if changing the table data structure to trie helps reducing the memory consumption and solves the concern around the hash collision. Recently Cilium introduced the [pure-go trie implementation](https://pkg.go.dev/github.com/cilium/cilium@v1.17.4/pkg/container/bitlpm) which we use for network policy engine. It also contains `CIDRTrie` which takes `netip.Prefix` as a key.
> 
> Unfortunately, I cannot try this soonish, but if you're interested in changing the table's implementation, it might be worth checking out.

Thanks for the pointer, a Trie sound a good idea based on CIDRTrie this should work
https://github.com/cilium/cilium/blob/v1.17.4/pkg/container/bitlpm/cidr.go#L20

fujita:
Very interesting. We tried radix long ago:
https://github.com/armon/go-radix

Anyway, I did some benchmarks:
https://github.com/fujita/bgp-table-bench

~/git/bgp-table-bench$ go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/fujita/bgp-table-bench
cpu: Intel(R) Core(TM) i9-14900
BenchmarkMapNetip-32           	       8	 133387192 ns/op
BenchmarkMapString-32          	       6	 187968740 ns/op
BenchmarkMapUint64-32          	      18	  64865437 ns/op
BenchmarkCIDRTrie-32           	      10	 104660166 ns/op
BenchmarkLookupMapNetip-32     	      12	  97008821 ns/op
BenchmarkLookupMapString-32    	       8	 125620556 ns/op
BenchmarkLookupMapUint64-32    	      37	  28796312 ns/op
BenchmarkLookupCIDRTrie-32     	      10	 101717153 ns/op
PASS
ok  	github.com/fujita/bgp-table-bench	15.814s

YutaroHayakawa:
Yeah, Cilium also heavily utilizes [go-immutable-radix](https://github.com/hashicorp/go-immutable-radix) which uses go-radix internally. However, it only supports byte-by-byte operations, but they need bit-by-bit operations. That's why they invented their own trie implementation.

Thanks for making a benchmark. I also tried with `-benchmem` and seems like the `MapUint64` performs well with caveat that it may have hash collision (in this benchmark with internet full route, we didn't have any collision).

CIDRTrie is generally compact and performs better than MapNetip in Update, but the Lookup is significantly slower. Makes sense for me. 

```
$ go test -bench . -benchmem
goos: linux
goarch: arm64
pkg: github.com/fujita/bgp-table-bench
BenchmarkMapNetip-4          	      10	 112955177 ns/op	201075721 B/op	    8203 allocs/op
BenchmarkMapString-4         	       6	 197159046 ns/op	143662042 B/op	 3012258 allocs/op
BenchmarkMapUint64-4         	      19	  61042123 ns/op	75522411 B/op	    8201 allocs/op
BenchmarkCIDRTrie-4          	      12	  96267939 ns/op	117400376 B/op	 1834384 allocs/op
BenchmarkLookupMapNetip-4    	      21	  53226109 ns/op	       0 B/op	       0 allocs/op
BenchmarkLookupMapString-4   	       7	 160932567 ns/op	32057301 B/op	 3004057 allocs/op
BenchmarkLookupMapUint64-4   	      30	  38993381 ns/op	       0 B/op	       0 allocs/op
BenchmarkLookupCIDRTrie-4    	       8	 127227362 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/fujita/bgp-table-bench	12.767s
```

nplanelcisco:
Looking at the performances, `uint64` is by far the best choice here, but could have some collision.
So I updated the code to detect collision, and save the item and log (warn) if that occur during an insert.
Collision will not affect the `Get()` as nlri will be compared

fujita:
I think that we need to panic if we hit collision. False advertising must never happen.

nplanelcisco:
> I think that we need to panic if we hit collision. False advertising must never happen.

I agree we must not false advertise. We don't need to panic as the current implementation detect hash collision and handle this case correctly [here](https://github.com/osrg/gobgp/pull/2984/files#diff-38b2af98135d32504ce23f970b3711c0c6e30b71ec3cac2b6b1be8f8f4f54439R115) and line 99

The current implementation is a `map[uint64][]*Destination`, so if a destination is already present with the same key we check the equality, then append it if not (and report it as a log.Warning). It's a bit slower than a map without list as we need to double check (compare) the NLRI before insert, but realistically  it's minimalist as it's just happen during an update (unlikely)

pkg/packet/bgp/bgp.go
```diff
@@ -1378,6 +1378,40 @@ type AddrPrefixInterface interface {
...
+		b[24] = T.Length
+		return b
+	}
+	return []byte(nlri.String())
+}
+
+func AddrPrefixCompare(a, b AddrPrefixInterface) int {
+	return bytes.Compare(addrPrefixSerialize(a), addrPrefixSerialize(b))
+}
+
```
fujita, line 1414: Why we can't use the existing Serialize function? Because of addpath id?

pkg/log/test_logger.go
```diff
@@ -0,0 +1,61 @@
+package log
+
```
fujita: Why we need a new logger implementation?

fujita:
The changes look good overall to me. I'm just unclear on the necessity of addrPrefixSerialize() and a new logger.
I think that we don't need collision stats or warnings because the code can handle it safely but if you prefer to have them, it's fine by me.

pkg/log/test_logger.go
nplanelcisco: it's just to catch the error message in the UT, I can remove it now as collision is part of Info().

nplanelcisco: I will remove this test logger

pkg/packet/bgp/bgp.go
nplanelcisco: We can't use Serialize directly because `addrPrefixSerialize()` serialize only the prefix but not the labels in case of `LabeledVPNIPAddrPrefix` `LabeledVPNIPv6AddrPrefix` basically I just moved this logic to bgp package so we could compare prefix now.

nplanelcisco:
> The changes look good overall to me. I'm just unclear on the necessity of addrPrefixSerialize() and a new logger. I think that we don't need collision stats or warnings because the code can handle it safely but if you prefer to have them, it's fine by me.

I renamed `addrPrefixOnlySerialize()` to avoid confusion with the object serialization, I dropped the collision test log message check
I think the warnings and stats would make sense as we don't expect to hit this and we should spent some time to investigate if that happen.
I found (and fix) an issue when labeled prefix is used, as the prefix is not part of the `table key` or `addrPrefixOnlySerialize()` but in the `prefix.Length` we were counting the numbers of labels.

pkg/packet/bgp/bgp.go
fujita: Understood but why you implement this in bgp.go? This is used by only the table package and looks like it can be implemented in the table package.

nplanelcisco: prefixes implementation are in `bgp` package, so it make sense to not expose internal mechanism (`addrPrefixOnlySerialize()`)  but only API that we need `AddrPrefixOnlyCompare()`
On the same logic it would make sense to have `AddrPrefixOnlyHash(a AddrPrefixInterface) uint64` in `bgp` package, like moving this code https://github.com/osrg/gobgp/pull/2984/files#diff-38b2af98135d32504ce23f970b3711c0c6e30b71ec3cac2b6b1be8f8f4f54439R65-R80

what would be the value to move this particular code in table ?
I would prefer this way, but if you really want I can move these bits back to `table` package

fujita: > prefixes implementation are in `bgp` package, so it make sense to not expose internal mechanism (`addrPrefixOnlySerialize()`) but only API that we need `AddrPrefixOnlyCompare()`
> 
> what would be the value to move this particular code in table ? I would prefer this way, but if you really want I can move these bits back to `table` package

I believe it's idiomatic in Go to keep the public API minimal in order to preserve flexibility and maintainability of the internal implementation.

fujita:
Pushed, thanks!
I'll move AddrPrefixCompare() later.

PR #2985, opened 2025-06-17
feat: add timestamp to watch event

maxime-peim:
when a lot of event are being pushed to watchers, it could be good to know when they were generated

### DIFF
--- a/docs/sources/bgp-ls.md
+++ b/docs/sources/bgp-ls.md
@@ -74,6 +74,7 @@ package main
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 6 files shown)


fujita:
I’m not opposed to this idea. But just to confirm — even though you're trying to improve the performance of the Watch API, you're also proposing to add the overhead of generating a timestamp every time a Watch response is created. Is that correct?

mpeim:
well it is already the case for some events so I though that it could be that case for all

fujita:
Pushed, thanks.

PR #2987, opened 2025-06-19
chore: move listener/socket code to its own package

maxime-peim:


fujita:
What's the motivation?

mpeim:
It is a first step to split the code base into smaller, more readable and improvable chuncks.
At least for me, it is easier to work with smaller files that ones with thousand of lines

fujita:
I agree. But does it need to be a public package? Wouldn't it be enough to use it internally within GoBGP?

mpeim:
Indeed, I can move it to the internals!

fujita:
It sounds good to me, assuming there’s no change in functionality.

mpeim:
there is no functionality changes in the first commit (it is more or less a copy/paste).
In the second commit (I just split it so it is obvious) I fix the accept loop.
in the third I add the tracking of the connections

fujita:
What's the motivation of the third patch?

maxime-peim:
> What's the motivation of the third patch?

It happen during my testing that TCP sockets did not get closed, so this would ensure that

fujita:
Pushed, thanks.

PR #2988, opened 2025-06-23
fix: fuzzing make linter happy

nplanelcisco:
and we must not check return error as it's legit execution path in fuzzing mode


fujita:
Thanks!

PR #2991, opened 2025-06-26
Count advertised-routes only when ESTABLISHED

yushoyamaguchi:
Cilium has this issue.
https://github.com/cilium/cilium/issues/39861
The problem is to display the number of advertised routes, even for disconnected peers.
The cause seems to be in `ListPeer()` in GoBGP.

`ListPeer()` is used for getting information about peers.
In this function, the number of advertised-routes should be shown only about ESTABLISHED peers, but currently advertised-routes of other state peers are counted.
In this commit, ListPeer() is changed it.

Honestly, I'm not sure that the _if-condition filter_ in this commit is necessary and sufficient to achieve this purpose.
I think that it may be better to insert _if-condition filter_ at an earlier stage, or insert _if-condition filter_ in `getBestFromLocal()`.
I want to discuss it more.
However, it is for sure that this patch resolves the above issue.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -3162,12 +3162,19 @@ func (s *BgpServer) ListPeer(ctx context.Context, r *api.ListPeerRequest, fn fun
 					c := afisafi.Config
 					if c.Family != nil && c.Family.Afi == api.Family_Afi(afi) && c.Family.Safi == api.Family_Safi(safi) {
 						flist := []bgp.Family{family}
-						received := uint64(peer.adjRibIn.Count(flist))
-						accepted := uint64(peer.adjRibIn.Accepted(flist))
+						peer.fsm.lock.RLock()
+						sesstionState := peer.fsm.state
+						peer.fsm.lock.RUnlock()
+						received := uint64(0)
+						accepted := uint64(0)
 						advertised := uint64(0)
-						if getAdvertised {
-							pathList, _ := s.getBestFromLocal(peer, flist)
-							advertised = uint64(len(pathList))
+						if sesstionState == bgp.BGP_FSM_ESTABLISHED {
+							received = uint64(peer.adjRibIn.Count(flist))
+							accepted = uint64(peer.adjRibIn.Accepted(flist))
+							if getAdvertised {
+								pathList, _ := s.getBestFromLocal(peer, flist)
+								advertised = uint64(len(pathList))
+							}
 						}
 						p.AfiSafis[i].State = &api.AfiSafiState{
 							Family:     c.Family,


yushoyamaguchi:
P.S.
`
ListPeer()` seems to be used in

1. API backend
2. BMP
3. metrics

part in GoBGP.
(Cilium uses this function by embedding pkg of `gobgp` and calls it directly. )

For all of these cases, I think `ListPeer()` should return 0 for the number of advertised-routes, for disconnected peers.

yushoyamaguchi:
@fujita 

I'm sorry for mentioning.
Could you assign some reviewer?

pkg/server/server.go
```diff
@@ -3165,7 +3165,7 @@ func (s *BgpServer) ListPeer(ctx context.Context, r *api.ListPeerRequest, fn fun
 						received := uint64(peer.adjRibIn.Count(flist))
 						accepted := uint64(peer.adjRibIn.Accepted(flist))
 						advertised := uint64(0)
-						if getAdvertised {
+						if getAdvertised && peer.fsm.state == bgp.BGP_FSM_ESTABLISHED {
```
fujita: - you cannot access to peer.fsm without fsm mutex.
-  If a peer isn't established, received and accepted are pointless too?

fujita: With LLGR enabled, even if the state isn't established, advertised-routes could be non zero?

yushoyamaguchi: Thank you for your comment.

I've added Lock and Unlock when accessing `peer.fsm`.

Also, I implement the procedure for initializing `received` and `accepted` as zero.
Actually, `received` and `accepted` become zero when Session-State is not ESTABLISHED, maybe because of `peer.adjRibIn` implementation. 
(These two variables is derived from `peer.adjRibIn`, unlike `advertised`. )
However, because I think these three variables should be treated in the same way when not ESTABLISHED, I've added this procedure.

yushoyamaguchi: I think `advertised` variable should represent the the number of advertised route which derive from **state of my RIB** (not neighboring router).
That's why I didn't handle the _LLGR_ case differently.
However, if you maintainers believe that neighbor statistics during _LLGR_ should be shown the same as in the _ESTABLISHED_ state, I have no hesitation in doing so.

fujita:
Pushed, thanks.

yushoyamaguchi:
@fujita 

I'm sorry, maybe it is not pushed and closed.
Is it intended?

fujita:
https://github.com/osrg/gobgp/commit/c867115bf487e3f6df465cdfa39c036f9316411e
?

yushoyamaguchi:
I'm so sorry, I've overlooked.

PR #2992, opened 2025-06-26
refactor: move peer/fsm code to its own package

maxime-peim:
- move some functions to a bgputils package
- move utility functions to utils package
- merge fsm and fsm handler
- peer fsm loop instead of creating handlers for each fsm state


fujita:
I agree that the design of Peer, fsm, and fsmHandler is in a poor state and needs significant cleanup. Before moving things into a separate package, I'd like to remove the incoming channel and reorganize the data with synchronization in mind.

maxime-peim:
Could I give it a try to remove the incoming channel on top of the changes that I have made here?

fujita:
I've already removed incoming
https://github.com/osrg/gobgp/pull/2993

maxime-peim:
> I've already removed incoming #2993

Oh nice, I can simply rebase my changes on it then

fujita:
Could you split out the obvious bug fixes from the commits in this pull request and submit them as a separate PR first? I think it would be better to decide on the overall design before starting the fsm refactoring — especially regarding how to handle data that's accessed from other goroutines.

maxime-peim:
Sure

fujita:
Also, let me know if you have any thoughts or concerns about the PR that removes the incoming channel.

fujita:
Can you send fixes in a different pull request?

For the FSM redesign, my initial goal is to eliminate the functions that are specific to each BGP state. Then we can reduce duplication (e.g., the code getting AdminState).

maxime-peim:
What do you mean by `eliminate the functions that are specific to each BGP state`?
Most of fixes are needed after some refactor.
Also it seems that the refactor reveal an issue that is not covered when two active peers try to actively connect and get a connection collision. GoBGP should implement: https://datatracker.ietf.org/doc/html/rfc4271#page-35 but it is not the case.

Edit: concerning removing state specific functions, do you want a single select case, and for each case execute it only if it is allowed in this state?

fujita:
> What do you mean by `eliminate the functions that are specific to each BGP state`? Most of fixes are needed after some refactor. Also it seems that the refactor reveal an issue that is not covered when two active peers try to actively connect and get a connection collision. GoBGP should implement: https://datatracker.ietf.org/doc/html/rfc4271#page-35 but it is not the case.

You mean that GoBGP needs to handle two connections concurrently, incoming and outgoing? Then, right. I know. We need two BGP state for incoming and outgoing.
 
> Edit: concerning removing state specific functions, do you want a single select case, and for each case execute it only if it is allowed in this state?

Yes.

maxime-peim:
> You mean that GoBGP needs to handle two connections concurrently, incoming and outgoing? Then, right. I know. We need two BGP state for incoming and outgoing.

Seems like why some test are taking long to get a peering established (and why with the refactor they never get established. Running some locally, I can see that both peer close the connexion coming from the other side)

> Yes.

I don't know if it is better. You will end up creating a function per event instead of per state.

fujita:
> I don't know if it is better. You will end up creating a function per event instead of per state.

I'll try and see how it works.

maxime-peim:
> > I don't know if it is better. You will end up creating a function per event instead of per state.
> 
> I'll try and see how it works.

do you think it will ease the fsm by inverting the switch cases?
is there something that you want to fix that I don't capture in my refactor?
is it worth for me to continue working on that?

fujita:
I've not checked the details of your work yet but I know that you have file per BGP state. I'll like to try a different design and then let me compare both.

Besides these, are there any other areas you think could be improved?

pkg/peering/definitions.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+type FSMOutgoingMsg struct {
+	Paths        []*table.Path
+	Notification *bgp.BGPMessage
+	StayIdle     bool
+}
+
+type AdminStateOperation struct {
+	State         oc.AdminState
```
fujita: Please don't use openconfig package internally.
I believe it was a major design mistake to use the data structures generated from OpenConfig internally.
The reasons are:

- We can’t use optimal data structures like atomic.
- Keeping up with changes in the OpenConfig specification requires large modifications.

Those structures should only be used for the purpose of loading configuration files.

maxime-peim: I agree but the other solution is to duplicate everything from the openconfig structure

fujita: Unfortunately, you're right. That's what we have to do.

fujita: Using the gRPC-generated structures is also an option. It would give us control over changes, but we wouldn’t be able to use optimal data structures.

maxime-peim: Let me write a go script that generate the appropriate data structure then

fujita: How we use optimal data structures with that approach? For example, we could use atomic for message stats.

maxime-peim: I think it should be done in another PR as a second step

fujita: Do you mean temporarily adding code that uses the OpenConfig structures? If so, I’d prefer not to go in that direction.

maxime-peim: let me see what I can do

maxime-peim: I thought about it again. I will revert my change on using the open config values in the PR so we can merge it. I will open another one later with other structures to use internally

fujita: No need to add that to the history. Would you mind keeping the pull request small — maybe around 5 commits? If there’s more work, feel free to open another PR. Reviewing more than 20 commits at once gets pretty tough.

PR #2996, opened 2025-07-01
multiples fixes

maxime-peim:


pkg/server/server.go
```diff
@@ -4406,7 +4406,16 @@ func (s *BgpServer) WatchEvent(ctx context.Context, r *api.WatchEventRequest, fn
...
+			case <-ctx.Done():
+				return
 			case ev := <-w.Event():
+				// we might receive an event after the context is done
+				// so we need to check the context again
+				select {
+				case <-ctx.Done():
+					return
+				default:
+				}
```
fujita, line 4416: Right after this check, the context might be canceled. Why we need to check again here?

mpeim: The `select` statement choose randomly a case if they are both available.
However, there, you really don't want to continue watching when the context is done.
Therefore you need to recheck if the context is done before processing the event and quit accordingly.

Take as and example the watch loop line 77 in TestMetrics, do you want to process possibly an other event that will trigger the closing of the channel, even though you clearly stopped the working context?
```go
watchCtx, watchCancel := context.WithCancel(context.Background())
stateCh := make(chan struct{})
err = s.WatchEvent(watchCtx, &api.WatchEventRequest{Peer: &api.WatchEventRequest_Peer{}}, func(r *api.WatchEventResponse, _ time.Time) {
	if peer := r.GetPeer(); peer != nil {
		if peer.Type == api.WatchEventResponse_PeerEvent_TYPE_STATE && peer.Peer.State.SessionState == api.PeerState_SESSION_STATE_ESTABLISHED {
			watchCancel()
			close(stateCh)
		}
	}
})
```

fujita: > Therefore you need to recheck if the context is done before processing the event and quit accordingly.

Even if we do select <- ctx.Done() before handling the event, there is still a possibility that the context gets cancelled right after the check, but before the event is processed — because cancellation is asynchronous by nature.
The check reduces the likelihood of processing after cancellation, but cannot completely eliminate it.

That's my point.

fujita: Just to clarify, I'm fine with reducing the likelihood of processing after cancellation. I think that ctx.Err() is a bit simpler though.

```
select {
			case <-ctx.Done():
				return
			case ev := <-w.Event():
				// we might receive an event after the context is done
				// so we need to check the context again
				if ctx.Err() != nil {
				    return
				}
```

mpeim: > > Therefore you need to recheck if the context is done before processing the event and quit accordingly.
> 
> Even if we do select <- ctx.Done() before handling the event, there is still a possibility that the context gets cancelled right after the check, but before the event is processed — because cancellation is asynchronous by nature. The check reduces the likelihood of processing after cancellation, but cannot completely eliminate it.
> 
> That's my point.

I totally agree but that is not the point of my changes. Here I want to prioritise watcher cancellation when we have the guarantee that it happen (as in my example).

mpeim: > Just to clarify, I'm fine with reducing the likelihood of processing after cancellation. I think that ctx.Err() is a bit simpler though.
> 
> ```
> select {
> 			case <-ctx.Done():
> 				return
> 			case ev := <-w.Event():
> 				// we might receive an event after the context is done
> 				// so we need to check the context again
> 				if ctx.Err() != nil {
> 				    return
> 				}
> ```

Indeed, will update to do like that

fujita:
Thanks!

PR #2997, opened 2025-07-01
fix: properly stop bgp (pre fsm refactor)

maxime-peim:
'handlefsmMsg' function sole possible purpose was when the handler context is cancelled, which only happen if the 'deleteNeighbor' function is called. so the responsability of ending properly the fsm is moved back to the peer with 'stopFSMHandler'. this allow us to quit properly the 'Serve' goroutine.

Should solve: https://github.com/osrg/gobgp/issues/3030

### DIFF
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -194,6 +194,9 @@ func TestFSMLoopMetrics(t *testing.T) {
 	require.NoError(err)
 	defer s.StopBgp(context.Background(), &api.StopBgpRequest{})
 
+	// wait to ensure we started BGP
+	time.Sleep(1 * time.Second)
+
 	// StartBgp counts as single management operation
 	metrics, err = registry.Gather()
 	require.NoError(err)


fujita:
> this allow us to quit properly the 'Serve' goroutine.

What does `quit properly` mean? This fixes a bug? Yeah, I know the stop routine is a mess and we need to clean up it though.

maxime-peim:
> > this allow us to quit properly the 'Serve' goroutine.
> 
> What does `quit properly` mean? This fixes a bug? Yeah, I know the stop routine is a mess and we need to clean up it though.

I mean that we actually stop the Serve function. Even when calling StopBgp does not makes the goroutine stops which is the point

fujita:
> > > this allow us to quit properly the 'Serve' goroutine.
> > 
> > 
> > What does `quit properly` mean? This fixes a bug? Yeah, I know the stop routine is a mess and we need to clean up it though.
> 
> I mean that we actually stop the Serve function. Even when calling StopBgp does not makes the goroutine stops which is the point

Did you mean that StopBGP() should wait for all the goroutines to finish?

maxime-peim:
No, but at least for `Serve` to end. Otherwise, it will loop indefinitely and what should it does?

fujita:
> No, but at least for `Serve` to end. Otherwise, it will loop indefinitely and what should it does?

I'd like to mimic the semantics of http.Serve() and http.Shutdown() in the standard net/http package.
That said, I think the current Stop() function should be renamed to Shutdown() to better reflect its behavior.

maxime-peim:
> > No, but at least for `Serve` to end. Otherwise, it will loop indefinitely and what should it does?
> 
> I'd like to mimic the semantics of http.Serve() and http.Shutdown() in the standard net/http package. That said, I think the current Stop() function should be renamed to Shutdown() to better reflect its behavior.

I don't get it then: (from https://pkg.go.dev/net/http#Server.Shutdown)
```
When Shutdown is called, [Serve](https://pkg.go.dev/net/http#Serve), [ListenAndServe](https://pkg.go.dev/net/http#ListenAndServe), and [ListenAndServeTLS](https://pkg.go.dev/net/http#ListenAndServeTLS) immediately return [ErrServerClosed](https://pkg.go.dev/net/http#ErrServerClosed). Make sure the program doesn't exit and waits instead for Shutdown to return.
```
then if you want to mimic what it is done by the http package, the Serve function should return.

pkg/server/fsm.go
```diff
@@ -274,7 +274,7 @@ func newFSM(gConf *oc.Global, pConf *oc.Neighbor, logger log.Logger) *fsm {
 		gConf:                gConf,
 		pConf:                pConf,
 		state:                bgp.BGP_FSM_IDLE,
-		outgoingCh:           channels.NewInfiniteChannel(),
+		outgoingCh:           make(chan any, 1024),
 		connCh:               make(chan net.Conn, 1),
```
fujita: This means trying to send to this channel could block? If so, we can't do this because if the send to the channel blocks, it could potentially stall the entire BGP daemon.

mpeim: Let me find another solution for the race condition then

fujita: The second commit looks fine. Feel free to create a pull request for that commit.

mpeim: I won't create a seperate PR, the commits solve the issue of properly stopping GoBGP

fujita:
The first 4 commits didn't pass CI, right? Given that this PR is a bug fix, I'd like to review those 4 commits individually and proceed with merging them.

pkg/server/server.go
```diff
@@ -1659,8 +1659,6 @@ func (s *BgpServer) handleFSMMessage(peer *peer, e *fsmMsg) {
 			}
 		}
 
-		cleanInfiniteChannel(peer.fsm.outgoingCh)
-		peer.fsm.outgoingCh = channels.NewInfiniteChannel()
```
fujita, line 1663: Why you remove them? When a peer is down, there some messages in the channel. If you don't remove them, once the peer becomes established again, they will be sent.
At least, I think we need to clean up during the state transition from Established to Idle.

fujita:
The following comment in the first commit:
>However, since we cancel the 'sendMessageLoop' context when returning
from the 'established' function,

Where the context is canceled?

pkg/server/fsm.go
```diff
@@ -1826,37 +1826,40 @@ func (h *fsmHandler) sendMessageloop(ctx context.Context, wg *sync.WaitGroup) er
...
+				}
+				if err := send(m.Notification); err != nil {
+					return
+				}
+			}
+		default:
+		}
+	}
+
 	for {
```
fujita, line 1858: Why moved? for readability?

pkg/server/server.go
```diff
@@ -134,6 +133,11 @@ type BgpServer struct {
 	uuidMap      map[string]uuid.UUID
 	logger       log.Logger
 	timingHook   FSMTimingHook
+	// manage lifecycle of the server
+	isServing     atomic.Bool
+	shutdownWG    *sync.WaitGroup
+	runningCtx    context.Context
+	runningCancel context.CancelFunc
```
fujita, line 138: Why do we need runningCtx stuff? We just wait for all the peers to stop?

mpeim:
> The first 4 commits didn't pass CI, right? Given that this PR is a bug fix, I'd like to review those 4 commits individually and proceed with merging them.

On the opposite, they pass the CI

fujita:
> > The first 4 commits didn't pass CI, right? Given that this PR is a bug fix, I'd like to review those 4 commits individually and proceed with merging them.
> 
> On the opposite, they pass the CI

Oops, I meant that the first 4 commits passed CI. Could we merge those 4 commits without adding new commits?

pkg/server/server.go
mpeim: I kind of agree but on the other hand those messages should have been send before stopping.
I think there is a much deeper issue: when you quit the established state, you stop the sendMessageLoop but you expect some paths to be sent out to the peer. This will never happen because we push the update to the outgoingCh that is not used anymore.
I have a fix for that

fujita: > I kind of agree but on the other hand those messages should have been send before stopping.

I'm not talking about a situation where we can choose when to stop the peer. I'm simply talking about a situation where messages remain in the outgoingCh when the remote peer disconnects the TCP connection for some reason. Then the peer becomes established again, these messages will be sent wrongly.

pkg/server/fsm.go
mpeim: yes

fujita:
Why you merge commits into one? The differences from the previous code become completely unclear, making it impossible for me to review and merge.

pkg/server/server.go
mpeim: Me too. For example: you don't receive a keepalive before the hold timer, then you may need to send path updates/notification to the peer to indicate that graceful restart mechanism is starting.
you need to drain the channel and send all the message that are in there, not throw them away

pkg/server/server.go
mpeim: to control the liveliness of the `Serve` function

pkg/server/server.go
fujita: We have to handle graceful restart differently (currently, GoBGP doesn't though). If gr isn't enabled, in such case, GoBGP sends a notification and clean up the messages in the channel.

pkg/server/fsm.go
fujita: Please don't mix changes for different purposes. If you improve the readability, please do in a different pull request.

pkg/server/server.go
fujita: I didn't quite follow your explanation. Could you please elaborate a bit more?

mpeim: Stopping BGP = quitting the `Serve` function. hence i use a context that I can cancel to quit the `Serve` function

mpeim: otherwise, how do stop the serve goroutine?

mpeim:
> Why you merge commits into one? The differences from the previous code become completely unclear, making it impossible for me to review and merge.

I needed to re-split them differently

pkg/server/server.go
mpeim: you don't get it: when `handleFSMMessage` is called, we propagate some routes to the peer (e.g. drops etc). if you just clear the outgoing channel you never send those messages though the socket

fujita:
> > Why you merge commits into one? The differences from the previous code become completely unclear, making it impossible for me to review and merge.
> 
> I needed to re-split them differently

Please write a commit message, why we need that change.

pkg/server/server.go
fujita: Sorry, I don't understand. If I remember correctly, it fixed the bug that wrong messages are sent when a peer becomes established again. I guess that you say that if we clear the channel, a notification will not send. It's a different bug. You need to create a different pull request.

mpeim:
> > > Why you merge commits into one? The differences from the previous code become completely unclear, making it impossible for me to review and merge.
> > 
> > 
> > I needed to re-split them differently
> 
> Please write a commit message, why we need that change.

done

fujita:
> > > > Why you merge commits into one? The differences from the previous code become completely unclear, making it impossible for me to review and merge.
> > > 
> > > 
> > > I needed to re-split them differently
> > 
> > 
> > Please write a commit message, why we need that change.
> 
> done

Hmm, I can't compile the first commit.
```
bee:~/git/gobgp$ go build ./cmd/gobgpd
# github.com/osrg/gobgp/v4/pkg/server
pkg/server/peer.go:653:39: undefined: sync
pkg/server/fsm.go:1205:5: h.msgCh undefined (type *fsmHandler has no field or method msgCh)
pkg/server/fsm.go:1210:5: h.msgCh undefined (type *fsmHandler has no field or method msgCh)
pkg/server/fsm.go:1283:4: h.msgCh undefined (type *fsmHandler has no field or method msgCh)
pkg/server/fsm.go:1336:21: h.msgCh undefined (type *fsmHandler has no field or method msgCh)
pkg/server/fsm.go:1555:4: h.msgCh undefined (type *fsmHandler has no field or method msgCh)
pkg/server/fsm.go:1614:21: h.msgCh undefined (type *fsmHandler has no field or method msgCh)
pkg/server/fsm.go:1851:21: too many arguments in call to h.callback
	have (*fsmMsg, bool)
	want (*fsmMsg)
pkg/server/fsm.go:2022:3: undefined: drainBufferedChannel
pkg/server/fsm.go:2073:9: fsm.h.sendPendingMessages undefined (type *fsmHandler has no field or method sendPendingMessages)
pkg/server/fsm.go:2073:9: too many errors
```
The code need to build and run properly after each patch in a pull request. Otherwise, debugging like `git biscet` doesn't work.

mpeim:
> The code need to build and run properly after each patch in a pull request. Otherwise, debugging like `git biscet` doesn't work.

I agree, I split it too quickly. I reverted it

fujita:
Looks like you added new commits again (unit test updates and watcher fix). Once you create a pull request, please only change the parts that were requested during the review. 

It’s getting difficult for me to review and merge this if new commits keep being added.
Could you let me know once the pull request includes only the commits that are truly necessary to fix the BGP shutdown issue?
I’ll hold off on reviewing it until then.

mpeim:
> Looks like you added new commits again (unit test updates and watcher fix). Once you create a pull request, please only change the parts that were requested during the review.
> 
> It’s getting difficult for me to review and merge this if new commits keep being added. Could you let me know once the pull request includes only the commits that are truly necessary to fix the BGP shutdown issue? I’ll hold off on reviewing it until then.

I updated the tests because of the way you wanted to stop `recv` / `send` function (with SetReadDeadline). It was not supported by the custom MockConnection so it needed to be updated.

I make changes based on the discussion we have here, and it is also hard for me to put changes in to the right place when you request changes. yesterday I found that small things that were added based here and there, and were not in the right commits or needed new commits, hence why I moved things around.

Those are the commits needed to be able to shutdown bgp, so please review them but keep in mind that it is as hard for me to keep with something coherent

fujita:
> > Looks like you added new commits again (unit test updates and watcher fix). Once you create a pull request, please only change the parts that were requested during the review.
> > It’s getting difficult for me to review and merge this if new commits keep being added. Could you let me know once the pull request includes only the commits that are truly necessary to fix the BGP shutdown issue? I’ll hold off on reviewing it until then.
> 
> I updated the tests because of the way you wanted to stop `recv` / `send` function (with SetReadDeadline). It was not supported by the custom MockConnection so it needed to be updated

Previously, you simply worked around it, you didn't?

mpeim:
> > > Looks like you added new commits again (unit test updates and watcher fix). Once you create a pull request, please only change the parts that were requested during the review.
> > > It’s getting difficult for me to review and merge this if new commits keep being added. Could you let me know once the pull request includes only the commits that are truly necessary to fix the BGP shutdown issue? I’ll hold off on reviewing it until then.
> > 
> > 
> > I updated the tests because of the way you wanted to stop `recv` / `send` function (with SetReadDeadline). It was not supported by the custom MockConnection so it needed to be updated
> 
> Previously, you simply worked around it, you didn't?

Yes but just adding a void `SetReadDeadline` to `MockConnection` does not work at all!
you end up with tests failing because they don't exit a `Read` call

fujita:
> > > > Looks like you added new commits again (unit test updates and watcher fix). Once you create a pull request, please only change the parts that were requested during the review.
> > > > It’s getting difficult for me to review and merge this if new commits keep being added. Could you let me know once the pull request includes only the commits that are truly necessary to fix the BGP shutdown issue? I’ll hold off on reviewing it until then.
> > > 
> > > 
> > > I updated the tests because of the way you wanted to stop `recv` / `send` function (with SetReadDeadline). It was not supported by the custom MockConnection so it needed to be updated
> > 
> > 
> > Previously, you simply worked around it, you didn't?
> 
> Yes but just adding a void `SetReadDeadline` to `MockConnection` does not work at all! you end up with tests failing because the don't exit a `Read` call

Please make sure all tests pass locally before creating a pull request with minimum changes.
If you make significant code changes during the review process, it becomes impossible to review and merge the pull request.

maxime-peim:
Can you please review, the current CI failure has nothing to do with the PR and passed on the previous run:
https://github.com/osrg/gobgp/actions/runs/16517444714/job/46711438623#step:4:8
```
Reading package lists...
E: Read error - read (5: Input/output error)
E: Read error - read (5: Input/output error)
E: The package lists or status file could not be parsed or opened.
```

fujita:
> Referenc



> Can you please review, the current CI failure has nothing to do with the PR and passed on the previous run: https://github.com/osrg/gobgp/actions/runs/16517444714/job/46711438623#step:4:8
> 
> ```
> Reading package lists...
> E: Read error - read (5: Input/output error)
> E: Read error - read (5: Input/output error)
> E: The package lists or status file could not be parsed or opened.
> ```

Looks like failure to load the docker image. I suspect that it's related with CI infrastructure issues (hopefully temporary).

maxime-peim:
it is ready to be reviewed

fujita:
I will not review or merge a pull request including 9 commits.
A pull request should contain only a single commit. This means the commit must pass CI. While a pull request might exceptionally have 2-3 commits, 9 is out of the question.

For example, the first commit is independent and meaningful on its own—it simply replaces a custom hack with the standard implementation, unrelated to the BGP shutdown bug. This commit alone would be suitable for a pull request.
Any commit that is meaningful and independent by itself can be submitted as a separate pull request.

Please split your work.

maxime-peim:
> I will not review or merge a pull request including 9 commits. A pull request should contain only a single commit. This means the commit must pass CI. While a pull request might exceptionally have 2-3 commits, 9 is out of the question.
> 
> For example, the first commit is independent and meaningful on its own—it simply replaces a custom hack with the standard implementation, unrelated to the BGP shutdown bug. This commit alone would be suitable for a pull request. Any commit that is meaningful and independent by itself can be submitted as a separate pull request.
> 
> Please split your work.

Know what you want. When I merged all the commits into one you weren't happy either.

There is 3 commits that are bug fixes for the CI that I have encountered during development of the PR:
ea15843f0f8441d87b684047f513968823b22fe0
1f2ca1e43da001fa67f67a624240b1f0534779ac
a2d4db40884f3e6e1ab3657361ec70c5292eabad

One commit is there because you asked me why I was still using the reflection in the `Serve` function, so I removed it:
9da26b47e2b33747d84f9d18f7ec9781df6a95df

The test commit is there to properly implement the `SetReadDeadline` function needed to shutdown the server the way you wanted.
2a1f6195f1e83c16ce1b950f704e2850539d4b10

Each commit compiles and implements in an incremental way the feature, or solve an issue encountered during development. It is split into small, managable and reviewable parts. GitHub permits to review commit per commit.

Now if I do open a PR per commit what would be the difference or the benefits for you and me?

maxime-peim:
If you have such rules in mind please state them from the beginning

fujita:
> If you have such rules in mind please state them from the beginning

https://github.com/fujita/gobgp/commit/db1babc0ccbff823c96287904991c93d9b0c557c

I thought I had communicated this repeatedly, but it seems that the special handling caused some confusion, which unlikely breaks the fundamental BGP code. Sorry about that. I've written up some guidelines, so please let me know if anything is unclear.

Regarding this pull request, if you prefer, I can extract some of the commits and create separate pull requests and merge them.

PR #2998, opened 2025-07-02
feat: Allow graceful restart in StopBgpRequest

rastislavs:
Before this change, it was not possible to stop `BgpServer` while allowing Graceful Restart to kick in on the remote peers, as we send a NOTIFICATION message which terminates GR to each peer upon `Stop()`. This change allows to skip sending the NOTIFICATION messages to GR-enabled peers if the `allow_graceful_restart` flag is set to true. Keeps the existing behavior if the flag is not set.

### DIFF
--- a/api/gobgp.pb.go
+++ b/api/gobgp.pb.go
@@ -1403,9 +1403,11 @@ func (*StartBgpResponse) Descriptor() ([]byte, []int) {
 }
 
 type StopBgpRequest struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	state protoimpl.MessageState `protogen:"open.v1"`
+	// Allows the Graceful Restart procedure on the remote peers by not sending a NOTIFICATION message to GR-enabled peers.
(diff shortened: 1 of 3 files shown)


fujita:
I think the functionality itself is good, but the Stop() function, like Serve(), should be called even when gRPC is not used. It feels a bit odd to take a gRPC request struct as an argument.

@mpeim Any thoughts?

rastislavs:
> I think the functionality itself is good, but the Stop() function, like Serve(), should be called even when gRPC is not used. It feels a bit odd to take a gRPC request struct as an argument.
> 
> @mpeim Any thoughts?

I get your point, I could omit the change in `Stop()` and have this available only via `StopBgp()`. The only disadvantage is that it won't be possible to stop the `apiServer` if `StopBgp()` needs to be used instead of `Stop()` in order to pass the `StopBgpRequest` - but maybe we can expose another stop method for it as a follow-up?

https://github.com/osrg/gobgp/blob/c8ed70dfa3130a9cb3c05ac311ec0cc8e57b0597/pkg/server/server.go#L175-L177

mpeim:
> I think the functionality itself is good, but the Stop() function, like Serve(), should be called even when gRPC is not used. It feels a bit odd to take a gRPC request struct as an argument.
> 
> @mpeim Any thoughts?

Well indeed I also find that passing the gRPC struct via the `Stop` method is not ideal

rastislavs:
Ack, removed the args from the `Stop()` function.

fujita:
Pushed, thanks.

PR #2999, opened 2025-07-02
fix(test): fix new waitState

maxime-peim:
by starting the watcher after adding the peer, we don't have the guaranty to receive the state event.
so it is needed to start the watcher before and wait for it after adding the peer.

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -22,6 +22,7 @@ import (
 	"runtime"
 	"slices"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/server/server_test.go
```diff
@@ -221,12 +222,14 @@ func TestListPolicyAssignment(t *testing.T) {
 }
 
 //nolint:errcheck // WatchEvent won't return an error here
-func waitState(s *BgpServer, state api.PeerState_SessionState, expectedFamilies ...bgp.Family) {
-	stateCh := make(chan any)
+func waitState(s *BgpServer, state api.PeerState_SessionState, expectedFamilies ...bgp.Family) *sync.WaitGroup {
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+
```
fujita, line 228: Why do you use WaitGroup instead of channel? I'm fine with that though.

fujita: > Why do you use WaitGroup instead of channel? I'm fine with that though.

Assuming you think that using a WaitGroup is better because it makes the intention of the code clearer.

fujita:
Thanks!

PR #3002, opened 2025-07-06
chore: remove invoke run for testing

maxime-peim:
the invoke library is only used to run somme commands. it prompts a lot of useless debugging logs when a test fails.


maxime-peim:
some tests seems to be really flaky

fujita:
Thanks!

PR #3003, opened 2025-07-07
chore: update go to 1.24.5

maxime-peim:
remove the osrg/quagga image for testing which is really old, and use golang images instead


maxime-peim:
how can we update `osrg/quagga` image @fujita ?

fujita:
> how can we update `osrg/quagga` image @fujita ?

on Dockerhub? I think that I could or we could move to GitHub registry.
However, the CI code depends a specific version of quagga. If you upgrade the version, you might also need to update https://github.com/osrg/gobgp/blob/master/test/lib/quagga.py

go.mod
```diff
@@ -1,63 +1,60 @@
 module github.com/osrg/gobgp/v4
 
+go 1.24.3
```
nplanelcisco: why not 1.24.5 it's as been released recently

fujita: Yeah, sounds better. 1.24.5 includes security fixes.

maxime-peim: yes, sorry, thought 1.24.3 was latest

maxime-peim:
@fujita updating some packages seems to reveal connection detection issues

fujita:
Ok, let's hold this until implementing connection collision.

fujita:
> Ok, let's hold this until implementing connection collision.

I'll try to implement it.

fujita:
These CI failures are not related to collision detection? The CI just succeeded with the same changes.
https://github.com/fujita/gobgp/actions/runs/16405676531

fujita:
> These CI failures are not related to collision detection? The CI just succeeded with the same changes. https://github.com/fujita/gobgp/actions/runs/16405676531

I created a pull request and the CI succeeded.
https://github.com/osrg/gobgp/pull/3024

Seems that 1.24.5 includes security fixes so I'll merge the pull request if there are no objections.

fujita:
> > Ok, let's hold this until implementing connection collision.
> 
> I'll try to implement it.

https://github.com/fujita/gobgp/tree/fsm-fix

I've been trying to fix bugs but the basic design was finished.

@maxime-peim @nplanelcisco

mpeim:
> > These CI failures are not related to collision detection? The CI just succeeded with the same changes. https://github.com/fujita/gobgp/actions/runs/16405676531
> 
> I created a pull request and the CI succeeded. #3024
> 
> Seems that 1.24.5 includes security fixes so I'll merge the pull request if there are no objections.

I am not at ease with that... 

> > > Ok, let's hold this until implementing connection collision.
> > 
> > 
> > I'll try to implement it.
> 
> https://github.com/fujita/gobgp/tree/fsm-fix
> 
> I've been trying to fix bugs but the basic design was finished.
> 
> @maxime-peim @nplanelcisco

Same here: https://github.com/maxime-peim/gobgp/tree/dev/fsm-collision

fujita:
> > > These CI failures are not related to collision detection? The CI just succeeded with the same changes. https://github.com/fujita/gobgp/actions/runs/16405676531
> > 
> > 
> > I created a pull request and the CI succeeded. #3024
> > Seems that 1.24.5 includes security fixes so I'll merge the pull request if there are no objections.
> 
> I am not at ease with that...

What's the problem? The CI passed.
 
> > > > Ok, let's hold this until implementing connection collision.
> > > 
> > > 
> > > I'll try to implement it.
> > 
> > 
> > https://github.com/fujita/gobgp/tree/fsm-fix
> > I've been trying to fix bugs but the basic design was finished.
> > @maxime-peim @nplanelcisco
> 
> Same here: https://github.com/maxime-peim/gobgp/tree/dev/fsm-collision

I need to check carefully but I want to avoid code duplication across states. Many states seem to handle similar events (e.g., graceful restart timer expiration) with nearly identical logic.

mpeim:
> > > > These CI failures are not related to collision detection? The CI just succeeded with the same changes. https://github.com/fujita/gobgp/actions/runs/16405676531
> > > 
> > > 
> > > I created a pull request and the CI succeeded. #3024
> > > Seems that 1.24.5 includes security fixes so I'll merge the pull request if there are no objections.
> > 
> > 
> > I am not at ease with that...

Well this PR does the same thing. Just rebased on top of master. There is no need to create another PR.


> > Same here: https://github.com/maxime-peim/gobgp/tree/dev/fsm-collision
> 
> I need to check carefully but I want to avoid code duplication across states. Many states seem to handle similar events (e.g., graceful restart timer expiration) with nearly identical logic.

Those are few lines, and they simplify the reading of the code.
I saw your version, I really don't want to reverse the two layers of switch/case (first layer: the state, second layer: the event), and to have a split brain that handle events (the `Serve` function, and the state go routine).
In your case you have the `Serve` function with a lot of nesting which, at least for me, is not readable.

Overall it complexifies *a lot* the code to avoid few lines duplication.

fujita:
> > > > > These CI failures are not related to collision detection? The CI just succeeded with the same changes. https://github.com/fujita/gobgp/actions/runs/16405676531
> > > > 
> > > > 
> > > > I created a pull request and the CI succeeded. #3024
> > > > Seems that 1.24.5 includes security fixes so I'll merge the pull request if there are no objections.
> > > 
> > > 
> > > I am not at ease with that...
> 
> Well this PR does the same thing. Just rebased on top of master. There is no need to create another PR.

Sure, if you rebase, I'll merge.

> > > Same here: https://github.com/maxime-peim/gobgp/tree/dev/fsm-collision
> > 
> > 
> > I need to check carefully but I want to avoid code duplication across states. Many states seem to handle similar events (e.g., graceful restart timer expiration) with nearly identical logic.
> 
> Those are few lines, and they simplify the reading of the code. I saw your version, I really don't want to reverse the two layers of switch/case (first layer: the state, second layer: the event), and to have a split brain that handle events (the `Serve` function, and the state go routine). In your case you have the `Serve` function with a lot of nesting which, at least for me, is not readable.
> 
> Overall it complexifies _a lot_ the code to avoid few lines duplication.

I have a different perspective. For me, data duplication is acceptable. For example, I think CapMap and RFMap could be duplicated between Peer and FSM. Since they are created once and never updated, acquiring locks per packet would negatively impact performance.

However, logic duplication should be avoided. If avoiding that makes the code more complex, that's necessary complexity for me. I believe existing OSS implementations like BIRD and FRR follow similar design patterns, and I don't consider this logic to be overly complex.

maxime-peim:
> I have a different perspective. For me, data duplication is acceptable. For example, I think CapMap and RFMap could be duplicated between Peer and FSM. Since they are created once and never updated, acquiring locks per packet would negatively impact performance.

I am not sure to get it. Last week you told us you were concern about memory usage in path for example, but data duplication is not an issue?
Both CapMap and RFMap are updated when you get a new established peering. They could change overtime.
Moreover, CapMap can be updated when you handle an error and disable a family.
Also I would add that, today, you already are locking the entire fsm to get access to those values.

> However, logic duplication should be avoided. If avoiding that makes the code more complex, that's necessary complexity for me. 

That is not correct. The only kind of event that comes for each states are :
- the fsm context cancellation
- a connection reception
- a transition request
- a graceful restart timer expiration
- an admin state change

This represent less that **20 lines of codes**, which are not always duplicated (since we don't handle them the same way in each state, e.g. we send a notification when the fsm is administratively shut down is some cases)

> I believe existing OSS implementations like BIRD and FRR follow similar design patterns, and I don't consider this logic to be overly complex.

I don't say the design is complex, just that the code in your changes is.

fujita:
> > I have a different perspective. For me, data duplication is acceptable. For example, I think CapMap and RFMap could be duplicated between Peer and FSM. Since they are created once and never updated, acquiring locks per packet would negatively impact performance.
> 
> I am not sure to get it. Last week you told us you were concern about memory usage in path for example, but data duplication is not an issue? Both CapMap and RFMap are updated when you get a new established peering. They could change overtime. Moreover, CapMap can be updated when you handle an error and disable a family. Also I would add that, today, you already are locking the entire fsm to get access to those values.

Yes, memory usage is important to me, but consuming u64 per Path Attribute versus consuming small maps per peer are completely different matters. Consider use cases where we hold multiple full routes.

You're right about CapMap. While the current implementation does lock the entire FSM, partial locking wouldn't improve performance.

> > However, logic duplication should be avoided. If avoiding that makes the code more complex, that's necessary complexity for me.
> 
> That is not correct. The only kind of event that comes for each states are :
> 
> * the fsm context cancellation
> * a connection reception
> * a transition request
> * a graceful restart timer expiration
> * an admin state change
> 
> This represent less that **20 lines of codes**, which are not always duplicated (since we don't handle them the same way in each state, e.g. we send a notification when the fsm is administratively shut down is some cases)
> 
> > I believe existing OSS implementations like BIRD and FRR follow similar design patterns, and I don't consider this logic to be overly complex.
> 
> I don't say the design is complex, just that the code in your changes is.

We have different perspectives here.

maxime-peim:
Lets discuss it on Thursday then

PR #3006, opened 2025-07-08
bgp,mrt: Replace ImplicitPrefix with HeaderContext for accurate Famil…

fujita:
…y handling

Replace the misleading ImplicitPrefix concept with HeaderContext structure that explicitly holds both Family and prefix information from RIB Entry Headers. This change addresses fundamental limitations in Family derivation and aligns terminology with RFC 6396.

Problems addressed:

1. Inaccurate Family derivation**: AddrPrefixInterface.AFI()/SAFI() methods cannot reliably provide correct Family information for complex address families where the same prefix structure may represent different address families.

2. Non-RFC terminology**: "ImplicitPrefix" is not a term used in RFC 6396. The RFC describes using "RIB Entry Header" context to provide AFI/SAFI/NLRI information that is omitted from the wire format (Section 4.3.4).

### DIFF
--- a/pkg/packet/mrt/mrt.go
+++ b/pkg/packet/mrt/mrt.go
@@ -391,7 +391,7 @@ type RibEntry struct {
 
 var errNotAllRibEntryBytesAvailable = errors.New("not all RibEntry bytes are available")
 
-func (e *RibEntry) DecodeFromBytes(data []byte, prefix ...bgp.AddrPrefixInterface) ([]byte, error) {
+func (e *RibEntry) DecodeFromBytes(data []byte, family bgp.Family, prefix ...bgp.AddrPrefixInterface) ([]byte, error) {
 	if len(data) < 8 {
 		return nil, errNotAllRibEntryBytesAvailable
 	}
@@ -419,7 +419,7 @@ func (e *RibEntry) DecodeFromBytes(data []byte, prefix ...bgp.AddrPrefixInterfac
 		case 0:
 			err = p.DecodeFromBytes(data)
 		case 1:
-			err = p.DecodeFromBytes(data, &bgp.MarshallingOption{ImplicitPrefix: prefix[0]})
+			err = p.DecodeFromBytes(data, &bgp.MarshallingOption{HeaderPrefix: &bgp.HeaderContext{Family: family, Prefix: prefix[0]}})
 		default:
 			return nil, fmt.Errorf("only one prefix should be used")
 		}
@@ -436,7 +436,7 @@ func (e *RibEntry) DecodeFromBytes(data []byte, prefix ...bgp.AddrPrefixInterfac
 	return data, nil
 }
 
-func (e *RibEntry) Serialize(prefix ...bgp.AddrPrefixInterface) ([]byte, error) {
+func (e *RibEntry) Serialize(family bgp.Family, prefix ...bgp.AddrPrefixInterface) ([]byte, error) {
 	pbuf := make([]byte, 0)
 	totalLen := 0
 	for _, pattr := range e.PathAttributes {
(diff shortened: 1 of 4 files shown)


pkg/packet/mrt/mrt.go
```diff
@@ -535,7 +536,7 @@ func (u *Rib) DecodeFromBytes(data []byte) error {
 		e := &RibEntry{
 			isAddPath: u.isAddPath,
 		}
-		data, err = e.DecodeFromBytes(data, prefix)
+		data, err = e.DecodeFromBytes(data, family, prefix)
```
nplanelcisco, line 539: if the `Safi` is important to keep (unicast vs multicast), 
why not saving `isMulticast bool` in `IP{,v6}AddrPrefix` types via `NewPrefixFromFamily()`  ?   (or the `Safi` directly as it's uint8)
This would avoid the option passing though all the code, ...

pkg/packet/mrt/mrt_test.go
```diff
@@ -335,8 +335,8 @@ func FuzzDecodeFromBytes(f *testing.F) {
 		(&MRTHeader{}).DecodeFromBytes(data)
 		(&Peer{}).DecodeFromBytes(data)
 		(&PeerIndexTable{}).DecodeFromBytes(data)
-		(&RibEntry{}).DecodeFromBytes(data)
-		(&RibEntry{isAddPath: true}).DecodeFromBytes(data)
+		(&RibEntry{}).DecodeFromBytes(data, bgp.RF_IPv4_UC)
+		(&RibEntry{isAddPath: true}).DecodeFromBytes(data, bgp.RF_IPv4_UC)
```
nplanelcisco, line 339: as the safi is explicit now for the RIB, could add these lines to cover the fuzzing
```
		(&RibEntry{}).DecodeFromBytes(data, bgp.RF_IPv4_MC)
		(&RibEntry{isAddPath: true}).DecodeFromBytes(data, bgp.RF_IPv4_MC)
		(&RibEntry{}).DecodeFromBytes(data, bgp.RF_IPv6_UC)
		(&RibEntry{isAddPath: true}).DecodeFromBytes(data, bgp.RF_IPv4_UC)
		(&RibEntry{}).DecodeFromBytes(data, bgp.RF_IPv6_MC)
		(&RibEntry{isAddPath: true}).DecodeFromBytes(data, bgp.RF_IPv6_MC)
		```

mpeim:
What RFC 6396 specifies is the following:
There is one exception to the encoding of BGP attributes for the BGP
MP_REACH_NLRI attribute (BGP Type Code 14) [[RFC4760](https://datatracker.ietf.org/doc/html/rfc4760)]. Since the AFI,
SAFI, and NLRI information is already encoded in the RIB Entry Header
or RIB_GENERIC Entry Header, only the Next Hop Address Length and
Next Hop Address fields are included. The Reserved field is omitted.
The attribute length is also adjusted to reflect only the length of
the Next Hop Address Length and Next Hop Address fields.

The AFI/SAFI is already in the Rib header, there is no need to add it another time in the MpReachNlri attribute as we already can derived its AFI/SAFI from the header.
The MarshallingOptions shouldn't exists at all

pkg/packet/mrt/mrt.go
fujita: I want to remove the code to get AFI/SAFI from AddrPrefixInterface and pass family explicitly everywhere. Then I reimplement IPAddrPrefixDefault with netip to use it for both ipv4 and ipv6. Then I remove IPv6AddrPrefix, LabeledVPNIPv6AddrPrefix, LabeledIPv6AddrPrefix, Encapv6NLRI.

fujita: Yeah, we could save something like `isMulticast`.  Then we might need to save something different in the future. I prefer to use Family explicitly.

fujita:
> The AFI/SAFI is already in the Rib header, there is no need to add it another time in the MpReachNlri attribute as we already can derived its AFI/SAFI from the header. The MarshallingOptions shouldn't exists at all

Yeah, Rib header has family but how do we pass the family to the parser for path attributes?

mpeim:
wouldn't it be clearer to remove the HeaderContext, parse the MpReachNLRI attribute with only the nexthop and complete in the rib entry deserialisation?

mpeim:
Also, there is no need to pass it for serialisation

PR #3007, opened 2025-07-08
improvement: MP REACH NLRI (de)serialisation

maxime-peim:
do not pass the whole family and NLRI to all path
attributes.


pkg/packet/mrt/mrt.go
```diff
@@ -408,24 +408,34 @@ func (e *RibEntry) DecodeFromBytes(data []byte, family bgp.Family, prefix ...bgp
...
+
+		// RFC 6396 4.3.4
+		mp, ok := p.(*bgp.PathAttributeMpReachNLRI)
+		if ok && len(prefix) == 0 {
+			return nil, fmt.Errorf("prefix is not provided for MP_REACH_NLRI")
+		} else if ok {
+			mp.AFI = prefix[0].AFI()
+			mp.SAFI = prefix[0].SAFI()
+			mp.Value = []bgp.AddrPrefixInterface{prefix[0]}
+		}
```
fujita, line 437: I prefer to initialize all fields when constructing the object, rather than modifying them later. That's why I put family and prefix to `MarshallingOption`. But that's just my personal preference. We're using Go after all, so if you think this approach is better, I'm happy to go with it.

pkg/packet/bgp/bgp.go
```diff
@@ -31,26 +31,22 @@ import (
...
-	Prefix AddrPrefixInterface
-}
-
 type MarshallingOption struct {
-	AddPath      map[Family]BGPAddPathMode
-	Attributes   map[BGPAttrType]bool
-	HeaderPrefix *HeaderContext
+	AddPath    map[Family]BGPAddPathMode
+	Attributes map[BGPAttrType]bool
+	MRT        bool
```
fujita, line 38: `Mrt` would be more idiomatic in Go than `MRT`?

pkg/packet/mrt/mrt.go
maxime-peim: Above all, it is simplier. You don't have to pass the family everywhere.

pkg/packet/bgp/bgp.go
maxime-peim: No: https://go.dev/wiki/CodeReviewComments#initialisms
as it is today in the MRT package, `MRT` is written in all-caps as it should

pkg/packet/mrt/mrt.go
```diff
@@ -408,24 +408,34 @@ func (e *RibEntry) DecodeFromBytes(data []byte, family bgp.Family, prefix ...bgp
...
 		}
+		err = p.DecodeFromBytes(data, options)
 		if err != nil {
 			return nil, err
 		}
+
+		// RFC 6396 4.3.4
+		mp, ok := p.(*bgp.PathAttributeMpReachNLRI)
+		if ok && len(prefix) == 0 {
+			return nil, fmt.Errorf("prefix is not provided for MP_REACH_NLRI")
```
nplanelcisco, line 432: it's pretty good as we detect miss configuration now :+1:

fujita:
Breaks flowspec CI? I'll merge once this passes the CI.

mpeim:
> Breaks flowspec CI? I'll merge once this passes the CI.

fixed

fujita:
> > Breaks flowspec CI? I'll merge once this passes the CI.
> 
> fixed

Looks like the unit test failed.

fujita:
Thanks!

PR #3010, opened 2025-07-09
fix: MP REACH NLRI without NLRI

maxime-peim:


pkg/packet/bgp/bgp.go
```diff
@@ -11803,15 +11803,14 @@ func (p *PathAttributeMpReachNLRI) String() string {
 }
 
 func NewPathAttributeMpReachNLRI(nexthop string, nlri []AddrPrefixInterface) *PathAttributeMpReachNLRI {
```
nplanelcisco, line 11805: would be better to change to optional field `NewPathAttributeMpReachNLRI(nexthop string, nlri ...AddrPrefixInterface)`

pkg/packet/bgp/vpls_test.go
```diff
@@ -64,13 +64,13 @@ func Test_VPLSNLRI_decoding(t *testing.T) {
...
+	m1 := &PathAttributeMpReachNLRI{}
 	err := m1.DecodeFromBytes(buf)
 	require.NoError(t, err)
 
-	ns := make([]AddrPrefixInterface, 0)
-	ns = append(ns, NewVPLSNLRI(NewRouteDistinguisherTwoOctetAS(65017, 104), 1, 1, 8, 800000))
-	m2 := NewPathAttributeMpReachNLRI("192.0.2.7", ns)
+	rd := NewRouteDistinguisherTwoOctetAS(65017, 104)
+	nlri := NewVPLSNLRI(rd, 1, 1, 8, 800000)
+	m2 := NewPathAttributeMpReachNLRI("192.0.2.7", []AddrPrefixInterface{nlri})
```
nplanelcisco, line 73: so you can do `NewPathAttributeMpReachNLRI("192.0.2.7")`

pkg/packet/bgp/bgp.go
nplanelcisco: could you change the prototype to optional
`NewPathAttributeMpReachNLRI(nexthop string, nlri ...AddrPrefixInterface)`

This would benefit the API, example line 73 (vpls_test.go)
`m2 := NewPathAttributeMpReachNLRI("192.0.2.7", nlri)`

mpeim: Sure!

fujita:
I think that the first argument of NewPathAttributeMpReachNLRI() should be family. Even if NLRI is empty, the family should be always set properly (e.g., EOR).

Maybe we should update the next hop argument as well. Right now, we can only handle a limited set of nexthop types. It's difficult to represent all types of nexthops as a single string. Should we introduce a Nexthop type?

fujita:
> I think that the first argument of NewPathAttributeMpReachNLRI() should be family. Even if NLRI is empty, the family should be always set properly (e.g., EOR).

Ah, my mistake. EoR was an MpUnreach attribute. So MpReach always have NLRI(s). But I still prefer to have family because some AddrPrefixInterface types return wrong family (SAFI). Instead, you think that we should fix AddrPrefixInterface to return correct family?

maxime-peim:
> > I think that the first argument of NewPathAttributeMpReachNLRI() should be family. Even if NLRI is empty, the family should be always set properly (e.g., EOR).
> 
> Ah, my mistake. EoR was an MpUnreach attribute. So MpReach always have NLRI(s). But I still prefer to have family because some AddrPrefixInterface types return wrong family (SAFI). Instead, you think that we should fix AddrPrefixInterface to return correct family?

That should be addressed in another PR, the point of this one is to fix when there is no nlri

fujita:
> > > I think that the first argument of NewPathAttributeMpReachNLRI() should be family. Even if NLRI is empty, the family should be always set properly (e.g., EOR).
> > 
> > 
> > Ah, my mistake. EoR was an MpUnreach attribute. So MpReach always have NLRI(s). But I still prefer to have family because some AddrPrefixInterface types return wrong family (SAFI). Instead, you think that we should fix AddrPrefixInterface to return correct family?
> 
> That should be addressed in another PR, the point of this one is to fix when there is no nlri

Yeah, as soon as CI passes, I’ll merge this. My question is that we should fix AddPrefixInterface to return correct family or passing family to this function.

maxime-peim:
> > > > I think that the first argument of NewPathAttributeMpReachNLRI() should be family. Even if NLRI is empty, the family should be always set properly (e.g., EOR).
> > > 
> > > 
> > > Ah, my mistake. EoR was an MpUnreach attribute. So MpReach always have NLRI(s). But I still prefer to have family because some AddrPrefixInterface types return wrong family (SAFI). Instead, you think that we should fix AddrPrefixInterface to return correct family?
> > 
> > 
> > That should be addressed in another PR, the point of this one is to fix when there is no nlri
> 
> Yeah, as soon as CI passes, I’ll merge this. My question is that we should fix AddPrefixInterface to return correct family or passing family to this function.

In my point of view it is always better to fix things instead of finding a workaround for it, we will probably introduce other issues by passing the family etc...

fujita:
> > > > > I think that the first argument of NewPathAttributeMpReachNLRI() should be family. Even if NLRI is empty, the family should be always set properly (e.g., EOR).
> > > > 
> > > > 
> > > > Ah, my mistake. EoR was an MpUnreach attribute. So MpReach always have NLRI(s). But I still prefer to have family because some AddrPrefixInterface types return wrong family (SAFI). Instead, you think that we should fix AddrPrefixInterface to return correct family?
> > > 
> > > 
> > > That should be addressed in another PR, the point of this one is to fix when there is no nlri
> > 
> > 
> > Yeah, as soon as CI passes, I’ll merge this. My question is that we should fix AddPrefixInterface to return correct family or passing family to this function.
> 
> In my point of view it is always better to fix things instead of finding a workaround for it, we will probably introduce other issues by passing the family etc...

From my perspective, the current design where AddrPrefixInterface provides AFI() and SAFI() methods feels more like a workaround. That's because AddrPrefixInterface and Family don't have a one-to-one relationship.

maxime-peim:
@fujita can you merge this please?
I won't fix the design issue in this PR as it was not what I was addressing here

fujita:
Indeed, they are related but should be done in different pull requests.
Merged, thanks.

PR #3011, opened 2025-07-10
fix: remove *testing.T from MockConnection

maxime-peim:
*testing.T can be access by other means than the MockConnection
the locking mechanism does not protect it from race condition
https://github.com/osrg/gobgp/actions/runs/16195741427/job/45721651605#step:4:1314

### DIFF
--- a/pkg/server/fsm_test.go
+++ b/pkg/server/fsm_test.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"errors"
 	"net"
-	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


fujita:
Do you mean to partially revert ae0f5d114?

maxime-peim:
> Do you mean to partially revert ae0f5d114?

Yes, see the CI attached. The *testing.T is not made to be accessed concurrently

fujita:
Understood, thanks.

PR #3012, opened 2025-07-11
make CanImportToVrf faster

bor1go:
We discovered that when using a large VRF table (with about a thousand import RTs) and a high number of routes (hundreds of thousands), restarting GoBGP can take up to half an hour. Profiling revealed that most of the time is spent in the CanImportToVrf function, which iterates over all import RTs and, for each route, converts route targets to strings and regular expressions. I have attached an SVG file with a functional graph illustrating the problem.

This PR addresses the issue by:

- Adding fast hash calculation for route targets.
- Using a map[hash] instead of a list to store import RTs in the VRF table.
- Adding a test for VRF-table functionality.

honey-cgw-dc-klg2-restart-1


pkg/packet/bgp/bgp.go
```diff
@@ -11963,6 +12023,20 @@ type ExtendedCommunityInterface interface {
...
+	return false
+}
+
+func SubTypeEqual(ec ExtendedCommunityInterface, st ExtendedCommunityAttrSubType) bool {
+	if _, ecSubType := ec.GetTypes(); ecSubType == st {
+		return true
+	}
+	return false
+}
+
```
fujita: Better to make them ExtendedCommunityInterface's methods? No?

pkg/packet/bgp/bgp.go
```diff
@@ -2375,6 +2377,64 @@ type RouteTargetMembershipNLRI struct {
...
+	for _, rt := range s {
+		key, err := ExtCommRouteTargetKey(rt)
+		if err != nil {
+			return nil, err
+		}
+		m[key] = rt
+	}
+	return m, nil
+}
+
```
fujita: Should move them to the table package? Is there any chance this might be used by packages other than table?　packet/bgp is for parsing and serializing.

internal/pkg/table/policy.go
```diff
@@ -4053,24 +4051,21 @@ func NewRoutingPolicy(logger log.Logger) *RoutingPolicy {
...
+		// match only with transitive community. see RFC7153
+		if !bgp.IsTransitiveType(x) {
+			continue
+		}
+		key, err := bgp.ExtCommRouteTargetKey(x)
+		if err != nil {
+			continue
+		}
+		if _, found := v.ImportRt[key]; found {
+			return true
```
fujita, line 4065: What happens if a hash collision occurs?

bor1go: Hashes generated from RouteTarget Serialize() ([]byte, error) functions do not produce collisions, as it returns unique 8-byte slices for each route target.

pkg/packet/bgp/bgp.go
bor1go: Hmm... Is it possible in Go? I tried, but it didn't work.
Having a lot of duplicate implementations is not a good idea, is it?

pkg/packet/bgp/bgp.go
bor1go: We use it in the `peerFilterPath(peer *peer, path, old *table.Path) *table.Path` function in the server package to filter RTCs, but most of the usage is actually in the table package. I can move it there if you think that's better.

fujita: Please make functions private unless there's a clear need for them to be used elsewhere. Public functions should only be added when absolutely necessary.

bor1go: Well, I moved these functions to table.go, cleaned some of them up and made private several of them.

pkg/packet/bgp/bgp.go
fujita: Ok, implementing for all ExtendedCommunity attributes might doesn't make sense. Then please move them to table.

bor1go: Done. I’ve moved the functions to policy.go, where they are used, and made them local.

fujita:
Pushed, thanks.

PR #3015, opened 2025-07-11
refactor(fsm): phase 1

maxime-peim:


pkg/server/fsm.go
```diff
@@ -847,13 +847,23 @@ func buildopen(gConf *oc.Global, pConf *oc.Neighbor) *bgp.BGPMessage {
...
+	}()
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-done:
+		if err != nil {
+			return nil, err
+		}
+		return buf, nil
 	}
```
fujita: Creating a new goroutine and notifying completion via a channel on every socket read doesn’t seem optimal in terms of performance.
Isn’t it more common for the caller of recvMessageWithError (i.e., recvMessageLoop) to spawn the goroutine, which waits for context cancellation, and calls conn.SetReadDeadline(time.Now())?

mpeim: what you suggest isn't better, you also end up creating a new goroutine (which basically waits as here as we wait for io).

fujita: Your approach creates go routine every read. If recvMessageLoop creates goroutine to do such, we don't need to do such.

mpeim: please see https://github.com/osrg/gobgp/pull/2997, I fixed according to your suggestion. thanks!

Issue #3017, opened 2025-07-12
Some BGP-LS bugs


reshad-equinix:
I am using the gRPC API to GoBGP which is being monitored by GoBMP. While the basics work, here's a few bugs I found:
- Node LS attribute (e.g. when node name is specified in gRPC client) does not get stored in GoBGP if SRBlock isn't specified.
- Serialization of node router id fails due to length mismatch for IPv4 addresses.
- Serialization of node name fails
- Node Flags  not set properly.

There might be other BGP-LS issues but those are the ones I ran into. I seem to have a working fit and will open PR.

reshad-equinix:
https://github.com/osrg/gobgp/issues/3017

PR #3018, opened 2025-07-12
Some BGP-LS fixes

reshad-equinix:
- Move setting of LS node attribute outside of srblock check
- Fixed a few types
- Convert IPv4 addresses to avoid serialization length issue
- Fixed setting of node flags

Addresses https://github.com/osrg/gobgp/issues/3017


fujita:
Thanks! Pushed with lint error fixes.
If you could add a unit test related to this change, that would be great.

PR #3021, opened 2025-07-16
Disable transparent MPTCP on listener

YutaroHayakawa:
Since 1.24, Go stdlib enables MPTCP for listeners by default. This makes setsockopt to call MPTCP-specific handler. This breaks the listener that enables TCP MD5 because it is not compatible with MPTCP. This is a workaround for that.

See: https://github.com/golang/go/issues/74643 for more details.

### DIFF
--- a/internal/pkg/netutils/listener.go
+++ b/internal/pkg/netutils/listener.go
@@ -150,6 +150,14 @@ func NewTCPListener(logger log.Logger, address string, port uint32, bindToDev st
 		Control: listenControl(logger, bindToDev),
 	}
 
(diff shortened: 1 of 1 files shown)


fujita:
Thanks!
Is anyone still using TCP MD5? Maybe we should support QUIC and recommend it instead.

YutaroHayakawa:
Yeah, some people are still using that. In fact, we noticed this via user's report (https://github.com/cilium/cilium/issues/40478). I think [TCP-AO](https://docs.kernel.org/networking/tcp_ao.html) is a direct successor of the MD5, but QUIC might be interesting as well.

PR #3026, opened 2025-07-21
Improve dest.Calculate() and dest.GetMultiBestPath() performance

nplanelcisco:
Improve performance by 10% ish by caching path attributes via a path attributes hash if needed

```
Before
BenchmarkMultiPath/Benchmark_Calculate-12                 598702              1837 ns/op             376 B/op         14 allocs/op
BenchmarkMultiPath/Benchmark_GetMultiBestPath-12         3420961               343.5 ns/op             0 B/op          0 allocs/op

After
BenchmarkMultiPath/Benchmark_Calculate-12         	  637652	      1764 ns/op	     376 B/op	      14 allocs/op
BenchmarkMultiPath/Benchmark_GetMultiBestPath-12  	 4771671	       298.0 ns/op	       0 B/op	       0 allocs/op

### DIFF
--- a/internal/pkg/table/destination.go
+++ b/internal/pkg/table/destination.go
@@ -692,12 +692,11 @@ func compareByMED(path1, path2 *Path) *Path {
 		firstAS := func(path *Path) uint32 {
 			if asPath := path.GetAsPath(); asPath != nil {
 				for _, v := range asPath.Value {
-					segType := v.GetType()
 					asList := v.GetAS()
 					if len(asList) == 0 {
 						continue
 					}
-					switch segType {
+					switch v.GetType() {
 					case bgp.BGP_ASPATH_ATTR_TYPE_CONFED_SET, bgp.BGP_ASPATH_ATTR_TYPE_CONFED_SEQ:
 						continue
 					}
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -29,6 +29,7 @@ import (
 	"github.com/osrg/gobgp/v4/pkg/config/oc"
 	"github.com/osrg/gobgp/v4/pkg/log"
 	"github.com/osrg/gobgp/v4/pkg/packet/bgp"
+	"github.com/segmentio/fasthash/fnv1a"
 )
 
 const (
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: internal/pkg/table/destination.go [hunk 1]
```diff
@@ -692,11 +692,11 @@ func compareByMED(path1, path2 *Path) *Path {
 		firstAS := func(path *Path) uint32 {
 			if asPath := path.GetAsPath(); asPath != nil {
 				for _, v := range asPath.Value {
-					segType := v.GetType()
 					asList := v.GetAS()
 					if len(asList) == 0 {
 						continue
 					}
+					segType := v.GetType()
 					switch segType {
```
fujita: `switch v.GetType() {`
?

internal/pkg/table/path.go
```diff
@@ -1034,19 +1031,41 @@ func (lhs *Path) Equal(rhs *Path) bool {
...
 
+	lhsPathAttrs := lhs.GetPathAttrs()
+	rhsPathAttrs := rhs.GetPathAttrs()
+	if len(lhsPathAttrs) == 0 && len(rhsPathAttrs) == 0 {
+		return lhs.EqualBySourceAndPathID(rhs)
+	}
+
+	if len(lhsPathAttrs) != len(rhsPathAttrs) {
+		return false
+	}
```
fujita, line 1046: Could you add a comment explaining why it's better to do these comparisons first?

nplanelcisco: done

fujita:
The zebra-nht failure doesn't look like a fluke?

nplanelcisco:
> The zebra-nht failure doesn't look like a fluke?

Seems a regression, I will take a look tomorrow

nplanelcisco:
> > The zebra-nht failure doesn't look like a fluke?
> 
> Seems a regression, I will take a look tomorrow

@fujita it's fix now, you can review/approve

fujita:
Thank you.
Just out of curiosity, what caused the zebra tests to fail?

nplanelcisco:
> Thank you. Just out of curiosity, what caused the zebra tests to fail?

I forgot to invalidate the `path.hashAttr` if a `add/delAttr` is called. What cause the issue is that Metrics (MED) was wrong for the path, as value changed but not the hash, so compare would return the wrong path.

I will add some tests for corner case like this and other behavior check (GetListAttrs() in order, ...)

fujita:
> > Thank you. Just out of curiosity, what caused the zebra tests to fail?
> 
> I forgot to invalidate the `path.hashAttr` if a `add/delAttr` is called. What cause the issue is that Metrics (MED) was wrong for the path, as value changed but not the hash, so compare would return the wrong path.

Understood.
 
> I will add some tests for corner case like this and other behavior check (GetListAttrs() in order, ...)

Great, there should be tests other than the zebra tests that can catch that bug. Thanks!

PR #3029, opened 2025-07-22
fix: implement connection collision detection

maxime-peim:


fujita:
These CI failures don't seem related to collision detection. I prefer each commit to pass CI individually to maintain git bisect functionality for debugging. I might accept a few failing commits as an exception, but not many.

maxime-peim:
> These CI failures don't seem related to collision detection. I prefer each commit to pass CI individually to maintain git bisect functionality for debugging. I might accept a few failing commits as an exception, but not many.

I will split the last commit into smaller ones. However, implementing the collision detection mechanism fixes the CI:
https://github.com/maxime-peim/gobgp/actions/runs/16449986377/job/46492362891

PR #3031, opened 2025-07-25
Reduce randomness in active connect retry delay

fujita:
In the current implementation, the wait time before the next active connection attempt can be up to twice the user-specified connect-retry interval. For example, if connect-retry is set to 10 seconds, the actual wait time can range from 10 to 20 seconds due to an added random delay.

While the retry algorithm for active connections is not defined in the BGP RFCs and is implementation-specific, BIRD simply uses the specified interval without adding randomness.

This patch adjusts the logic to reduce the delay and make it closer to the configured interval. Specifically, it uses the connect-retry interval plus a random delay of up to 2000 milliseconds. The 2000 milliseconds was chosen arbitrarily, based on the assumption that most GoBGP use cases are within data centers, where long timeouts are undesirable.

This change is expected to reduce the likelihood of peer connection timeouts in CI environments.

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -38,7 +38,7 @@ import (
 )
 
 const (
-	minConnectRetryInterval = 1
+	minConnectRetryInterval = 2
 )
 
 type fsmStateReasonType uint8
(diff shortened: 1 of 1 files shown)


mpeim:
I agree that we shouldn't wait 2x the configured value, however the RFC states [1]:
```
   To minimize the likelihood that the distribution of BGP messages by a
   given BGP speaker will contain peaks, jitter SHOULD be applied to the
   timers associated with MinASOriginationIntervalTimer, KeepaliveTimer,
   MinRouteAdvertisementIntervalTimer, and ConnectRetryTimer.  A given
   BGP speaker MAY apply the same jitter to each of these quantities,
   regardless of the destinations to which the updates are being sent;
   that is, jitter need not be configured on a per-peer basis.

   The suggested default amount of jitter SHALL be determined by
   multiplying the base value of the appropriate timer by a random
   factor, which is uniformly distributed in the range from 0.75 to 1.0.
   A new random value SHOULD be picked each time the timer is set.  The
   range of the jitter's random value MAY be configurable.
```

So the value to be used should be something like: `time * (0.75 + rand(0, 0.25))`

[1]https://datatracker.ietf.org/doc/html/rfc4271#section-10

fujita:
Ok, make sense. Updated the commit.

maxime-peim:
lgtm

PR #3034, opened 2025-07-26
fix: update fsm test

maxime-peim:
update fsm test by using standard net.Pipe to correctly implement all net.Conn features (SetReadDeadline, etc...)


fujita:
Thanks a lot!

PR #3035, opened 2025-07-26
fix: miscellaneous fixes encountered in CI

maxime-peim:


fujita:
pushed, thanks a lot!

PR #3036, opened 2025-07-26
fix: remove incoming channels field and reflection

maxime-peim:
remove unused incoming channels from BGP server and use a plain select/case instead of reflection in the Serve function.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"net"
 	"net/netip"
-	"reflect"
 	"slices"
 	"strconv"
 	"strings"
@@ -116,7 +115,6 @@ type BgpServer struct {
 	apiServer    *server
 	bgpConfig    oc.Bgp
 	acceptCh     chan net.Conn
-	incomings    []*channels.InfiniteChannel
 	mgmtCh       chan *mgmtOp
 	policy       *table.RoutingPolicy
 	listeners    []*netutils.TCPListener
@@ -189,19 +187,6 @@ func (s *BgpServer) Stop() {
 
 	if s.apiServer != nil {
 		s.apiServer.grpcServer.Stop()
-	}
-}
-
-func (s *BgpServer) addIncoming(ch *channels.InfiniteChannel) {
-	s.incomings = append(s.incomings, ch)
-}
-
-func (s *BgpServer) delIncoming(ch *channels.InfiniteChannel) {
-	for i, c := range s.incomings {
-		if c == ch {
-			s.incomings = append(s.incomings[:i], s.incomings[i+1:]...)
-			return
-		}
 	}
 }
 
@@ -314,7 +299,7 @@ func (s *BgpServer) startFsmHandler(peer *peer) {
 			}
 
 			cleanInfiniteChannel(fsm.outgoingCh)
-			if s.shutdownWG != nil && len(s.incomings) == 0 {
+			if s.shutdownWG != nil {
 				s.shutdownWG.Done()
 			}
 			return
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -444,57 +433,27 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
...
-			cases[i] = reflect.SelectCase{
-				Dir:  reflect.SelectRecv,
-				Chan: reflect.ValueOf(s.incomings[i-firstPeerCaseIndex].Out()),
-			}
-		}
-
-		chosen, value, _ := reflect.Select(cases)
 		tStart := time.Now()
-		s.shared.mu.Lock()
-		switch chosen {
```
fujita, line 476: roaManager.HandleROAEvent doesn't need the mutex? Why you need to move it from here?

maxime-peim: before the `reflect.Select` was blocking until an event occurs, and you acquire the lock only when an event occurs. without a vanilla select/case, you cannot just lock before the select because you probably will be blocking until an event occurs, hence holding a lock and for nothing.

concerning ROA events, the BGP server and the ROA manager share nothing. so why locking the bgp server event loop when an ROA event will not compete to access memory.

fujita: > before the `reflect.Select` was blocking until an event occurs, and you acquire the lock only when an event occurs. without a vanilla select/case, you cannot just lock before the select because you probably will be blocking until an event occurs, hence holding a lock and for nothing.

Just to clarify — my question was about why the mutex operations were moved out of Serve(). Locking code shouldn't be changed just for reasons like making the code simpler or cleaner.

Please make the change in a straightforward manner, like this:

```
	s.listeners = make([]*netutils.TCPListener, 0, 2)

	for {
		tStart := time.Now()
		select {
		case op := <-s.mgmtCh:
			tWait := tStart.Sub(op.timestamp)
			s.shared.mu.Lock()
			s.handleMGMTOp(op)
			s.shared.mu.Unlock()
			s.timingHook.Observe(FSMMgmtOp, time.Since(tStart), tWait)
		case conn := <-s.acceptCh:
			// NOTE: it would be useful to use kernel metrics such as SO_TIMESTAMPING to record time we got
			// first SYN packet in TCP connection. For now we skip tWait for accept events, message/mgmt op
			// delays should be enough to analyze FSM loop.
			s.shared.mu.Lock()
			s.passConnToPeer(conn)
			s.shared.mu.Unlock()
			s.timingHook.Observe(FSMAccept, time.Since(tStart), 0)
		case ev := <-s.roaManager.ReceiveROA():
			tWait := tStart.Sub(ev.timestamp)
			s.shared.mu.Lock()
			s.roaManager.HandleROAEvent(ev)
			s.shared.mu.Unlock()
			s.timingHook.Observe(FSMROAEvent, time.Since(tStart), tWait)
		}
	}
``` 

> concerning ROA events, the BGP server and the ROA manager share nothing. so why locking the bgp server event loop when an ROA event will not compete to access memory.

Really? roaManager.HandleROAEvent might update roaTable. roaTable can be accessed in handleFSMMessage() (table policy evaluation).

I might misunderstand. If you think so, please create a new commit with a new pull request to remove the lock, with the commit message that explains why removing that lock should be safe.

A critical change such as removing a lock should be made explicitly in its own commit.  I don't think It's the right approach to sneak it into another commit like this.

fujita:
The unit test failure is fluke or related with this change?

maxime-peim:
This does not seem to be related. Can't you restart a test manually instead of restart the whole CI?

fujita:
Thanks!

Issue #3030, opened 2025-07-23
panic: sync: negative WaitGroup counter


rastislavs:
After bumping to the latest master version, hitting a new panic caused by negative `shutdownWG` WaitGroup in `BgpServer`.  Most likely caused by this change: https://github.com/osrg/gobgp/commit/f1de13828eb46b9f123292298caa473d1aac5b5d

It is happening in our tests, which stop the server right after removing a peer / updating a peer that is handled by delete+add:

```
panic: sync: negative WaitGroup counter

goroutine 88 [running]:
sync.(*WaitGroup).Add(0x4000a0c900?, 0x8?)
	/usr/local/go/src/sync/waitgroup.go:64 +0x104
sync.(*WaitGroup).Done(0x4000979200?)
	/usr/local/go/src/sync/waitgroup.go:89 +0x20
github.com/osrg/gobgp/v4/pkg/server.(*BgpServer).handleFSMMessage.(*BgpServer).startFsmHandler.func7(0x4000a0a480, 0x1)
	/github.com/osrg/gobgp/v4/pkg/server/server.go:319 +0x5d8
github.com/osrg/gobgp/v4/pkg/server.(*fsmHandler).loop(0x40005bc620, {0x3e38ea0, 0x4000b22370}, 0x0?)
	/github.com/osrg/gobgp/v4/pkg/server/fsm.go:2058 +0x91c
created by github.com/osrg/gobgp/v4/pkg/server.newFSMHandler in goroutine 84
	/github.com/osrg/gobgp/v4/pkg/server/fsm.go:410 +0x1d8
```

fujita:
@rastislavs The testing and bug report on the latest version are really helpful, thank you.
https://github.com/osrg/gobgp/pull/2997 will fix this, I think. I'm reviewing it now.

If possible, could you provide the exact procedure that might trigger the bug so we can add it to our unit tests?

rastislavs:
@fujita It was caught [by this Cilium unit-test](https://github.com/cilium/cilium/blob/1222c37dff49abf46de2e986faa1368835bd0415/pkg/bgpv1/manager/reconcilerv2/neighbor_test.go#L278-L351), which starts gobgp server, adds peers, updates and removes them and stops the server after it (only occasionally, not on every test run).

I can retest after the potential fix is merged.

fujita:
@rastislavs
Sorry for the long delay, but this should be fixed in the latest code. When you have some time, please check and let me know.

[...]

PR #3037, opened 2025-07-26
fix: properly stop bgp

maxime-peim:
https://github.com/osrg/gobgp/issues/3030

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -381,6 +381,8 @@ func (fsm *fsm) sendNotification(code, subType uint8, data []byte, msg string) (
 	return fsm.sendNotificationFromErrorMsg(e.(*bgp.MessageError))
 }
 
+type fsmCallback func(*fsmMsg)
+
 type fsmHandler struct {
 	fsm              *fsm
 	conn             net.Conn
@@ -391,24 +393,22 @@ type fsmHandler struct {
 	sentNotification *bgp.BGPMessage
 	ctx              context.Context
 	ctxCancel        context.CancelFunc
-	wg               *sync.WaitGroup
-	callback         func(*fsmMsg, bool)
+	callback         fsmCallback
 }
 
-func newFSMHandler(fsm *fsm, outgoing *channels.InfiniteChannel, callback func(*fsmMsg, bool)) *fsmHandler {
+func newFSMHandler(fsm *fsm, outgoing *channels.InfiniteChannel, wg *sync.WaitGroup, callback func(*fsmMsg)) *fsmHandler {
 	ctx, cancel := context.WithCancel(context.Background())
 	h := &fsmHandler{
 		fsm:              fsm,
 		stateReasonCh:    make(chan fsmStateReason, 2),
 		outgoing:         outgoing,
 		holdTimerResetCh: make(chan bool, 2),
-		wg:               &sync.WaitGroup{},
 		ctx:              ctx,
 		ctxCancel:        cancel,
 		callback:         callback,
 	}
-	h.wg.Add(1)
-	go h.loop(ctx, h.wg)
+	wg.Add(1)
+	go h.loop(ctx, wg)
 	return h
 }
 
@@ -1849,7 +1849,7 @@ func (h *fsmHandler) recvMessageloop(ctx context.Context, wg *sync.WaitGroup) er
 	for {
 		fmsg, err := h.recvMessageWithError()
 		if fmsg != nil && ctx.Err() == nil {
-			h.callback(fmsg, false)
+			h.callback(fmsg)
 		}
 		if err != nil {
 			return nil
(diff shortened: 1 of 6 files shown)


pkg/server/fsm.go
```diff
@@ -992,7 +993,9 @@ func (h *fsmHandler) recvMessageWithError() (*fsmMsg, error) {
 	}
 
 	headerBuf, err := readAll(h.conn, bgp.BGP_HEADER_LENGTH)
-	if err != nil {
+	if errors.Is(err, os.ErrDeadlineExceeded) {
+		return nil, nil
+	} else if err != nil {
```
fujita: Why this change is necessary?

pkg/server/fsm.go
```diff
@@ -1807,32 +1826,51 @@ func (h *fsmHandler) sendMessageloop(ctx context.Context, wg *sync.WaitGroup) er
...
-				if m.Notification != nil {
-					if m.StayIdle {
-						// current user is only prefix-limit
-						// fix me if this is not the case
-						_ = h.changeadminState(adminStatePfxCt)
-					}
-					if err := send(m.Notification); err != nil {
+			// send all remaining messages
+			// before returning
+			for {
```
fujita: I think writes might block for a long time (e.g., socket buffer is full), which seems inappropriate when the context is canceled?

pkg/server/fsm.go
maxime-peim: you don't want to send a state reason when you cancel the context (hence setting the read deadline).
you are already shutting down the peer

pkg/server/fsm.go
maxime-peim: Agreed, we should add a write deadline there. nevertheless, we are stopping manually the peer, we want to do that properly by sending all the messages/notifications needed.

fujita: Please add the code to set a write deadline to avoid waiting for a long time. It would be ideal to send the all but we need to set some deadline.

Also I think that we should send a notification first (I think that other BGP implementations do the same). We don't know if we call send all here so sending the notification should be the priority. Once the peer gets the notification, then closes the connection. So I'm not sure we need to send pending messages after the notification. If gr is enabled, we might be better to keep these messages until gr timer expires.

pkg/server/fsm.go
fujita: Please add that comment in either place.

pkg/server/fsm.go
fujita: Ah, we already removed the peer so gr configuration doesn't matter, I guess.

pkg/server/fsm.go
```diff
@@ -1807,32 +1830,51 @@ func (h *fsmHandler) sendMessageloop(ctx context.Context, wg *sync.WaitGroup) er
...
```
fujita: You don't explain why we have to send all remaining messages. When the peer was removed (or the bgp stopped), why we have to do?

Please write a comment about **why**, not what.

fujita:
Please don't add a new commit during review process.

maxime-peim:
> Please don't add a new commit during review process.

@fujita we discussed a matter (the WriteDeadline) that has *nothing* to do with the content of the previous commits.
So either I change the content and commit message of a previous commit, you won't be happy because I change the content of a commit, or, as I did, I add another commit on top with a clear message on what it addresses and left untouched the previously reviewed commits.

In both case you don't seems ok with that, so how should I address your comments?

maxime-peim:
I am okay merging all the commits into one, it would be harder for you to review and the logical changes would be mixed together, but it would follow your rules

fujita:
> > Please don't add a new commit during review process.
> 
> @fujita we discussed a matter (the WriteDeadline) that has _nothing_ to do with the content of the previous commits. So either I change the content and commit message of a previous commit, you won't be happy because I change the content of a commit, or, as I did, I add another commit on top with a clear message on what it addresses and left untouched the previously reviewed commits.
> 
> In both case you don't seems ok with that, so how should I address your comments?


There seems to be a misunderstanding.

`CONTRIBUTING.md` says
>Please treat your pull request as "frozen" once it has been created. Adding new commits or modifying the code unprompted makes review much harder. Only changes based on review comments should be made.

You cannot modify the code unprompted. But you can modify the code based on review comments. As I commented on the first commit, without write deadline, writing a socket might take long time. You can modify the first commit to avoid that.

If you interpreted that document differently than I intended, please send a pull request to improve that document.

To clarify another issue with the first commit, I don't understand **why** we have to send remaining messages. The peer is removed so even if we send update messages to a remote peer, the remote peer just drops them when the peer is down.
When the peer is removed, a notification is send to fsm via fsm.notificaiton channel, not h.outgoing channel.

Indeed, I might misunderstand. So please explain in a commit message why we have to send messages in h.outgoing channel when the peer is removed.

maxime-peim:
Done. Please see the first commit.

pkg/server/fsm.go
```diff
@@ -1807,32 +1822,53 @@ func (h *fsmHandler) sendMessageloop(ctx context.Context, wg *sync.WaitGroup) er
...
-						// current user is only prefix-limit
-						// fix me if this is not the case
-						_ = h.changeadminState(adminStatePfxCt)
-					}
-					if err := send(m.Notification); err != nil {
+			// send all remaining messages before
+			// returning. This is to be sure to send
+			// the notification and drop paths that are
+			// generally sent when the peer is brought down.
+			for {
```
fujita: The above code is executed only when the peer is removed. I looked into the BIRD and FRR implementations, and I understand that they send notifications immediately and don't send remaining messages. I believe that this is the expected behavior in such situations.

The notification should be de-configured, which doesn't exist in outgoing channel. It's queued in fsm.notification channel, which will be sent by goroutine which executes `established`. So sendMessageloop() should return immediately. Please remove the above change.

maxime-peim: > The above code is executed only when the peer is removed.

No, I updated the code so that the io functions context expires when we quit the `established` state.
The notification being pushed into the channel when the hold timer is expired has a chance of not being sent without that code.

maxime-peim: Let's recap that order of events:
- the hold timer expires
- you add a notification to the outgoing channel
- you cancel the context
- the sendMessageloop can exit without having sent the notification that is still in the channel.

What was done before:
- the hold timer expires
- you add a notification to the outgoing channel
- you wait for the sendMessageloop to send back a state reason
- you push an error to outgoing channel (**forcing all remaining messages to be sent**)
- the sendMessageloop ends once you reached the error in the channel

what I did achieve the same thing that was done before in a simpler way

maxime-peim: Either we move the notification sending out of the sending channel or it stays like I did.

fujita: > Let's recap that order of events:
> 
> * the hold timer expires

The commit title is 'when context is cancelled'

How hold timer is related with the title?

maxime-peim: you cancel the io context whenever you quit the function.

fujita: Ok, I was talking about context used in established() argument.

With this patch, please explain what will happen in order after deleteNeighbor() is called.

fujita: > Let's recap that order of events:
> 
> * the hold timer expires
> * you add a notification to the outgoing channel
> * you cancel the context
> * the sendMessageloop can exit without having sent the notification that is still in the channel.
> 
> What was done before:
> 
> * the hold timer expires
> * you add a notification to the outgoing channel
> * you wait for the sendMessageloop to send back a state reason
> * you push an error to outgoing channel (**forcing all remaining messages to be sent**)
> * the sendMessageloop ends once you reached the error in the channel
> 
> what I did achieve the same thing that was done before in a simpler way

Note that both behaviors are wrong. If GR isn't enabled, when a hold timer expired in established state, you send a notification immediately (IIRC, BIRD does that and discard pending messages).

I know that the current implementation about hold timer is wrong. But this pull request is about "fix: properly stop bgp". I think that it means that fixing StopBgp (essentially, deleteNeighbor).

maxime-peim: I am not talking about GR.

```go
case <-holdTimer.C:
	fsm.lock.RLock()
	fsm.logger.Warn("hold timer expired",
		log.Fields{
			"Topic": "Peer",
			"Key":   fsm.pConf.State.NeighborAddress,
			"State": fsm.state.String(),
		})
	fsm.lock.RUnlock()
	m := bgp.NewBGPNotificationMessage(bgp.BGP_ERROR_HOLD_TIMER_EXPIRED, 0, nil)
	h.outgoing.In() <- &fsmOutgoingMsg{Notification: m}
	fsm.lock.RLock()
	s := fsm.pConf.GracefulRestart.State
	fsm.lock.RUnlock()
	// Do not return hold timer expired to server if graceful restart is enabled
	// Let it fallback to read/write error or fsmNotificationSent handled above
	// Reference: https://github.com/osrg/gobgp/issues/2174
	if !s.Enabled {
		return bgp.BGP_FSM_IDLE, newfsmStateReason(fsmHoldTimerExpired, m, nil)
	}
```
You don't send the notification straight away. This is what I am telling you. You push the notification in the outgoing channel. Hence you have to make sure that this notification is sent before stopping the peer.

fujita: Yeah, I think that sending notification to outgoing is the wrong approach. We can directly send it to socket like we do when the fsm's context is canceled.

Why do you keep talking about hold timer expire? Why fixing deleteNeighbor() needs to fix hold timer?

maxime-peim: Properly stopping bgp also means behaving properly when a peer comes down, whatever the reason.

maxime-peim: I split that in another PR if you prefer

fujita: Yes, please. please don't create hold timer fix pr yet. We already got a bug report about deleteNeighbor() so that's the priority. Once the issue is fixed, we can start to work on holding timer.

nplanelcisco:
@fujita could you review this PR

fujita:
> @fujita could you review this PR

I think that the first change in this pull request is under reviewing.

https://github.com/osrg/gobgp/pull/3051

We have the bug report related with the this area. Would it be okay if I try to fix?
https://github.com/osrg/gobgp/issues/3030

mpeim:
this has been addressed in #3104

PR #3042, opened 2025-07-27
pkg/packet/bgp: use netip for IPAddrPrefix and IPv6AddrPrefix

fujita:
Replace uses of net.IP with netip.Prefix for IPAddrPrefix and IPv6AddrPrefix.

When GoBGP was started, the net/netip package did not exist and net.IP was the standard type for representing IP addresses. However, with the introduction of netip.Addr in Go 1.18, we now have a more efficient and type-safe alternative.

Although netip.Addr may use slightly more memory for IPv4 addresses (as it always reserves space for IPv6), it eliminates the need for memory allocation on the heap, which is expected to improve performance, especially in high-throughput paths.

This patch replaces uses of net.IP with netip.Prefix type implemented by using netip.Addr for IPAddrPrefix and IPv6AddrPrefix.

We should use netip for IpAddrPrefixDefault however it means that this patch becomes much larger since we also update all labeled addrprefix types. There will be some temporary code duplication, but I plan to gradually proceed with the replacement.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -1383,13 +1383,13 @@ func addrPrefixOnlySerialize(nlri AddrPrefixInterface) []byte {
 	switch T := nlri.(type) {
 	case *IPAddrPrefix:
 		b := make([]byte, 5)
-		copy(b, T.Prefix.To4())
-		b[4] = T.Length
+		copy(b, T.Prefix.Addr().AsSlice())
+		b[4] = uint8(T.Prefix.Bits())
 		return b
 	case *IPv6AddrPrefix:
 		b := make([]byte, 17)
-		copy(b, T.Prefix.To16())
-		b[16] = T.Length
+		copy(b, T.Prefix.Addr().AsSlice())
+		b[16] = uint8(T.Prefix.Bits())
 		return b
 	case *LabeledVPNIPAddrPrefix:
 		b := make([]byte, 13)
@@ -1542,15 +1542,61 @@ func (r *IPAddrPrefixDefault) MarshalJSON() ([]byte, error) {
 	})
 }
 
-type IPAddrPrefix struct {
-	IPAddrPrefixDefault
+// Once we convert all the users of IPAddrPrefixDefault to use IPAddrPrefixDefaultNetip,
+// this will be renamed.
+type IPAddrPrefixDefaultNetip struct {
+	PrefixDefault
+	Prefix  netip.Prefix
 	addrlen uint8
 }
 
-func (r *IPAddrPrefix) DecodeFromBytes(data []byte, options ...*MarshallingOption) error {
+func (r *IPAddrPrefixDefaultNetip) decodePrefix(data []byte, bitlen uint8) error {
 	if r.addrlen == 0 {
-		r.addrlen = 4
+		r.addrlen = net.IPv4len
 	}
+
+	if len(data) < 1 {
+		eCode := uint8(BGP_ERROR_UPDATE_MESSAGE_ERROR)
+		eSubCode := uint8(BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST)
+		return NewMessageError(eCode, eSubCode, nil, "prefix misses length field")
+	}
+
+	bytelen := (int(bitlen) + 7) / 8
+	if len(data) < bytelen {
+		return NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_INVALID_NETWORK_FIELD, nil, "network bytes is short")
+	}
+	if bitlen > r.addrlen*8 {
+		return NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_INVALID_NETWORK_FIELD, nil, "network bit length is too long")
+	}
+
+	var b [16]byte
+	copy(b[:], data[:bytelen])
+	// clear trailing bits in the last byte. rfc doesn't require
+	// this but some bgp implementations need this...
+	rem := bitlen % 8
+	if rem != 0 {
+		mask := 0xff00 >> rem
+		lastByte := b[bytelen-1] & byte(mask)
+		b[bytelen-1] = lastByte
+	}
+	addr, _ := netip.AddrFromSlice(b[:r.addrlen])
+	r.Prefix = netip.PrefixFrom(addr, int(bitlen))
+	return nil
+}
+
+func (r *IPAddrPrefixDefaultNetip) serializePrefix() []byte {
+	byteLen := (r.Prefix.Bits() + 7) / 8
+	buf := make([]byte, byteLen)
+	copy(buf, r.Prefix.Addr().AsSlice()[:byteLen])
+
+	return buf
+}
+
+type IPAddrPrefix struct {
+	IPAddrPrefixDefaultNetip
+}
+
+func (r *IPAddrPrefix) DecodeFromBytes(data []byte, options ...*MarshallingOption) error {
 	f := RF_IPv4_UC
 	if r.addrlen == 16 {
 		f = RF_IPv6_UC
@@ -1567,8 +1613,8 @@ func (r *IPAddrPrefix) DecodeFromBytes(data []byte, options ...*MarshallingOptio
 		eSubCode := uint8(BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST)
 		return NewMessageError(eCode, eSubCode, nil, "prefix misses length field")
 	}
-	r.Length = data[0]
-	return r.decodePrefix(data[1:], r.Length, r.addrlen)
+
+	return r.decodePrefix(data[1:], data[0])
 }
 
 func (r *IPAddrPrefix) Serialize(options ...*MarshallingOption) ([]byte, error) {
@@ -1584,12 +1630,9 @@ func (r *IPAddrPrefix) Serialize(options ...*MarshallingOption) ([]byte, error) 
 			return nil, err
 		}
 	}
-	buf = append(buf, r.Length)
-	pbuf, err := r.serializePrefix(r.Length)
-	if err != nil {
-		return nil, err
-	}
-	return append(buf, pbuf...), nil
+	buf = append(buf, uint8(r.Prefix.Bits()))
+
+	return append(buf, r.serializePrefix()...), nil
 }
 
 func (r *IPAddrPrefix) AFI() uint16 {
@@ -1601,18 +1644,36 @@ func (r *IPAddrPrefix) SAFI() uint8 {
 }
 
 func (r *IPAddrPrefix) Len(options ...*MarshallingOption) int {
-	return 1 + (int(r.Length)+7)/8
+	return 1 + (r.Prefix.Bits()+7)/8
 }
 
-func NewIPAddrPrefix(length uint8, prefix string) *IPAddrPrefix {
+func (r *IPAddrPrefix) String() string {
+	return r.Prefix.String()
+}
+
+func (p *IPAddrPrefix) Flat() map[string]string {
+	return map[string]string{
+		"Prefix":    p.Prefix.Addr().String(),
+		"PrefixLen": fmt.Sprintf("%d", p.Prefix.Bits()),
+	}
+}
+
+func (r *IPAddrPrefix) MarshalJSON() ([]byte, error) {
+	return json.Marshal(struct {
+		Prefix string `json:"prefix"`
+	}{
+		Prefix: r.Prefix.String(),
+	})
+}
+
+func NewIPAddrPrefix(bits uint8, prefix string) *IPAddrPrefix {
 	p := &IPAddrPrefix{
-		IPAddrPrefixDefault{
-			Length: length,
+		IPAddrPrefixDefaultNetip: IPAddrPrefixDefaultNetip{
+			addrlen: net.IPv4len,
 		},
-		4,
 	}
 	// TODO: pass the error to the caller
-	_ = p.decodePrefix(net.ParseIP(prefix).To4(), length, 4)
+	_ = p.decodePrefix(net.ParseIP(prefix).To4(), bits)
 	return p
 }
 
@@ -1628,25 +1689,16 @@ func (r *IPv6AddrPrefix) AFI() uint16 {
 	return AFI_IP6
 }
 
-func (r *IPv6AddrPrefix) String() string {
-	prefix := r.Prefix.String()
-	if isIPv4MappedIPv6(r.Prefix) {
-		prefix = "::ffff:" + prefix
-	}
-	return prefix + "/" + strconv.FormatUint(uint64(r.Length), 10)
-}
-
-func NewIPv6AddrPrefix(length uint8, prefix string) *IPv6AddrPrefix {
+func NewIPv6AddrPrefix(bits uint8, prefix string) *IPv6AddrPrefix {
 	p := &IPv6AddrPrefix{
 		IPAddrPrefix{
-			IPAddrPrefixDefault{
-				Length: length,
+			IPAddrPrefixDefaultNetip: IPAddrPrefixDefaultNetip{
+				addrlen: net.IPv6len,
 			},
-			16,
 		},
 	}
 	// TODO: pass the error to the caller
-	_ = p.decodePrefix(net.ParseIP(prefix), length, 16)
+	_ = p.decodePrefix(net.ParseIP(prefix).To16(), bits)
 	return p
 }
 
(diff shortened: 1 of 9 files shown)


maxime-peim:
> Although netip.Addr may use slightly more memory for IPv4 addresses (as it always reserves space for IPv6)

From https://pkg.go.dev/net/netip: 
Compared to the [net.IP](https://pkg.go.dev/net#IP) type, [Addr](https://pkg.go.dev/net/netip#Addr) type takes less memory, is immutable, and is comparable (supports == and being a map key). 

Also, `net.IP` also allocates memory for IPv6 when building an IPv4: https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/net/ip.go;l=53

> it eliminates the need for memory allocation on the heap

That is totally false. When the variable needs to outlive the scope of the function it is declared in, it is always declared on the heap.

pkg/packet/bgp/bgp.go
```diff
@@ -1565,8 +1566,36 @@ func (r *IPAddrPrefix) DecodeFromBytes(data []byte, options ...*MarshallingOptio
...
-	r.Length = data[0]
-	return r.decodePrefix(data[1:], r.Length, r.addrlen)
+	bitlen := data[0]
+	data = data[1:]
+
+	nlriError := func(msg string) error {
+		eCode := uint8(BGP_ERROR_UPDATE_MESSAGE_ERROR)
+		eSubCode := uint8(BGP_ERROR_SUB_INVALID_NETWORK_FIELD)
+		return NewMessageError(eCode, eSubCode, nil, msg)
+	}
```
maxime-peim: Why declaring a function?
Simply:
```go
NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_INVALID_NETWORK_FIELD, nil, msg)
```
where needed, as it is done everywhere else

pkg/packet/bgp/bgp.go
```diff
@@ -1565,8 +1566,36 @@ func (r *IPAddrPrefix) DecodeFromBytes(data []byte, options ...*MarshallingOptio
...
+	// this but some bgp implementations need this...
+	rem := bitlen % 8
+	if rem != 0 {
+		mask := 0xff00 >> rem
+		lastByte := b[bytelen-1] & byte(mask)
+		b[bytelen-1] = lastByte
+	}
+	addr, _ := netip.AddrFromSlice(b[:r.addrlen])
+	r.Prefix = netip.PrefixFrom(addr, int(bitlen))
+	return nil
```
maxime-peim: It should be moved in a separate function for future use with `netip`

pkg/packet/bgp/bgp.go
```diff
@@ -1582,11 +1611,13 @@ func (r *IPAddrPrefix) Serialize(options ...*MarshallingOption) ([]byte, error)
...
-	pbuf, err := r.serializePrefix(r.Length)
-	if err != nil {
-		return nil, err
-	}
+	bits := uint8(r.Prefix.Bits())
+	buf = append(buf, bits)
+
+	byteLen := (int(bits) + 7) / 8
+	pbuf := make([]byte, byteLen)
+	copy(pbuf, r.Prefix.Addr().AsSlice()[:byteLen])
```
maxime-peim: It should be moved in a separate function for future use with netip

pkg/packet/bgp/bgp.go
```diff
@@ -1599,19 +1630,48 @@ func (r *IPAddrPrefix) SAFI() uint8 {
...
+		}
+		buf := append([]byte{length}, addr.AsSlice()[:addrlen]...)
+		err := p.DecodeFromBytes(buf)
+		if err != nil {
+			return nil
+		}
+		return p
+	} else {
+		return &IPAddrPrefix{}
 	}
```
maxime-peim: to have less nested code:

```go
if prefix == "" {
	return &IPAddrPrefix{}
}

addr := netip.MustParseAddr(prefix)
addrlen := net.IPv4len
if addr.Is6() {
	addrlen = net.IPv6len
}

p := &IPAddrPrefix{
	addrlen: uint8(addrlen),
}
buf := append([]byte{length}, addr.AsSlice()[:addrlen]...)
err := p.DecodeFromBytes(buf)
if err != nil {
	return nil
}
return p
```

pkg/packet/bgp/bgp.go
```diff
@@ -1599,19 +1630,48 @@ func (r *IPAddrPrefix) SAFI() uint8 {
...
 }
 
 func NewIPAddrPrefix(length uint8, prefix string) *IPAddrPrefix {
-	p := &IPAddrPrefix{
-		IPAddrPrefixDefault{
-			Length: length,
-		},
-		4,
+	if prefix != "" {
+		addr := netip.MustParseAddr(prefix)
```
maxime-peim: you shouldn't use the `Must` version to avoid panic-ing

pkg/packet/bgp/bgp.go
```diff
@@ -1599,19 +1630,48 @@ func (r *IPAddrPrefix) SAFI() uint8 {
...
+		}
+
+		p := &IPAddrPrefix{
+			addrlen: uint8(addrlen),
+		}
+		buf := append([]byte{length}, addr.AsSlice()[:addrlen]...)
+		err := p.DecodeFromBytes(buf)
+		if err != nil {
+			return nil
+		}
```
maxime-peim: once the prefix decoding is moved to its own function, you should directly use it as it was done before

pkg/packet/bgp/bgp.go
```diff
@@ -1599,19 +1630,48 @@ func (r *IPAddrPrefix) SAFI() uint8 {
...
-	p := &IPAddrPrefix{
-		IPAddrPrefixDefault{
-			Length: length,
-		},
-		4,
+	if prefix != "" {
+		addr := netip.MustParseAddr(prefix)
+		addrlen := net.IPv4len
+		if addr.Is6() {
+			addrlen = net.IPv6len
```
maxime-peim: this is a big change, you allow ipv6 addresses to be decoded from there, which was not the case before

pkg/packet/bgp/bgp.go
```diff
@@ -1626,26 +1686,14 @@ func (r *IPv6AddrPrefix) AFI() uint16 {
...
-}
-
 func NewIPv6AddrPrefix(length uint8, prefix string) *IPv6AddrPrefix {
-	p := &IPv6AddrPrefix{
-		IPAddrPrefix{
-			IPAddrPrefixDefault{
-				Length: length,
-			},
-			16,
+	p := *NewIPAddrPrefix(length, prefix)
```
maxime-peim: you don't need to dereference the return value, but you should check that it is not nil.

pkg/packet/bgp/bgp.go
fujita: I thought that it was more readable but yeah probably not. I'll change.

pkg/packet/bgp/bgp.go
fujita: You mean it should be a separate function and it will be used other `IPAddrPrefix` structures like `LabeledVPNIPAddrPrefix`?
If so, as I wrote in the commit message, this structure will be `IPAddrPrefixDefault`, will be embedded in other structures. So no need to make this a separate function.

pkg/packet/bgp/bgp.go
fujita: Ditto.

pkg/packet/bgp/bgp.go
fujita: Yeah, I'll fix.
Note that I plan to make `NewIPAddrPrefix()` take only a valid string.

pkg/packet/bgp/bgp.go
fujita: NewSomePrefix APIs don't return an error. So for me, you must pass a valid string to those APIs.

pkg/packet/bgp/bgp.go
fujita: As I wrote above, I plan to make `NewIPAddrPrefix()` take only a valid string. So this function will not need DecodeFromBytes stuff. It just calls `netip.MustParseAddr` and then `PrefixFrom`.

Note that I plan to change the `DecodeFromBytes` API for AddrPrefix.
As a premise, I believe that the same struct should be used for both v6 and v4 like netip, and I'm planning to make that change.

Currently, we create an uninitialized `IPAddrPrefix`, set the addrlen for v6 cases, and then call `DecodeFromBytes`. Instead, I like to have something like `func IPAddrPrefixFromBytes(family, []byte) (IPAddrPrefix, error)`.  The API interface will be aligned with netip's API. Unfortunately, Go doesn't enforce RAII though.

`netip.Addr` has the information about ipv4 or v6 so we don't need addrlen stuff.

I'm still wondering how to handle EOR with the above API changes.

pkg/packet/bgp/bgp.go
fujita: Yeah, but this change will not break users' existing code (If so, I'll write something in a commit message or do in a separate commit). This change just allows temporarily `NewIPv6AddrPrefix()` to use this function.

pkg/packet/bgp/bgp.go
fujita: Ah, will fix.
About nil checking, as I wrote above, this function is supposed to panic with an invalid input.

fujita:
> > Although netip.Addr may use slightly more memory for IPv4 addresses (as it always reserves space for IPv6)
> 
> From https://pkg.go.dev/net/netip: Compared to the [net.IP](https://pkg.go.dev/net#IP) type, [Addr](https://pkg.go.dev/net/netip#Addr) type takes less memory, is immutable, and is comparable (supports == and being a map key).
> 
> Also, `net.IP` also allocates memory for IPv6 when building an IPv4: https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/net/ip.go;l=53

Ah, good. Then there are no disadvantages to migrating from net to netip.
 
> > it eliminates the need for memory allocation on the heap
> 
> That is totally false. When the variable needs to outlive the scope of the function it is declared in, it is always declared on the heap.

Hmm, I was talking about the number of heap allocations. Each use of net.IP may require one heap allocation.

```
type A struct {
   len uint32
   add netip.Addr
}

type B struct {
   len uint32
   addr net.IP
}
```
If we have something like above, there could be two heap allocations for struct B. On the other hand, struct A do one heap allocation. As you said, if net.IP always allocates enough space for v6, we could do one heap allocation for struct B too though.

pkg/packet/bgp/bgp.go
maxime-peim: It is a bit weird to update an other type to become the new default, would it be clearer to have a new type `IPAddrPrefixDefaultNetip` to be used incrementally in other structures?

pkg/packet/bgp/bgp.go
maxime-peim: I am not comfortable with that. If a user mistype a prefix from the CLI for example, I don't think it is a good idea to make the whole gobgp process panic. You should handle and report an error.

pkg/packet/bgp/bgp.go
maxime-peim: as stated before, panic-ing is not a good solution, errors happens and should be reported and not stop the whole process

pkg/packet/bgp/bgp.go
fujita: The current API is designed in that way. New*AddrPrefix() uses net.Parse() so I think that if you pass a bad input, you get Prefix in AddrPrefix is nil. Probably it will panic shortly.

We could change All New*AddrPrefix API to return error but I'm not sure it's a good idea. This API has two types of use cases.
- unit tests
- input from API

For unit tests cases, returning an error is just verbose. Including CLI, API input should be checked in the API layer and we do because the current API can't handle invalid input.

pkg/packet/bgp/bgp.go
fujita: As I wrote above, I believe that the current API panics with an invalid input. So I don't change the semantics.

pkg/packet/bgp/bgp.go
fujita: If you prefer, I'm fine with returning error. It should be done in a different commit because we change the API interface.

pkg/packet/bgp/bgp.go
fujita: Then once I finish, I remove IPAddrPrefixDefault and rename IPAddrPrefixDefaultNetip to IPAddrPrefixDefault?
Yeah, it might work. Let me try.

pkg/packet/bgp/bgp.go
maxime-peim:
```go
func NewIPAddrPrefix(length uint8, prefix string) *IPAddrPrefix {
	p := &IPAddrPrefix{
		IPAddrPrefixDefault{
			Length: length,
		},
		4,
	}
	// TODO: pass the error to the caller
	_ = p.decodePrefix(net.ParseIP(prefix).To4(), length, 4)
	return p
}
```
if `prefix` is not valid, you get a `nil` value from `ParseIP`, `To4` returns `nil` as well and an error is returned.
As I wrote in a previous PR, `TODO: pass the error to the caller`, we should handle that error to avoid panic. I don't think that explicitly panic is the right way to do it, you should pass the error to the caller where it should be handled or passed upper.

pkg/packet/bgp/bgp.go
maxime-peim: That is not true, the current does not panic directly, you assumed it will panic at some point but it will not in the `New*Prefix` function

fujita: > That is not true, the current does not panic directly, you assumed it will panic at some point but it will not in the `New*Prefix` function

Ok, I keep that semantics, it will panic at some point. From user point of view, I think that panicking immediately is better though.

pkg/packet/bgp/bgp.go
fujita: Ok, added this to the TODO list
https://github.com/osrg/gobgp/issues/3043

fujita:
@maxime-peim As you suggested, I go with the approach to add `IPAddrPrefixDefaultNetip` temporarily. As a result, I should have addressed all of your review comments.

pkg/packet/bgp/bgp.go
```diff
@@ -1540,15 +1540,61 @@ func (r *IPAddrPrefixDefault) MarshalJSON() ([]byte, error) {
 	})
 }
 
-type IPAddrPrefix struct {
-	IPAddrPrefixDefault
+// Once we convert all the users of IPAddrPrefixDefault to use IPAddrPrefixDefaultNetip,
+// this is renamed.
```
maxime-peim: `this will be renamed`

pkg/packet/bgp/bgp.go
```diff
@@ -1540,15 +1540,61 @@ func (r *IPAddrPrefixDefault) MarshalJSON() ([]byte, error) {
...
+
+	bytelen := (int(bitlen) + 7) / 8
+	if len(data) < bytelen {
+		return NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_INVALID_NETWORK_FIELD, nil, "network bytes is short")
 	}
+	if bitlen > r.addrlen*8 {
+		return NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_INVALID_NETWORK_FIELD, nil, "network bit length is too long")
+	}
+
+	var b [16]byte
```
maxime-peim, line 1572: why do you differ from `IPAddrPrefixDefault.decodePrefix` there?
since it is exactly the same code, and just the two last lines differ when setting the `Prefix` field to the right type, can't you create an underlying function that takes `data []byte, bitlen uint8, addrlen uint8` and returns the parsed buffer?
this would avoid a lot of code duplication

pkg/packet/bgp/bgp.go
```diff
@@ -1540,15 +1540,61 @@ func (r *IPAddrPrefixDefault) MarshalJSON() ([]byte, error) {
...
+// this is renamed.
+type IPAddrPrefixDefaultNetip struct {
+	PrefixDefault
+	Prefix  netip.Prefix
 	addrlen uint8
 }
 
-func (r *IPAddrPrefix) DecodeFromBytes(data []byte, options ...*MarshallingOption) error {
+func (r *IPAddrPrefixDefaultNetip) decodePrefix(data []byte, bitlen uint8) error {
 	if r.addrlen == 0 {
```
maxime-peim, line 1554: it is weird to check `addrlen` here and in `DecodeFromBytes`, both checks should be done in `DecodeFromBytes`

pkg/packet/bgp/bgp.go
fujita: Create an underlying function means creating the common code to void duplication? If so, that's exactly what we should not do. The point is that we should not change IPAddrPrefixDefault so that there is no risk to break the users of IPAddrPrefixDefault. We can convert step by step.

I could do the exact function but this patch should not change IPAddrPrefixDefault. In the end, the duplication will be removed.

pkg/packet/bgp/bgp.go
fujita: Oops, will fix.

pkg/packet/bgp/bgp.go
fujita: Why weird? IPAddrPrefixDefaultNetip owns addrlen so checking it inside the method is ok for me. Once we merge v6 and v4 structures, it will be removed. So it doesn't matter much.

pkg/packet/bgp/bgp.go
maxime-peim:
```go
func decodePrefix(data []byte, bitlen uint8, addrlen uint8) ([]byte, error) {
	bytelen := (int(bitlen) + 7) / 8
	if len(data) < bytelen {
		eCode := uint8(BGP_ERROR_UPDATE_MESSAGE_ERROR)
		eSubCode := uint8(BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST)
		return nil, NewMessageError(eCode, eSubCode, nil, "network bytes is short")
	}
	if bitlen > addrlen*8 {
		eCode := uint8(BGP_ERROR_UPDATE_MESSAGE_ERROR)
		eSubCode := uint8(BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST)
		return nil, NewMessageError(eCode, eSubCode, nil, "network bit length is too long")
	}
	b := make([]byte, addrlen)
	copy(b, data[:bytelen])
	// clear trailing bits in the last byte. rfc doesn't require
	// this but some bgp implementations need this...
	rem := bitlen % 8
	if rem != 0 {
		mask := 0xff00 >> rem
		lastByte := b[bytelen-1] & byte(mask)
		b[bytelen-1] = lastByte
	}
	return b, nil
}

func (r *IPAddrPrefixDefault) decodePrefix(data []byte, bitlen, addrlen uint8) error {
	b, err := decodePrefix(data, bitlen, addrlen)
	r.Prefix = b
	return err
}

func (r *IPAddrPrefixDefaultNetip) decodePrefix(data []byte, bitlen, addrlen uint8) error {
	b, err := decodePrefix(data, bitlen, addrlen)
	if err != nil {
		return err
	}
	addr, _ := netip.AddrFromSlice(b[:r.addrlen])
	r.Prefix = netip.PrefixFrom(addr, int(bitlen))
	return nil
}
```

The behavior of `IPAddrPrefixDefault` does not change and you avoid copying the same code

maxime-peim: This cannot breaks the usage of `IPAddrPrefixDefault` since you do exactly the same thing.

pkg/packet/bgp/bgp.go
maxime-peim:
```go
func (r *IPAddrPrefix) DecodeFromBytes(data []byte, options ...*MarshallingOption) error {
	f := RF_IPv4_UC
	switch r.addrlen {
		case 0:
			r.addrlen = net.IPv4len
		case net.IPv6len:
			f = RF_IPv6_UC
		default:
			return ...
	}

	if IsAddPathEnabled(true, f, options) {
		var err error
		data, err = r.decodePathIdentifier(data)
		if err != nil {
			return err
		}
	}
	if len(data) < 1 {
		eCode := uint8(BGP_ERROR_UPDATE_MESSAGE_ERROR)
		eSubCode := uint8(BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST)
		return NewMessageError(eCode, eSubCode, nil, "prefix misses length field")
	}
	r.Length = data[0]
	return r.decodePrefix(data[1:], r.Length, r.addrlen)
}
```

So you don't have to check and modify the addrlen value in decode prefix, and you keep the same signature for `decodePrefix` as it was before.

maxime-peim: See: 
```go
if r.addrlen == 0 {
	r.addrlen = 4
```
move from `DecodeFromBytes` to `decodePrefix` for no reason

pkg/packet/bgp/bgp.go
fujita: This is not the approach I want to pursue, as I explained above. With this approach, after converting all the users of IPAddrPrefixDefault, we need to do more than renaming. Once we remove old IPAddrPrefixDefault, there is no reason why we keep independent decodePrefix() so we better to incorporate decodePrefix() into IPAddrPrefixDefaultNetip decodePrefix() and I might make a mistake, which breaks all the users of IPAddrPrefixDefaultNetip.

Of course, there's still a possibility of making mistakes even with just renaming, but those mistakes are more likely to result in compile errors.

So I prefer to share nothing between IPAddrPrefixDefault and IPAddrPrefixDefaultNetip. In the end, all the duplication will be removed anyway.

maxime-peim: You'll just have to make the `decodePrefix` function a method of `(r *IPAddrPrefixDefaultNetip)`, this is not a big deal. But anyway do as you like, I still think it is a bad idea, you could introduce behavioral changes by note writing the same code when it should be

fujita: Not exactly same. For example, I drop the bitLen argument in serializePrefix() because netip.Prefix keeps bitLen itself. I adjusted the code for netip.Prefix.

PR #3044, opened 2025-07-29
fix: close the stream properly, otherwise the last few paths are dropped

maeglin89273:
Without properly closing the stream, gobgpd will lose some prefixes.


fujita:
Can you fix the lint error?

fujita:
Thanks!

PR #3051, opened 2025-07-30
fix: stop io loop when context is cancelled

maxime-peim:
If NegotiatedHoldTime is zero we would never be able to send a message as the write deadline would be set to now. Remove this write deadline and add one only when we are stopping the send goroutine.

In case of hold timer expiration, we need to send the notification straight away and avoid pushing it to the outgoing channel, since there is a race between the closing of the sendMessageloop and the sending of the notification.

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -17,6 +17,7 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
@@ -992,7 +993,13 @@ func (h *fsmHandler) recvMessageWithError() (*fsmMsg, error) {
 	}
 
 	headerBuf, err := readAll(h.conn, bgp.BGP_HEADER_LENGTH)
-	if err != nil {
+	if errors.Is(err, os.ErrDeadlineExceeded) {
+		// we set a read deadline when we cancel the FSM handler context,
+		// so this is expected when the FSM is shutting down.
+		// We dont' send a state reason here because the FSM is already
+		// shutting down.
+		return nil, nil
+	} else if err != nil {
 		sendToStateReasonCh(fsmReadFailed, nil)
 		return nil, err
 	}
@@ -1020,7 +1027,9 @@ func (h *fsmHandler) recvMessageWithError() (*fsmMsg, error) {
 	}
 
 	bodyBuf, err := readAll(h.conn, int(hd.Len)-bgp.BGP_HEADER_LENGTH)
-	if err != nil {
+	if errors.Is(err, os.ErrDeadlineExceeded) {
+		return nil, nil
+	} else if err != nil {
 		sendToStateReasonCh(fsmReadFailed, nil)
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/server/fsm.go
```diff
@@ -1843,31 +1858,40 @@ func (h *fsmHandler) sendMessageloop(ctx context.Context, wg *sync.WaitGroup) er
...
 
-	go h.sendMessageloop(ctx, &wg)
-	go h.recvMessageloop(ctx, &wg)
+	go h.sendMessageloop(ioCtx, wg)
+	go h.recvMessageloop(ioCtx, wg)
+
+	defer func() {
+		// for to stop the recv goroutine
+		h.conn.SetReadDeadline(time.Now())
+		h.conn.SetWriteDeadline(time.Now().Add(time.Second))
```
fujita: Why we set write deadline here? It should be done before sending a notification to socket directly?

pkg/server/fsm.go
```diff
@@ -1955,8 +1979,11 @@ func (h *fsmHandler) established(ctx context.Context) (bgp.FSMState, *fsmStateRe
 					"State": fsm.state.String(),
 				})
 			fsm.lock.RUnlock()
+
 			m := bgp.NewBGPNotificationMessage(bgp.BGP_ERROR_HOLD_TIMER_EXPIRED, 0, nil)
-			h.outgoing.In() <- &fsmOutgoingMsg{Notification: m}
+			b, _ := m.Serialize(h.fsm.marshallingOptions)
+			_, err := h.conn.Write(b)
```
fujita, line 1986: When established() hits holdTimer.C and send a notification directly, sendMessageloop goroutine is still sending? 
When a hold timer expires, what BGP should do is, to stop sending update messages anymore, send a notification, and close the connection.

pkg/server/fsm.go
maxime-peim: leftover from the version where we pushed into the channel

pkg/server/fsm.go
maxime-peim: I followed what you did when the global context get cancelled.
when the notification is sent, the function will soon returns and stop the sending loop, those are the same assumption you followed for the `case <-ctx.Done()`

fujita: Currently, sendMessageloop and recvMessageloop also use the context that is used by established loop. So once it is canceled, both sendMessageloop and recvMessageloop will be canceled.

Your change makes sure that  until established loop exits (after sending a notification), sendMessageloop is keep sending. It makes the situation worse.

If you don't introduce ioCtx, I'm fine with the change because it doesn't worsen.
If you introduce ioCtx, you could move sending a notification to the defer or when holdtimer expires, the established stops both loop goroutines and waits, and then send a notification.

fujita:
I still think that we should stop sendMessageloop first then send a notification to make sure that "A NOTIFICATION message is sent when an error condition is detected. The BGP connection is closed immediately after it is sent." in RFC 4271.

However, since the peer will close the connection after receiving the NOTIFICATION, it shouldn’t really be an issue in practice. Also, this code will be changed soon, so after talking directly with @nplanelcisco, I decided to merge this pull request.

Thanks!

PR #3054, opened 2025-07-30
fix: race condition in TestWatch

maxime-peim:


fujita:
Cancel the context for WatchEvent() to stop goroutine that WatchEvent() created, right?

fujita:
rebased and pushed, thanks.

PR #3056, opened 2025-07-31
pkg/table: make AddrPrefixOnlyCompare private

fujita:
Move AddrPrefixOnlyCompare from bgp to table package; make it private function.

Public API interface is written in stone. It cannot be changed without the major version updated. Currently, AddrPrefixOnlyCompare is used only by table package so let's keep it in the table package. We can always make it public in the future when needed.

### DIFF
--- a/internal/pkg/table/table.go
+++ b/internal/pkg/table/table.go
@@ -16,6 +16,7 @@
 package table
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


fujita:
@nplanelcisco 

Now IPAddrPrefix, IPv6AddrPrefix, LabeledVPNIPAddrPrefix, and LabeledVPNIPv6AddrPrefix use netip.Prefix so I think that you can compare them directly; no need to serialize IP prefix bytes.

It might improve the performance; addrPrefixOnlySerialize() allocates heap and it will be freed immediately. We can avoid it. The compiler is cleaver enough to avoid such even with the current code?

nplanelcisco:
> @nplanelcisco
> 
> Now IPAddrPrefix, IPv6AddrPrefix, LabeledVPNIPAddrPrefix, and LabeledVPNIPv6AddrPrefix use netip.Prefix so I think that you can compare them directly; no need to serialize IP prefix bytes.
> 
> It might improve the performance; addrPrefixOnlySerialize() allocates heap and it will be freed immediately. We can avoid it. The compiler is cleaver enough to avoid such even with the current code?

LGTM

nplanelcisco (approved):

fujita:
Thanks!

PR #3057, opened 2025-07-31
pkg/packet/bgp: use netip for LabeledIPAddrPrefix and LabeledIPv6Addr…

fujita:
…Prefix

Use IpAddrPrefixDefaultNetip instead of IpAddrPrefixDefault.

### DIFF
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -1351,13 +1351,13 @@ func nlriToIPNet(nlri bgp.AddrPrefixInterface) *net.IPNet {
 		}
 	case *bgp.LabeledIPAddrPrefix:
 		return &net.IPNet{
-			IP:   T.Prefix.To4(),
-			Mask: net.CIDRMask(int(T.Length)-T.Labels.Len()*8, 32),
+			IP:   T.Prefix.Masked().Addr().AsSlice(),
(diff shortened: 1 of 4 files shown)


internal/pkg/table/path.go
```diff
@@ -1350,13 +1350,13 @@ func nlriToIPNet(nlri bgp.AddrPrefixInterface) *net.IPNet {
 		}
 	case *bgp.LabeledIPAddrPrefix:
 		return &net.IPNet{
-			IP:   T.Prefix.To4(),
-			Mask: net.CIDRMask(int(T.Length)-T.Labels.Len()*8, 32),
+			IP:   T.Prefix.Addr().AsSlice(),
```
nplanelcisco: I think we shall have extra fence by `T.Prefix.Masked().Addr().AsSlice()`

fujita: Added. If it was created by `DecodeFromBytes()`, this extra fence isn't necessary but indeed, safer with other cases.

Once this conversion is done, I'll remove this function (we can use `netip.Prefix` directly) so I leave other NLRI cases alone.

nplanelcisco:
LGTM

nplanelcisco (approved):

fujita:
@nplanelcisco Thanks!

PR #3059, opened 2025-07-31
pkg/table: simplify tableKey()

fujita:
Use ipnet.Prefix.MarshalBinary(), which returns Addr.MarshalBinary() with an additional byte appended containing the prefix bits.

### DIFF
--- a/internal/pkg/table/table.go
+++ b/internal/pkg/table/table.go
@@ -52,21 +52,21 @@ func tableKey(nlri bgp.AddrPrefixInterface) addrPrefixKey {
 	h := fnv1a.Init64
 	switch T := nlri.(type) {
 	case *bgp.IPAddrPrefix:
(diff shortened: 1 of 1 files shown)


fujita:
@nplanelcisco

nplanelcisco:
it's looks good at the first look but it's less efficient as `Mashalbinary` do allocation as well
A small benchmark
Before
BenchmarkTableKeyWithLabels-12    	  478888	      **2266 ns/op**	     864 B/op	      **36 allocs/op**
After (this patch)
BenchmarkTableKeyWithLabels-12    	  398408	      2718 ns/op	    1088 B/op	      50 allocs/op

fujita:
@nplanelcisco
Thanks for the benchmark results! Looks like Marshalbinary does more than the current code like checking, which might make the difference. Let's keep the current code.

PR #3060, opened 2025-07-31
pkg/packet/bgp: use netip for IPAddressSpecificExtended

fujita:
Replace net.IP with netip.Addr for IPAddressSpecificExtended.


nplanelcisco (approved):

fujita:
Thanks for the review!

PR #3062, opened 2025-07-31
pkg/packet/bgp: use netip for PathAttributeOriginatorId

fujita:
Replace net.IP with netip.Addr for PathAttributeOriginatorId.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -11542,7 +11542,7 @@ func NewPathAttributeCommunities(value []uint32) *PathAttributeCommunities {
 
 type PathAttributeOriginatorId struct {
 	PathAttribute
-	Value net.IP
+	Value netip.Addr
 }
 
 func (p *PathAttributeOriginatorId) DecodeFromBytes(data []byte, options ...*MarshallingOption) error {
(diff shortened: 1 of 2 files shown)


pkg/packet/bgp/bgp.go
```diff
@@ -11576,20 +11576,20 @@ func NewPathAttributeCommunities(value []uint32) *PathAttributeCommunities {
...
 		return err
 	}
-	if p.Length != 4 {
+	if p.Length != 4 || len(value) != 4 {
 		eCode := uint8(BGP_ERROR_UPDATE_MESSAGE_ERROR)
 		eSubCode := uint8(BGP_ERROR_SUB_ATTRIBUTE_LENGTH_ERROR)
 		return NewMessageError(eCode, eSubCode, nil, "originator id length isn't correct")
 	}
-	p.Value = value
+	p.Value, _ = netip.AddrFromSlice(value)
```
nplanelcisco, line 11559: why not returning the error here ?

nplanelcisco: ... it will never fail as boundaty `len(value) != 4` is already tested

fujita: Right, added the comment here.

I added the `len(value)` check but it's not necessary. `p.PathAttribute.DecodeFromBytes` guarantees that it returns data with the same length as p.Length. So I dropped that change.

fujita:
Added `check Is4()` comment to `NewPathAttributeOriginatorId()` like `NewPathAttributeClusterList()`

PR #3063, opened 2025-08-01
pkg/packet/bgp: use netip for PathAttributeMpReachNLRI

fujita:
Replace net.IP with netip.Addr for PathAttributeMpReachNLRI.


nplanelcisco (approved):

fujita:
@nplanelcisco thanks!

PR #3065, opened 2025-08-01
table key benchmark

nplanelcisco:


nplanelcisco:
@fujita

fujita:
This adds benchmark for generating tablekey for LabeledIPv6AddrPrefix, right?

nplanelcisco:
> This adds benchmark for generating tablekey for LabeledIPv6AddrPrefix, right?

yes but not only
```
			prefixes = append(prefixes, bgp.NewLabeledVPNIPAddrPrefix(lengthv4, prefixv4, l, rd))
			prefixes = append(prefixes, bgp.NewLabeledIPAddrPrefix(lengthv4, prefixv4, l))
			prefixes = append(prefixes, bgp.NewLabeledVPNIPv6AddrPrefix(lengthv6, prefixv6, l, rd))
			prefixes = append(prefixes, bgp.NewLabeledIPv6AddrPrefix(lengthv6, prefixv6, l))

fujita:
Added a commit message and pushed, thanks.

Is there room for performance improvement? I think that we could use just serialized bytes for a struct for label.

nplanelcisco:
> Added a commit message and pushed, thanks.
> 
> Is there room for performance improvement? I think that we could use just serialized bytes for a struct for label.

I don't think there are or maybe marginal, that just the test I used to bench your previous code (netip.Marshal())

PR #3067, opened 2025-08-02
packet/bgp: remove unnecessary double family checking in ValidateAttr…

fujita:
…ibute

Remove family checking done twice in ValidateAttribute().

### DIFF
--- a/pkg/packet/bgp/validate.go
+++ b/pkg/packet/bgp/validate.go
@@ -92,17 +92,17 @@ func ValidateAttribute(a PathAttributeInterface, rfs map[Family]BGPAddPathMode, 
 	eSubCodeUnknown := uint8(BGP_ERROR_SUB_UNRECOGNIZED_WELL_KNOWN_ATTRIBUTE)
 	eSubCodeMalformedAspath := uint8(BGP_ERROR_SUB_MALFORMED_AS_PATH)
 
-	checkPrefix := func(l []AddrPrefixInterface) error {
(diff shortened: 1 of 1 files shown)


pkg/packet/bgp/validate.go
```diff
@@ -92,17 +92,16 @@ func ValidateAttribute(a PathAttributeInterface, rfs map[Family]BGPAddPathMode,
 	eSubCodeUnknown := uint8(BGP_ERROR_SUB_UNRECOGNIZED_WELL_KNOWN_ATTRIBUTE)
 	eSubCodeMalformedAspath := uint8(BGP_ERROR_SUB_MALFORMED_AS_PATH)
 
-	checkPrefix := func(l []AddrPrefixInterface) error {
+	checkPrefix := func(family Family, l []AddrPrefixInterface) error {
 		for _, prefix := range l {
-			rf := NewFamily(prefix.AFI(), prefix.SAFI())
-			if _, ok := rfs[rf]; !ok {
-				return NewMessageError(0, 0, nil, fmt.Sprintf("Address-family %s not available for this session", rf))
+			if _, ok := rfs[family]; !ok {
```
nplanelcisco: we could put that outside of the loop, we only need to check once

fujita: Good point, done. It could return earlier with non flow spec family but it would be better to do in a different commit.

nplanelcisco (approved):

fujita:
@nplanelcisco thanks!

PR #3069, opened 2025-08-02
packet/mrt: Add Family argument to NewRib

fujita:
Add Family argument to NewRib function in order to avoid getting Family from NLRI. The NLRI type and Family do not have a one-to-one correspondence, we can only obtain an incomplete value for Family.

### DIFF
--- a/pkg/packet/mrt/mrt_test.go
+++ b/pkg/packet/mrt/mrt_test.go
@@ -187,7 +187,7 @@ func TestMrtRib(t *testing.T) {
 	e2 := NewRibEntry(2, uint32(time.Now().Unix()), 0, p, false)
 	e3 := NewRibEntry(3, uint32(time.Now().Unix()), 0, p, false)
 
-	r1 := NewRib(1, bgp.NewIPAddrPrefix(24, "192.168.0.0"), []*RibEntry{e1, e2, e3})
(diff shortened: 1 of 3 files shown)


pkg/packet/mrt/mrt_test.go
```diff
@@ -187,7 +187,7 @@ func TestMrtRib(t *testing.T) {
 	e2 := NewRibEntry(2, uint32(time.Now().Unix()), 0, p, false)
 	e3 := NewRibEntry(3, uint32(time.Now().Unix()), 0, p, false)
 
-	r1 := NewRib(1, bgp.NewIPAddrPrefix(24, "192.168.0.0"), []*RibEntry{e1, e2, e3})
+	r1 := NewRib(1, bgp.RF_IPv4_UC, bgp.NewIPAddrPrefix(24, "192.168.0.0"), []*RibEntry{e1, e2, e3})
```
nplanelcisco, line 190: I think rib family should be addressed by prefix parsing as suggested here https://github.com/osrg/gobgp/pull/3070/files#r2251531518

nplanelcisco: @fujita here my proposal instead of this PR
we could remove addrlen from IpAddrDefaultPrefix
and handle multicast correctly

It's a 2 commits in this PR, we can split if needed
https://github.com/osrg/gobgp/pull/3081

PR #3070, opened 2025-08-03
apiutil: Add Family argument to NewPath function

fujita:
Add Family argument to NewPath function in order to avoid getting Family from NLRI. The NLRI type and Family do not have a one-to-one correspondence, we can only obtain an incomplete value for Family.

### DIFF
--- a/pkg/apiutil/util.go
+++ b/pkg/apiutil/util.go
@@ -161,7 +161,7 @@ func NewDestination(dst *api.Destination) *Destination {
 	return &Destination{Paths: l}
 }
 
-func NewPath(nlri bgp.AddrPrefixInterface, isWithdraw bool, attrs []bgp.PathAttributeInterface, age time.Time) (*api.Path, error) {
+func NewPath(family bgp.Family, nlri bgp.AddrPrefixInterface, isWithdraw bool, attrs []bgp.PathAttributeInterface, age time.Time) (*api.Path, error) {
 	n, err := MarshalNLRI(nlri)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/apiutil/util.go
```diff
@@ -175,7 +175,7 @@ func NewPath(nlri bgp.AddrPrefixInterface, isWithdraw bool, attrs []bgp.PathAttr
 		Pattrs:     a,
 		Age:        tspb.New(age),
 		IsWithdraw: isWithdraw,
-		Family:     ToApiFamily(nlri.AFI(), nlri.SAFI()),
```
nplanelcisco, line 178: in which case getting the family from the nlri/prefix is not the right way ?

fujita: RF_IPv4_MC and RF_IPv6_MC.

nplanelcisco: why not doing it in the prefix itself ? something like
```
func NewIPAddrPrefix(bits uint8, prefix string) *IPAddrPrefix
   p := netip.ParsePrefix("....")
   return &IPAddrPrefix{ isMulticast: p.Addr().IsMulticast() , .... }

func (ip4 *IPAddrPrefix) SAFI() {
    if ip4.isMulticast {
       return RF_IPv4_MC
    }
   return RF_IPv4_UC
}

fujita: We can. The point to discuss is where to store the relationship between Family and NLRI.

Since the number of AddrPrefixInterfaces increases proportionally to both the number of routes and the number of peers, in a full route environment with 10 peers, if NLRI increases by 4 bytes, memory usage would increase by 40MB. Storing Family information in NLRI does not seem ideal.

Note that Paths in the same Destination should share the NLRI. We need to remove PathID info (that is, PrefixDefault) from NLRI before that.

nplanelcisco: I think we can do this way https://github.com/osrg/gobgp/pull/3081

PR #3071, opened 2025-08-03
pkg/packet/bmp: use netip

fujita:
Replace net.IP with netip.Addr for bmp code.


fujita:
@nplanelcisco thanks!

PR #3072, opened 2025-08-03
pkg/packet/rtr: use netip

fujita:
Replace net.IP with netip.Addr for rtr code.


pkg/packet/rtr/rtr.go
```diff
@@ -250,19 +250,20 @@ func (m *RTRIPPrefix) Serialize() ([]byte, error) {
...
 	}
 	return data, nil
 }
 
-func NewRTRIPPrefix(prefix net.IP, prefixLen, maxLen uint8, as uint32, flags uint8) *RTRIPPrefix {
+func NewRTRIPPrefix(prefix netip.Addr, prefixLen, maxLen uint8, as uint32, flags uint8) *RTRIPPrefix {
 	var pduType uint8
 	var pduLen uint32
-	if prefix.To4() != nil && prefixLen <= 32 {
+	// TODO: return error
```
nplanelcisco: TODO: return error if !prefix.IsValid()

fujita: Indeed, we need to add IsValid() checking to all the APIs that takes netip.Addr.
Once I finish netip conversion, I'll update the APIs to return an error with the checking.

fujita:
@nplanelcisco thanks!

PR #3073, opened 2025-08-03
pkg/packet/mrt: use netip

fujita:
Replace net.IP with netip.Addr for mrt code.


fujita:
@nplanelcisco thanks!

PR #3077, opened 2025-08-06
pkg/packet/bgp: use netip for LsTLVSrv6SIDInfo

fujita:
Replace net.IP with netip.Addr for LsTLVSrv6SIDInfo.


fujita:
@nplanelcisco thanks!

PR #3078, opened 2025-08-06
pkg/packet/mrt: use netip for BGP4MPHeader and GeoPeer

fujita:
Replace net.IP with netip.Addr for BGP4MPHeader, GeoPeer, and GeoPeerTable.

### DIFF
--- a/pkg/packet/mrt/mrt.go
+++ b/pkg/packet/mrt/mrt.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"net"
 	"net/netip"
 	"time"
 
@@ -604,7 +603,7 @@ func (u *Rib) String() string {
 
 type GeoPeer struct {
 	Type      uint8
-	BgpId     net.IP
+	BgpId     netip.Addr
 	Latitude  float32
 	Longitude float32
 }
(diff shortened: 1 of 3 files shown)


pkg/packet/mrt/mrt.go
```diff
@@ -627,31 +626,34 @@ func (p *GeoPeer) DecodeFromBytes(data []byte) ([]byte, error) {
...
-	copy(buf[1:5], bgpId)
+	copy(buf[1:5], p.BgpId.AsSlice())
 	binary.BigEndian.PutUint32(buf[5:9], math.Float32bits(p.Latitude))
 	binary.BigEndian.PutUint32(buf[9:13], math.Float32bits(p.Longitude))
 	return buf, nil
 }
 
-func NewGeoPeer(bgpid string, latitude float32, longitude float32) *GeoPeer {
+func NewGeoPeer(bgpid netip.Addr, latitude float32, longitude float32) (*GeoPeer, error) {
+	if !bgpid.Is4() {
```
nplanelcisco, line 639: shall we enforce `&& !bgpid.IsUnspecified()` ?  (line 703 as well)

pkg/packet/mrt/mrt.go
```diff
@@ -775,40 +779,40 @@ func (m *BGP4MPHeader) serialize() ([]byte, error) {
...
 }
 
-func newBGP4MPHeader(peeras, localas uint32, intfindex uint16, peerip, localip string, isAS4 bool) (*BGP4MPHeader, error) {
+func newBGP4MPHeader(peeras, localas uint32, intfindex uint16, peerip, localip netip.Addr, isAS4 bool) (*BGP4MPHeader, error) {
 	var af uint16
-	paddr := net.ParseIP(peerip).To4()
-	laddr := net.ParseIP(localip).To4()
-	if paddr != nil && laddr != nil {
+
+	if !peerip.IsValid() || !localip.IsValid() {
```
nplanelcisco, line 797: same here

pkg/packet/mrt/mrt.go
fujita: This is a serialization/deserialization library, so I think that it’s better if users can construct whatever binary they want — as long as the values don’t cause GoBGP or the library to crash.

PR #3080, opened 2025-08-07
Make containscidr private

fujita:


internal/pkg/table/table.go
```diff
@@ -619,6 +620,13 @@ func (t *Table) GetKnownPathListWithMac(id string, as uint32, rt bgp.ExtendedCom
 	return paths
 }
 
+// ContainsCIDR checks if one IPNet is a subnet of another.
+func containsCIDR(n1, n2 *net.IPNet) bool {
```
nplanelcisco, line 624: I feel the whole things (table/ipnet) need to be changed to use netip.Prefix.Contains() 
let's merge this one add this to the todolist

nplanelcisco (approved):

internal/pkg/table/table.go
fujita: Indeed, we don't need to convert netip.Prefix to net.IPnet. The priority is removing the usage of net.IP* from public APIs. We can clean up the internal usages even after 4.0 release.

fujita:
@nplanelcisco thanks!

Issue #3043, opened 2025-07-28
Planed BGP packet library changes for 4.0


fujita:
- [x] Add NLRIFromSlice APIs
   Allocating incomplete AddrPrefix structures and calling AddrPrefix.DecodeFromBytes() isn't good. API to return a fully initialized object is aligned with netip API. We should do the same with other types like Attribute.
   This means that we remove `DecodeFromBytes` from `AddrPrefixInterface`. 
- [x] Remove IPv6 option in `MarshallingOption`.
- [x] Prohibiting passing null strings to NewIPAddrPrefix() for the same reason.
- [x] Rename AddrPrefixInterface to NLRI
   Some of address families are not related with Prefix (IP prefix) at all so AddrPrefixInterface is confusing.
- [x] Use the same IPAddrPrefix structure for v4/v6 (for example, IPAddrPrefix and IPv6AddrPrefix).
- [x] Remove AFI() and SAFI() from AddrPrefixInterface.
   There is not a one-to-one relationship between types implementing AddrPrefixInterface and Family. Usually, a single type supports multiple Families.
- [x] Make New*AddrPrefix() return an error.
- [x] Make New*AddrPrefix() take netip.Prefix instead of string
- [x] Remove mutex
   packet library is supposed to serialize and deserialize bytes. Locking should be done by the users of this package.
- [ ] Make serialize function take a buffer.
   Currently, serializer function allocates heap again and again (by append() function) and return a buffer. For performance reasons, it would be desirable to accept a buffer as an argument and avoid heap allocations. This is a major change, so it's unclear if it will make it into 4.0.
- [x] Replace net package with netip.
   Internal usage is fine but API should use netip.

mayuresh82:
When is v4 coming out ? The offical examples are already referencing v4 : https://github.com/osrg/gobgp/blob/master/docs/sources/lib.md

fujita:
> When is v4 coming out ? The offical examples are already referencing v4 : https://github.com/osrg/gobgp/blob/master/docs/sources/lib.md

I planed at the end of summer but it might be postponed. The master branch uses v4 but is not released:
https://github.com/osrg/gobgp/releases

vincentbernat:
Hey!

Do you plan for a migration guide? The checklist looks almost OK, but it does not consider accumulated effects (`AddrPrefixInterface` becomes `NLRI`, which then becomes `PathNLRI` from the user point of view).

fujita:
[...]

PR #3081, opened 2025-08-07
remove IPAddrPrefixDefault addrlen, use netip.Prefix and options to know/set the ip family

nplanelcisco:
and handle multicast correctly

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -37,6 +37,8 @@ type MarshallingOption struct {
 	AddPath    map[Family]BGPAddPathMode
 	Attributes map[BGPAttrType]bool
 	MRT        bool
+
+	isIPv6 bool
 }
 
 func IsMRTSerialization(options []*MarshallingOption) bool {
@@ -75,6 +77,18 @@ func IsAttributePresent(attr BGPAttrType, options []*MarshallingOption) bool {
 		if o := opt.Attributes; o != nil {
 			_, ok := o[attr]
 			return ok
+		}
+	}
+	return false
+}
+
+func isIPv6(options []*MarshallingOption) bool {
+	for _, opt := range options {
+		if opt == nil {
+			continue
+		}
+		if opt.isIPv6 {
+			return true
 		}
 	}
 	return false
@@ -1459,15 +1473,13 @@ func (p *PrefixDefault) serializeIdentifier() ([]byte, error) {
 
 type IPAddrPrefixDefault struct {
 	PrefixDefault
-	Prefix  netip.Prefix
-	addrlen uint8
+	Prefix netip.Prefix
 }
 
-func (r *IPAddrPrefixDefault) decodePrefix(data []byte, bitlen uint8) error {
-	if r.addrlen == 0 {
-		r.addrlen = net.IPv4len
+func (r *IPAddrPrefixDefault) decodePrefix(data []byte, bitlen uint8, addrlen int) error {
+	if addrlen != 4 && addrlen != 16 {
+		return NewMessageError(BGP_ERROR_UPDATE_MESSAGE_ERROR, BGP_ERROR_SUB_INVALID_NETWORK_FIELD, nil, "invalid address length")
 	}
-
 	if len(data) < 1 {
 		eCode := uint8(BGP_ERROR_UPDATE_MESSAGE_ERROR)
 		eSubCode := uint8(BGP_ERROR_SUB_MALFORMED_ATTRIBUTE_LIST)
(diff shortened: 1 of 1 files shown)


fujita:
I had planned to remove addrlen when consolidating the separate structs for v4 and v6, but since you’ve started working on it, let's discuss the `DecodeFromBytes` and `Serialize` functions interface first.

With this change, since MarshallingOption becomes required to parse IPv6 addresses correctly, is the functional options pattern still appropriate? Better to use a normal argument?

I think that`Family` type is better than ipv6 `bool` type.  Then we don't need to do the following stuff every time serializing a prefix.

```
	f := RF_IPv4_VPN
	isMulticast := l.Prefix.Addr().IsMulticast()
	isIPv6 := l.Prefix.Addr().Is6()
	switch {
	case !isMulticast && !isIPv6:
		f = RF_IPv4_VPN
	case isMulticast && !isIPv6:
		f = RF_IPv4_VPN_MC
	case !isMulticast && isIPv6:
		f = RF_IPv6_VPN
	case isMulticast && isIPv6:
		f = RF_IPv6_VPN_MC
	}
```

fujita:
Furthermore, checking the AddPath map every time during serialization is not efficient.
When serializing an MP_REACH message containing multiple prefixes, it will only include a single family, so the map can be checked once at the beginning and the result reused for each prefix serialization.

nplanelcisco:
> I had planned to remove addrlen when consolidating the separate structs for v4 and v6, but since you’ve started working on it, let's discuss the `DecodeFromBytes` and `Serialize` functions interface first.
> 
> With this change, since MarshallingOption becomes required to parse IPv6 addresses correctly, is the functional options pattern still appropriate? Better to use a normal argument?
> 
> I think that`Family` type is better than ipv6 `bool` type. Then we don't need to do the following stuff every time serializing a prefix.
> 
> ```
> 	f := RF_IPv4_VPN
> 	isMulticast := l.Prefix.Addr().IsMulticast()
> 	isIPv6 := l.Prefix.Addr().Is6()
> 	switch {
> 	case !isMulticast && !isIPv6:
> 		f = RF_IPv4_VPN
> 	case isMulticast && !isIPv6:
> 		f = RF_IPv4_VPN_MC
> 	case !isMulticast && isIPv6:
> 		f = RF_IPv6_VPN
> 	case isMulticast && isIPv6:
> 		f = RF_IPv6_VPN_MC
> 	}
> ```

Sure,

Marshaling option is only used in the `DecodeFromBytes` path and will be passed along the way from labeled vpn prefix to IPAddrPrefix, so keeping the isv6 in the MarshalingOption would make sense here. we can put the switch case in an helper `GetFamily(prefix netip.Prefix) Family`.
I don't think changing the `DecodeFromBytes` API to add family parameter could be justified here as it's like optional, so MarshallingOption make sense in this way

On the `Serialize` side l.Prefix already contain all the family needed info and avoid us to store twice the same info in the IPAddrPrefix (if we add Family)

nplanelcisco:
> Furthermore, checking the AddPath map every time during serialization is not efficient. When serializing an MP_REACH message containing multiple prefixes, it will only include a single family, so the map can be checked once at the beginning and the result reused for each prefix serialization.

Agreed, I tried to have minimal impact here on changes, and think it would be another PR as the marshaling option, should be accessed other way, like O(1) by merging MarshalingOptions in a single object, so no need to iterate on the options list afterwards

nplanelcisco:
the problem I don't like with the isAddPath lookup (in DecodeFromBytes) is the prefix is not aware on what to expect 
in case of multicast ipv6 (or v4) I would need to lookup twice unicast and multicast family, but this is based only to the fact that both unicast and multicast families entries are always in sync as the check is basically a logical or

nplanelcisco:
overall I think we need to talk on our next sync, via sharing some code, the conversation/ideas would probably be more effective

fujita:
> overall I think we need to talk on our next sync, via sharing some code, the conversation/ideas would probably be more effective

Ok, I try to implement some ideas.

btw, the second commit to use `Addr` to see if multicast or not doesn't work for EOR path object. The fundamental problem is that EOR path should not have nlri. We should not use uninitialized nlri in any place but put uninitialized nlri to EOR path to get family when we convert the path to gRPC path. I'll fix `Path.GetFamily()` not to use nlri.AFI/SAFI(). Then we don't need AFI() and SAFI() in AddrPrefixInterface in the first place.

fujita:
Pushed, thanks!

As I wrote:
https://github.com/osrg/gobgp/issues/3043

I'll remove `DecodeFromBytes` from `AddPrefixInterface` so I think that I'll remove isIPv6 in `MarshallingOption` but removing `addrlen` is the first step.

PR #3085, opened 2025-08-15
packet/bgp: Add NLRIFromSlice to build initialized NLRI from bytes

fujita:
This avoids the current pattern of creating an uninitialized IPAddrPrefix (with an invalid netip.Addr) and then mutating it via parsing, which is error-prone.

The name intentionally avoids “AddrPrefix”, as some families are not prefix-based (FlowSpec, RTC, EVPN, etc.). The API follows the net/netip FromSlice convention.


fujita:
@nplanelcisco Thanks!

PR #3087, opened 2025-08-18
don't answer for rt's while waiting for rtc eor

bor1go:
1. Add a test for correct message sequencing: VPN paths must precede VPN EoR.  
While in the rtc deferral-time == 0 case, the VPN EoR is sent immediately after peering (because the RT list is empty), it is important to verify that with rtc deferral-time > 0, the VPN EoR is sent after all VPN routes. This test case checks for the correct order of route advertisement versus End-of-RIB signaling.

2. Fix double-sending of paths with rtc deferral-time > 0.  
The test also tracks the number of received messages and has revealed that each VPN route is sent twice: once as a reply to an RT, and then again in the full set of VPN paths after RTC EoR is received.  
To address this duplication, new VPN paths should not be sent in response to RTs during the initial state. In this model, GoBGP waits for RTC EoR before sending the complete list of VPN paths. In the propagateUpdate function if a route is added during the initial exchange, GoBGP still announces it immediately to the peer, and not as an RT response.  
As a result, for VPN paths already present when the session is established, GoBGP will not send them before either RTC EoR is received or the rtc deferral-time expires. This behavior integrates well with graceful restart and RTC EoR signaling, but otherwise may introduce a delay up to the configured rtc deferral-time. That's why:

3. Always send RTC EoR, regardless of graceful-restart.  
Previously, RTC EoR was sent only if graceful-restart was enabled. According to RFC 4684, section 6:  
> "As a hint that initial RT membership exchange is complete, implementations SHOULD generate an End-of-RIB marker, as defined in [8], for the Route Target membership (AFI, SAFI), regardless of whether graceful-restart is enabled on the BGP session."

RTC EoR is now sent unconditionally after the initial RT membership exchange.

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -188,6 +188,7 @@ type fsm struct {
 	adminStateCh         chan adminStateOperation
 	h                    *fsmHandler
 	rfMap                map[bgp.Family]bgp.BGPAddPathMode
+	rtcEORWait           bool
 	capMap               map[bgp.BGPCapabilityCode][]bgp.ParameterCapabilityInterface
 	recvOpen             *bgp.BGPMessage
 	peerInfo             *table.PeerInfo
--- a/pkg/server/peer.go
+++ b/pkg/server/peer.go
@@ -297,6 +297,31 @@ func (peer *peer) isDynamicNeighbor() bool {
 	return peer.fsm.pConf.Config.NeighborAddress == "" && peer.fsm.pConf.Config.NeighborInterface == ""
 }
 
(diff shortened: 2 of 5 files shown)


bor1go:
@fujita could you please take a look at this PR, if possible?
I tried to explain the reason above ^^^. In other words, with rtc deferral-time>0, vpn routes are sent twice during the initial exchange. This PR fixes this.

fujita:
Thanks, pushed.

I tried to think of a way to use the existing `MpGracefulRestart.State.EndOfRibReceived` instead of adding a new `rtcEORWait`, but I couldn't come up with a good idea.

PR #3088, opened 2025-08-20
expose LlgrRunning flag in API

myaut:
We implemented and external monitor which checks if some peers are in active GR (PeerRestarting) or LLGR (LonglivedRestarting), so we added:
- exposing LonglivedRestarting in gRPC API to distinguish GR and LLGR
- changes to GR test that checks that both flags are correctly returned by gRPC API
- a small fix to advertised fields which might incorrectly be retained if config is changed to disabled GR

### DIFF
--- a/proto/api/gobgp.proto
+++ b/proto/api/gobgp.proto
@@ -920,7 +920,6 @@ message GracefulRestart {
   bool peer_restarting = 9;
   bool local_restarting = 10;
   string mode = 11;
-  bool longlived_running = 12;
 }
 
 message MpGracefulRestartConfig {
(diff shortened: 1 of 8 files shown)


proto/api/gobgp.proto
```diff
@@ -920,6 +920,7 @@ message GracefulRestart {
   bool peer_restarting = 9;
   bool local_restarting = 10;
   string mode = 11;
+  bool longlived_running = 12;
```
fujita: This should be per family? If so, it should be added to`MpGracefulRestartState`?

fujita:
Exporting those states sounds reasonable. But these states are per family? We need to clean up `MpGraceful` and `Gracefulrestart` messages a bit?

proto/api/gobgp.proto
myaut: I reworked it to per-family flag and implemented same flag for GR (`MpGracefulRestartState`). Please take a look.

fujita:
Pushed, thanks.

PR #3089, opened 2025-08-21
Move peerinfo to server

fujita:


pkg/server/fsm.go
```diff
@@ -2017,6 +2009,17 @@ func (h *fsmHandler) loop(ctx context.Context, wg *sync.WaitGroup) error {
...
 	case bgp.BGP_FSM_ESTABLISHED:
+		// Allow updates from host loopback addresses if the BGP connection
+		// with the neighbour is both dialed and received on loopback
+		// addresses.
+		fsm.lock.RLock()
+		remoteTCP := fsm.conn.RemoteAddr().(*net.TCPAddr)
+		remoteAddr, _ := netip.AddrFromSlice(remoteTCP.IP)
+		localTCP := fsm.conn.LocalAddr().(*net.TCPAddr)
+		localAddr, _ := netip.AddrFromSlice(localTCP.IP)
+		h.allowLoopback = remoteAddr.Is4() && localAddr.Is4() && remoteAddr.IsLoopback() && localAddr.IsLoopback()
```
nplanelcisco, line 2056: why not supporting ipv6 loopback ?
`h.allowLoopback = remoteAddr.BitLen() == localAddr.BitLen() && remoteAddr.IsLoopback() && localAddr.IsLoopback()`

fujita: This is for testing multiple GoBGP instances on localhost https://github.com/osrg/gobgp/commit/16a9c953939e003447bed4b3a80ac817c02e299a

In IPv4, we can pick two different addresses from 127.0.0.0/8, have two peers bind to those addresses, and establish a TCP connection over the loopback device. In IPv6, however, the loopback address is limited to ::1, so with the current implementation (which distinguishes peers by IP address) this likely would not work (I've not tried).

nplanelcisco: yeah good point, it's another topic as we probably need to distinguishes per router id

fujita:
@nplanelcisco thanks, I added why we allow only v4 here to the commit message.

PR #3093, opened 2025-08-22
feat(netip) apiutil use netip.Addr

nplanelcisco:


pkg/server/grpc_server.go
```diff
@@ -216,13 +217,17 @@ func toPathAPI(binNlri []byte, binPattrs [][]byte, anyNlri *api.NLRI, anyPattrs
...
 		Filtered:        path.Filtered,
 		Validation:      path.Validation,
 	}
+	if path.PeerID.IsValid() {
+		p.SourceId = path.PeerID.String()
+	}
+	if path.PeerAddress.IsValid() {
+		p.NeighborIp = path.PeerAddress.String()
+	}
 	return p
```
fujita, line 231: What path has invalid id and address? Locally generated path?

nplanelcisco: path.PeerID and PeerAddress are only valid if `path.GetSource() != nil` https://github.com/osrg/gobgp/blob/master/pkg/server/server.go#L4454

As source is optional, this code is to avoid to expose "Invalid Address" string in case of invalid `netip.Addr{}`

fujita: I think that it's cleaner to make a path always have valid source. But it should be done in a different patch. I'll merge this.

fujita:
Thanks!

PR #3098, opened 2025-08-23
Netip peerinfo

fujita:


internal/pkg/table/destination.go
```diff
@@ -785,15 +785,15 @@ func compareByNeighborAddress(path1, path2 *Path) *Path {
...
 		return path1
 	}
 	p2 := path2.GetSource().Address
-	if p2 == nil {
+	if !p2.IsValid() {
 		return path2
 	}
 
-	cmp := bytes.Compare(p1, p2)
+	cmp := bytes.Compare(p1.AsSlice(), p2.AsSlice())
```
nplanelcisco: IIRC `p1.Compare(p2)`

pkg/server/server.go
```diff
@@ -551,7 +550,7 @@ func filterpath(peer *peer, path, old *table.Path) *table.Path {
 				// the advertisement received SHOULD be ignored.
 				for _, clusterID := range path.GetClusterList() {
 					peer.fsm.lock.RLock()
-					rrClusterID := peer.fsm.peerInfo.RouteReflectorClusterID
+					rrClusterID := net.ParseIP(string(peer.fsm.pConf.RouteReflector.State.RouteReflectorClusterId))
```
nplanelcisco, line 553: `net.IP(peer.fsm.pConf.RouteReflector.State.RouteReflectorClusterId.AsSlice())`

pkg/server/server.go
```diff
@@ -878,11 +877,11 @@ func (s *BgpServer) notifyPrePolicyUpdateWatcher(peer *peer, pathList []*table.P
...
 		Message:      msg,
-		PeerAS:       peer.fsm.peerInfo.AS,
-		LocalAS:      peer.fsm.peerInfo.LocalAS,
-		PeerAddress:  peer.fsm.peerInfo.Address,
+		PeerAS:       peer.fsm.pConf.State.PeerAs,
+		LocalAS:      peer.fsm.pConf.Config.LocalAs,
+		PeerAddress:  net.ParseIP(peer.fsm.pConf.State.NeighborAddress),
 		LocalAddress: net.ParseIP(l),
-		PeerID:       peer.fsm.peerInfo.ID,
+		PeerID:       net.ParseIP(peer.fsm.pConf.State.RemoteRouterId).To4(),
```
nplanelcisco, line 884: `net.IP(xyz.AsSlice())`

pkg/server/server.go
```diff
@@ -2616,7 +2600,7 @@ func (s *BgpServer) AddVrf(ctx context.Context, r *api.AddVrfRequest) error {
 
 		pi := &table.PeerInfo{
 			AS:      s.bgpConfig.Global.Config.As,
-			LocalID: net.ParseIP(s.bgpConfig.Global.Config.RouterId).To4(),
+			LocalID: netip.MustParseAddr(s.bgpConfig.Global.Config.RouterId),
```
nplanelcisco, line 2603: `MustParseAddr` is not a good idea here as it can panic if routerId is bad (but should not)

pkg/server/server_test.go
```diff
@@ -956,7 +957,7 @@ func TestNumGoroutineWithAddDeleteNeighbor(t *testing.T) {
 }
 
 func newPeerandInfo(t *testing.T, myAs, as uint32, address string, rib *table.TableManager) (*peer, *table.PeerInfo) {
```
nplanelcisco: we could drop `table.PeerInfo` here as `peer.fsm.peerInfo` is set

pkg/server/server.go
fujita: Indeed, `peer.fsm.pConf.RouteReflector.State` should be used here. But `AsSlice()`? `peer.fsm.pConf.RouteReflector.State.RouteReflectorClusterId` is string.

pkg/server/server.go
fujita: Ditto. `peer.fsm.pConf.State.RemoteRouterId` is string.

pkg/server/server.go
fujita: https://github.com/osrg/gobgp/blob/08a60b3b69539ab52f8286e0ed81f17c29f9b01b/pkg/server/server.go#L2513

If `s.bgpConfig.Global.Config.RouterId` is invalid, GoBGP server can't start. So `MustParseAddr` here should be safe.

internal/pkg/table/destination.go
fujita: Nice, it should work. I'll fix.

pkg/server/server_test.go
fujita: You meant that we don't need to return PeerInfo, right? The caller can use `peer.fsm.peerInfo`.

Yeah, we could clean up the tests a bit with the change. I'll try.

pkg/server/server.go
nplanelcisco: ah yes it's another patch having  RouteReflectorClusterId as `netip.Addr`

fujita:
@nplanelcisco Thanks!

PR #3104, opened 2025-08-27
fix: properly stop bgp

nplanelcisco:
'handlefsmMsg' function sole possible purpose was when the handler context is cancelled, which only happen if the 'deleteNeighbor' function is called. so the responsability of ending properly the fsm is moved back to the peer with 'stopFSMHandler'. this allow us to quit properly the 'Serve' goroutine.

rebase of #3037

fix issue #3030

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -122,7 +122,6 @@ type BgpServer struct {
 	globalRib    *table.TableManager
 	rsRib        *table.TableManager
 	roaManager   *roaManager
-	shutdownWG   *sync.WaitGroup
 	watcherMap   map[watchEventType][]*watcher
 	zclient      *zebraClient
 	bmpManager   *bmpClientManager
@@ -131,6 +130,11 @@ type BgpServer struct {
 	uuidMap      map[string]uuid.UUID
 	logger       log.Logger
 	timingHook   FSMTimingHook
+	// manage lifecycle of the server
+	isServing     atomic.Bool
+	shutdownWG    *sync.WaitGroup
+	runningCtx    context.Context
+	runningCancel context.CancelFunc
 }
 
 func NewBgpServer(opt ...ServerOption) *BgpServer {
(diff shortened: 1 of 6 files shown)


pkg/server/server.go
```diff
@@ -239,84 +244,7 @@ func (s *BgpServer) mgmtOperation(f func() error, checkActive bool) (err error)
...
-					PeerAS:      fsm.pConf.State.PeerAs,
-					PeerAddress: net.ParseIP(fsm.pConf.State.NeighborAddress),
-					PeerID:      net.ParseIP(fsm.pConf.State.RemoteRouterId).To4(),
-					State:       bgp.BGP_FSM_IDLE,
-					Timestamp:   time.Now(),
-					StateReason: &fsmStateReason{
-						Type: fsmDeConfigured,
-					},
-				})
-			}
```
fujita, line 298: Why the above feature is removed?

pkg/server/server.go
```diff
@@ -429,11 +357,33 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
 }
 
 func (s *BgpServer) Serve() {
+	if s.isServing.Swap(true) {
+		s.logger.Warn("server is already serving",
+			log.Fields{
+				"Topic": "BgpServer",
+			})
+	}
```
fujita, line 365: should this function return or panic here? The function can't continue, right?

fujita:
The major design change (a handler continues to run until the peer is de-configured or BgpServer is killed) looks fine because now the state transition happens synchronously.

pkg/server/server.go
```diff
@@ -3475,14 +3429,11 @@ func (s *BgpServer) deleteNeighbor(c *oc.Neighbor, code, subcode uint8, sendNoti
...
 
-	n.stopPeerRestarting()
 	if sendNotification {
 		n.fsm.notification <- bgp.NewBGPNotificationMessage(code, subcode, nil)
 	}
-	n.fsm.h.ctxCancel()
-
-	delete(s.neighborMap, addr)
 	s.propagateUpdate(n, n.DropAll(n.configuredRFlist()))
+	s.stopNeighbor(n, -1, nil)
```
fujita, line 3436: Why the oldState argument is -1?

pkg/server/server.go
fujita: Ok, stopNeighbor() calls broadcastPeerState() so it could be redundant.

fujita:
I rebased and merged this, thanks!
I'll create a pull request to call stopNeighbor with the appropriate state.

PR #3105, opened 2025-08-27
withdraw old path even if the new one will be omited

bor1go:


bor1go:
@fujita , hello! 
Could you please take a look, if possible?

When peer2 advertises RT1 to peer1, and peer1 has path1 with RT1 and RT2, peer1 sends this path to peer2.
If path1 is then updated to path1' with only RT2, path1' replaces path1 in the GRT and should be withdrawn from peer2.
However, this withdrawal was skipped because path1' does not have RT1.
This PR for fix and test.

fujita:
The fix and the added tests look great, thanks!

PR #3112, opened 2025-08-31
packet/bgp: Update NewIPAddrPrefix()

fujita:
Update NewIPAddrPrefix()

- use netip.Prefix
- return error

### DIFF
--- a/cmd/gobgp/global.go
+++ b/cmd/gobgp/global.go
@@ -2087,7 +2087,12 @@ func parsePath(rf bgp.Family, args []string) (*api.Path, error) {
 			if ip.To4() == nil {
 				return nil, fmt.Errorf("invalid ipv4 prefix")
 			}
-			nlri = bgp.NewIPAddrPrefix(uint8(ones), ip.String())
+			addr, ok := netip.AddrFromSlice(ip.To4())
+			if !ok {
+				return nil, fmt.Errorf("invalid ipv4 address")
+			}
+			prefix := netip.PrefixFrom(addr, ones)
+			nlri, _ = bgp.NewIPAddrPrefix(prefix)
 		} else {
 			if ip.To16() == nil {
 				return nil, fmt.Errorf("invalid ipv6 prefix")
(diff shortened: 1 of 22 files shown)


cmd/gobgp/global.go
```diff
@@ -2087,7 +2087,12 @@ func parsePath(rf bgp.Family, args []string) (*api.Path, error) {
 			if ip.To4() == nil {
 				return nil, fmt.Errorf("invalid ipv4 prefix")
 			}
-			nlri = bgp.NewIPAddrPrefix(uint8(ones), ip.String())
+			addr, ok := netip.AddrFromSlice(ip.To4())
```
nplanelcisco: why not `netip.ParsePrefix(args[0])` line 2081

internal/pkg/table/table_test.go
```diff
@@ -146,8 +151,9 @@ func TestTableKey(t *testing.T) {
 func BenchmarkTableKey(b *testing.B) {
 	rd := bgp.NewRouteDistinguisherTwoOctetAS(1, 2)
 	esi, _ := bgp.ParseEthernetSegmentIdentifier([]string{"lacp", "aa:bb:cc:dd:ee:ff", "100"})
+	nlri1, _ := bgp.NewIPAddrPrefix(netip.MustParsePrefix("192.168.1.0/24"))
```
nplanelcisco, line 154: shall we introduce
```
func bgp.MustIPAddrPrefix(prefix string) {
 nlri, _ := bgp.NewIPAddrPrefix(netip.MustParsePrefix(prefix)
 return nlri
}
```
I think this would we useful for current gobgp users that will move to `apiutil`

cmd/gobgp/global.go
fujita: Oops, fixed.

internal/pkg/table/table_test.go
fujita: I want to avoid such. I understand that such functions can make user migration easier. However, since providing many functions in a library makes future changes to the library itself more difficult, For me a library should only provide a minimal set of functions.

fujita:
@nplanelcisco btw, NewIPAddrPrefix() should check if the prefix is v4. But I finished and plan to merge the patch to remove IPv6AddrPrefix and use IPAddrPrefix for both v4 and v6 shortly so I leave it alone.
https://github.com/osrg/gobgp/commit/c7435cf25e803a29143ad51627f0bdc654c03186

PR #3138, opened 2025-09-12
fix: fix route stuck on eBGP filterpath

Ivan-Pokhabov:
I found bug with eBGP withdraw handling (scenario you can see in test or in comment)

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -551,9 +551,24 @@ func filterpath(peer *peer, path, old *table.Path) *table.Path {
 		// Do not filter local (static) routes with as-path loop
 		// if configured to bypass these checks in the peer
 		// as-path options config.
-		if path.IsLocal() && !peer.allowAsPathLoopLocal() {
-			return nil
-		} else if !path.IsLocal() {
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -551,9 +551,24 @@ func filterpath(peer *peer, path, old *table.Path) *table.Path {
...
+				//
+				// Example: routers from AS A: A1 and A2, AS B: B1
+				// A1 <eBGP> B1, A2 <eBGP> B1, no iBGP
+				// All of them announce prefix P
+				// Paths on A1, A2: [local, B1]
+				// Paths on B1: [local, A1, A2]
+				// When B1 receive local prefix withdraw it choose new best
+				// for example from A1. Then it send. explicit withdraw for A1
+				// and try to implicit withdraw for other peers. But we have AS-Loop for
+				// A2 and path stuck. So in this case B2 should send explicit withdraw for old path.
```
fujita: > So in this case B2 should send explicit withdraw for old path.

`B1`?

Ivan-Pokhabov: Yeah, `B1`, missclick

fujita:
Thanks!

PR #3156, opened 2025-09-22
gobgpd: support certificate reloading

felixhuettner:
TLS on the grpc api can be used for authentication purposes. Previously gobgpd would only load the certificates on startup. If the server certificate of gobgpd or the client CA file changed the user would need to restart gobgpd.
When using short lived certificates (e.g. a few days) for security purposes this requires regular restarts of the gobgp daemon.

In order to make using such short lived certificates easier we introduce the "--tls-reload-interval" command line option. If set then the certificates will be reloaded from disk after the reload interval expires.

### DIFF
--- a/cmd/gobgpd/main.go
+++ b/cmd/gobgpd/main.go
@@ -18,8 +18,6 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"io"
 	"net/http"
@@ -42,6 +40,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
+	"github.com/osrg/gobgp/v4/internal/pkg/grpcutil"
 	"github.com/osrg/gobgp/v4/internal/pkg/version"
 	"github.com/osrg/gobgp/v4/pkg/config"
 	"github.com/osrg/gobgp/v4/pkg/metrics"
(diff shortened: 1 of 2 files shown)


fujita:
This should be a new gRPC API?

felixhuettner:
I can also add this via gRPC if desired. Then i would add something like
```
rpc ReloadCertificates(ReloadCertificatesRequest) returns (ReloadCertificatesResponse);
```
where the request and response objects are empty.
I guess then also a cli command would be helpful.

The drawback i see here is that this requires grpc authentication to work. If e.g. the grpc server certificate expired and only afterwards the file is updated there is no way to recover from this. You would then need to restart gobgpd.

In case of the timing implementation (or also something like inotify if desired) such situations are automatically recovered.

Let me know what you prefer

fujita:
> The drawback i see here is that this requires grpc authentication to work. If e.g. the grpc server certificate expired and only afterwards the file is updated there is no way to recover from this. You would then need to restart gobgpd.

Good point.

> In case of the timing implementation (or also something like inotify if desired) such situations are automatically recovered.

The idea of periodically checking feels to me like an outdated practice from ancient software, and I’d prefer to avoid it. fnotify seems somewhat better than that. I think it would be more desirable to make gobgp command support reload for consistency with other commands, but if we don’t use gRPC, then we’d have to implement notifications to the process via a Unix domain socket or signals, which doesn’t seem like a very good approach.

I have no good idea.

felixhuettner:
We could also combine the approaches:
* Add a cli + gRPC command to reload the certificates, that allows to reload on a user defined change
* Use fnotify to automatically reload on file changes

fnotify should here cover nearly all cases. However if you e.g. store your certificate on NFS (whyever you would do so) then you could still manually reload them.

Or the more simple version would be combining cli + gRPC and the timing approach. But other than being more simple it has no benefit from my perspective

fujita:
Ok, let's go with the fnotify approach.

We can add a gRPC API for this in the future if it ever becomes necessary.

felixhuettner:
ok, sounds good. Then i'll update this PR when i am ready

fujita:
Please also add documentation about gRPC and certificates.

felixhuettner:
I will close this for now, as i am not sure when i will work on this. Thanks a lot

PR #3203, opened 2025-10-20
Parallel connections

fujita:


maxime-peim:
Hi @fujita 
It seems there can still be an issue with your approach, let's say we have two peers A and B that initiate a connection toward there other one:
Let's name fA and fB the main fsm goroutine of A and B, and oA and oB the outgoing connection goroutines:
- oA and oB will try to connect.
- fA and fB (in active state) accept the incoming connection, send an open message, and go to opensent state.
- oA and oB also send an open message and go to opensent state.
- you have then the four goroutines in opensent state
- oA and oB wait to receive the corresponding open message from the other side.
- fA and fB receive there open message before oA and oB
- fA and fB check the outgoingConnectionCh and there is no outgoing connection since oA and oB are still waiting for there open message
- fA and fB stop their outgoing connection manager
- fA just closed the connection it has with oB and fB just closed it has with oA

pkg/server/fsm.go
```diff
@@ -206,6 +207,150 @@ func (s *adminStateRaw) Store(state adminState) {
...
+			case conn = <-connCh:
+				wg.Wait()
+				initializeConn(fsm, conn)
+
+				fsm.lock.Lock()
+				open := buildopen(fsm.gConf, fsm.pConf)
+				fsm.lock.Unlock()
+				b, _ := open.Serialize()
+
+				if _, err := conn.Write(b); err != nil {
```
maxime-peim: if the other side of the connection close, we will continue the loop, recreate a new channel, restart a new connectLoop (even though the previous one) is still running.
If the other side always close our connection, we will eat up memory and spawn more and more goroutine trying to connect to the peer.

pkg/server/fsm.go
```diff
@@ -206,6 +207,150 @@ func (s *adminStateRaw) Store(state adminState) {
...
+				conn.SetReadDeadline(time.Now())
+				conn.Close()
+				wg.Wait()
+				return
+			case reason := <-reasonCh:
+				fsm.logger.Debug("outgoing connection IO error", slog.String("reason", reason.String()))
+				conn.Close()
+				wg.Wait()
+				state = bgp.BGP_FSM_CONNECT
+				continue
```
maxime-peim, line 312: same here, the `recvCh` and `reasonCh` are not closed

pkg/server/fsm.go
```diff
@@ -206,6 +207,150 @@ func (s *adminStateRaw) Store(state adminState) {
...
+				wg.Wait()
+				nextState, _, notif := ocm.fsm.handleOpen(fmsg)
+				if nextState != bgp.BGP_FSM_OPENCONFIRM {
+					if notif != nil {
+						_ = fsm.sendNotification(conn, notif)
+					} else {
+						conn.Close()
+					}
+					state = bgp.BGP_FSM_CONNECT
+					continue
```
maxime-peim, line 323: same

pkg/server/fsm.go
fujita: I can't follow. At L.281, the connectLoop goroutine finished?

pkg/server/fsm.go
fujita: Who references recvCh and reaconCh? If nobody, GC reclaims them?

pkg/server/fsm.go
```diff
@@ -206,6 +207,150 @@ func (s *adminStateRaw) Store(state adminState) {
...
+			connCh := make(chan net.Conn, 1)
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go ocm.fsm.h.connectLoop(ocm.ctx, connCh, &wg)
+			select {
+			case <-ocm.ctx.Done():
+				wg.Wait()
+				return
+			case conn = <-connCh:
+				wg.Wait()
```
kerart: Hey @fujita, do you really need to spawn a goroutine for `ocm.fsm.h.connectLoop`
I see that you are doing `wg.Wait()` as the first step of all the cases. So, the only consequence is that if the connection is received before the context is Done during the run of `connectLoop`, you would still proceed with the connection.

In my view, it would be even better to check the context after the run of `connectLoop` entirely, don't you think?

fujita: Ah, indeed. You are right. We don't need to use goroutine here (and connCh channel) because context can cancel `connectLoop()` properly.

I'll update the code.

fujita:
> Hi @fujita It seems there can still be an issue with your approach, let's say we have two peers A and B that initiate a connection toward there other one: Let's name fA and fB the main fsm goroutine of A and B, and oA and oB the outgoing connection goroutines:
> 
> * oA and oB will try to connect.
> * fA and fB (in active state) accept the incoming connection, send an open message, and go to opensent state.
> * oA and oB also send an open message and go to opensent state.
> * you have then the four goroutines in opensent state
> * oA and oB wait to receive the corresponding open message from the other side.
> * fA and fB receive there open message before oA and oB
> * fA and fB check the outgoingConnectionCh and there is no outgoing connection since oA and oB are still waiting for there open message
> * fA and fB stop their outgoing connection manager
> * fA just closed the connection it has with oB and fB just closed it has with oA

Theoretically, it could happen, but I think it’s quite unlikely in practice.
Both fA and fB would have to receive their Open messages almost simultaneously.
In addition, the four FSMs have sent open messages the oA and oB would need to receive their Open messages before fA and fB have received them.
Even if this case does occur and both connections are dropped, the FSMs will restart from the IDLE state, and thanks to the random delay in the connect loop, the future attempt is expected to succeed.

So I think that the current code works but I'll fix incoming FSM to stop outgoing connection manager only when the manager is in CONNECT state.

fujita:
rebased and pushed.

Issue #3198, opened 2025-10-18
Fatal error in zclient: concurrent map iteration and map write


ponedo:
Hi. I am using zebra api to redistribute many (several hundreds) routes into GoBGP and running into fatal error below:

```
fatal error: concurrent map iteration and map write

goroutine 58 [running]:
internal/runtime/maps.fatal({0xf3d112?, 0xc000441d30?})
        /opt/hostedtoolcache/go/1.24.3/x64/src/runtime/panic.go:1058 +0x18
internal/runtime/maps.(*Iter).Next(0xc0002c5b28?)
        /opt/hostedtoolcache/go/1.24.3/x64/src/internal/runtime/maps/table.go:683 +0x86
github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).propagateUpdateToNeighbors(0xc00044c008, 0xc0003219c0, 0x0, 0xc0004581e0, {0xc0005240e0, 0x1, 0x1}, 0x1)
        /home/runner/work/gobgp/gobgp/pkg/server/server.go:1397 +0x1cf
github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).propagateUpdate(0xc00044c008, 0x0, {0xc0002c5f38, 0x1, 0x1?})
        /home/runner/work/gobgp/gobgp/pkg/server/server.go:1365 +0x665
github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).addPathList(0xc00044c008, {0x0?, 0xc0002c5ed8?}, {0xc0002c5f38, 0x1, 0x1})
        /home/runner/work/gobgp/gobgp/pkg/server/server.go:2319 +0x56
github.com/osrg/gobgp/v3/pkg/server.(*zebraClient).loop(0xc0003ea0a0)
        /home/runner/work/gobgp/gobgp/pkg/server/zclient.go:382 +0x51e
created by github.com/osrg/gobgp/v3/pkg/server.newZebraClient in goroutine 56
        /home/runner/work/gobgp/gobgp/pkg/server/zclient.go:532 +0x710

goroutine 1 [runnable]:
github.com/osrg/gobgp/v3/pkg/config.addNeighbors({0x106d768, 0x1787c00}, 0xc00044c008, {0xc0005a2000?, 0xd3f800?, 0xc0001334a0?})
        /home/runner/work/gobgp/gobgp/pkg/config/config.go:183 +0xcc
[...]

Issue #2153, opened 2019-09-01
fatal error: concurrent map iteration and map write | github.com/osrg/gobgp/internal/pkg/table.(*Table).Info(0xc00013a1d0, 0xcb8ca7, 0x6, 0xc000000000, 0x40e901)


amitn:
Importing full RIB and calling `gobgp global rib summary` results in a crash:

`gobgp mrt inject global -d  /tmp/latest-bview`

Where `latest-bview` is from [data.ris.ripe.net/rrc00/latest-bview.gz](http://data.ris.ripe.net/rrc00/latest-bview.gz)

```
goroutine 25 [running]:
runtime.throw(0xcd3d08, 0x26)
        /usr/lib/golang/src/runtime/panic.go:617 +0x72 fp=0xc000242bf0 sp=0xc000242bc0 pc=0x42fa32
runtime.mapiternext(0xc000242cc0)
        /usr/lib/golang/src/runtime/map.go:860 +0x597 fp=0xc000242c78 sp=0xc000242bf0 pc=0x411417
github.com/osrg/gobgp/internal/pkg/table.(*Table).Info(0xc00013a1d0, 0xcb8ca7, 0x6, 0xc000000000, 0x40e901)
        /labhome/amitn/storage/gobgp/internal/pkg/table/table.go:440 +0x97 fp=0xc000242d30 sp=0xc000242c78 pc=0xa4da37
github.com/osrg/gobgp/internal/pkg/table.(*TableManager).TableInfo(0xc000136660, 0xcb8ca7, 0x6, 0x0, 0x10001, 0xc000134c60, 0xc000242e68, 0x40e0b9)
        /labhome/amitn/storage/gobgp/internal/pkg/table/table_manager.go:363 +0x168 fp=0xc000242d98 sp=0xc000242d30 pc=0xa522c8
github.com/osrg/gobgp/pkg/server.(*BgpServer).getRibInfo.func1(0xc0000e5b00, 0x0)
        /labhome/amitn/storage/gobgp/pkg/server/server.go:2577 +0x91 fp=0xc000242e70 sp=0xc000242d98 pc=0xac56c1
github.com/osrg/gobgp/pkg/server.(*BgpServer).handleMGMTOp(0xc0000e5b00, 0xc0072c4100)
        /labhome/amitn/storage/gobgp/pkg/server/server.go:217 +0x34 fp=0xc000242ea8 sp=0xc000242e70 pc=0xa93ad4
[...]

PR #3207, opened 2025-10-21
Fix fatal error in zclient

ponedo:
This pull request is solution of issue https://github.com/osrg/gobgp/issues/3198.

Only one line is modified. I just replace unsafe `addPathList` with contention-safe `addPathStream` in zclient.go.

The solution resembles that in issue https://github.com/osrg/gobgp/issues/2153 (commit https://github.com/osrg/gobgp/commit/07e70de118842609a8e1e4740fdb82179f4bdaca). Very simple but useful. Feel like it can be merged instantly.


fujita:
Thanks

Issue #3214, opened 2025-10-28
RIB showing unspecified next-hops when only link-local hop is provided by neighbour


TimVosch:
Hi,

I'll preface this by saying I do not have advanced knowledge about BGP specs.

I am running on DN42, my peer is using BIRD, and I am running an IPv6 only stack. My gobgp shows an established connection with 900+ accepted routes. The peer provides me with the following output for troubleshooting:

<details><summary>BIRD Details running on peer</summary>
<pre>
Name       Proto      Table      State  Since         Info
AS4242420700_v4 BGP        ---        start  2025-10-26 11:06:58  Connect       
  BGP state:          Connect
    Neighbor address: 192.168.222.127
    Neighbor AS:      4242420700
    Local AS:         4242423914
  Channel ipv4
    State:          DOWN
    Table:          t_bgp_v4
    Preference:     100
    Input filter:   (unnamed)
    Output filter:  (unnamed)
    Import limit:   5000
      Action:       block

Name       Proto      Table      State  Since         Info
AS4242420700_v6 BGP        ---        up     2025-10-26 23:08:31  Established   
  BGP state:          Established
    Neighbor address: fe80::ade1%4242420700
    Neighbor AS:      4242420700
    Local AS:         4242423914
    Neighbor ID:      10.188.100.10
    Local capabilities
      Multiprotocol
        AF announced: ipv6
      Route refresh
      Graceful restart
      4-octet AS numbers
      Enhanced refresh
      Long-lived graceful restart
    Neighbor capabilities
      Multiprotocol
        AF announced: ipv6
      Route refresh
      4-octet AS numbers
      Hostname: ENTRY
    Session:          external AS4
    Source address:   fe80::ade0
    Hold timer:       77.333/90
    Keepalive timer:  0.746/30
    Send hold timer:  160.370/180
  Channel ipv6
    State:          UP
    Table:          t_bgp_v6
    Preference:     100
    Input filter:   (unnamed)
    Output filter:  (unnamed)
    Import limit:   5000
      Action:       block
    Routes:         0 imported, 926 exported, 0 preferred
    Route change stats:     received   rejected   filtered    ignored   accepted
      Import updates:              0          0          0          0          0
      Import withdraws:            0          0        ---          0          0
      Export updates:          27429         12          0        ---      27417
      Export withdraws:          473        ---        ---        ---        473
    BGP Next hop:   :: fe80::ade0
</pre>
</details>

[...]

PR #3215, opened 2025-10-29
WIP: fix(#3214): tablemanager use link-local nexthop if global is unspecified

TimVosch:
Work In Progress. Showcasing a quick fix for #3214.

When BIRD peers using only a link-local address it supplies this in a 32byte next-hop as `:: LL` (i.e. `::   fe80::ade0`). FRR supports this format as well ([see this issue](https://github.com/FRRouting/frr/issues/6259)).

Though not entirely related, there's a draft RFC about this [here](https://www.ietf.org/archive/id/draft-ietf-idr-linklocal-capability-02.html).

This PR adds logic to the table manager that: if the nexthop is unspecified and the LinkLocalNexthop is valid and actually a local-link, then uses the LinkLocalNextHop for the path.

Some questions and statements floating in my mind while working on this:

A new [RFC Draft](https://www.ietf.org/archive/id/draft-ietf-idr-linklocal-capability-02.html) for LinkLocal only capable peers state that a peer that advertises ONLY a LinkLocal Nexthop MUST encode this in 16 bytes nexthop. 
This would mean that while decoding the nexthop in [bgp.go](https://github.com/TimVosch/gobgp/blob/fix-3214-ll-only/pkg/packet/bgp/bgp.go) the `BGP_ATTR_NHLEN_IPV6_GLOBAL` actually means `BGP_ATTR_NHLEN_IPV6_GLOBAL_OR_LINKLOCAL`.

This might also - at least for me - require a concrete definition for `PathAttributeMpReachNLRI.NextHop` In [bgp.go](https://github.com/TimVosch/gobgp/blob/4c62dacd66a8e81c5d149d85f6eb0eedd1ab3017/pkg/packet/bgp/bgp.go#L11809-L11810). Is the field NextHop only Global or can it also be the LinkLocal? What is the LinkLocalNextHop field in that case.
**2025-10-31**: [path.go](https://github.com/osrg/gobgp/blob/master/internal/pkg/table/path.go) derives whether its local from the fact it uses a LocalAddress vs Address of a peer. This would argue in favor of NextHop being global address only.

### DIFF
--- a/internal/pkg/table/table_manager.go
+++ b/internal/pkg/table/table_manager.go
@@ -90,6 +90,10 @@ func ProcessMessage(m *bgp.BGPMessage, peerInfo *PeerInfo, timestamp time.Time, 
 		nexthop := reach.Nexthop
 		family := bgp.NewFamily(reach.AFI, reach.SAFI)
 
+		if nexthop.IsUnspecified() && reach.LinkLocalNexthop.IsValid() && !reach.LinkLocalNexthop.IsUnspecified() {
+			nexthop = reach.LinkLocalNexthop
+		}
+
 		for _, nlri := range reach.Value {
 			// when build path from reach
 			// reachAttrs might not contain next_hop if `attrs` does not have one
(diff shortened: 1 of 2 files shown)


fujita:
Fixed in master by da33165a,
acd4aba2 and 9fd6457f, so I am closing this.

The fix took a different shape. It keeps both addresses in the
attribute and picks the link-local one where the next hop is used,
instead of overwriting the global one. A peer that shares the link then
still gets the next hop in the form it was received, which RFC 2545
section 3 asks for.

Your patch also showed the reason the RIB held `::`. gobgp dropped the
link-local address when it built a path from a received UPDATE.

PR #3216, opened 2025-10-30
fix: remove logging on connection closing

maxime-peim:
The FSM handling of a TCP connection is pretty messy, and a connection can be closed mutliple times, examples:
- in established state, if the fsm context is done, sendNotification will close the connection, then from the fsm loop, logging an error
- in any state, if a reason is received on the reasonCh, we close the connection, which get closed in the fsm loop
- in the established state, if the hold time expires, sendNotification closes the connection, then push to the reasonCh which closes the connection, and then the fsm loop closes the connection...

### DIFF
--- a/pkg/server/fsm.go
+++ b/pkg/server/fsm.go
@@ -1997,10 +1997,7 @@ func (h *fsmHandler) loop(ctx context.Context, wg *sync.WaitGroup) {
 	default:
 	}
 	if fsm.conn != nil {
-		err := fsm.conn.Close()
-		if err != nil {
-			fsm.logger.Error("failed to close existing tcp connection", slog.String("State", oldState.String()))
-		}
+		fsm.conn.Close()
 	}
 	close(fsm.connCh)
 	cleanInfiniteChannel(fsm.outgoingCh)


fujita:
Closing a socket multiple times causes any real problem?

mpeim:
There a log printed each time at the end of the fsm loop, this is kind of annoying to get an error when you shouldn't.

maxime-peim:
> There a log printed each time at the end of the fsm loop, this is kind of annoying to get an error when you shouldn't.
> 
> 

To put it differently, the error is currently not logged, we just tell the user that an error happen when closing the connection.
So, currently, since we always double close, we always log an error. This is telling us nothing then.
Furthermore, if there is really an error when closing the connection it would be on the first close and we will never log it.

I have updated the PR so that if an error when closing the connection it will be returned every time we close the connection afterward, hence correctly logging the error if it has happened.

If we actually don't care about the error, then we shouldn't log something.

fujita:
> If we actually don't care about the error, then we shouldn't log something.

So let's just remove the logging code since the problem is just annoying to get an error.  I don't think that OS would return an error that a users need to investigate when closing a tcp socket.

fujita:
> > If we actually don't care about the error, then we shouldn't log something.
> 
> So let's just remove the logging code since the problem is just annoying to get an error. I don't think that OS would return an error that a users need to investigate when closing a tcp socket.

Just to make my intention clear, I don’t think we need a new mechanism to avoid closing TCP sockets multiple times, as proposed in this PR. Let’s simply remove that logging code.

fujita:
rebased and pushed. thanks.

PR #3217, opened 2025-10-30
fix: drain channel

maxime-peim:
when a channel is closed, we can still receive indefinitely nil element from it. Check https://go.dev/play/p/VZ6rskaydvF
This is what make sometimes the unittest to fail.


fujita:
rebased and pushed, thanks!

PR #3237, opened 2025-11-03
fix: fsm connection write deadline

maxime-peim:
A deadline of 1 second is set in the connection initialisation. This deadline is never changed until the connection gets in the established state.
So, if the connection doesn't send its open message and its first keepalive under 1 second we never get a peering.
Under heavy load this is an issue, while testing with 10k peers without routes, we never connect most of the peers. Instead, it is preferrable to set a deadline of 1s for each write.


fujita:
10k peer? It should be fun. Thanks!

PR #3242, opened 2025-11-19
server: Avoid overwriting log fields in adj table

rastislavs:
Do not provide peer-specific logger when creating adj table, but use parent logger instead.
Peer-specific logger contains "Topic" and "Key" log fields, which are also set when logging table events, but with different values. This is unnecessary and can be even deemed invalid in more strict logging setups.

(This was detected by the CI in Cilium, and seems to be the only occurrence of log fields overwrite in the GoBGP codebase)


fujita:
make sense, pushed. Thanks!

PR #3243, opened 2025-11-19
config: Fix RouteReflectorClusterId setting if configured

rastislavs:
If `RouteReflectorClusterId` is configured for a `Neighbor` (`.RouteReflector.Config.RouteReflectorClusterId` is valid), it should be propagated into `.RouteReflector.State.RouteReflectorClusterId`, as that is the source of truth for the server / table operations. Before this fix, it was set only by the defaulting logic, when not explicitly configured.


fujita:
Nice catch, thanks!

PR #3253, opened 2025-11-26
table: Fix regexp for parsing large community

rastislavs:
If the large community value is provided as a regexp including start/end of string anchors to `ParseLargeCommunityRegexp`, e.g. `"^1111:2222:3333$"`, the start/end anchors are duplicated in `ParseLargeCommunityRegexp`, e.g. to `"^^1111:2222:3333$$"`.

This fixes the `_regexpCommunityLarge` so that it matches only the large community values without any prefixes/suffixes, which avoids the problem.


fujita:
Thanks!

PR #3254, opened 2025-11-27
README.md: add GoBGP.nix external reference

DerDennisOP:
Would be nice if we could get this through.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Try [a binary release](https://github.com/osrg/gobgp/releases/latest).
 ### Externals
 
 - [Tutorial: Using GoBGP as an IXP connecting router](http://www.slideshare.net/shusugimoto1986/tutorial-using-gobgp-as-an-ixp-connecting-router)
+- [GoBGP.nix: A NixOS module for GoBGP. Containing a working FRR implementation and a rich set of Options](https://github.com/wavelens/gobgp.nix)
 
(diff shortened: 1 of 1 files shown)


fujita:
Nikpkgs has GoBGP?

DerDennisOP:
Yeah, but the package is missing quite a lot, e.g. command completion is not installed and it does not work with frr (because nixpkgs has a newer version of frr). With this module the Zebra integration works. Also this version has the correct types for all options and a better config structure: see https://wavelens.github.io/gobgp.nix and https://github.com/wavelens/gobgp.nix/blob/main/module/config/types.nix. Getting a such big module upstream to nixpkgs is going to be hard, and not really welcome, because the maintainability is a little bit harder for many options and obviously its not allowed to downgrade packages (frr) to make another tool work. 

If I may ask is there a gobgp config validator (e.g. gobgb --config [file] and it returns the correct exit code)? It would be useful to check the config before deployment automatically.

fujita:
Let’s ask the maintainer for their opinion.
@higebu

DerDennisOP:
he does not seem to response.

side note about the config validator: there is --dry-run, but this needs all networking interfaces available, i want to validate the config on other machines.

fujita:
Please add the description about the difference with Nikpkgs.

fujita:
merged.

PR #3260, opened 2025-12-15
Revert "path attributes hash, fast compare"

nplanelcisco:
As it break bgp packer and make benchmark slow and failed to sync all peers to connect and send routes.

This reverts commit 20f69ea2ea611d8b1f80cd4efffa5416ed644e4b.


fujita:
Thanks, can you explain why the commit breaks the packer code?

nplanelcisco:
let's close this PR, I will propose a fix instead of this one

PR #3261, opened 2025-12-17
fix: data race in toConfig() when reading peer AfiSafis configuration

okharch:
## Problem

A data race was detected by the Go race detector in production:

```
WARNING: DATA RACE
Write at 0x00c0002f82cb by goroutine 9:
  capabilitiesFromConfig()
      pkg/server/fsm.go:1056
  (*BgpServer).toConfig()
      pkg/server/server.go:758

Previous read at 0x00c0002f82c8 by goroutine 15:
  (*BgpServer).toConfig()
      pkg/server/server.go:735
```

The race occurs when `capabilitiesFromConfig()` writes to `AfiSafis[i].MpGracefulRestart.State.Advertised` while another goroutine is iterating over the same slice elements.

## Root Cause

The original `toConfig()` function had two issues:

### 1. Shallow copy of AfiSafis

Only the slice header was copied, not the elements. This allowed concurrent access to the underlying array:

```go
peerAfiSafis := peer.fsm.pConf.AfiSafis  // shallow copy - still points to same elements!
```

### 2. Multiple lock sections with gaps

The function used 3 separate `RLock` + 1 `Lock` sections with unlocked gaps between them:

```go
// OLD: 4 separate lock sections with race-prone gaps
peer.fsm.lock.Lock()    // 1st lock
conf := *peer.fsm.pConf
peerAfiSafis := peer.fsm.pConf.AfiSafis  // shallow copy!
peer.fsm.lock.Unlock()

// ... iterate over peerAfiSafis WITHOUT lock (RACE!) ...

peer.fsm.lock.Lock()    // 2nd lock  
conf.State.LocalCapabilityList = capabilitiesFromConfig(peer.fsm.pConf)
peer.fsm.lock.Unlock()

peer.fsm.lock.Lock()    // 3rd lock
conf.State.SessionState = ...
peer.fsm.lock.Unlock()

if state == bgp.BGP_FSM_ESTABLISHED {
    peer.fsm.lock.Lock()  // 4th lock
    // read recvOpen
    peer.fsm.lock.Unlock()
}
```

### 3. Unintended side effect

`capabilitiesFromConfig(peer.fsm.pConf)` was called with the **live** peer config, which has a side effect of modifying `peer.fsm.pConf.AfiSafis[i].MpGracefulRestart.State.Advertised`. However:

- The returned config already had a **stale** copy of AfiSafis (copied before this call)
- This side effect was not intentional for a **read** operation like `toConfig()`
- The `Advertised` flag is properly set by `buildopen()` when FSM actually sends OPEN messages

The code comment in `capabilitiesFromConfig()` confirms this:
```go
// Update advertised flag on sending new OPEN message. Retain advertised flag on PeerDown
// event since remote peer might treat it as Graceful Restart and in this case, GR caps
// are still "advertised" to it. However, if config is changed to disabled, reset it.
```

The flag should be updated when **sending OPEN messages** (via `buildopen()` in FSM), not when reading config.

## Reproducing the Race

Unit tests are included that reproduce the race on unfixed code:

```bash
go test -race -count=1 ./pkg/server/... -run TestRace_SoftResetPeerAndWatch
```

**Test `TestRace_SoftResetPeerAndWatch`** creates concurrent goroutines that:
- Call `newWatchEventPeer()` which triggers `buildopen()` → `capabilitiesFromConfig()` → WRITES to AfiSafis
- Call `toConfig()` directly which READS from AfiSafis
- Use API methods (`ListPeer`, `DisablePeer`, `EnablePeer`) to exercise real code paths

### 1. Add `copyAfiSafis()` helper for deep copy

```go
// Caller must hold peer.fsm.lock.RLock().
func (peer *peer) copyAfiSafis() []oc.AfiSafi {
    afiSafis := make([]oc.AfiSafi, len(peer.fsm.pConf.AfiSafis))
    copy(afiSafis, peer.fsm.pConf.AfiSafis)
    return afiSafis
}
```

### 2. Consolidate `toConfig()` to single lock section

```go
// NEW: Single lock section - consistent snapshot
peer.fsm.lock.Lock()
conf := *peer.fsm.pConf
peerCapMap := peer.fsm.capMap
conf.AfiSafis = peer.copyAfiSafis()  // deep copy!
state := peer.State()
conf.State.SessionState = ...
conf.State.AdminState = ...
conf.State.Flops = ...
if state == bgp.BGP_FSM_ESTABLISHED {
    // read recvOpen atomically with state check
}
peer.fsm.lock.Unlock()

// Safe to work on copy outside lock
for i := range conf.AfiSafis {
    // modify our copy
}
conf.State.LocalCapabilityList = capabilitiesFromConfig(&conf)  // operates on COPY
```

### 3. Remove unintended side effect

`capabilitiesFromConfig(&conf)` now operates on the **local copy**, not the live peer config. 

The `Advertised` flag is still correctly set in the right places:
- `buildopen()` in FSM (fsm.go:282, 904) - when actually sending OPEN messages
- `newWatchEventPeer()` (server.go:833) - for watch events

### 4. Fix other race locations

- `peer.go`: `updatePrefixLimitConfig()` now uses `copyAfiSafis()`
- `peer.go`: `handleUpdate()` now uses `copyAfiSafis()` for prefix limit check
- `server.go`: `handleFSMMessage()` EOR processing now uses `copyAfiSafis()`

### 5. Fix minor efficiency bug

`capAddPathFromConfig()` was called twice in `capabilitiesFromConfig()`:
```go
// Before (bug):
if c := capAddPathFromConfig(pConf); c != nil {
    caps = append(caps, capAddPathFromConfig(pConf))  // called twice!
}

// After (fixed):
if c := capAddPathFromConfig(pConf); c != nil {
    caps = append(caps, c)  // reuse result
}
```

### Race verification tests (should PASS after fix)

```bash
go test -race -count=1 ./pkg/server/... -run "TestRace_"
```

- `TestRace_SoftResetPeerAndWatch` - Tests toConfig race
- `TestRace_UpdatePrefixLimitConfig` - Tests updatePrefixLimitConfig race  
- `TestRace_HandleUpdatePrefixLimit` - Tests handleUpdate race
- `TestRace_HandleFSMMessageEOR` - Tests handleFSMMessage race

### Full test suite

```bash
go test -race -count=1 ./pkg/server/...
```

All tests pass.

## References

This follows the locking guidelines from CONTRIBUTING.md:

> "consider copying data before releasing locks to minimize contention"

## Files Changed

- `pkg/server/peer.go` - Add `copyAfiSafis()`, fix `updatePrefixLimitConfig()`, fix `handleUpdate()`
- `pkg/server/server.go` - Fix `toConfig()`, fix `handleFSMMessage()`
- `pkg/server/fsm.go` - Fix duplicate `capAddPathFromConfig()` call
- `pkg/server/race_softreset_test.go` - Race reproducer test
- `pkg/server/race_afisafis_test.go` - Additional race verification tests

### DIFF
--- a/pkg/server/race_afisafis_test.go
+++ b/pkg/server/race_afisafis_test.go
@@ -25,10 +25,13 @@ func TestRace_UpdatePrefixLimitConfig(t *testing.T) {
 	defer s.StopBgp(context.Background(), &api.StopBgpRequest{})
 
 	var testPeer *peer
-	s.mgmtOperation(func() error {
+	err := s.mgmtOperation(func() error {
 		testPeer = s.neighborMap[peerAddrIP]
 		return nil
 	}, true)
+	if err != nil {
+		t.Fatalf("mgmtOperation failed: %v", err)
+	}
 
 	if testPeer == nil {
 		t.Fatal("Could not get internal peer object")
@@ -86,10 +89,13 @@ func TestRace_HandleUpdatePrefixLimit(t *testing.T) {
 	defer s.StopBgp(context.Background(), &api.StopBgpRequest{})
 
 	var testPeer *peer
-	s.mgmtOperation(func() error {
+	err := s.mgmtOperation(func() error {
 		testPeer = s.neighborMap[peerAddrIP]
 		return nil
 	}, true)
+	if err != nil {
+		t.Fatalf("mgmtOperation failed: %v", err)
+	}
 
 	if testPeer == nil {
 		t.Fatal("Could not get internal peer object")
@@ -150,10 +156,13 @@ func TestRace_HandleFSMMessageEOR(t *testing.T) {
 	defer s.StopBgp(context.Background(), &api.StopBgpRequest{})
 
 	var testPeer *peer
-	s.mgmtOperation(func() error {
+	err := s.mgmtOperation(func() error {
 		testPeer = s.neighborMap[peerAddrIP]
 		return nil
 	}, true)
+	if err != nil {
+		t.Fatalf("mgmtOperation failed: %v", err)
+	}
 
 	if testPeer == nil {
 		t.Fatal("Could not get internal peer object")
@@ -205,4 +214,3 @@ func TestRace_HandleFSMMessageEOR(t *testing.T) {
 	close(stop)
 	wg.Wait()
 }
-
--- a/pkg/server/race_softreset_test.go
+++ b/pkg/server/race_softreset_test.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2024 Cisco Systems, Inc.
+// Copyright (C) 2026 Cisco Systems, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


fujita:
Can you fix lint errors?

PR #3274, opened 2025-12-29
fix: rtc deferral timer expired bugs

Ivan-Pokhabov:
PR bug fix:

1. RTC timer expired for some families may work after RTC EOR received, so we will not send anything even EOR. So in this case I want to offer make this once for all families
2. When our connection flapping we don't reset out deferral funcs so we can start send routes to early because of old timer func

### DIFF
--- a/pkg/server/fsm_test.go
+++ b/pkg/server/fsm_test.go
@@ -60,19 +60,23 @@ type MockConnection struct {
 	net.Conn
 	remote net.Conn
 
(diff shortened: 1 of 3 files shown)


Ivan-Pokhabov:
@fujita, hello! I resolved and fixed mistakes that copilot find, can you check the PR, please?

fujita:
Pushed, thanks!

PR #3275, opened 2025-12-30
Properly handle peer's zone information

YutaroHayakawa:
When we propagate the fsm.pConf.Transport.State.RemoteAddress => peer.peerInfo => path.info.source, we drop the IPv6 zone information with the operation like .WithZone("") or .AsSlice() in several places. As a result, some path-related API calls return wrong peer address when the IPv6 link-local address is used (e.g. BGP Unnumbered).

ListPath/WatchEvent for paths

```
{
  "2001:db8::/64": [
    {
      "Family": 0,
      "nlri": {
        "prefix": "2001:db8::/64"
      },
      ...
      "stale": false,
      "peer-id": "10.0.0.1",
      "peer-address": "fe80::ac84:abff:fea4:2ed8", <= Missing zone
      "RemoteID": 0,
      "LocalID": 0
    }
  ]
}
```

This commit fixes it to propagate zone information properly

```
{
  "2001:db8::/64": [
    {
      "Family": 0,
      "nlri": {
        "prefix": "2001:db8::/64"
      },
      ...
      "stale": false,
      "peer-id": "10.0.0.1",
      "peer-address": "fe80::ac84:abff:fea4:2ed8%eth0", <= Contains zone
      "RemoteID": 0,
      "LocalID": 0
    }
  ]
}
```

### DIFF
--- a/internal/pkg/table/path.go
+++ b/internal/pkg/table/path.go
@@ -1091,7 +1091,7 @@ func (path *Path) MarshalJSON() ([]byte, error) {
 		Withdrawal bool                         `json:"withdrawal,omitempty"`
 		Validation string                       `json:"validation,omitempty"`
 		SourceID   net.IP                       `json:"source-id,omitempty"`
-		NeighborIP net.IP                       `json:"neighbor-ip,omitempty"`
+		NeighborIP netip.Addr                   `json:"neighbor-ip,omitempty"`
 		Stale      bool                         `json:"stale,omitempty"`
 		UUID       string                       `json:"uuid,omitempty"`
 		ID         uint32                       `json:"id,omitempty"`
@@ -1101,7 +1101,7 @@ func (path *Path) MarshalJSON() ([]byte, error) {
 		Age:        path.GetTimestamp().Unix(),
 		Withdrawal: path.IsWithdraw,
 		SourceID:   path.GetSource().ID.AsSlice(),
-		NeighborIP: path.GetSource().Address.AsSlice(),
+		NeighborIP: path.GetSource().Address,
 		Stale:      path.IsStale(),
 		ID:         path.remoteID,
 	})
--- a/internal/pkg/table/table_manager.go
+++ b/internal/pkg/table/table_manager.go
@@ -263,7 +263,7 @@ func (manager *TableManager) handleMacMobility(path *Path) []*Path {
 		return nil
 	}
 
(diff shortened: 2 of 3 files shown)


fujita:
This remoteAddr will be stored in PeerInfo and used widely. Should we keep the zone information internally?

We currently use the remote address without the zone as the neighborMap key, right? If so, what’s the intended guideline for when to keep vs strip the zone?

YutaroHayakawa:
> We currently use the remote address without the zone as the neighborMap key, right?

I don't think so. Since b5e996b70482de383266942d686fbe1f8ca5cb31, we use netip.Addr for neighborMap's key and netip.Addr contains zone information ([this](https://github.com/osrg/gobgp/commit/b5e996b70482de383266942d686fbe1f8ca5cb31#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R3244) `MustParseAddr` can return the Addr with zone).

> If so, what’s the intended guideline for when to keep vs strip the zone?

I think for the IPv6 link-local neighbor, we should never drop zone information because, it's possible to have multiple peer with same link-local address in different zone in theory. Without the zone, we cannot identify the neighbor correctly.

The case we can drop the zone is when we encode link-local address to the MP_REACH_NLRI's NextHop field as we cannot encode zone in there.

fujita:
> > We currently use the remote address without the zone as the neighborMap key, right?
> 
> I don't think so. Since [b5e996b](https://github.com/osrg/gobgp/commit/b5e996b70482de383266942d686fbe1f8ca5cb31), we use netip.Addr for neighborMap's key and netip.Addr contains zone information ([this](https://github.com/osrg/gobgp/commit/b5e996b70482de383266942d686fbe1f8ca5cb31#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R3244) `MustParseAddr` can return the Addr with zone).

Ah, you are right.

> > If so, what’s the intended guideline for when to keep vs strip the zone?
> 
> I think for the IPv6 link-local neighbor, we should never drop zone information because, it's possible to have multiple peer with same link-local address in different zone in theory. Without the zone, we cannot identify the neighbor correctly.
> 
> The case we can drop the zone is when we encode link-local address to the MP_REACH_NLRI's NextHop field as we cannot encode zone in there.

Sounds reasonable.

Merged, thanks a lot!

PR #3276, opened 2025-12-31
Properly handle local zone information

YutaroHayakawa:
When we populate the fsm.pConf.Transport.State.LocalAddress, we drop the IPv6 zone information with .WithZone(""). As a result, the api.Peer.Transport.LocalAddress has wrong local address when the IPv6 link-local address is used (e.g. BGP Unnumbered).

```
$ gobgp neighbor -j | jq '.[0].transport'
{
  "local_address": "fe80::ac84:abff:fea4:2ed8" <= Missing zone
}
```

This commit fixes it.

```
$ gobgp neighbor -j | jq '.[0].transport'
{
  "local_address": "fe80::ac84:abff:fea4:2ed8%eth0" <= Contains zone
}
```


fujita:
The changes look good. Please resolve the conflicts.
Thanks!

fujita:
Resolved the conflict and pushed, thanks a lot!

PR #3277, opened 2025-12-31
bmp: Use netip.AddrPort to identify client

YutaroHayakawa:
Identify BMP client with netip.AddrPort instead of string addr:port. Also piggybacking some minor cleanups.

- Eliminate unnecessary conversion that converts netip.Addr to string and parse it using netip.MustParseAddr().
- Replace netip.MustParseAddr("0.0.0.0") with netip.IPv4Unspecified().


fujita:
Nice work, thanks a lot!

PR #3278, opened 2025-12-31
Migrate path.SetNexthop to netip types

YutaroHayakawa:
The path.SetNexthop() still takes net.IP as an argument. Migrating it to netip.Addr simplifies good amount of code. Also migrating PolicyOptions.OldNextHop and NextHopSet.list to use netip types as some callsites pass them to path.SetNexthop().


fujita:
This has been on my to-do list for a while. I really appreciate you making it happen!

PR #3285, opened 2026-01-08
Add CompareAndSwap for adminStateRaw

nplanelcisco:
Aim to have safer code, as this state is atomic, in the idea to remove the fsm lock and relaying on atomic


fujita:
make sense, thanks.

PR #3286, opened 2026-01-08
Call initializeConn directly under lock and remove useless fsm.lock

nplanelcisco:


pkg/server/fsm.go
```diff
@@ -208,7 +208,6 @@ func (s *adminStateRaw) Store(state adminState) {
 }
 
 func initializeConn(fsm *fsm, conn net.Conn) {
-	fsm.lock.Lock()
```
fujita, line 211: setPeerConnTTL reads config parameters such as fsm.pConf.TtlSecurity.Config.TtlMin? Could it race with updateNeighbor() which might update those values?

pkg/server/server.go
```diff
@@ -1368,9 +1368,7 @@ func (s *BgpServer) handleFSMMessage(peer *peer, e *fsmMsg) {
 		peer.fsm.lock.Lock()
 		oldState := bgp.FSMState(peer.fsm.pConf.State.SessionState.ToInt())
 		peer.fsm.pConf.State.SessionState = oc.IntToSessionStateMap[int(nextState)]
-		peer.fsm.lock.Unlock()
```
fujita, line 1371: This looks fine. If you create a separate pull request, I'll merge.

pkg/server/fsm.go
fujita: Scratch this comment, I misunderstood.

fujita:
Merged, thanks.

PR #3288, opened 2026-01-08
fsm: mostly read peer.fsm.pConf

nplanelcisco:
The logic is to use atomic.Value to store the pointer
and use fsm.lock Mutex for a write transaction in the fsm.pConf

Most fields are read-only, so they can be accessed an atomic copy via fsm.pConf.Load().Field.

For write fields, state messages statistics and timer state update are written via the transaction under lock

ref/idea from https://pkg.go.dev/sync/atomic@master#example-Value-ReadMostly
fix https://github.com/osrg/gobgp/pull/3261
include race test from https://github.com/osrg/gobgp/pull/3261 thanks @okharch


pkg/config/oc/bgp_configs.go
```diff
@@ -2672,6 +2672,7 @@ func (lhs *Transport) Equal(rhs *Transport) bool {
 // State information relating to the timers used for the BGP
 // neighbor or group.
 type TimersState struct {
+	Padding_for_i386 uint16
```
nplanelcisco: @fujita does it still make sense (required ?) to support x86_32 ?
atomic need to be 64bits aligned

nplanelcisco:
moving to draft as we can probably solve the whole read mosly peer configuration like [this](https://pkg.go.dev/sync/atomic@master#example-Value-ReadMostly) in golang 
WDYT @fujita ?

fujita:
> moving to draft as we can probably solve the whole read mosly peer configuration like [this](https://pkg.go.dev/sync/atomic@master#example-Value-ReadMostly) in golang WDYT @fujita ?

Yeah, need to think through the details, but it seems likely it would work nicely.

Making fsm.pConf an atomic pointer might be the simplest. For the frequently updated pConf.State.Messages and Timers, instead of mutating pConf directly, store them on fsm and update via atomic.AddUint64 etc., and have ListPeer() use those values.

This should fix the problem about https://github.com/osrg/gobgp/pull/3261

pkg/config/oc/bgp_configs.go
fujita: I think that float64 is 8‑byte aligned, so within a struct the compiler inserts any needed padding automatically. Typically you don’t need to add padding by hand.

nplanelcisco: If I don't add the padding it's aligned on 32bits on x86_32, but golang atomic require 64bits alignment, else panic at runtime

fujita: Hmm, however, I don't think that it's a good idea to add a padding:
- we should not modify bgp_configs.go by hand
- there is not atomic functions for float64 (such as AddFloat64). we can modify it atomically but it's hacky.

We often update only UpdateRecvTime. So I think that we had better to handle UpdateRecvTime differently. I think that adding UpdateRecvTime to TimersState is a mistake. Probably we should move it somewhere else in the next major update.

nplanelcisco: The limitation here is golang in 32bits and atomic https://github.com/golang/go/issues/36606

I agree it's not ideal to add padding by hand, so hence my initial question, does it still make sense to build x86_32 ? do you anyone using this target today ?

For float64 I think atomic.Value shall be used

fujita: I'm not sure there is anyone who still use x86_32 but I prefer to avoid atomic.Value for float. I prefer to store only UpdateRecvTime in a different place.

nplanelcisco: we talking about 2 differents things here
1/ `UpdateRecvTime` is `int64` so it's not a problem, furthermore the padding is needed for the state messages counters as well because `x86_32` align `int64` on 32 bits
2/ I use atomic.Value for the whole configuration as it's read mostly, please review the PR

I will try to find another way to align these fields

nplanelcisco: @fujita I removed the atomics for the moment, and just focus this PR on the mostly read behavior that I would like to introduce

pkg/server/fsm.go
```diff
@@ -350,11 +351,26 @@ func (ocm *outgoingConnManager) stop() {
 	}
 }
 
+type pConfAccess struct {
+	conf atomic.Value // oc.Neighbor
+}
+
+func (p *pConfAccess) Load() oc.Neighbor {
+	return *p.conf.Load().(*oc.Neighbor)
+}
```
fujita, line 375: Needs to do deep copy here?

pkg/server/server.go
```diff
@@ -623,11 +615,11 @@ func (s *BgpServer) prePolicyFilterpath(peer *peer, path, old *table.Path) (*tab
...
 		// When the local address contains zone, we need to strip it
 		// because BGP nexthop cannot contain zone information.
-		options.OldNextHop = peer.fsm.pConf.Transport.State.LocalAddress.WithZone("")
+		options.OldNextHop = conf.Transport.State.LocalAddress.WithZone("")
 	} else {
 		options.OldNextHop = path.GetNexthop()
 	}
-	path = table.UpdatePathAttrs(peer.fsm.logger, peer.fsm.gConf, peer.fsm.pConf, peer.peerInfo, path)
+	path = table.UpdatePathAttrs(peer.fsm.logger, peer.fsm.gConf, &conf, peer.peerInfo, path)
 	peer.fsm.lock.Unlock()
```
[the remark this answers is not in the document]
nplanelcisco: lock is been acquired on line 608

fujita:
Is the intent here to avoid read locks by deep‑copying pConf, while still guarding writes with fsm.lock to prevent lost updates? If so, could we clarify that deep‑copying (especially slices/maps like AfiSafis) is required for lock‑free reads?

pkg/server/server.go
```diff
@@ -756,29 +747,20 @@ func (s *BgpServer) toConfig(peer *peer, getAdvertised bool) *oc.Neighbor {
...
 	if state == bgp.BGP_FSM_ESTABLISHED {
-		peer.fsm.lock.Lock()
 		buf, _ := peer.fsm.recvOpen.Serialize()
 		// need to copy all values here
 		conf.State.ReceivedOpenMessage, _ = bgp.ParseBGPMessage(buf)
-		conf.State.RemoteRouterId = peer.fsm.pConf.State.RemoteRouterId
-		peer.fsm.lock.Unlock()
 	}
+
 	return &conf
```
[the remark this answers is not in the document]
nplanelcisco: in golang the object have a reference counter

pkg/server/server.go
```diff
@@ -1397,39 +1378,40 @@ func (s *BgpServer) handleFSMMessage(peer *peer, e *fsmMsg) {
...
 			s.propagateUpdate(peer, peer.DropAll(dropFamilies))
 
-			peer.fsm.lock.Lock()
-			if peer.fsm.pConf.Config.PeerAs == 0 {
-				peer.fsm.pConf.State.PeerAs = 0
+			if conf.Config.PeerAs == 0 {
+				peer.fsm.lock.Lock()
+				conf = peer.fsm.pConf.Load()
+				conf.State.PeerAs = 0
+				peer.fsm.pConf.Store(&conf)
```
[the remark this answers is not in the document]
nplanelcisco: as there are the same type, the variable can be re-used

pkg/server/server.go
```diff
@@ -3406,17 +3385,19 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
 		return needsSoftResetIn, fmt.Errorf("neighbor that has %v doesn't exist", addr)
 	}
 
-	if !peer.fsm.pConf.ApplyPolicy.Equal(&c.ApplyPolicy) {
+	original := peer.fsm.pConf.Load()
+	conf := peer.fsm.pConf.Load()
```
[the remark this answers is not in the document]
nplanelcisco: the original will be used as backup if some API call failed

pkg/server/fsm.go
nplanelcisco: thanks, I added copies here and an unit test

fujita: https://github.com/osrg/gobgp/pull/3261

The above already implemented it?

nplanelcisco:
> Is the intent here to avoid read locks by deep‑copying pConf, while still guarding writes with fsm.lock to prevent lost updates? If so, could we clarify that deep‑copying (especially slices/maps like AfiSafis) is required for lock‑free reads?

yes and yes slices need to be deep copied, thanks there no maps in the config
A better way would to split the pConf in the definition of fsm, basically copy/paste oc.Neighbor definition in fsm type, so we will have a better granularity and avoid deep copy all slices each load, WDYT ?

pkg/server/fsm.go
```diff
@@ -258,9 +258,8 @@ func (ocm *outgoingConnManager) run(ch chan<- outgoingConn) {
 	}()
 
 	fsm := ocm.fsm
-	fsm.lock.Lock()
-	isPassive := fsm.pConf.Transport.Config.PassiveMode
-	fsm.lock.Unlock()
+	conf := fsm.pConf.Load()
+	isPassive := conf.Transport.Config.PassiveMode
```
fujita, line 263: We copy pConf to read isPassive? we need pCOnfAccess's method to read pConf without copying?

nplanelcisco: yes we do a copy just for this flag, hence the idea of the split

Furthermore, in this case we shouldn't call the `go ocm.run()` in `newOutGoingConnManager`

pkg/server/server.go
```diff
@@ -4688,39 +4671,38 @@ func (s *BgpServer) watch(opts ...WatchOption) (w *watcher) {
 				for _, rf := range peer.configuredRFlist() {
 					peer.fsm.lock.Lock()
 					_, y := peer.fsm.capMap[bgp.BGP_CAP_FOUR_OCTET_AS_NUMBER]
```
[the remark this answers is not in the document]
nplanelcisco: protected by  peer.fsm.lock.Lock()

pkg/server/peer.go
```diff
@@ -566,8 +556,12 @@ func (peer *peer) isPrefixLimit(k bgp.Family, c *oc.PrefixLimitConfig) bool {
 
 func (peer *peer) updatePrefixLimitConfig(c []oc.AfiSafi) (bool, error) {
 	peer.fsm.lock.Lock()
-	x := peer.fsm.pConf.AfiSafis
+	conf := peer.fsm.pConf.Load()
+	conf.AfiSafis = c
+	peer.fsm.pConf.Store(&conf)
 	peer.fsm.lock.Unlock()
+
+	x := conf.AfiSafis
```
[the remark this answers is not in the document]
nplanelcisco: I revert to the old code behavior by assign it at the end

pkg/server/fsm.go
fujita: We use [ReadMostly](https://pkg.go.dev/sync/atomic#example-Value-ReadMostly) pattern, right? If so, we don't need copying to read pConf?

nplanelcisco: @fujita I moved the implementation/access in ReadOnly() ReadCopy() Update()
WDYT ?

fujita:
I like the ReadOnly()/ReadCopy() split; it’s a good direction.

One concern: ReadCopy() is taken without holding fsm.lock in updateNeighbor(). If another goroutine updates counters/timers under fsm.lock and calls pConf.Update() in between, the later Update() from updateNeighbor() can overwrite those newer changes with the older snapshot. That would silently roll back message counters, timers, GR state, etc.

Should we take fsm.lock before ReadCopy and hold it until Update() (or otherwise merge changes) to avoid lost updates?

Merging changes would be complicated so I prefer to hold the lock between ReadCopy() and Update(). We can improve the approach later in different commits.

nplanelcisco:
> I like the ReadOnly()/ReadCopy() split; it’s a good direction.
> 
> One concern: ReadCopy() is taken without holding fsm.lock in updateNeighbor(). If another goroutine updates counters/timers under fsm.lock and calls pConf.Update() in between, the later Update() from updateNeighbor() can overwrite those newer changes with the older snapshot. That would silently roll back message counters, timers, GR state, etc.
> 
> Should we take fsm.lock before ReadCopy and hold it until Update() (or otherwise merge changes) to avoid lost updates?
> 
> Merging changes would be complicated so I prefer to hold the lock between ReadCopy() and Update(). We can improve the approach later in different commits.

@fujita You're right it was missing a lock around the first update, I reviewed and move the lock on a wider scope and change the logic of updatePrefixLimitConfig().

pkg/server/server.go
```diff
@@ -3448,24 +3437,25 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
...
 		err = s.setAdminState(addr, "", adminStatePfxCt)
 	}
 	if err != nil {
 		peer.fsm.logger.Error("failed to update prefixLimit", slog.String("Err", err.Error()))
 		// rollback to original state
-		peer.fsm.pConf = original
+		peer.fsm.pConf.Update(original)
+		peer.fsm.lock.Unlock()
 		return needsSoftResetIn, err
 	}
```
fujita, line 3465: If peer.updatePrefixLimitConfig succeed, No Update() call?

pkg/server/server.go
```diff
@@ -3448,24 +3437,25 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
...
+	isLimit, err := peer.updatePrefixLimitConfig(&conf, c.AfiSafis)
 	if isLimit && err == nil {
 		err = s.setAdminState(addr, "", adminStatePfxCt)
 	}
 	if err != nil {
 		peer.fsm.logger.Error("failed to update prefixLimit", slog.String("Err", err.Error()))
 		// rollback to original state
-		peer.fsm.pConf = original
+		peer.fsm.pConf.Update(original)
+		peer.fsm.lock.Unlock()
```
fujita, line 3461: Do we need this rollback? It means that we lost some earlier updates?

If we like to make updateNeighbor atomic, after all the updates to copy is successful, we call pConf.Update()?

nplanelcisco: yes line 3404 (safiafi), but we can drop it with original
in favor of to update at the end
I will update

nplanelcisco: I added some comment when rolling back the old ApplyPolicy

pkg/server/server.go
```diff
@@ -3437,33 +3422,45 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
...
 		}
-		if err = s.deleteNeighbor(peer.fsm.pConf, bgp.BGP_ERROR_CEASE, sub, true); err != nil {
+		// we don't hold fsm lock during del/add peer, deleteNeighbor do read only config
+		peer.fsm.pConf.Update(&conf)
+		peer.fsm.lock.Unlock()
+
+		if err = s.deleteNeighbor(&conf, bgp.BGP_ERROR_CEASE, sub, true); err != nil {
+			// rollback to original ApplyPolicy
+			peer.fsm.pConf.Update(original)
+
```
fujita, line 3432: Update() is called without the lock held. We can't hold the lock during deleteNeighbor()?

pkg/server/server.go
```diff
@@ -3437,33 +3422,45 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
...
 			peer.fsm.logger.Error("failed to delete neighbor", slog.String("Err", err.Error()))
 			return needsSoftResetIn, err
 		}
 		err = s.addNeighbor(c)
 		if err != nil {
+			// rollback to original ApplyPolicy
+			peer.fsm.pConf.Update(original)
+
```
fujita, line 3440: Ditto.

pkg/server/server.go
```diff
@@ -3437,33 +3422,45 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
 		} else if original.Config.PeerAs != c.Config.PeerAs {
 			sub = bgp.BGP_ERROR_SUB_PEER_DECONFIGURED
 		}
-		if err = s.deleteNeighbor(peer.fsm.pConf, bgp.BGP_ERROR_CEASE, sub, true); err != nil {
+		// we don't hold fsm lock during del/add peer, deleteNeighbor do read only config
+		peer.fsm.pConf.Update(&conf)
```
fujita, line 3426: Do we need to call Update() here? We simply call deleteNeighbor so no need to update?

nplanelcisco: I think we want to save the ApplyPolicy before del/add Neighbor, hence the update here

pkg/server/server.go
nplanelcisco: We can't keep the fsm lock here as deleteNeighbor will do in propagateNeighbor so we will deadlock
the conf is only accessed read only in deleteNeighbor so it's not a problem (comment line 3425)

I rollback the original config on failure case, the update is atomic
Note we are already under server.shared.Write lock, so no reentrance in other API call or fsmHandler can be called
I can move back to the previous behavior that keep the new ApplyPolicy, but doesn't sound good/logic

fujita:
Pushed, thanks!

This improves the performance? We copy State.Messages every time we receive a message.

PR #3289, opened 2026-01-08
remove debug log as it's too noisy an on the hot path

nplanelcisco:
The issue here is on high load, previous pprof showed, that slog lock cost some time even if log level was set to info


fujita:
Thanks!

PR #3290, opened 2026-01-09
internal/destination: remove unused code

nplanelcisco:
used only in local package test, so field can be accessed


fujita:
pushed, thanks.

PR #3291, opened 2026-01-09
internal/table/destination: make it private as it's only used internally

nplanelcisco:
depend on https://github.com/osrg/gobgp/pull/3290


fujita:
Thanks!

PR #3303, opened 2026-01-21
Scaling up gobgp, handleFSMMessage re-entrant

nplanelcisco:
This PR will enable `handleFSMMessage` function to be re-entrant, this will allow peers to run in parallel
Notes:
 o adj tables are accessed under s.shared.Lock()
 o I ran some benchmark and fuzzing adding/delete peers and path dynamically under `-race`
 o ran on staging for a week with 2500 peers / 28 routes per peer

Depend on https://github.com/osrg/gobgp/pull/3332
Depend on https://github.com/osrg/gobgp/pull/3349

image

image

### DIFF
--- a/pkg/server/peer.go
+++ b/pkg/server/peer.go
@@ -96,6 +96,11 @@ func newDynamicPeer(g *oc.Global, neighborAddress string, pg *oc.PeerGroup, loc 
 	return newPeer(g, &conf, bgp.BGP_FSM_ACTIVE, loc, policy, logger)
 }
 
+// pathIDSet is the set of add-path local identifiers advertised for a destination.
+// Values of this type are stored in peer.sentPaths and MUST only be accessed
+// under the propagation bucket lock for the corresponding prefix.
+type pathIDSet map[uint32]struct{}
+
 type peer struct {
 	tableId           string
 	fsm               *fsm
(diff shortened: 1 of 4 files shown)


fujita:
I think that adding RWMutex to destination consumes too much memory. If we enable concurrent access to a table, sharding would be an option.
Making handleFSMMessage read lock needs more work. For example, stopNeighbor() manipulates neighborMap.

nplanelcisco:
> I think that adding RWMutex to destination consumes too much memory. If we enable concurrent access to a table, sharding would be an option. Making handleFSMMessage read lock needs more work. For example, stopNeighbor() manipulates neighborMap.

Thank you,
o fixed the stopNeighbor() by calling it under write lock
o limited the number of destination mutex by allocating lock bucket based on NLRI
o adding some FSM Handler race tests
o made EVPNMacNLRIs thread safe

internal/pkg/table/destination.go
```diff
@@ -136,14 +137,33 @@ func NewPeerInfo(g *oc.Global, p *oc.Neighbor, AS, localAS uint32, ID, localID n
 	}
 }
 
+// This global will be parameterizable in the future
+const destinationBucketLocksSize = uint32(2048)
+
```
fujita: The bucket count of 2048 seems excessive. Since no goroutine sleeps while holding a destination lock in the current design, the maximum number of goroutines contending for a lock at any instant is bounded by GOMAXPROCS?

fujita:
The bucket-based destination locking and the handleFSMMessage re-entrancy looks good improvements. Would it be possible to split this into two separate PRs?

PR 1: Destination lock granularity (table layer only)

- Bucket-based RWMutex on destination
- Thread-safe EVPNMacNLRIs
- TableManager.mu

PR 2: Make handleFSMMessage re-entrant (server layer)

- sharedData: sync.Mutex → sync.RWMutex
- Read-lock handleFSMMessage, upgrade to write lock only for stopNeighbor
- peer.mu for prefixLimitWarned and llgrEndChs

I think that PR 1 is self-contained and easier to review, and merging it.

internal/pkg/table/destination.go
fujita: I withdraw this comment. I'm not sure which value is appropriate here. I guess that It depends on the configuration and the routes that peers receive. It doesn't consume much memory so 2048 is fine by me.

nplanelcisco:
> The bucket-based destination locking and the handleFSMMessage re-entrancy looks good improvements. Would it be possible to split this into two separate PRs?
> 
> PR 1: Destination lock granularity (table layer only)
> 
> * Bucket-based RWMutex on destination
> * Thread-safe EVPNMacNLRIs
> * TableManager.mu
> 
> PR 2: Make handleFSMMessage re-entrant (server layer)
> 
> * sharedData: sync.Mutex → sync.RWMutex
> * Read-lock handleFSMMessage, upgrade to write lock only for stopNeighbor
> * peer.mu for prefixLimitWarned and llgrEndChs
> 
> I think that PR 1 is self-contained and easier to review, and merging it.

I splitted in two PRs
https://github.com/osrg/gobgp/pull/3332
https://github.com/osrg/gobgp/pull/3303

fujita:
Pushed, thanks!

PR #3306, opened 2026-01-29
atomic peer.peerInfo by using atomic.Pointer[table.PeerInfo]

nplanelcisco:


fujita:
Pushed, thanks! Sorry for the delay.

PR #3307, opened 2026-01-29
fix(StopBgp): if server is not serving, we cannot stop BGP server

nplanelcisco:
If the server is not serving, we don't want the caller to be stuck as the managment chan will never be read.


fujita:
Merged, sorry for the delay. I'm finally back.

PR #3310, opened 2026-02-04
Adj-Rib: fix leak memory on withdraw unknown path

taitov:
If you try to delete an unknown path from adj-rib, a key with an empty slice will be created, which will live until you delete adj-rib.


taitov:
Hi @fujita ! Please take a look at this little fix.

fujita:
Nice catch! The Copilot comment is reasonable but trivial. So I pushed this fix.

PR #3329, opened 2026-03-06
packet/bgp: Enhance opaque extended community printout

rastislavs:
The current Stringer outputs opaque extended community as a decimal value, which is hard to distinguish among other extended communities as well as hard to decode (packet analyzers usually print out opaque values in hex, not decimals). This makes the printout a bit more human-friendly.


rastislavs:
Ah, probably not a good idea, as it affects some ext communities where decimal makes sense (like RouteTargetMembershipNLRI)

PR #3334, opened 2026-03-10
Fix nil pointer panic in TestGRPCAddPathUpdatesUUIDMap

tkjaer:
This PR fixes a panic in `TestGRPCAddPathUpdatesUUIDMap`

Use `require.NoError` instead of `assert.NoError` for the AddPath gRPC call so the test stops immediately on connection failure rather than panicking on a `nil` response.


fujita:
Pushed, thanks!

PR #3349, opened 2026-03-19
feat: consistent route update

nplanelcisco:
Lock these path to be serialized in sync
* Initial establishment when not local restarting
* GR sync when local restarting ends
* GR sync when EOR received during local restarting
* RTC EOR wait

depend on
https://github.com/osrg/gobgp/pull/3347
https://github.com/osrg/gobgp/pull/3397

### DIFF
--- a/pkg/server/peer.go
+++ b/pkg/server/peer.go
@@ -111,6 +111,8 @@ type peer struct {
 	longLivedRunning    atomic.Bool
 	// Route Target Membership handler after import policy (for constrained VPN distribution).
 	rtmHandler *table.RouteTargetMembershipHandler
+	// Route refresh in progress, during an established session or route refresh, this need to be atomic to avoid out of order updates
+	routeRefreshInProgress sync.RWMutex
 }
 
 func newPeer(g *oc.Global, conf *oc.Neighbor, state bgp.FSMState, loc *table.TableManager, policy *table.RoutingPolicy, logger *slog.Logger) *peer {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -949,7 +949,19 @@ func (s *BgpServer) getPossibleBest(peer *peer, family bgp.Family) []*table.Path
 	return peer.localRib.GetBestPathList(peer.TableID(), peer.AS(), []bgp.Family{family})
 }
 
-func (s *BgpServer) getBestFromLocal(peer *peer, rfList []bgp.Family, addEOR bool) ([]*table.Path, []*table.Path) {
+func (s *BgpServer) getBestFromLocalCallback(peer *peer, rfList []bgp.Family, addEOR bool, routeRefresh bool, fn func([]*table.Path, []*table.Path)) {
+	if routeRefresh {
+		peer.routeRefreshInProgress.Lock()
+		defer peer.routeRefreshInProgress.Unlock()
+	} else {
+		peer.routeRefreshInProgress.RLock()
+		defer peer.routeRefreshInProgress.RUnlock()
+	}
+
+	s.getBestFromLocalCallbackLocked(peer, rfList, addEOR, fn)
+}
+
+func (s *BgpServer) getBestFromLocalCallbackLocked(peer *peer, rfList []bgp.Family, addEOR bool, fn func([]*table.Path, []*table.Path)) {
 	pathList := []*table.Path{}
 	filtered := []*table.Path{}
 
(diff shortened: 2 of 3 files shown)


fujita:
Before diving into the implementation details, I'd like to raise a concern about the interaction between shard-level dump tracking and concurrent update propagation.

Consider this scenario when combined with [parallel handleFSMMessage](https://github.com/osrg/gobgp/pull/3303)

```
Goroutine A (dump for peer X):     Goroutine B (update from peer Y):
──────────────────────────────     ──────────────────────────────────
shard[3] RLock
snapshot (contains pathA)
markShardDumped(3)
shard[3] RUnlock
                                   table.update(withdraw pathA in shard[3])
                                   propagateUpdateToNeighbors:
                                     peer X: isShardDumped(3)? yes → propagate
                                     → withdraw pathA enters peer X's outgoing channel
...still dumping other shards...
sendfsmOutgoingMsg(dump result)
  → pathA enters peer X's outgoing channel
```

fujita:
As an alternative to shard-level bitmap tracking, a simpler approach might be to queue all update propagations during the dump and replay them after the dump result is sent:

```
type peerDumpState struct {
    pendingCh *channels.InfiniteChannel
}

// In propagateUpdateToNeighbors:
targetPeer.dumpMu.Lock()
if targetPeer.dumpState != nil {
    targetPeer.dumpState.pendingCh.In() <- path
    targetPeer.dumpMu.Unlock()
    continue
}
targetPeer.dumpMu.Unlock()
// normal propagation

// Starting a dump:
peer.dumpMu.Lock()
peer.dumpState = &peerDumpState{pendingCh: channels.NewInfiniteChannel()}
peer.dumpMu.Unlock()

// dump shard by shard...
sendfsmOutgoingMsg(peer, dumpResult)

// Ending the dump:
peer.dumpMu.Lock()
pending := peer.dumpState.pendingCh
peer.dumpState = nil
peer.dumpMu.Unlock()
// drain pending and send
```

fujita:
@nplanelcisco Just following up on our offline discussion. I think locking all shard mutexes for the duration of the dump is the simplest approach, as you proposed. Since this only happens on peer establishment and ROUTE-REFRESH, the frequency is low enough that the performance impact should be acceptable.

fujita:
During the discussion, I was thinking of taking all shard mutexes for the duration of the initial dump to prevent concurrent table modifications. Could you explain why you chose to introduce a new per-peer routeRefreshInProgress mutex instead? I'm curious about the trade-offs you considered.

With this design, if a peer updates a route while the dump is in progress (after the shard has already been read but before routeRefreshInProgress is unlocked), the same route could be sent twice to the refreshing peer: once as part of the initial snapshot and again when the blocked propagation resumes after the unlock. Is this duplicate UPDATE considered acceptable?

fujita:
The Graceful Restart sync paths take only a Read Lock, which allows concurrent propagateUpdateToNeighbors to interleave with the initial route advertisement. Since GR sync also needs to send all routes followed by EOR (RFC 4724), the same Write Lock used for initial establishment seems appropriate here?

fujita:
Note that we need another PR before #3303; The watch() function performs initial RIB dumps via some initial options by calling w.notify() repeatedly in a loop. With parallel handleFSMMessage (PR #3303), concurrent calls to notifyWatcher can interleave live update events between those dump events in the watcher channel, potentially delivering an inconsistent snapshot to BMP receivers.

nplanelcisco:
> Note that we need another PR before #3303; The watch() function performs initial RIB dumps via some initial options by calling w.notify() repeatedly in a loop. With parallel handleFSMMessage (PR #3303), concurrent calls to notifyWatcher can interleave live update events between those dump events in the watcher channel, potentially delivering an inconsistent snapshot to BMP receivers.

I looked more deeply and that's right we need to serialize watcher init/registration and notifyWatcher()
I addressed [here](https://github.com/osrg/gobgp/pull/3397)

fujita:
PR #3397, which fixes the race condition in the watcher's initial dump, has been merged. Once the Graceful Restart concern is resolved, I think this is ready to merge.

nplanelcisco:
> The Graceful Restart sync paths take only a Read Lock, which allows concurrent propagateUpdateToNeighbors to interleave with the initial route advertisement. Since GR sync also needs to send all routes followed by EOR (RFC 4724), the same Write Lock used for initial establishment seems appropriate here?

To resume I Lock these path to be serialized in sync
* Initial establishment when not local restarting
* GR sync when local restarting ends
* GR sync when EOR received during local restarting
* RTC EOR wait

fujita:
Looks fine, pushed.

PR #3361, opened 2026-04-01
server: Use State instead of Config to determine GR-enabled peers in StopBgp

rastislavs:
As we aim to skip sending the NOTIFICATION messages to GR-enabled peers if the `allow_graceful_restart` flag is set to true in `StopBgpRequest`, we should not rely on peer's config, but the actual state (that's where `isGracefulRestartEnabled` looks it up) - as the graceful restart may be configured, but not enabled if the remote side is not supporting it.


fujita:
Thanks!

PR #3364, opened 2026-04-05
gobgpd: Fix printing of help message

gibmat:
Currently, running `gobgpd --help` displays the nicely formatted help message, then also logs that string as a parser error and exits with an error. Instead, detect when the help message was printed and exit cleanly.

### DIFF
--- a/cmd/gobgpd/main.go
+++ b/cmd/gobgpd/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


cmd/gobgpd/main.go
```diff
@@ -81,6 +82,12 @@ func main() {
 	}
 	_, err := flags.Parse(&opts)
 	if err != nil {
+		if flagsErr, ok := errors.AsType[*flags.Error](err); ok {
```
fujita: I'd prefer not to bump the minimum Go version to 1.26 just yet — it was only recently released. Could you replace errors.AsType with errors.As so it works with Go 1.24?

gibmat: Sure, done!

fujita:
Thanks!

PR #3365, opened 2026-04-07
fix BMP map issue

sephiroth8189:
The Protobuf values for api.AddBmpRequest_MonitoringPolicy range from 0 to 5 (including UNSPECIFIED = 0; policies are 1–5), while the keys in oc.IntToBmpRouteMonitoringPolicyTypeMap range from 0 to 4 (with no UNSPECIFIED). AddBmp directly looks up the table using int(r.Policy), causing the entire file to be misaligned (for example, selecting PRE results in POST).


fujita:
Nice catch!

But you can't modify `bgp_configs.go`‎ by hand. It is generated the yang files. Please drop the changes to `bgp_configs.go`‎ .

sephiroth8189:
@fujita  I have reverted the changes about configs. and add a test about mapping

fujita:
Squashed the commits (we don't need the revert commit in the history) and pushed. Thanks!

PR #3370, opened 2026-04-10
fix: ipv6 link local address with zone

### LINKED ISSUE #3369: BUG: can not connet to bgp peer with password via ipv6 link local address
# how to reproduce
configure bgp peer as link local address with password.
for example: peer address "fe80::4850:31ff:fe01:fc55%eth0"  with password "helloworld"

# what do I expect to see
established a connection to peer

# what did I see
tcp connection can not be established.
I found there is no TCP_MD5SIG_EXT option at tcp syn so the target peer just drop these packets.
This will not happen for ipv4 nor ipv6 global scope addresses.

# reason
setting sin6_scope_id(Ifindex) is REQUIRED when using TCP MD5 signatures with IPv6 link-local addresses — in this example, the interface index of `eth0`

jayl1e:
# Why
fix bug at parsing ipv6 link local address  with zone as peer address

# How
- use `net/netip` package instead of `net` package to parse ip address. It can handle address like `fe80::4850:31ff:fe01:fc55%eth0`
- TCPMD5Sig.Ifindex will be populated as needed.

Fix: #3369


fujita:
https://github.com/osrg/gobgp/actions/runs/24251575285/job/71251309266?pr=3370

The CI failed. This might just be flaky, but the test uses v4 TCPMD5Sig so it's possible this change broke something?

jayl1e:
PTAL
There used to be a bug for cidr, now it gets fixed. Thanks

fujita:
squashed and pushed, thanks!

PR #3368, opened 2026-04-09
feat: rework RTC handling with per-peer RT membership index

Ivan-Pokhabov:
Replace the O(n) RT-scan logic in propagateUpdate with a structured
  peerRTMIndex (rtHash → peerID → {rtmKey set}) that enables O(1)
  lookup of per-peer RT interest.

  Add support for RTC import policy filtering, making route-target
  membership filtering more flexible by allowing default import policies
  to apply to RTC routes.

  Key changes:
  - Add peerRTMIndex to TableManager; expose UpdateRTC() and PeerHasRT()
  - Convert rtcHandler to a sync.RWMutex-protected struct for safe concurrent
  - Fix rtcHandler not being updated on destination delete
  - Export ExtCommRouteTargetKey / NlriRouteTargetKey
  - Remove adjRts / rtCounter / HasRTinRtcTable — superseded by peerRTMIndex
  - Add tests for ADD-PATH tracking, multi-RT withdrawal, and policy-based RTC filtering

PR #3374, opened 2026-04-14
Replace rtCounter with adjRTSet

Ivan-Pokhabov:
Counter semantics track only how many paths share an RT hash, not which
specific (AS, pathID) pairs are present. Set semantics make each entry
explicit: add is idempotent for the same key, and has() returning false
means every registered (AS, pathID) pair was explicitly removed.

Also make it more concurent safe

part of #3368


fujita:
rebased and pushed, thanks.

PR #3381, opened 2026-04-17
feat: (table/server) add VPN path RT index for O(1) RTC candidate lookup

Ivan-Pokhabov:
Add VPNPathIndex — thread-safe per-Table index of paths by Route Target.
  Replaces the O(n·m) GetBestPathList + ext-community scan in the RTC block
  with a direct hash lookup via TableManager.GetPathsByRT

  Add two early exits via rtmSet (before any path computation):
  - duplicate announce: RT already known → skip (replaces GetPathListWithSource scan)
  - ADD-PATH or different AS withdraw: RT still held by another path-id/as → skip

  Extract RTC handling into processRTCMembership + helpers.
  
  Final part of #3368

### DIFF
--- a/internal/pkg/table/rtc.go
+++ b/internal/pkg/table/rtc.go
@@ -98,6 +98,109 @@ func (s *rtmSet) reset() {
 	s.m = make(map[uint64]map[rtmKey]struct{})
 }
 
+// vpnPathKey identifies a VPN path without relying on *Path pointer identity.
+// info points to the root originInfo shared by a path and all its clones, so
+// Register/Unregister match even when the path was cloned in the pipeline.
+type vpnPathKey struct {
+	info   *originInfo
+	pathID uint32
+}
+
+func makeVPNPathKey(path *Path) vpnPathKey {
+	return vpnPathKey{info: path.OriginInfo(), pathID: path.remoteID}
+}
+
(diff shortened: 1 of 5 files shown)


Ivan-Pokhabov:
Hello, @fujita! Can you rewiew last part of RTC improvements PR?

internal/pkg/table/rtc.go
```diff
@@ -98,6 +98,97 @@ func (s *rtmSet) reset() {
 	s.m = make(map[uint64]map[rtmKey]struct{})
 }
 
+// vpnRTEntry holds the set of paths indexed under a single RT hash.
+type vpnRTEntry struct {
+	paths map[*Path]struct{}
+}
```
fujita, line 116: Using *Path as a map key is a bit fragile — it requires register/unregister to receive the exact same pointer. What happens if the path is cloned somewhere in the pipeline? Would register/unregister still match? Is there a more robust key we could use (e.g., NLRI key + path ID)?

Ivan-Pokhabov: Yeah, I think it is a not a very good solution. So now I change implementation: key changed from *Path to *OriginInfo that shares between path and its clones and PathID for add-path

fujita:
Pushed, thanks.

Issue #3382, opened 2026-04-17
`gobgp rpki server` Panics on the Documented No-arg Usage


zyk2507:
`gobgp rpki server` is documented as the command to verify RPKI server sessions in [this tutorial](https://github.com/osrg/gobgp/blob/master/docs/sources/rpki.md).
The current CLI prints the session list and then panics with an index-out-of-range error as shown below.

```
root@fra:~# gobgp rpki server
Session                 State  Uptime     #IPv4/IPv6 records
*.*.*.*:8082       Up     00:00:06   2449/2478
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.newRPKICmd.func1(0xe2405567800?, {0x155eba0, 0x0, 0xda1fd1?})
        github.com/osrg/gobgp/cmd/gobgp/rpki.go:149 +0x665
github.com/spf13/cobra.(*Command).execute(0xe2405773208, {0x155eba0, 0x0, 0x0})
        github.com/spf13/cobra/command.go:1019 +0xafb
github.com/spf13/cobra.(*Command).ExecuteC(0xe240575a008)
        github.com/spf13/cobra/command.go:1148 +0x465
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra/command.go:1071
main.main()
        github.com/osrg/gobgp/cmd/gobgp/main.go:32 +0xce
```

The issue exists in version 4.4.0 and the snapshot of 97ca4e7.

PR to fix: #3383

PR #3383, opened 2026-04-17
fix: rpki server listing panics

zyk2507:
The PR is to fix the problem described in #3382.
The root cause is that in `cmd/gobgp/rpki.go`, `newRPKICmd()` calls `showRPKIServer(args)` when `len(args) == 0 || len(args) == 1`,
but execution continues into:
```go
addr := net.ParseIP(args[0])
switch args[1] {
```
So:

- the 0-arg path reaches args[0]
- the 1-arg path can reach args[1]

`showRPKIServer(args)` already implements the intended display behavior for both:

0 args: list all RPKI servers
1 arg: show details for the specified server

My proposed fix return immediately after a successful showRPKIServer(args) call for the 0-arg and 1-arg paths.

### DIFF
--- a/cmd/gobgp/rpki.go
+++ b/cmd/gobgp/rpki.go
@@ -143,6 +143,7 @@ func newRPKICmd() *cobra.Command {
 				if err != nil {
 					exitWithError(err)
 				}
+				return
 			} else if len(args) != 2 {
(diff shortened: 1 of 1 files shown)


fujita:
Thanks!

PR #3388, opened 2026-04-21
feat: add partial BFD protocol implementation for BGP peer failure detection

Ivan-Pokhabov:
Implements RFC 5880/5881 BFD (Bidirectional Forwarding Detection) for
fast detection of BGP peer failures. BFD is configured per peer-group
or neighbor and triggers a hard reset when a BFD session expires.

Authored-by: Timur Aitov <timonbl4@gmail.com>
Co-authored-by: Ivan-Pokhabov <vanek3372@gmail.com>

### DIFF
--- a/api/gobgp.pb.go
+++ b/api/gobgp.pb.go
@@ -6839,6 +6839,7 @@ type Peer struct {
 	GracefulRestart *GracefulRestart       `protobuf:"bytes,9,opt,name=graceful_restart,json=gracefulRestart,proto3" json:"graceful_restart,omitempty"`
 	AfiSafis        []*AfiSafi             `protobuf:"bytes,10,rep,name=afi_safis,json=afiSafis,proto3" json:"afi_safis,omitempty"`
 	TtlSecurity     *TtlSecurity           `protobuf:"bytes,11,opt,name=ttl_security,json=ttlSecurity,proto3" json:"ttl_security,omitempty"`
+	Bfd             *BfdConfig             `protobuf:"bytes,12,opt,name=bfd,proto3" json:"bfd,omitempty"`
 	unknownFields   protoimpl.UnknownFields
 	sizeCache       protoimpl.SizeCache
 }
@@ -6950,6 +6951,13 @@ func (x *Peer) GetTtlSecurity() *TtlSecurity {
 	return nil
 }
 
+func (x *Peer) GetBfd() *BfdConfig {
+	if x != nil {
+		return x.Bfd
+	}
+	return nil
+}
+
 type PeerGroup struct {
 	state           protoimpl.MessageState `protogen:"open.v1"`
 	ApplyPolicy     *ApplyPolicy           `protobuf:"bytes,1,opt,name=apply_policy,json=applyPolicy,proto3" json:"apply_policy,omitempty"`
@@ -6963,6 +6971,7 @@ type PeerGroup struct {
 	GracefulRestart *GracefulRestart       `protobuf:"bytes,9,opt,name=graceful_restart,json=gracefulRestart,proto3" json:"graceful_restart,omitempty"`
 	AfiSafis        []*AfiSafi             `protobuf:"bytes,10,rep,name=afi_safis,json=afiSafis,proto3" json:"afi_safis,omitempty"`
 	TtlSecurity     *TtlSecurity           `protobuf:"bytes,11,opt,name=ttl_security,json=ttlSecurity,proto3" json:"ttl_security,omitempty"`
+	Bfd             *BfdConfig             `protobuf:"bytes,12,opt,name=bfd,proto3" json:"bfd,omitempty"`
 	unknownFields   protoimpl.UnknownFields
 	sizeCache       protoimpl.SizeCache
 }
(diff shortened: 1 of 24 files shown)


Ivan-Pokhabov:
Hello, @fujita! Can you please review this PR? A lot of this changes is regen api and test. I think that we don't need to split it so its provided one feature

fujita:
have you considered using openconfig-bfd?

OpenConfig has a standard BFD model (openconfig-bfd). Could you share why you chose to define BfdConfig independently rather than importing it? 

Please split into separate PRs. This is too large.

- internal/pkg/netutils/sockopt_* — generic socket helpers (SetUDPTTLSockopt, SetReuseAddrSockopt).
- pkg/packet/bfd/ — BFD packet codec. Standalone with its own tests, no dependency on the rest.

Ivan-Pokhabov:
> have you considered using openconfig-bfd?
> 
> OpenConfig has a standard BFD model (openconfig-bfd). Could you share why you chose to define BfdConfig independently rather than importing it?
> 

We are not trying to model or implement BFD as a full standalone control plane. As we think  GoBGP only needs a very small slice of BFD: enough to treat it as a liveness signal next to a BGP session (essentially “is the peer still reachable on the path we already use for BGP”), with a handful of knobs that map directly to that use case.

The complete openconfig-bfd model describes a rich, general-purpose BFD subsystem (interfaces, peers, session details, micro-BFD, echo-related knobs, extensive operational state, and so on). That breadth is appropriate for a router that implements BFD end-to-end as its own product surface, but it is far beyond what GoBGP needs or intends to support in the near term.

Ivan-Pokhabov:
> Please split into separate PRs. This is too large.
> 
> * internal/pkg/netutils/sockopt_* — generic socket helpers (SetUDPTTLSockopt, SetReuseAddrSockopt).
> * pkg/packet/bfd/ — BFD packet codec. Standalone with its own tests, no dependency on the rest.

Ok, we will split it, thanks

fujita:
"openconfig-bfd is too large" is not by itself a reason to avoid it — importing a YANG model doesn't mean implementing every leaf; you only map the fields you actually use.

The more relevant question is whether the structure fits. openconfig-bfd is interface-scoped (/bfd/interfaces/interface/config/ holds timers; peers under it are read-only discovery state). Seems that your BfdConfig is peer-scoped (per-neighbor multiplier / rx_interval / tx_interval). Could you explain why peer-scoped configuration is needed?

Ivan-Pokhabov:
> "openconfig-bfd is too large" is not by itself a reason to avoid it — importing a YANG model doesn't mean implementing every leaf; you only map the fields you actually use.
> 
> The more relevant question is whether the structure fits. openconfig-bfd is interface-scoped (/bfd/interfaces/interface/config/ holds timers; peers under it are read-only discovery state). Seems that your BfdConfig is peer-scoped (per-neighbor multiplier / rx_interval / tx_interval). Could you explain why peer-scoped configuration is needed?

You're right that we don't have to implement every leaf. The structural mismatch is the issue: openconfig-bfd is interface-scoped — timers live at /bfd/interfaces/interface/config/, and discovered peers under it are read-only state. GoBGP has no concept of "interface" in its config model; the unit is always a BGP neighbor by address. Forcing that mapping would require a synthetic interface entry per neighbor, which is more confusing than an honest peer-scoped config.

fujita:
I agree that using openconfig-bfd directly doesn't fit well here — its interface-scoped structure doesn't map naturally to GoBGP's neighbor-centric config model.

That said, I'd like us to introduce BFD configuration in a way that's consistent with GoBGP's existing OpenConfig conventions: specifically, the config/state container split, using grouping and augment in the gobgp YANG module, and placing it under the neighbor and peer-group trees as we do for other gobgp-specific extensions.

I'd also like to align the leaf names and units with openconfig-bfd (e.g., desired-minimum-tx-interval in microseconds rather than tx-interval in milliseconds), so that the schema stays close to the standard and remains easier to migrate or integrate in the future.

https://github.com/fujita/gobgp/tree/bfd-config

Could you take a look and let me know what you think?

Ivan-Pokhabov:
> https://github.com/fujita/gobgp/tree/bfd-config
> 
> Could you take a look and let me know what you think?

Could we keep a port leaf (default 3784) in the config to allow non-standard deployments? Everything else in your schema looks good.

fujita:
> > https://github.com/fujita/gobgp/tree/bfd-config
> > Could you take a look and let me know what you think?
> 
> Could we keep a port leaf (default 3784) in the config to allow non-standard deployments? Everything else in your schema looks good.

Sure, a port leaf makes sense for non-standard deployments. Could you open a PR directly against master that incorporates my PoC schema and adds the port leaf on top?

Please include a note in the YANG leaf description that setting a non-default port deviates from RFC 5881, which mandates destination UDP port 3784.

Ivan-Pokhabov:
> Please include a note in the YANG leaf description that setting a non-default port deviates from RFC 5881, which mandates destination UDP port 3784.

What do you think about this PR, but with port leaf? Or something wrong with it?

fujita:
@Ivan-Pokhabov Can we close this, right?

Ivan-Pokhabov:
> @Ivan-Pokhabov Can we close this, right?

Yeah, it's done

PR #3394, opened 2026-04-24
fix(rtc): fix VPN path indexing — track best path (no-add-path) or all path-IDs (add-path)

Ivan-Pokhabov:
The previous syncVPNIndex compared KnownPathList snapshots and registered every path present after an update. For non-add-path sessions this caused two peers advertising the same NLRI to both appear in vpnIdx, so RTC would redistribute duplicate paths to subscribing peers instead of only the best one.

Fix the indexing logic:
- Extend destination.Calculate to return the displaced old path (from explicitWithdraw / implicitWithdraw), giving callers precise before/after information without needing snapshot comparison.
- Replace syncVPNIndex with updateVPNIdx that branches on the session type:
  - ADD-PATH (RemoteID != 0): each (source, path-ID) pair is distinct; the exact old path is unregistered and the new one registered.
  - non-add-path: only the best path per NLRI is tracked; changes are detected by comparing OldKnownPathList[0] with KnownPathList[0].
- Refactor rtcVPNCandidates to accept isWithdraw bool instead of *Path to reduce coupling.

Add TestTableVPNPathIndexPathCount covering both scenarios end-to-end.

### DIFF
--- a/internal/pkg/table/destination.go
+++ b/internal/pkg/table/destination.go
@@ -250,19 +250,20 @@ func (dd *destination) GetMultiBestPath(id string) []*Path {
 // paths from known paths. Also, adds new paths to known paths.
 // INTERNAL USE ONLY: Caller MUST hold the appropriate shard lock.
 // This method must NEVER be called on snapshot destinations.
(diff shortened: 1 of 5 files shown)


fujita:
Pushed, thanks.

PR #3395, opened 2026-04-24
feat(netutils): add SetUDPTTLSockopt and SetReuseAddrSockopt helpers

Ivan-Pokhabov:
Add two UDP socket option helpers needed for BFD transport:
- SetUDPTTLSockopt: sets TTL/hop-limit to 255 per RFC 5881
- SetReuseAddrSockopt: enables SO_REUSEADDR on the BFD server socket

Also moves setSockOptString from the platform-neutral utils.go into sockopt_linux.go where it is only compiled on Linux.

Part of #3388


fujita:
Merged, thanks.

PR #3396, opened 2026-04-24
feat(bfd): add RFC 5880 BFD packet codec

Ivan-Pokhabov:
Implements binary marshal/unmarshal for BFD control packets per RFC 5880. The codec is self-contained with no dependency on the rest of the codebase and covers the mandatory 24-byte header fields used in single-hop sessions.

Part of #3388


fujita:
Pushed, thanks!

PR #3397, opened 2026-04-24
watcher: serialize new watcher registration

nplanelcisco:
To avoid out of order or duplicated events
When registering a watcher, first dump the state, then register for the nexts notifyWatcher().


fujita:
Looks good preparation, thanks.

PR #3399, opened 2026-04-27
feat(bfd): parse and encode Diagnostic field per RFC 5880 §4.1

Ivan-Pokhabov:
Add DiagnosticType enum (9 codes + reserved range) and expose it as a
  Diagnostic field in BFDHeader. UnmarshalBinary now reads the lower 5
  bits of byte 0; MarshalBinary writes them back and validates that
  Version ≤ 7 and Diagnostic ≤ 31.

  Needed to surface local/remote diagnostic codes in BFD session state.
  
  Part of #3388


fujita:
Pushed, thanks.

PR #3400, opened 2026-04-27
feat(bfd): align BFD schema with OpenConfig conventions.

Ivan-Pokhabov:


tools/pyang_plugins/gobgp.yang
```diff
@@ -1449,4 +1449,216 @@ module gobgp {
...
+      description
+        "Enable BFD liveness detection for this BGP neighbor.";
+    }
+
+    leaf port {
+      type uint16;
+      default 3784;
+      description
+        "Destination UDP port for BFD control packets (RFC 5881).";
+    }
```
fujita, line 1522: The description should make clear that using a non-default port deviates from RFC 5881, which explicitly mandates destination UDP port 3784.

tools/pyang_plugins/gobgp.yang
```diff
@@ -1449,4 +1449,216 @@ module gobgp {
...
+    leaf desired-minimum-tx-interval {
+      type uint32;
+      units microseconds;
+      default 1000000;
+      description
+        "Desired minimum transmission interval for BFD control packets.";
+      reference "RFC 5880 §6.8.7";
+    }
+
+    leaf required-minimum-rx-interval {
```
fujita: We should use the OpenConfig-compatible name required-minimum-receive to stay consistent with openconfig-bfd.

fujita (changes requested):
gobgp.pb.go is a generated file and should only be committed alongside changes to gobgp.proto. Since gobgp.proto is not modified in this PR, the gobgp.pb.go changes should be dropped. This PR should be scoped to the YANG and config changes only, which means util.go changes are also out of scope and should be removed.

Ivan-Pokhabov:
> gobgp.pb.go is a generated file and should only be committed alongside changes to gobgp.proto. Since gobgp.proto is not modified in this PR, the gobgp.pb.go changes should be dropped.

I think you missed, gobgp.proto changes is in the PR

> This PR should be scoped to the YANG and config changes only, which means util.go changes are also out of scope and should be removed.

Ok, I agree to delete changes for util.go, but maybe we will keep changes with API?

tools/pyang_plugins/gobgp.yang
Ivan-Pokhabov: My fault, I will fix it

tools/pyang_plugins/gobgp.yang
Ivan-Pokhabov: I will do it thanks

fujita:
Could you split the gRPC API changes (gobgp.proto, gobgp.pb.go, util.go) into a separate PR? This PR should be scoped to the YANG and config changes only.

On the API design, a couple of naming suggestions for consistency:
- BfdConfig should probably be BfdPeerConfig to match the BfdPeerState naming and make the per-peer scope explicit, especially if BfdServerState is also defined.
- BfdServerState collects server-wide socket statistics (not per-session), so it won't have a corresponding YANG definition. Since it is API-only, BfdState might be a cleaner name.

Ivan-Pokhabov:
> Could you split the gRPC API changes (gobgp.proto, gobgp.pb.go, util.go) into a separate PR? This PR should be scoped to the YANG and config changes only.
> 
> On the API design, a couple of naming suggestions for consistency:
> 
> * BfdConfig should probably be BfdPeerConfig to match the BfdPeerState naming and make the per-peer scope explicit, especially if BfdServerState is also defined.
> * BfdServerState collects server-wide socket statistics (not per-session), so it won't have a corresponding YANG definition. Since it is API-only, BfdState might be a cleaner name.

Ok, I will make separate PRs

Ivan-Pokhabov:
> Could you split the gRPC API changes (gobgp.proto, gobgp.pb.go, util.go) into a separate PR? This PR should be scoped to the YANG and config changes only.

I fix PR, Can you check it?

tools/pyang_plugins/gobgp.yang
```diff
@@ -1449,4 +1449,219 @@ module gobgp {
...
+        "Nanoseconds since the Unix epoch at which the last session failure occurred.";
+    }
+
+    leaf failure-transitions {
+      type uint64;
+      description "Number of UP-to-DOWN transitions.";
+    }
+
+    leaf local-discriminator {
+      type string;
```
fujita: RFC 5880 defines this as 32-bit fields. Since they are used as identifiers (zero is "unknown"), uint32 is the natural representation? My PoC had the same mistake — apologies for the confusion.

tools/pyang_plugins/gobgp.yang
```diff
@@ -1449,4 +1449,219 @@ module gobgp {
...
+      description "Unique local discriminator for this BFD session.";
+      reference "RFC 5880 §4.1";
+    }
+
+    leaf remote-discriminator {
+      type string;
```
fujita: Ditto.

tools/pyang_plugins/gobgp.yang
```diff
@@ -1449,4 +1449,219 @@ module gobgp {
...
+      description
+        "Enable BFD liveness detection for this BGP neighbor.";
+    }
+
+    leaf port {
+      type uint16;
+      default 3784;
+      description
+        "Destination UDP port for BFD control packets.
+        using a non-default port deviates from RFC 5881,
```
fujita: Please capitalize it to "Using".

fujita (changes requested):

tools/pyang_plugins/gobgp.yang
Ivan-Pokhabov: Not a big deal, I fixed

tools/pyang_plugins/gobgp.yang
Ivan-Pokhabov: Fixed

tools/pyang_plugins/gobgp.yang
Ivan-Pokhabov: Fixed

fujita:
Pushed, thanks.

PR #3403, opened 2026-04-29
correct sub-TLV length validation in BGP-LS decoders

MyLi2tlePony:
The default branch of the sub-TLV decoder loops in `LsLinkNLRI`, `LsPrefixV4NLRI`, `LsPrefixV6NLRI` and `LsTLVNodeDescriptor` compared the sub-TLV length against the parent `l.Length` field, which has different semantics from the remaining buffer:

- For `LsLinkNLRI` / `LsPrefixV*NLRI`, `l.Length` includes the 9-byte NLRI header, while the `tlv` slice already excludes it. The check `sub.Len() > l.Length` was therefore too permissive and could let a malformed sub-TLV trigger an out-of-bounds slice on the subsequent `tlv = tlv[sub.Len():]`.
- In all four locations `l.Length` was decremented as a side effect of skipping unknown sub-TLVs, but it was not decremented in the success path that handles known sub-TLVs. This left the field in an inconsistent state and broke any later use of `Len()` or serialization.

## Changes

- Replace the check with `sub.Len() > len(tlv)` (the actual remaining buffer).
- Stop mutating `l.Length` during decoding.
- Add table-driven tests covering each of the four decoders to verify that unknown sub-TLVs are correctly skipped and that the parent `Length` field is preserved.

## Related

- https://github.com/osrg/gobgp/pull/3358

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -5161,11 +5161,10 @@ func (l *LsLinkNLRI) DecodeFromBytes(data []byte) error {
 			subTLV = &LsTLVIPv6NeighborAddr{}
 
 		default:
-			if uint16(sub.Len()) > l.Length {
+			if sub.Len() > len(tlv) {
 				return malformedAttrListErr("sub-TLV length exceeds parent TLV length")
 			}
 			tlv = tlv[sub.Len():]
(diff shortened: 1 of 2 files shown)


fujita:
Thanks!

PR #3405, opened 2026-04-30
refactor peerInfo and UpdatePathAttr

Ivan-Pokhabov:


internal/pkg/table/destination.go
```diff
@@ -123,16 +151,42 @@ func (i *PeerInfo) String() string {
 
 func NewPeerInfo(g *oc.Global, p *oc.Neighbor, AS, localAS uint32, ID, localID netip.Addr, addr, localAddr netip.Addr) *PeerInfo {
 	return &PeerInfo{
-		AS:                      AS,
-		LocalAS:                 localAS,
+		PeerType:                p.State.PeerType,
 		ID:                      ID,
-		LocalID:                 localID,
+		AS:                      p.Config.PeerAs,
```
fujita: How ignore AS argument? This breaks dynamic neighbor.

internal/pkg/table/destination.go
```diff
@@ -123,16 +151,42 @@ func (i *PeerInfo) String() string {
...
 		Address:                 addr,
+		LocalAS:                 p.Config.LocalAs,
```
fujita: Ditto

internal/pkg/table/destination.go
```diff
@@ -107,9 +133,11 @@ func (lhs *PeerInfo) Equal(rhs *PeerInfo) bool {
 }
 
 func (i *PeerInfo) String() string {
-	if !i.Address.IsValid() {
+	const peerTypeUnspecified = oc.PeerType("")
+	if i.PeerType == peerTypeUnspecified {
```
fujita, line 137: Why we need to change here?

Ivan-Pokhabov: The change is needed because we added NewPeerGroupInfo, which does not set the Address field. With the old !i.Address.IsValid() check, a peer group's PeerInfo would be printed as "local", which is incorrect. PeerType is always set for peer groups (INTERNAL/EXTERNAL), so it's a more reliable

internal/pkg/table/destination.go
Ivan-Pokhabov: Fixed, thanks

internal/pkg/table/destination.go
Ivan-Pokhabov: Fixed, thanks

fujita:
Pushed, thanks.

PR #3406, opened 2026-04-30
fix bgp-ls mt-id rib collision

MyLi2tlePony:
## Problem

In ISIS-MT deployments a router advertises the same physical link/prefix
once per topology (e.g. classic IPv4/MT-0 alongside SRv6/MT-2). The
advertisements differ only by the Multi-Topology ID sub-TLV (RFC 7752
§3.2.1.5, type 263). Before this change BGP-LS NLRI decoders did not parse
that sub-TLV, so `String()` — which is used as the RIB destination key —
was identical for MT-distinct advertisements. Result: the second NLRI
implicit-withdrew the first, and only one topology survived in the RIB.

## Fix

- Parse `LS_TLV_MULTI_TOPO_ID` in `LsLinkNLRI`, `LsPrefixV4NLRI` and
  `LsPrefixV6NLRI` decoders.
- Add `MultiTopoIDs map[uint16]struct{}` to `LsLinkDescriptor` and
  `LsPrefixDescriptor`; populate from the new sub-TLV in `ParseTLVs`.
- Append a sorted ` MT: [v1 v2 …]` suffix in `String()` so MT-distinct
  NLRIs get distinct keys. Empty MT set → no suffix (backward compatible).

## Compatibility

- No wire-format change; only decoder, struct fields and string formatting.
- For non-MT NLRIs `String()` output is byte-for-byte identical — existing
  RIB keys and CLI output are unaffected.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -27,6 +27,7 @@ import (
 	"net/netip"
 	"reflect"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks.

PR #3410, opened 2026-05-09
feat: add local caps and peer-group to watchEventPeer

Ivan-Pokhabov:


pkg/server/grpc_server.go
```diff
@@ -438,13 +451,16 @@ func (s *server) watchEvent(ctx context.Context, r *api.WatchEventRequest, fn fu
...
+								PeerGroup:       p.State.PeerGroup,
 								RemoteCap:       remoteCaps,
+								LocalCap:        localCaps,
 							},
 							Transport: &api.Transport{
 								LocalAddress: p.Transport.LocalAddress.String(),
 								LocalPort:    p.Transport.LocalPort,
 								RemotePort:   p.Transport.RemotePort,
 							},
+							AfiSafis: afiSafis,
```
fujita: This sets the result of negotiatedRFList(). However, ListPath() sets the result of locally configured families.
So it's inconsistent.
I think that it would be better to let the field alone. You can get the info from LocalCaps and RemoteCaps.

fujita:
Pushed, thanks.

PR #3418, opened 2026-05-11
fix(cmd/gobgp): allow multiple RTs in mup route commands

higebu:
MUP route parsers restricted `rt` to a single value, while other VPN/EVPN/Flowspec
parsers and the documented syntax (`[rt <rt>...]` in srv6_mup.md) allow multiple.
Switch all four MUP parsers to `paramList`.


fujita:
Thanks!

PR #3419, opened 2026-05-11
fix(server): compare ECMP paths against best path, not predecessor

sergei-ilisavskii:
## Summary

- Fix incorrect ECMP best-path marking in `ListPath()` when `UseMultiplePaths` is enabled: compare each candidate path against `knownPathList[0]` (the actual best) instead of `knownPathList[i-1]` (its predecessor)
- The pairwise comparison was intransitive — paths with lower LOCAL_PREF could be marked `Best` if they happened to equal their neighbor in the sorted list
- Add regression test with 3 paths (LP=200, LP=100, LP=100) that catches the bug

## The Bug

`knownPathList` is sorted best-first. The ECMP loop compared `path.Compare(knownPathList[i-1])` instead of `path.Compare(knownPathList[0])`. With paths `[LP=200, LP=100, LP=100]`:

| Index | LOCAL_PREF | Old code compared against | Result | Correct? |
|-------|-----------|--------------------------|--------|----------|
| 0 | 200 | — | Best ✓ | ✓ |
| 1 | 100 | [0] LP=200 | not Best | ✓ |
| 2 | 100 | [1] LP=100 | **Best** | **✗** |

After fix, index 2 compares against [0] (LP=200) → not Best ✓

## Test plan

- [x] New `TestListPathEnableMultipath_DifferentLocalPref` — fails before fix, passes after
- [x] Existing `TestListPathEnableMultipath` — still passes
- [x] Full `go test ./pkg/server/` — passes (3 pre-existing env-dependent failures unrelated to this change)

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -2940,7 +2940,7 @@ func (s *BgpServer) ListPath(r apiutil.ListPathRequest, fn func(prefix bgp.NLRI,
 						case api.TableType_TABLE_TYPE_LOCAL, api.TableType_TABLE_TYPE_GLOBAL:
 							p.Best = true
 						}
-					} else if s.bgpConfig.Global.UseMultiplePaths.Config.Enabled && path.Compare(knownPathList[i-1]) == 0 {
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks!

PR #3420, opened 2026-05-11
server: use configured local address for nexthop when set

sergei-ilisavskii:
## Summary

When `Transport.Config.LocalAddress` is set to a specific (non-wildcard) address, override `Transport.State.LocalAddress` with it instead of using the TCP connection's local address. This ensures the configured address is advertised as the BGP nexthop.

This is needed for deployments where the BGP listener binds to an internal address (e.g., VRFs over GENEVE tunnels) but the configured local address is what should be advertised as the nexthop. Without this fix, the TCP connection's local address (the listener's bind address) is used, resulting in incorrect nexthop advertisement.

## Background

This issue was originally addressed in #2827 and #2839 (both closed). The fix was applied to downstream forks and evolved with IPv6 improvements, but was inadvertently removed during two upstream refactoring commits:
- `075aab46` ("convert PeerInfo to use netip instead of net", Aug 2025) removed the override from `handleFSMMessage`
- `0c74e9c6` ("server: remove access to fsm.conn in server", Oct 2025) removed the override from `toConfig`

## Approach

The override is applied in `server.go:handleFSMMessage()` when a session reaches ESTABLISHED, consistent with the original PRs. It uses the standard `ReadCopy → modify → Update` pattern under `fsm.lock` to persist the overridden address to `pConf` before `NewPeerInfo` is created. All downstream consumers (`toConfig`, `prePolicyFilterpath`/nexthop rewrite, watch events) read from `Transport.State` and automatically get the correct value.

The override only applies when the configured address is valid and non-unspecified (not `0.0.0.0` or `::`). IPv6 zone information is stripped since BGP nexthop attributes cannot carry zone info.

## Test plan

- [x] Added `TestLocalAddressOverride` integration test with 3 sub-tests that establish a real two-server BGP session and verify the overridden address via the `ListPeer` API:
  - No config address set (TCP address used as-is)
  - Wildcard IPv4 `0.0.0.0` (no override)
  - Specific IPv4 with `BindInterface` set (override applied)
- [x] `go build ./...` passes
- [x] `go vet ./pkg/server/...` passes
- [x] `go test -race ./pkg/server/...` passes
- [x] Existing tests unaffected

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1635,6 +1635,19 @@ func (s *BgpServer) handleFSMMessage(peer *peer, e *fsmMsg) {
 		drainChannel(peer.fsm.outgoingCh.Out())
 
 		if nextState == bgp.BGP_FSM_ESTABLISHED {
+			// Override the local address from the TCP connection with the
+			// configured local address when it is a specific (non-wildcard)
+			// address. This is needed for deployments where the BGP listener
+			// binds to an internal address (e.g. GENEVE tunnel endpoint) but
+			// the configured address is what should be advertised as nexthop.
+			peer.fsm.lock.Lock()
+			if cfgAddr := peer.fsm.pConf.ReadOnly().Transport.Config.LocalAddress; cfgAddr.IsValid() && !cfgAddr.IsUnspecified() {
+				localOverride := peer.fsm.pConf.ReadCopy()
+				localOverride.Transport.State.LocalAddress = cfgAddr.WithZone("")
+				peer.fsm.pConf.Update(&localOverride)
+			}
+			peer.fsm.lock.Unlock()
+
 			conf := peer.fsm.pConf.ReadOnly()
 			peerInfo := table.NewPeerInfo(peer.fsm.gConf, conf,
 				conf.State.PeerAs, conf.Config.LocalAs,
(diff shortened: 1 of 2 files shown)


sergei-ilisavskii:
I need to do additional tests, will come back with results soon

fujita:
I am closing this as stale: draft since May, conflicts with master, and you never posted the extra tests you mentioned. Reopen it or send a new PR if you want to continue.

The regression is real. 832b2d99 (Aug 2024) added the override, and 075aab46 and 0c74e9c6 dropped it in v4.0.0. Your analysis is correct.

But an export policy should already cover your case:

```toml
[[policy-definitions]]
  name = "nexthop"
  [[policy-definitions.statements]]
    name = "statement1"
    [policy-definitions.statements.actions]
      route-disposition = "accept-route"
    [policy-definitions.statements.actions.bgp-actions]
      set-next-hop = "192.168.1.100"
```

Attach it with export-policy-list on the neighbor. The export policy runs after UpdatePathAttrs, so it replaces the address taken from the TCP session, and SetNexthop() rewrites MP_REACH_NLRI too. Please tell me if that is not enough.

That is also why I see the override as low priority. The two addresses can differ only for a passive peer with BindInterface set.

Two notes if you do resend:

- Transport.State.LocalAddress also feeds the BMP peer-up message and the MRT BGP4MP record, and RFC 7854 4.10 defines that field as "The local IP address associated with the peering TCP session". Change only the value  passed to table.NewPeerInfo.

- Transport.Config.LocalAddress is never the zero value. default.go sets it to 0.0.0.0, to "::", or to the interface link-local for an unnumbered peer, so your guard fires for every unnumbered peer and strips the zone.

PR #3424, opened 2026-05-18
fsm: Do not emit warnings upon sending notification

rastislavs:
Warnings should point to potentially incorrect behavior. However, sending a notfication is completely expected during session lifetime and emiting a warning tends to point to incorrect direction when debugging in that case (e.g. emits a warning when a peer is deconfigured).


fujita:
Make sense, thanks.

PR #3425, opened 2026-05-18
table: Do not emit warnings upon no-op withdrawals

rastislavs:
As already described in the explicitWithdraw func comment, if policies are used, some paths may not be installed into the table, which causes warnings upon explicit withdrawal.

As this is a known situation under this scenario, we should not emit warnings and pollute logs with non-helpful warnings when policies are in use.


fujita:
Pushed, thanks.

Probably we should add a mechanism to record what routes to be sent peer peer.

PR #3428, opened 2026-05-25
fix(table): preserve last action for the same path in coalesced updates

rastislavs:
As of c86f89ea33d3723107dd1114b96df7005ef423be outgoing BGP UPDATE messages are coalesced before packing them. The packers serialize withdrawals before announcements, so if an announcement and withdrawal for the same NLRI/path-id are coalesced in one packing pass, the final wire order can invert the original action sequence.

Collapse the path list by local NLRI/path-id before handing it to the family packers, keeping only the final action.
This makes announce-then-withdraw produce only a withdrawal, and withdraw-then-announce produce only an announcement.


fujita:
Good catch, thanks a lot!

PR #3434, opened 2026-05-28
server: withdraw policy-filtered routes on soft reset out

rastislavs:
When an export policy change causes a previously advertised path to be filtered, soft reset out must send an explicit withdrawal, otherwise the neighbor can keep using stale routes.

Track advertised paths for all peers so that soft reset can distinguish routes that were actually sent from routes that were already filtered.

For VRF peers, normalize filtered VPN candidates to the local unicast form before checking advertised state, matching the path key used when the route was originally sent.

Add regression coverage for regular, add-path, and VRF soft-reset withdrawals.

### DIFF
--- a/pkg/server/peer.go
+++ b/pkg/server/peer.go
@@ -248,17 +248,28 @@ func (peer *peer) getRoutesCount(family bgp.Family, dstPrefix string) uint8 {
 	return 0
 }
 
+func (peer *peer) advertisedPathID(path *table.Path) uint32 {
+	if peer.isAddPathSendEnabled(path.GetFamily()) {
+		return path.LocalID()
+	}
+	return 0
+}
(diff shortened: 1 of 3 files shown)


fujita:
Thanks, merged.

One concern for a follow-up: resetAdvertisedRoutes acquires all 2048 propagation bucket locks sequentially, which pauses propagation to all peers on every session reset. This could be worth revisiting for busy environments.

rastislavs:
> One concern for a follow-up: resetAdvertisedRoutes acquires all 2048 propagation bucket locks sequentially, which pauses propagation to all peers on every session reset. This could be worth revisiting for busy environments.

Follow-up: https://github.com/osrg/gobgp/pull/3437

PR #3435, opened 2026-05-29
fsm: Do not emit warnings upon receiving notification

rastislavs:
Similar to 78d1c5678d5423958845d835e8c8939a2f0028f1 but in the opposite direction.
Warnings should point to potentially incorrect behavior. However, receiving a notfication is expected
during normal operation and emiting a warning tends to point to incorrect direction when debugging in that case.


fujita:
Pushed, thanks.

PR #3436, opened 2026-06-01
server: use sync.Map.Clear() in resetAdvertisedRoutes

fujita:
sync.Map.Clear() was added in Go 1.23 and go.mod requires Go 1.24.5, so the Range+Delete pattern is no longer necessary.


fujita:
@rastislavs

PR #3437, opened 2026-06-01
server: avoid global propagation locks in resetAdvertisedRoutes

rastislavs:
Follow-up to https://github.com/osrg/gobgp/pull/3434

Use the per-peer routeRefreshInProgress lock as the reset barrier for advertised-route bookkeeping instead of taking every propagation bucket lock.

Publish the peer down state before clearing advertised routes so later live propagation does not rebuild RIB-out state for a peer that is no longer advertising. Keep live propagation guarded by needToAdvertise before touching sentPaths/sendMaxPathFiltered.


fujita:
Pushed, thanks.

PR #3439, opened 2026-06-02
Implement RFC 8654 Extended Message Support for BGP

### LINKED ISSUE #2204: RFC 8654 support
https://tools.ietf.org/html/rfc8654

paunadeu:
Implements [RFC 8654 (Extended Message Support for BGP)](https://www.rfc-editor.org/rfc/rfc8654.html). The inbound length check in `pkg/server/fsm.go` has carried a `TODO: RFC 8654` since v4.0.0, and #2204 has tracked the work since 2019.

### What changes on the wire

A speaker advertises Capability Code 6 with a zero-length value inside its OPEN on every session. When the peer advertises the same capability, the FSM lifts the per-message length cap from 4096 to 65535 octets for UPDATE, NOTIFICATION and ROUTE-REFRESH (per Section 6); OPEN and KEEPALIVE keep the legacy 4096-octet ceiling. The negotiation result is the AND of both advertisements, as Section 4 requires.

### Always-on advertisement

Per maintainer review feedback, the capability is advertised unconditionally — no `disable-extended-message` knob, no YANG/proto config field, no surface area in `bgp_configs.go`. The shape follows the 4-byte AS Number capability: every session gets the cap in OPEN, and Section 4's "AND of both sides" rule on the receiver-side negotiation guarantees the safety property the operator-facing knob used to provide.

This also keeps the change YANG-clean: `bgp_configs.go` is untouched, `tools/pyang_plugins` does not need to learn a new openconfig path, and `oc.NeighborConfig` / `oc.PeerGroupConfig` keep their existing shape.

### Commit layout

Three logical commits:

1. `packet/bgp: add the BGP Extended Message capability (RFC 8654)`: capability code, struct, dispatch, the `BGP_MAX_EXTENDED_MESSAGE_LENGTH` constant, the `MarshallingOption.ExtendedMessage` flag, the serialise-side per-message-type length gate, and the packet-layer test matrix (round-trip, dispatch case, name map, length cap for each message type, helper nil-safety).
2. `api, apiutil: surface the BGP Extended Message capability`: `ExtendedMessageCapability` oneof entry on `api.Capability.cap` (field 14, mirroring `RouteRefreshCapability`), the regenerated `api/*.pb.go` from `buf generate`, and the `apiutil.MarshalCapability` / `unmarshalCapability` glue so `ListPeer` does not fall back to `CapUnknown`. No `bgp_configs.go` change, no proto field on `PeerConf`.
3. `server, table: advertise the cap, negotiate it, and gate the per-session length`: the `capabilitiesFromConfig` unconditional append (matching the `NewCapFourOctetASNumber` call site), the FSM `extendedMessage atomic.Bool`, the `handleOpen` AND'd negotiation, the read-loop length gate (removing the TODO that has sat in this code since v4.0.0), the `MarshallingOption.ExtendedMessage` plumbing on every FSM send-side call, the outbound UPDATE-packer budget in `internal/pkg/table/message.go`, and a server-package integration test covering the symmetric negotiation case and the apiutil round-trip.

### CI

On Go 1.25 (latest stable, matching the CI `setup-go@v6 stable` matrix):

- `go test -race -timeout 240s ./...`: green
- `golangci-lint run`: 0 issues
- `buf lint proto/` + `buf format --diff --exit-code`: green
- `buf generate --template proto/buf.gen.yaml proto` + `git diff --exit-code`: green. The `api/*.pb.go` output matches the byte-for-byte output of `buf generate` against the BSR plugins the upstream CI uses.
- `GOARCH=386 CGO_ENABLED=1 go test ./...`: green
- crossbuild `freebsd/openbsd/darwin/windows × amd64`: built
- `tools/spell-check/scspell.sh` and `tools/grep_avoided_functions.sh`: clean
- `markdownlint $(find . -name '*.md')`: clean

### Production exercise

The patch runs against a mixed-vendor BGP-LS lab built with containerlab: 42 FRR routers, 3 Nokia SR OS routers, 1 Juniper Junos router, and 3 GoBGP test peers, around 150 BGP sessions in total. Without the patch, every Nokia and Juniper BGP-LS UPDATE that crossed 4096 octets reset the session with NOTIFICATION code 1 / subcode 2, and the lab dropped 81 IS-IS links from its topology snapshot. With the patch in place, every established peer carries the capability on both its local and remote capability lists, no Bad-Message-Length notifications appear in a 24h run, and the link count holds at 81 against the same wire data v3.37.0 used to handle.

The rework you see here was also smoke-tested back on that same lab before the force-push: all 3 Nokia BGP-LS sessions established with the capability on both sides, 126 UPDATEs received on the busiest Nokia peer without any NOTIFICATION, and zero error/disconnect events in `pkg/server` logs across a fresh restart.

Closes #2204.

### DIFF
--- a/api/capability.pb.go
+++ b/api/capability.pb.go
@@ -91,7 +91,7 @@ func (x AddPathCapabilityTuple_Mode) Number() protoreflect.EnumNumber {
 
 // Deprecated: Use AddPathCapabilityTuple_Mode.Descriptor instead.
 func (AddPathCapabilityTuple_Mode) EnumDescriptor() ([]byte, []int) {
-	return file_api_capability_proto_rawDescGZIP(), []int{9, 0}
+	return file_api_capability_proto_rawDescGZIP(), []int{10, 0}
 }
 
 type Capability struct {
@@ -111,6 +111,7 @@ type Capability struct {
 	//	*Capability_RouteRefreshCisco
 	//	*Capability_Fqdn
 	//	*Capability_SoftwareVersion
+	//	*Capability_ExtendedMessage
 	Cap           isCapability_Cap `protobuf_oneof:"cap"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
@@ -270,6 +271,15 @@ func (x *Capability) GetSoftwareVersion() *SoftwareVersionCapability {
 	return nil
 }
 
+func (x *Capability) GetExtendedMessage() *ExtendedMessageCapability {
+	if x != nil {
+		if x, ok := x.Cap.(*Capability_ExtendedMessage); ok {
+			return x.ExtendedMessage
+		}
+	}
+	return nil
+}
+
 type isCapability_Cap interface {
 	isCapability_Cap()
 }
@@ -326,6 +336,10 @@ type Capability_SoftwareVersion struct {
 	SoftwareVersion *SoftwareVersionCapability `protobuf:"bytes,13,opt,name=software_version,json=softwareVersion,proto3,oneof"`
 }
 
+type Capability_ExtendedMessage struct {
+	ExtendedMessage *ExtendedMessageCapability `protobuf:"bytes,14,opt,name=extended_message,json=extendedMessage,proto3,oneof"`
+}
+
 func (*Capability_Unknown) isCapability_Cap() {}
 
 func (*Capability_MultiProtocol) isCapability_Cap() {}
@@ -351,6 +365,8 @@ func (*Capability_RouteRefreshCisco) isCapability_Cap() {}
 func (*Capability_Fqdn) isCapability_Cap() {}
 
 func (*Capability_SoftwareVersion) isCapability_Cap() {}
+
+func (*Capability_ExtendedMessage) isCapability_Cap() {}
 
 type MultiProtocolCapability struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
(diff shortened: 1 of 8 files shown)


fujita:
Great.

You can't modify bgp_configs.go, which is generated by the YANG files.
https://github.com/osrg/gobgp/blob/master/tools/pyang_plugins/README.md

However, I think that you can drop the configuration; always advertise this capability as the 4bytes-AS capability.

paunadeu:
Hey @fujita,
working on it! 

Thanks

paunadeu:
Reworked per the maintainer feedback at https://github.com/osrg/gobgp/pull/3439#issuecomment-4612093743. The capability now advertises unconditionally on the OPEN, same shape as the 4-byte AS Number cap; `bgp_configs.go` is untouched and there is no proto field on `PeerConf`. The receiver-side negotiation flag on the FSM still gates whether messages above 4096 octets are accepted, so the on-wire behaviour the operator-facing knob used to provide falls out of the Section 4 AND'd negotiation by itself.

The PR body has been updated to match, and the previous "commit 2 surfaces the disable knob" entry is gone. The diff is now 3 files smaller (no `oc.NeighborConfig` knob, no `oc.PeerGroupConfig` knob, no `Equal()` plumbing, no `newConfigFromAPIStruct` / `oc.NewPeerFromConfigStruct` glue).

Validated end-to-end on the same containerlab BGP-LS topology I had used for the original PR (3 Nokia + 42 FRR + 1 Juniper, ~150 BGP sessions): every Nokia peer reaches `BGP_FSM_ESTABLISHED` with `extended-message` on both its local and remote capability lists, the busiest Nokia BGP-LS peer takes 126 UPDATEs without a NOTIFICATION, and the gobgp logs are clean. Happy to add anything else you want to see covered.

fujita:
Thanks.

PR #3441, opened 2026-06-03
Feature/fix bmp

sephiroth8189:
## Problem
In multipath scenarios, BMP Local-RIB monitoring did not clearly show per-path changes (which path was updated vs withdrawn).
## Changes
- Added multipath delta fields to bestpath events:
  - `UpdatePathList` (new or changed paths)
  - `WithdrawPathList` (removed paths)
- Kept existing fields (`PathList`, `MultiPathList`) for backward compatibility.
- Generated multipath deltas at the update source (`table.Update`) .
- Unified path selection logic for BMP and WatchEvent/monitor:
  1. use `UpdatePathList`/`WithdrawPathList` first  
  2. fallback to `MultiPathList`  
  3. fallback to `PathList`
- Enabled Add-Path serialization for Local-RIB BMP messages.
- Added Local-RIB Peer Up Add-Path capability advertisement(only when multipath is enabled).
## Impact
- Multipath changes are now visible as explicit update/withdraw events.
- Non-multipath behavior remains unchanged.
- Zebra path is not affected.

by the way:  There still has a minor issue with the gobgp monitor: when the path-id does not change when route update, only a single "update" message is generated, making it impossible to distinguish whether it is a "replace" or an "add." To resolve this issue, the monitor's output needs to be modified. Maybe need a new word replace.

### DIFF
--- a/internal/pkg/table/destination.go
+++ b/internal/pkg/table/destination.go
@@ -530,6 +530,41 @@ type Update struct {
 	OldKnownPathList []*Path
 }
 
+// GetMultiBestPathDiff returns multipath delta as update and withdraw lists.
+// update includes both newly added and changed paths.
+func (u *Update) GetMultiBestPathDiff(id string) (update []*Path, withdraw []*Path) {
+	oldM := getMultiBestPath(id, u.OldKnownPathList)
+	newM := getMultiBestPath(id, u.KnownPathList)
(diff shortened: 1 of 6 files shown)


sephiroth8189:
@fujita  Could you please check if these can be merged?

fujita:
Pushed, thanks!

PR #3443, opened 2026-06-04
table: fix ext-community-set match-set-options ALL and INVERT

s7urmhond:
ExtCommunityCondition.Evaluate iterated route extended communities in the outer loop and configured patterns in the inner loop. For MATCH_OPTION_ALL this gave the inverse of the OpenConfig semantics: "every route community matches some pattern" instead of "every configured pattern matches some route community". As a result, match-set-options: all rejected routes that carried extended communities not listed in the set.

MATCH_OPTION_INVERT was affected by the same nesting. The early break fired only for ANY, so result was reset on every outer iteration and the final value reflected only the last route community.

Swap the loop nesting and add INVERT to the early-break condition. Mirrors the community-set fix from #859 (commit e06bb04f).


fujita:
Pushed, thanks!

PR #3444, opened 2026-06-05
packet/bgp: keep Link Local/Remote Identifier in LsLinkDescriptor.String when addresses are also present

paunadeu:
## Summary

`LsLinkDescriptor.String` falls through to the IPv4 or IPv6 address form whenever the interface and neighbor address sub-TLVs (RFC 7752 Table 5, types 259/260 or 261/262) are populated, and never appends the Link Local/Remote Identifier sub-TLV (RFC 7752 Table 5, type 258) even when the descriptor carries it.

That fall-through is correct under a strict reading of RFC 7752 Section 3.2.2, which calls the two shapes mutually exclusive. RFC 9086 Section 5.2 relaxes that for the BGP Egress Peer Engineering case. A PeerAdj-SID Link NLRI MUST include the Link Local/Remote Identifier sub-TLV and MAY also include the interface/neighbor address sub-TLVs alongside it. The companion PeerNode-SID Link NLRI for the same eBGP session reuses the same addresses but omits type 258, so the two NLRIs are distinct prefixes in adj-RIB-in because of the identifier sub-TLV.

When both shapes appear together the existing String form collapses the two prefixes to the same human-readable key. Callers that key on String (table-layer dedup, RIB walkers, log correlation loops) see one prefix where the wire carries two and silently drop the second flavour.

## Fix

Append `(L:<local>,R:<remote>)` to the address form whenever the Link Local/Remote Identifier sub-TLV is set alongside an address sub-TLV. The standalone identifier-only case is untouched, the all-UNKNOWN fallback is untouched, and the MT-ID suffix still appends at the end.

## Reproduction

Verified live against a Nokia SR OS 25.10 producer that emits both a PeerNode-SID Link NLRI (addresses only) and a PeerAdj-SID Link NLRI (addresses + Link Local/Remote Identifier 3/0) on the same eBGP session.

Before the patch both NLRIs collapse to `192.0.2.1->192.0.2.2`. The receiver's String-keyed cache stores only the last-seen flavour and a periodic reconcile silently withdraws the other. After the patch the PeerAdj NLRI renders as `192.0.2.1->192.0.2.2(L:3,R:0)`.

## Tests

`Test_LsLinkDescriptor_StringIncludesLinkID` covers three NLRI shapes:

- addresses only (PeerNode-SID NLRI shape)
- addresses + Link ID 3/0 (PeerAdj-SID NLRI shape Nokia emits)
- addresses + Link ID 1/1 (distinct identifier values)

Same shape as `Test_LsLinkNLRI_MultiTopoID`: every distinct sub-TLV combination must produce a distinct String. The addresses-only case asserts backward compatibility (no `(L:` suffix appears).

## References

- RFC 7752 Table 5 (Link Descriptor sub-TLVs)
- RFC 7752 Section 3.2.2 (IGP Link Descriptors)
- RFC 9086 Section 5.2 (PeerAdj-SID Link NLRI format)

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -5051,9 +5051,26 @@ func (l *LsLinkDescriptor) String() string {
 	default:
 		base = "UNKNOWN"
 	}
-	// MT-ID participates in the destination key (RFC 7752 §3.2.1.5).
-	// Two ISIS-MT advertisements of the same physical link must not
-	// collide in the RIB.
+
+	// The Link Local/Remote Identifier sub-TLV (Type 258, RFC 7752
+	// Table 5) coexists with the interface/neighbor address sub-TLVs
+	// in some NLRI shapes. RFC 9086 Section 5.2 mandates Type 258 on
+	// PeerAdj-SID Link NLRIs and lets the producer also include the
+	// IPv4 or IPv6 address sub-TLVs alongside. PeerNode-SID Link
+	// NLRIs for the same eBGP session may share the addresses but
+	// omit Type 258. The two NLRIs are distinct prefixes in
+	// adj-RIB-in because of Type 258's presence, so the String form
+	// must show the identifier whenever it is set or two distinct
+	// prefixes collide in any caller keyed on this output.
+	if l.LinkLocalID != nil && l.LinkRemoteID != nil &&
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks!

PR #3446, opened 2026-06-05
packet/bgp: fix TunnelEncapSubTLVSRCandidatePathName decode over-read silently dropping PathAttributeTunnelEncap

paunadeu:
## Summary

`TunnelEncapSubTLVSRCandidatePathName.DecodeFromBytes` reads `value[1:t.Len()]`. `value` is the body slice returned by `TunnelEncapSubTLV.DecodeFromBytes` (length `t.Length`). `t.Len()` adds the three-byte sub-TLV header because Candidate Path Name uses type 129 (>= 0x80, RFC 9012 Section 2.4.4 / IANA SR Policy Tunnel Encapsulation Sub-TLVs). The upper bound therefore exceeds the body length by three.

When the sub-TLV sits in front of another sub-TLV inside the same Tunnel-Encap attribute (the SR-Policy install shape: Preference + BSID + Candidate Path Name + Segment List), the over-read consumes the next sub-TLV's header bytes, corrupts its parse, and the receive engine silently drops the entire PathAttributeTunnelEncap. The receiver's `apiutil.Path.Attrs` slice surfaces every attribute except type 23.

## Fix

Read `value[1:]` (the body is one Reserved byte followed by the Candidate Path Name octet string). Replace the unreachable "length is too short" guard `int(t.Length) < t.Len()` (always false because `t.Len() = 3 + t.Length`) with a strict check on `t.Length < 1` so a zero-length body is rejected at parse time.

## Reproducer

`TestSRPolicy_TunnelEncap_SubTLVMatrix` runs a two-`BgpServer` in-process SR-Policy SAFI round-trip across seven sub-TLV shapes. Pre-fix output:

\`\`\`
PASS  pref_only
PASS  pref_segList
PASS  pref_bsid_segList
FAIL  pref_cpName_segList               <- Candidate Path Name triggers strip
PASS  pref_unk130_segList
FAIL  pref_bsid_cpName_segList
FAIL  pref_bsid_cpName_unk130_segList
\`\`\`

Post-fix every shape passes.

## Tests

- \`Test_TunnelEncapSubTLVSRCandidatePathName_RoundTrip\` (packet layer) covers four name lengths, including the empty name, and asserts the decoder neither over-reads nor mutates trailing bytes.
- \`Test_TunnelEncapSubTLVSRCandidatePathName_RejectsZeroLength\` locks in the mandatory Reserved byte.
- \`TestSRPolicy_TunnelEncap_SubTLVMatrix\` (server layer) is the seven-shape matrix above.

## References

- RFC 9012 Section 2.4.4 (Tunnel Encapsulation attribute layout for SR-Policy)
- IANA \"BGP Tunnel Encapsulation Sub-TLVs\" registry (type 129 = SR Candidate Path Name)

### DIFF
--- a/pkg/server/sr_policy_tunnel_encap_bisect_test.go
+++ b/pkg/server/sr_policy_tunnel_encap_bisect_test.go
@@ -1,0 +1,243 @@
+package server
+
+import (
+	"context"
+	"net"
+	"net/netip"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	api "github.com/osrg/gobgp/v4/api"
+	"github.com/osrg/gobgp/v4/pkg/apiutil"
+	"github.com/osrg/gobgp/v4/pkg/config/oc"
+	"github.com/osrg/gobgp/v4/pkg/packet/bgp"
+)
+
+// TestSRPolicy_TunnelEncap_SubTLVMatrix runs the same two-server SR-
+// Policy SAFI round-trip across every relevant combination of inner
+// sub-TLVs (Preference, BSID, Candidate-Path-Name, Unknown type 130,
+// Segment List). The test asserts that PathAttributeTunnelEncap
+// survives the receive WatchEvent callback for each combination.
+//
+// The shape mirrors RFC 9012 Section 2.4.4 (Tunnel Encapsulation
+// sub-TLV layout) plus the IANA "BGP Tunnel Encapsulation Sub-TLVs"
+// registry: type 12 (SR Preference), 13 (SR Binding SID), 128 (SR
+// Segment List), 129 (SR Candidate Path Name), 130 (SR Policy Name
+// / Unknown). Every combination that includes the Candidate Path
+// Name sub-TLV in front of another sub-TLV inside the same Tunnel
+// Encap attribute reproduces the silent PathAttributeTunnelEncap
+// drop on the receive side; combinations without it round-trip
+// cleanly.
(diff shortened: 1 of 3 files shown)


pkg/server/sr_policy_tunnel_encap_bisect_test.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+var testTunnelEncapPortCounter int
+
+func testTunnelEncapCounter() int {
+	testTunnelEncapPortCounter++
+	return testTunnelEncapPortCounter * 2
+}
+
```
fujita, line 218: var testTunnelEncapPortCounter atomic.Int32

func testTunnelEncapCounter() int32 {
    return testTunnelEncapPortCounter.Add(1) * 2
}
?

paunadeu: Done — applied verbatim in 3e84458d, force-pushed to the branch.

Counter is now `var testTunnelEncapPortCounter atomic.Int32` with `testTunnelEncapCounter() int32` returning `testTunnelEncapPortCounter.Add(1) * 2`. Dropped the now-redundant outer `int32(...)` cast on `senderPort` since `testTunnelEncapCounter()` already returns `int32`.

`go test -race -count=1 -run TestSRPolicy_TunnelEncap_SubTLVMatrix ./pkg/server/` passes in 15.6s. Thanks for the catch.

fujita:
pushed, thanks.

PR #3451, opened 2026-06-10
Support BGP and BFD Listeners to Device

ChrisPortman:
Connect the existing `StartBgpRequest.Global.BindToDevice`, to a new corresponding option in the configuration file.  This allows for the **listeners** for BGP and BFD to bind to a device.

Similarly, apply the `BgpConfigSet.Neighbor.Transport.Config.BindInterface`, when set, to the neighbor's corresponding outbound BFD connection.


ChrisPortman:
This is not ready, I need to work config changes in via the YANG generator

PR #3452, opened 2026-06-11
Use BindToDevice/Interface configurations for inbound/outbound BFD connections

ChrisPortman:
This PR contains 2 commits:

1. Exposes the API.Global.BindInterface option via the configuration file providing the ability to configure the BGP listener sockets for incomming BGP connections to a specific (likely VRF interface).  For BGP, this was already in the code base, just not accessible from the configuration file.  For BFD, additional functionality has been added such that if BGP listeners are interface bound, BFD listeners will be too.
2. If a neighbors configuration includes a Transport.Config.BindInterface, in addition to binding the interface for outbound BGP connections for that neighbor, use it for outbound BFD connections as well.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -372,7 +372,7 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
 		// register BFD for the dynamic neighbor too (explicit neighbors do this in addNeighbor): the
 		// BFD config is inherited from the peer group. Without this, BFD never runs for dynamic peers.
 		if s.bfdServer != nil && conf.Bfd.Config.Enabled {
-			if err := s.bfdServer.AddPeer(context.Background(), addr, conf.Bfd.Config, ""); err != nil {
+			if err := s.bfdServer.AddPeer(context.Background(), addr, conf.Bfd.Config, s.bgpConfig.Global.Config.BindToDevice); err != nil {
 				s.logger.Warn("failed to add BFD peer for dynamic neighbor",
 					slog.String("Topic", "Peer"),
 					slog.String("Key", addr.String()),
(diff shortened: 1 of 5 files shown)


fujita:
Sorry, the global BindToDevice/BindInterface changes overlap with another PR that was just merged — please remove those.

ChrisPortman:
> Sorry, the global BindToDevice/BindInterface changes overlap with another PR that was just merged — please remove those.

Awesome thanks.

I've updated this to remove changes to the config file.  As such, I've also collapsed it into a single commit dealing with using the same interface bindings for BGP for related BFD.

ChrisPortman:
Just a quick update to fix the failing tests

ChrisPortman:
Another rebase, and subsequent test fixup

ChrisPortman:
sorry.  Had to fix a lint, and a test failure observed with `-race`

pkg/server/server.go
```diff
@@ -3846,7 +3851,11 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
 		peer.fsm.pConf.Update(&conf)
 		peer.fsm.lock.Unlock()
 		if bfdConfigChanged {
-			err = s.updateBfdPeer(addr, original.Bfd.Config, c.Bfd.Config)
```
fujita, line 3849: bfdConfigChanged is false even if the interface is changed because the bfd config isn't changed.

pkg/server/server.go
```diff
@@ -372,7 +372,7 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
 		// register BFD for the dynamic neighbor too (explicit neighbors do this in addNeighbor): the
 		// BFD config is inherited from the peer group. Without this, BFD never runs for dynamic peers.
 		if s.bfdServer != nil && conf.Bfd.Config.Enabled {
-			if err := s.bfdServer.AddPeer(context.Background(), addr, conf.Bfd.Config); err != nil {
+			if err := s.bfdServer.AddPeer(context.Background(), addr, conf.Bfd.Config, ""); err != nil {
```
fujita: Why don't you support dynamic neighbors?

pkg/server/server.go
ChrisPortman: I think Ive addressed this with the review feedback commit.  I tried to capture this in the `TestUpdatePeer`, but it doesnt seem to cover this.

pkg/server/server.go
ChrisPortman: Sorry, oversight on my part being new to the code base, and initially not reasoning about what would make sense.  But on review, it would make sense to use the global BindToDevice, given it would be that device that received the new dynamic peer connection?

I've updated this in the review feedback commit

fujita:
Squashed and merged, thanks.

ChrisPortman:
> Squashed and merged, thanks.

Thankyou!

PR #3453, opened 2026-06-12
zebra: set ifindex for IPv6 link-local next-hops

### LINKED ISSUE #3299: Implement "Link-Local Next Hop Capability for BGP" draft
Usually, implementations send both global + link-local addresses even for unnumbered or IPv6 link-local peerings, encoding the the link-local address as a global unicast address (which is not right) or a global unicast address is absolutely empty (`::`), leading to inconsistent handling of next-hops.

**Additional context**
Draft link => https://datatracker.ietf.org/doc/html/draft-ietf-idr-linklocal-capability

DerDennisOP:
A link-local next-hop (`fe80::/10`) is ambiguous to the kernel without an outgoing interface, so zebra cannot install the route. This derives the outgoing ifindex from the path source address's zone (handling both numeric and named zones) and sets it on the nexthop.

Closes #3299.

### DIFF
--- a/pkg/server/zclient.go
+++ b/pkg/server/zclient.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
+	"net"
 	"net/netip"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


fujita:
Thanks!

PR #3454, opened 2026-06-12
Add BGP route aggregation

### LINKED ISSUE #3259: How to aggregate Routes?
I don't know how I can aggregate routes?

DerDennisOP:
Implements `aggregate-address` for IPv4/IPv6 unicast, configurable via the config file (`[[global.aggregates]]`), gRPC, and CLI (`gobgp global aggregate`). Supports `summary-only`, `as-set` (RFC 4271 AS_SET / ATOMIC_AGGREGATE / AGGREGATOR), and filtering contributors through a policy. The engine is event-driven off RIB updates, so it adds no per-update RIB scan. Closes #3259.

`pkg/config/oc/bgp_configs.go` was hand-edited to match the `gobgp.yang` change because I couldn't get pyang set up locally.

### DIFF
--- a/api/gobgp.pb.go
+++ b/api/gobgp.pb.go
@@ -12774,6 +12774,374 @@ func (x *BfdState) GetUnknownPeer() uint64 {
 	return 0
 }
 
+type AggregateAddress struct {
+	state         protoimpl.MessageState `protogen:"open.v1"`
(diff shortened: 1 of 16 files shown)


DerDennisOP:
what about the license header? i just straight out copied that? No SPDX?

PR #3456, opened 2026-06-15
bfd: support multihop (RFC 5883) and coexist with a host bfdd via specific-address binds

fivetime:
## What

Two related improvements to the native BFD server's listening sockets:

1. **Listen on the RFC 5883 multihop control port (UDP 4784) in addition to 3784.**
   The server only bound the single-hop control port (`s.config.Port`, 3784).
   A peer configured with `BfdPeerConfig.Port = 4784` (multihop, RFC 5883) sends
   its control packets to 4784, so the server never received the returns and the
   session could not come up. The server now also binds 4784; `rxPacket`
   dispatches by source IP, so a single receive path handles both single-hop and
   multihop peers.

2. **Bind the control sockets to the configured listen addresses, not the wildcard.**
   The sockets were bound to `:3784` / `:4784` (wildcard). On a host already
   running a system `bfdd` that owns `0.0.0.0:3784/4784`, the embedded GoBGP's
   returns were delivered to that wildcard listener instead of GoBGP. By binding
   each control port on each address in `Global.Config.LocalAddressList`, the
   kernel's most-specific-match UDP demux delivers packets destined to a GoBGP
   address to GoBGP, so an embedded GoBGP can run BFD on a host that also runs a
   system bfdd. When no listen address is configured it keeps the wildcard bind
   (unchanged behaviour).

## Why

These let an embedded GoBGP run **multihop** BFD (controller several hops from its
peers) and coexist with a host `bfdd` — neither was possible before.

## Testing

- `go test ./pkg/server/ -run BFD` passes.
- Verified end-to-end: an embedded GoBGP controller brought three BIRD peers to
  multihop-BFD `Up` over a routed (non-adjacent) path, on a host also running FRR
  `bfdd`, without touching bfdd.

## Notes

`bfdServer.Start` gained a variadic `listenAddrs ...string` parameter, so existing
call sites are source-compatible.

### DIFF
--- a/pkg/server/bfd_server.go
+++ b/pkg/server/bfd_server.go
@@ -56,6 +56,14 @@ type bfdServer struct {
 
 	config *oc.BfdConfig
 
+	// listenAddrs are the global BGP listen addresses. When non-empty the BFD
+	// server binds its control sockets to each SPECIFIC address (e.g.
+	// 10.0.0.1:4784) instead of the wildcard (:4784). A specific bind wins the
+	// kernel's most-specific-match UDP demux over any wildcard listener already on
+	// the port (e.g. a host bfdd owning 0.0.0.0:3784/4784), so an embedded GoBGP
+	// can run BFD on a host that also runs a system bfdd. Empty → wildcard bind.
+	listenAddrs []string
+
 	udpServers []*net.UDPConn
 
 	peersMutex sync.RWMutex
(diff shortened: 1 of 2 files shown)


fivetime:
Closing this conflicting PR in favor of the rebased replacement: https://github.com/osrg/gobgp/pull/3479

PR #3458, opened 2026-06-18
bgp-ls: accept prefix-length 0 in LsTLVIPReachability (default route)

Vadims06:
## Problem

GoBGP rejects a BGP-LS IPv4/IPv6 Topology Prefix NLRI whose *IP Reachability Information* TLV (type 265) carries **prefix-length 0** (the default route `0.0.0.0/0` or `::/0`), sending a NOTIFICATION and resetting the session in a loop.

Observed with a Huawei NE40E advertising `default-route-advertise always` into BGP-LS. The session reached ESTABLISHED but immediately dropped:

```
Session will be reset due to malformed BGP message
Error="Incorrect IP reachability Info length"
sent notification Code=3(update) Subcode=10(invalid network field)
```

## Root cause

`LsTLVIPReachability.DecodeFromBytes` in `pkg/packet/bgp/bgp.go` had two overly strict guards:

1. `if len(value) < 2` — rejects a 1-byte value, but for prefix-length 0 the TLV value is exactly 1 byte (the `0x00` prefix-length octet with zero prefix bytes following).
2. `if value[0] > 128 || value[0] == 0` — explicitly rejects prefix-length 0.

A default route is a valid BGP-LS Prefix NLRI per [RFC 7752 §3.2.3.2](https://tools.ietf.org/html/rfc7752#section-3.2.3.2): prefix-length 0 → 0 prefix bytes → TLV value length 1.

`toPrefix()` had the same latent bug: `(int(0)-1)/8+1` evaluates to 1 (not 0) due to Go integer truncation, producing a wrong `bytes` count masked only by the `if i >= len(l.Prefix) { break }` guard inside the copy loop.

## Fix

- Relax the length guard to `< 1`.
- Remove the `value[0] == 0` rejection.
- Compute `ll` as 0 when prefix-length is 0.
- Apply the same guard to `toPrefix()` for correctness.

## Test

Added a table-driven case for the exact captured TLV bytes `{0x01, 0x09, 0x00, 0x01, 0x00}` (type 265, len 1, value `0x00`), verifying decode to `prefix_length=0, prefix=[]`.


fujita:
Pushed, thanks!

PR #3459, opened 2026-06-18
Support VRF for TCP MD5

YutaroHayakawa:
Copying the commit message from the first commit for convenience.

===

Linux has per-VRF TCP MD5 key registry.

https://github.com/torvalds/linux/commit/6b102db50cdde3ba2f78631ed21222edf3a5fb51

Setting unix.TCPMD5Sig.Ifindex to the VRF's ifindex makes the key scoped within the VRF. GoBGP already has a support for binding Listener/Neighbor socket to VRF, so what we need to do is propagating the interface ifindex to SetTCPMD5SigSockopt call.

Actually, we already have a logic to set unix.TCPMD5Sig.Ifindex, but the current usage is wrong. https://github.com/osrg/gobgp/pull/3370 added it as a part of the MD5 support for IPv6 link-local address peering (a.k.a. BGP Unnumbered).

It derives Ifindex from the IPv6 zone information, but this doesn't have any effect as Ifindex only accepts VRF device's ifindex. Also, it uses unix.TCP_MD5SIG instead of unix.TCP_MD5SIG_EXT and it doesn't set unix.TCP_MD5SIG_FLAG_IFINDEX flag, so the kernel even ignores the parameter.

Remove the current logic and support setting bindInterface derived from the (Dynamic)Neighbor.Transport.Config.BindInterface configuration. Note that technically it is still possible to specify non-VRF interface to this parameter, so the new implementation always checks the interface type via netlink and if the interface is not VRF device, skip setting the parameter.

As a minor cleanup, always use unix.TCP_MD5SIG_EXT instead of unix.TCP_MD5SIG as EXT is a superset of the non-EXT one. We believe the risk of having Linux kernel version that doesn't have EXT support is low these days as it is v4.13 which is almost 10 years old.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -618,3 +618,25 @@ jobs:
           sudo apt-get install python3-setuptools
           sudo pip3 install -r test/pip-requires.txt
           PYTHONPATH=test python3 test/scenario_test/mup_test.py --gobgp-image gobgp -x -s
+
+  tcp-md5:
+    name: tcp-md5
+    runs-on: ubuntu-24.04
+    needs: build
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/download-artifact@v7
+        with:
+          path: artifact
+      # ubuntu-24.04's system python is v3.12 which breaks the test.
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+      - name: test
+        run: |
+          docker load < artifact/gobgp.tar
+          sudo apt-get update
+          sudo apt-get install -y linux-modules-extra-$(uname -r)
+          sudo modprobe vrf
+          python -m pip install -r test/pip-requires.txt
+          PYTHONPATH=test python test/scenario_test/tcp_md5_test.py --gobgp-image gobgp -x -s
--- a/internal/pkg/netutils/sockopt.go
+++ b/internal/pkg/netutils/sockopt.go
@@ -24,8 +24,8 @@ import (
 	"syscall"
 )
 
-func SetTCPMD5SigSockopt(l *net.TCPListener, address string, key string) error {
-	return SetTcpMD5SigSockopt(l, address, key)
+func SetTCPMD5SigSockopt(l *net.TCPListener, bindInterface string, address string, key string) error {
+	return SetTcpMD5SigSockopt(l, bindInterface, address, key)
 }
 
 func SetTCPTTLSockopt(conn net.Conn, ttl int) error {
--- a/internal/pkg/netutils/sockopt_bsd.go
+++ b/internal/pkg/netutils/sockopt_bsd.go
@@ -27,7 +27,7 @@ const (
 	ipv6MinHopCount = 73   // Generalized TTL Security Mechanism (RFC5082)
 )
 
-func SetTcpMD5SigSockopt(l *net.TCPListener, address string, key string) error {
+func SetTcpMD5SigSockopt(l *net.TCPListener, bindInterface string, address string, key string) error {
 	sc, err := l.SyscallConn()
 	if err != nil {
 		return err
--- a/internal/pkg/netutils/sockopt_darwin.go
+++ b/internal/pkg/netutils/sockopt_darwin.go
@@ -24,7 +24,7 @@ import (
 	"syscall"
 )
 
-func SetTcpMD5SigSockopt(l *net.TCPListener, address string, key string) error {
+func SetTcpMD5SigSockopt(l *net.TCPListener, bindInterface string, address string, key string) error {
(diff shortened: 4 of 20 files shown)


YutaroHayakawa:
This is still a draft (also seems like the new scenario test I introduced is failing somehow).

YutaroHayakawa:
Now the tests are passing. The problem was the iproute2 (ip and ss commands) in `osrg/quagga` images are old which doesn't support creating VRF device and displaying MD5 status of the socket respectively. I worked around it by executing netshoot's iproute2 within the target container's netns.

The fundamental solution against this problem is upgrading `osrg/quagga` side, but I couldn't find the source Dockerfile. Looking at the `docker history`, the image looks not that complicated. It's consist of base Quagga image and some golang bits and supervisord are added on top of it. The Quagga version is pretty outdated and also it doesn't have ARM build, so doesn't work on Apple-sillicon-based system. I'm willing to work on upgrading it to the latest FRR and add ARM build whenever I have time.

There was also a Python version dependency issue (nose depends on imp that only exists < 3.12). For the moment, I used setup-python action to use the desired Python version. In the future, we may want to stop relying on the CI environment's system python and use Python version management tool such as `uv`.

fujita:
> The fundamental solution against this problem is upgrading `osrg/quagga` side, but I couldn't find the source Dockerfile. Looking at the `docker history`, the image looks not that complicated. It's consist of base Quagga image and some golang bits and supervisord are added on top of it. The Quagga version is pretty outdated and also it doesn't have ARM build, so doesn't work on Apple-sillicon-based system. I'm willing to work on upgrading it to the latest FRR and add ARM build whenever I have time.

osrg/quagga is used by other CI tests including zebra. Upgrading it will require updating the ZAPI implementation in GoBGP as well, since FRR's ZAPI differs across versions. Also, vtysh output format changes in FRR may break the existing CI tests that parse its output.

> There was also a Python version dependency issue (nose depends on imp that only exists < 3.12). For the moment, I used setup-python action to use the desired Python version. In the future, we may want to stop relying on the CI environment's system python and use Python version management tool such as `uv`.

Probably we wan to migrate to pytest.

fujita (approved):

fujita:
Merged, thanks!

fujita:
Could you describe your use case? If you need peers in different Linux VRFs within a single GoBGP instance, would tcp_l3mdev_accept=1 with a VRF-unbound listen socket be more appropriate?

YutaroHayakawa:
> osrg/quagga is used by other CI tests including zebra. Upgrading it will require updating the ZAPI implementation in GoBGP as well, since FRR's ZAPI differs across versions. Also, vtysh output format changes in FRR may break the existing CI tests that parse its output.

I see... Yeah then it would be better to not upgrading it at least for the Zebra tests for now.

> Probably we wan to migrate to pytest.

Yeah agree. Seems like `nose` is not maintained actively right now.

> Could you describe your use case? If you need peers in different Linux VRFs within a single GoBGP instance, would tcp_l3mdev_accept=1 with a VRF-unbound listen socket be more appropriate?

My use case is per-VRF GoBGP instance and exchanging IPv4/Unicast and IPv6/Unicast (not VPN) routes over VRF. So far GoBGP identifies the neighbor by IP address internally (`neighborMap`'s key is `netip.Addr`). Also, the LocRIB is not isolated per-VRF. Therefore, when there are overlapping Prefix/NeighborIPs among VRFs, per-neighbor VRF binding doesn't work correctly.

I think the proper VRF support requires eliminating this kind of the logic that identifies something with IP address which I suspect exists everywhere.

fujita:
> My use case is per-VRF GoBGP instance and exchanging IPv4/Unicast and IPv6/Unicast (not VPN) routes over VRF. So far GoBGP identifies the neighbor by IP address internally (`neighborMap`'s key is `netip.Addr`). Also, the LocRIB is not isolated per-VRF. Therefore, when there are overlapping Prefix/NeighborIPs among VRFs, per-neighbor VRF binding doesn't work correctly.
> 
> I think the proper VRF support requires eliminating this kind of the logic that identifies something with IP address which I suspect exists everywhere.

The LocRIB isolation issue can be addressed with the existing peer VRF configuration. When a peer is configured with a VRF, GoBGP internally converts its IPv4/IPv6 unicast routes to VPNv4/VPNv6 via ToGlobal() before storing them in the global RIB, and converts them back via ToLocal() on the way out. This means two VRFs with the same prefix are stored as distinct NLRIs distinguished by the RD, giving you isolated LocRIBs without needing a per-VRF RIB structure — essentially the same isolation MPLS VPN provides, but for plain unicast peers.

For overlapping neighbor IPs across VRFs, you are right that the current neighborMap keyed solely on netip.Addr won't work. Some additional key component (e.g. VRF name or interface) would be needed there.

PR #3462, opened 2026-06-26
Add support RTC route target matching policy condition

themiron:
Add a bgp-condition that matches the Route Target carried in a Route Target Constrain (RFC 4684, rtc family) NLRI against a referenced ext-community-set, honoring any/all/invert match-set-options. Unlike match-ext-community-set, which matches the path's extended-communities attribute, this matches the Route Target embedded in the NLRI key itself.

The condition is exposed through OpenConfig (match-route-target-prefix), the gRPC API (Conditions.route_target_prefix) and the gobgp CLI (policy statement condition route-target-prefix <ext-community-set>).

### DIFF
--- a/api/gobgp.pb.go
+++ b/api/gobgp.pb.go
@@ -10895,6 +10895,7 @@ type Conditions struct {
 	Origin            OriginType             `protobuf:"varint,13,opt,name=origin,proto3,enum=api.OriginType" json:"origin,omitempty"`
 	LocalPrefEq       *LocalPrefEq           `protobuf:"bytes,14,opt,name=local_pref_eq,json=localPrefEq,proto3" json:"local_pref_eq,omitempty"`
 	MedEq             *MedEq                 `protobuf:"bytes,15,opt,name=med_eq,json=medEq,proto3" json:"med_eq,omitempty"`
+	RouteTargetPrefix *MatchSet              `protobuf:"bytes,16,opt,name=route_target_prefix,json=routeTargetPrefix,proto3" json:"route_target_prefix,omitempty"`
 	unknownFields     protoimpl.UnknownFields
 	sizeCache         protoimpl.SizeCache
 }
@@ -11030,6 +11031,13 @@ func (x *Conditions) GetLocalPrefEq() *LocalPrefEq {
 func (x *Conditions) GetMedEq() *MedEq {
 	if x != nil {
 		return x.MedEq
+	}
+	return nil
+}
+
+func (x *Conditions) GetRouteTargetPrefix() *MatchSet {
+	if x != nil {
+		return x.RouteTargetPrefix
 	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


fujita:
RTC NLRIs are prefixes in the BGP sense — the Route Target is the NLRI itself, not a path attribute. Wouldn't it be more consistent to extend the prefix matching mechanism rather than introducing a new condition type? I understand reusing ExtCommunitySet matchers is simpler to implement, but from a conceptual standpoint, matching on an RTC NLRI feels like prefix matching.

Something like the following:

```
[[defined-sets.prefix-sets]]
  prefix-set-name = "ps1"
  [[defined-sets.prefix-sets.prefix-list]]
     rtc-prefix = "RT=65000:100"

  [[defined-sets.prefix-sets.prefix-list]]
    rtc-prefix = "AS:65000"

  [[defined-sets.prefix-sets.prefix-list]]
    rtc-prefix = "*"
```

We can support something like "mpls-vpn-prefix", in general "family-prefix" style.

themiron:
@fujita thank you for that mention, yes indeed jt seems more logical from architectural and rfc point of view. 

Actually, we have initially implemented it exactly as prefix match here internally for gobgp v3 codebase with ip-prefix string reuse, so ipv6 prefix becomes container for 96-bit rtm nlri - perfect gate including as subfield - using in prod env.

But, for current v4 codebase ip-prefix field gets hard translated into netip.Prefix with no supported ability to keep rtc nature inside the prefix to keep back translation from binary to String() form, even with abusing “rtc” zone for netip.Prefix - not supported by stdlib, only for netip.Addr.
And that means - we need expand Prefix{} with dedicated rtc-prefix field, exactly as you have suggested above. prefix with ip-prefix is declared in general yang outside bgp-conditions, so it leads to openconfig yang https://github.com/openconfig/public/blob/3f7f6753ea5c4978e542b7289194ad61f66e39bc/release/models/policy/openconfig-routing-policy.yang#L320

I’m was not sure how to solve thia blocker w/o changing openconfig standard, so… for v4 upstream selected to abuse ext-comms-set and skip rtm:as matching since it’s not really used internally in gobgp rtc support anyway.

May I ask you to point me in right ideological direction for expanding Prefix? that’s the only left bit, technically I’ve got everything required already.

p.s Our implemented prefix-style of specifying rtc is stright-forwarded to match existing rtm format, example:
as:asn:local/len, where 0:0:0/0 is for default, asn:0:0/32 for asn-only matching, and the rest 64 bits are for rt including subtype bits, i.e:
* asn:65000:0/64 for 2-octet rt asn
* asn:65000.1:0/80 for 4-octet/ip rt asn

Quite simple, do you thing complex prefix formats with “rt:…” & “as:…” should be implemented, taking into account that partial rtc prefix mask (skip local value bits) is a must

themiron:
seems something like below in local gobgp.yang model could help, going this way
```
  deviation "/rpol:routing-policy/rpol:defined-sets/rpol:prefix-sets/" +
            "rpol:prefix-set/rpol:prefixes/rpol:prefix/rpol:config/rpol:ip-prefix" {
    deviate replace { mandatory false; }
  }
  augment ".../rpol:prefix/rpol:config" {
    leaf rtc-prefix { type string; ... }
  }
```

themiron:
@fujita hi, reworked in https://github.com/osrg/gobgp/pull/3485 as prefix-set

PR #3469, opened 2026-06-29
fix: register bfd for dynamic neighbours

frezbo:
Use doing dynamic neighbour discovery and BFD we need to dynamically add/remove peers.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -369,6 +369,16 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
 		}
 
 		s.neighborMap[addr] = peer
+		// register BFD for the dynamic neighbor too (explicit neighbors do this in addNeighbor): the
+		// BFD config is inherited from the peer group. Without this, BFD never runs for dynamic peers.
+		if s.bfdServer != nil && conf.Bfd.Config.Enabled {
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -1541,6 +1551,10 @@ func (s *BgpServer) stopNeighbor(peer *peer, oldState bgp.FSMState, e *fsmMsg) {
 	if s.neighborMap[key] == peer {
 		delete(s.neighborMap, key)
 	}
+	// deregister BFD for stopped dynamic neighbors (registered in the accept path); no-op if none.
+	if s.bfdServer != nil {
+		_ = s.bfdServer.DeletePeer(context.Background(), key)
+	}
```
fujita, line 1563: If you call bfdServer.DeletePeer in `stopNeighbor()`,  when we delete a static neighbor, we call `bfdServer.DeletePeer()` twice. `deleteNeighbor()` calls it directly, and then calls it via `stopNeighbor()`.

I think that we should remove `bfdServer.DeletePeer()` in `deleteNeighbor()`. `stopNeighbor()` is called for both static and dynamic peers.

frezbo: good point, fixed

fujita:
Pushed, thanks.

PR #3478, opened 2026-06-30
Support Config Via `io.Reader`

ChrisPortman:
This change implement support for configuring via a `io.Reader` that yields the equivalent of the standard configuration file.  This helps those using GoBGP as a library do things like embedding standard GoBGP configuration in a larger configuration file.  It also helps if we want to source the configuration from somewhere other than a file.

### DIFF
--- a/pkg/config/oc/bgp_configs_test.go
+++ b/pkg/config/oc/bgp_configs_test.go
@@ -17,7 +17,6 @@ package oc
 
 import (
 	"bufio"
-	"fmt"
 	"net/netip"
 	"os"
 	"path"
@@ -25,8 +24,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/go-viper/mapstructure/v2"
-	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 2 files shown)


hiSandog:
Accepting config from an `io.Reader` is a useful API improvement, but it should define error context carefully. When parsing fails, callers need the same file/line or source-location detail they get from path-based config, even if the source is not a file. A small test using a reader with invalid config would ensure the new path does not return a less actionable parse error than the existing file-loading path.

ChrisPortman:
> Accepting config from an `io.Reader` is a useful API improvement, but it should define error context carefully. When parsing fails, callers need the same file/line or source-location detail they get from path-based config, even if the source is not a file. A small test using a reader with invalid config would ensure the new path does not return a less actionable parse error than the existing file-loading path.

Thanks for the feedback.  I've added a test that processes invalid config.  It tests that the resulting error is equal to an expected error that I captured using main.

fujita:
In general I try to avoid adding new public API surface, since every exported function becomes something we need to keep backward-compatible indefinitely. That said, I think this one is reasonable.

One question: is there actual code on your end that calls ReadConfig directly (e.g. embedding GoBGP config inside a larger config file, as described in the PR description)? I'd like to confirm this is backed by a real use case and not just a speculative addition.

ChrisPortman:
> One question: is there actual code on your end that calls ReadConfig directly (e.g. embedding GoBGP config inside a larger config file, as described in the PR description)? I'd like to confirm this is backed by a real use case and not just a speculative addition.

Sorry for the tardy reply, I've been AFK for a week or so.

Yeah, basically I have an application that uses GoBGP as a library for network integration and state monitoring.  At the moment the application runs with 2 separate configuration files, one specifically for the bgp config and the other for the rest of the application config.  Ideally with this change in place, I'll merge the BGP config into the other under a top level key.  Then ultimately, that whole config will be sourced from a configuration server over the network.

fujita:
Pushed, thanks.

PR #3481, opened 2026-06-30
bfd: listen on RFC 5883 multihop port

fivetime:
## Summary
- listen on the RFC 5883 multihop BFD control port (4784) in addition to the configured single-hop port
- keep compatibility with upstream interface/device binding by applying listenAddrs alongside listenInterface
- add regression coverage for single-hop and multihop port handling, multihop session establishment, and Linux bind-to-device + multihop behavior

## Tests
- go test ./pkg/server -run 'Test_BfdServer(AcceptsSingleHopAndMultihopPorts|EstablishesMultihopSessionOnRFC5883Port|ListenAddrsBindToDeviceAndMultihopPort)$' -race -v\n- go test -race -timeout 240s ./pkg/server\n- go test ./pkg/server\n- golangci-lint run ./pkg/server

### DIFF
--- a/pkg/server/bfd_server.go
+++ b/pkg/server/bfd_server.go
@@ -20,6 +20,13 @@ import (
 	"github.com/osrg/gobgp/v4/pkg/packet/bfd"
 )
 
+// bfdMultihopPort is the RFC 5883 (multihop BFD) UDP control port. The server
+// listens on it in addition to the single-hop control port (s.config.Port, the
+// RFC 5881 port 3784) so it can receive returns from multihop peers; a peer runs
+// multihop by sending to this port (its BfdPeerConfig.Port = 4784). Dispatch is
+// by source IP, so a single rxPacket path handles both single-hop and multihop.
+const bfdMultihopPort uint16 = 4784
+
 type bfdServerStats struct {
 	rxPacket      atomic.Uint64
 	rxDrop        atomic.Uint64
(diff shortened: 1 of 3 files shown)


pkg/server/bfd_server.go
```diff
@@ -219,11 +234,11 @@ func (s *bfdServer) loop() {
...
 
 func (s *bfdServer) start() bool {
-	if s.udpServer == nil {
+	if len(s.udpServers) == 0 {
 		s.startServer()
 	}
 
-	return s.udpServer != nil
+	return len(s.udpServers) > 0
 }
```
fujita, line 242: Even if only one of the several (host, port) combinations bound successfully. Is that intentional?

fivetime: Good question — partial success was only partly intentional, and you're right that the old check was too loose. I've split the two ports by role in 3b822b11:

- The single-hop control port (RFC 5881) is now **required**: it must bind on every configured listen address. If it can't, `startServer` rolls back every listener it opened (including any multihop socket that did bind) and reports failure, so the caller retries instead of silently running without the ability to receive single-hop returns.
- The multihop port (RFC 5883/4784) stays **best-effort**: it is commonly already owned by a host BFD daemon, so a bind failure there is logged but does not, on its own, fail startup.

So "only one combination bound" no longer counts as success unless that set covers the full single-hop port. Added a regression test that occupies the single-hop port and asserts the server refuses to come up and tears down the multihop socket it bound.

pkg/server/bfd_server.go
```diff
@@ -50,7 +57,15 @@ type bfdServer struct {
 
 	config *oc.BfdConfig
 
-	udpServer       *net.UDPConn
+	// listenAddrs are the global BGP listen addresses. When non-empty the BFD
+	// server binds its control sockets to each SPECIFIC address (e.g.
+	// 10.0.0.1:4784) instead of the wildcard (:4784). A specific bind wins the
+	// kernel's most-specific-match UDP demux over any wildcard listener already on
+	// the port (e.g. a host bfdd owning 0.0.0.0:3784/4784), so an embedded GoBGP
+	// can run BFD on a host that also runs a system bfdd. Empty → wildcard bind.
```
fujita, line 65: Please remove all the non ascii characters. Just use `->` instead.

PR #3482, opened 2026-07-01
bgp: fix BGP-LS Link/Prefix NLRI key collision for IS-IS L1/L2

geforcebtw:
LsLinkNLRI.String(), LsPrefixV4NLRI.String() and LsPrefixV6NLRI.String() did not include Protocol-ID or Identifier in their output. Since tableKey() hashes String() for LS families, two NLRIs differing only by Protocol-ID (e.g. IS-IS Level-1 vs Level-2) on the same physical link produced identical keys, causing one to overwrite the other in the RIB.

Fix 1: add Protocol-ID and Identifier to the three String() methods, consistent with the existing LsNodeNLRI.String() behavior.

Fix 2: add an explicit case *bgp.LsAddrPrefix in tableKey() that hashes the full serialized NLRI bytes, making the RIB key independent of String() formatting.

Verified with a 3-router ContainerLab topology (Juniper cRPD) running IS-IS dual-level: Wireshark capture confirms the headend sends both L1 and L2 Link NLRIs in the same BGP UPDATE; after this patch both entries appear in 'gobgp global rib -a ls' instead of one silently dropping the other.

### DIFF
--- a/internal/pkg/table/table.go
+++ b/internal/pkg/table/table.go
@@ -87,6 +87,16 @@ func tableKey(nlri bgp.NLRI) addrPrefixKey {
 		h = fnv1a.AddBytes64(h, serializedRD)
 		h = fnv1a.AddBytes64(h, T.Prefix.Addr().AsSlice())
 		h = fnv1a.AddBytes64(h, []byte{uint8(T.Prefix.Bits())})
+        case *bgp.LsAddrPrefix:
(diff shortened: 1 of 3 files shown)


antoinekh:
go test now ok

antoinekh:
golangci-lint should be ok as well

fujita:
Thanks

PR #3483, opened 2026-07-03
bfd: derive detection time from remote timers

fivetime:
## Summary
- compute BFD Detection Time from the remote Detect Mult and the negotiated receive interval per RFC 5880 Section 6.8.4
- avoid duplicate peer resets when an already-down session receives a queued expiry tick
- add regression tests for slower remote transmit intervals and already-down expiry handling

## Tests
- go test ./pkg/server -run 'Test_RxPacketDetectionTimeFromRemote|Test_ExpiryDoesNotResetAlreadyDownPeer|Test_ResetPeer|Test_StateUpDown' -race -v\n- go test ./pkg/server\n- golangci-lint run ./pkg/server

### DIFF
--- a/pkg/server/bfd_peer.go
+++ b/pkg/server/bfd_peer.go
@@ -291,7 +291,18 @@ func (p *bfdPeer) rxPacket(h *bfd.BFDHeader) {
 
 	p.stats.rxPacket.Add(1)
 
-	// NOTE: remote DesiredMinTxInterval and RequiredMinRxInterval ignored
+	// RFC 5880 Section 6.8.4: Detection Time is the remote Detect Mult
+	// multiplied by the negotiated receive interval, i.e. the greater of our
+	// RequiredMinRxInterval and the remote DesiredMinTxInterval.
+	negotiatedRx := p.rxInterval
+	if remoteTx := time.Duration(h.DesiredMinTxInterval) * time.Microsecond; remoteTx > negotiatedRx {
+		negotiatedRx = remoteTx
+	}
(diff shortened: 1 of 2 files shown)


pkg/server/bfd_peer.go
```diff
@@ -291,7 +291,18 @@ func (p *bfdPeer) rxPacket(h *bfd.BFDHeader) {
...
+	// multiplied by the negotiated receive interval, i.e. the greater of our
+	// RequiredMinRxInterval and the remote DesiredMinTxInterval.
+	negotiatedRx := p.rxInterval
+	if remoteTx := time.Duration(h.DesiredMinTxInterval) * time.Microsecond; remoteTx > negotiatedRx {
+		negotiatedRx = remoteTx
+	}
+	remoteMult := time.Duration(h.DetectTimeMultiplier)
+	if remoteMult == 0 {
+		remoteMult = time.Duration(p.multiplier)
+	}
```
fujita: The RFC states that `If the Detect Mult field is zero, the packet MUST be discarded.`

fivetime: Good catch, thanks. Fixed in b6ebb21a: `rxPacket` now discards any packet whose Detect Mult is zero per RFC 5880 Section 6.8.6 (counted in a new `invalidMultiplier` stat), right next to the discriminator check, so an invalid packet can no longer drive a state transition or reset the detection timer. I also removed the local-multiplier fallback, which is now unreachable, and added a regression test (`Test_RxPacketZeroMultiplierDiscarded`).

fujita:
Merged, thanks.

PR #3485, opened 2026-07-04
table: add rtc-prefix match for prefix-set policy conditions

themiron:
Add a prefix-set entry form that matches the Route Target carried in a Route Target Constrain (RFC 4684, rtc family) NLRI against the prefix-set trie, honoring masklength-range over the 96-bit NLRI key (<origin-as>:<route-target>[/masklen]). Unlike match-ext-community-set, which matches the path's extended-communities attribute, this matches the Route Target embedded in the NLRI key itself, and is mutually exclusive with ip-prefix within a prefix-set entry.

The rtc-prefix field is exposed through OpenConfig (prefix-set prefix rtc-prefix), the gRPC API (api.Prefix rtc_prefix) and the gobgp CLI (policy prefix-set prefix <as>:<rt>[/len]). RTC and IPv4 sets are kept apart since both share AFI_IP; an RTC set only matches RTC paths and vice-versa.

Promote the RT trie-key helper to pkg/packet/bgp as the shared allocation-free path (stack-serialized, netip.Prefix returned by value), keeping the policy evaluation hot path at 0 allocs/op for IPv4, IPv6 and RTC. Measured (13th-gen i7-13700H, benchtime=10s):

  BenchmarkPrefixConditionEvaluate/IPv4     30.47 ns/op    0 B/op    0 allocs/op
  BenchmarkPrefixConditionEvaluate/IPv6     25.98 ns/op    0 B/op    0 allocs/op
  BenchmarkPrefixConditionEvaluate/RTC      50.09 ns/op    0 B/op    0 allocs/op
  BenchmarkExtCommRouteTargetKey             4.87 ns/op    0 B/op    0 allocs/op


themiron:
hi @fujita , what’s your opinion regarding this approach? once prefix-set base boilerplate is landed, matching performance can be even more case-optimized I believe, ie for as/rt-as-only w/o full trie check

pkg/server/grpc_server.go
```diff
@@ -1345,12 +1363,32 @@ func newPrefixFromApiStruct(a *api.Prefix) (*table.Prefix, error) {
...
+			return nil, err
+		}
+		ipPrefix = prefix
+	case a.RtcPrefix != "":
+		nlri, masklen, err := bgp.ParseRouteTargetMembershipNLRI(a.RtcPrefix)
+		if err != nil {
+			return nil, err
+		}
+		rtcPrefix = fmt.Sprintf("%s/%d", nlri.String(), masklen)
+	default:
```
fujita, line 1374: Can this handle the wildcard, &api.Prefix{RtcPrefix: "0:0:0/0", ...}?

fujita:
The change to the yang looks nice. I prefer to use `oneof` for the proto file but it breaks the api package compatibility so your change is fine (we should wait for the major version update).

I have one comment about the wildcard match.

pkg/server/grpc_server.go
themiron: now yes, have dropped "default" token from rtc announces representation (that I have previously added to distinguish "0:0:0/0" from "0:0:0/n") -> so rtm prefixes are canonicalized now with "/length" except full "/96" - they are preserved for backward compatibility as length-less.
at the same time, not sure is is really required, so ping please if there's no need for such compatibility - I'll make all rtc announces suffixed with length w/o exceptions.

themiron:
> The change to the yang looks nice. I prefer to use `oneof` for the proto file but it breaks the api package compatibility so your change is fine (we should wait for the major version update).
> 
> I have one comment about the wildcard match.

Yes, I couldn't use one of w/o api breakage, one possible way was to keep ip-prefix but add rtc-prefix as oneof as new - that was looking quite weird in code from my opinion, so I've dropped that.

Just rebased and added new commit for rtm string representation round trips, more tests and fix partial rtm announce lost

pkg/server/grpc_server.go
fujita: I think that we dropped "default" so the compatibility is already gone. Prefixes of other families also always use the format, prefix/length. So I changed:

https://github.com/osrg/gobgp/commit/93dd9a68851097627615749f1882922891efb675

fujita:
Pushed, thanks.

Am I right that internal/pkg/table/rtc.go needs the same treatment before partial lengths actually do anything useful?

themiron:
thank you!

> Am I right that internal/pkg/table/rtc.go needs the same treatment before partial lengths actually do anything useful?

yes, internal rtc subscription logic still has major limitation - only full route target affects path filtration. that means:
1. no as filtering can happen, 111:65000:123/96 matches paths with rt:65000:123 from any as
2. partial rtm 111:65000:0/64 matches paths with rt:65000:0 as direct 64-bit key, but not the rest the prefix covers
therefore for now can be used correctly mainly for reannounces (i.e for route reflector / route server)

to make internal filtration to work, rtc logic needs to be ~trie (probably per as) and not plain map[rt] - seems like quite major change.

PR #3486, opened 2026-07-06
fix: clear ribout cache on peer down

sephiroth8189:
Drop cached BMP route-monitoring state when a peer leaves ESTABLISHED so
identical routes are reported again after the session comes back.


sephiroth8189:
@fujita  can you check this? I think this is an issue.

fujita:
I think that this doesn't work with GR enabled; when the peer is down, the GR is enabled, routes are not removed.

I think that withdrawn generated by DropAll() needs to passed to bmp.

PR #3487, opened 2026-07-07
bmp: defer GR peer down/up and clear ribout cache on session loss

sephiroth8189:
Skip BMP Peer Down while a peer is in graceful restart and suppress the redundant Peer Up on GR recovery by tracking GR-pending peers. Clear the ribout deduplication cache on real session loss so routes can be resent after flap. Keep INIT replay Peer Up behavior for BMP session bootstrap.
By the way,  To limit blast radius, only the BMP file is changed; GR-pending state is in-memory and lost on BMP session restart, which may cause a redundant Peer Up on GR recovery in that edge case.

### DIFF
--- a/pkg/server/bmp.go
+++ b/pkg/server/bmp.go
@@ -83,6 +83,54 @@ func (r ribout) update(p *table.Path) bool {
 	return true
 }
 
+func (r ribout) dropPeer(peer netip.Addr) {
+	for key, paths := range r {
+		n := paths[:0]
+		for _, p := range paths {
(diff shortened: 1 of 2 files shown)


sephiroth8189:
@fujita GR processing has been added, please check it again

fujita:
I prefer to avoid adding GR concept to BMP code so fixed the bug in a different way.
https://github.com/osrg/gobgp/commit/d0f1080575c109428e08467b60fcf5d92d3644fd

PR #3488, opened 2026-07-08
flowspec: add draft-cui packet-content (content-filter) component

hagaisela:
Add BGP FlowSpec support for the draft-cui-idr-content-filter-flowspec packet-content match (component type 254), matching the DNOS wire contract:

  [type=254][ptype/otype:1B][offset:2B BE][content-len:1B][content:N][mask:N]

- pkg/packet/bgp: FlowSpecContentFilter (Type()/Serialize()/DecodeFromBytes), FLOW_SPEC_TYPE_CONTENT_FILTER=254, name map + serialize test.
- api: FlowSpecContentFilter message (type/flags/offset/content/mask) and regenerated attribute.pb.go.
- pkg/apiutil: marshal/unmarshal FlowSpecContentFilter to/from the gRPC API.
- cmd/flowspec-announce: helper to originate/withdraw a single content-filter rule for testing against DNOS.

The ptype/otype byte (ptype high nibble = AFI, otype low nibble = offset base) is carried verbatim from the API so the originator controls the address-family and offset-base semantics.


fujita:
Looks like the draft-cui-idr-content-filter-flowspec is still an individual draft. I'd prefer to wait until the draft is adopted by the WG.

PR #3489, opened 2026-07-11
Propagate BindToDevice to GetBgp

YutaroHayakawa:
The GetBgp currently doesn't return BindToDevice value. Fix it.


fujita:
Thanks!

PR #3490, opened 2026-07-13
packet/bgp: fix SRv6InformationSubTLV sub-sub-TLV decode over-read

Mounika2456:
go test, two SRv6 Information sub-TLVs packed in one SRv6 L3 Service TLV:

    prefix_sid_multi_subtlv_test.go:38: Not equal:
                expected: 1
                actual  : 2
    first Information sub-TLV must not absorb the sibling's bytes as its own sub-sub-TLVs

`SRv6InformationSubTLV.DecodeFromBytes` walks its Sub-Sub-TLVs over `data[p:]`, bounded by the whole slice the caller hands it. `SRv6ServiceTLV`/`SRv6L3ServiceAttribute` pass the entire enclosing Service TLV body (not this sub-TLV alone), so once a second Information sub-TLV follows, the first reads past its own declared `Length` and consumes the sibling's type/length header as a sub-sub-TLV. The sibling is then parsed a second time by the outer loop, so its SID Structure ends up attached to the wrong parent (or the whole attribute is rejected when the borrowed bytes do not form a valid sub-sub-TLV).

Clamp the loop to `subTLVHdrLen + Length`, the end of this sub-TLV's body, and reject a declared length that runs past the buffer.

### DIFF
--- a/pkg/packet/bgp/prefix_sid.go
+++ b/pkg/packet/bgp/prefix_sid.go
@@ -403,11 +403,19 @@ func (s *SRv6InformationSubTLV) DecodeFromBytes(data []byte) error {
 	p += 2
 	// reserved byte
 	p++
-	if p+3 > len(data) {
+	// Sub-Sub-TLVs are bounded by this sub-TLV's declared Length, not by
+	// the caller's remaining buffer, which may still hold sibling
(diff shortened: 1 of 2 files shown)


fujita:
Thanks!

PR #3491, opened 2026-07-14
packet/mrt: fix attribute-length truncation in parseRibEntry

Mounika2456:
go test, a RIB entry declaring a 3-byte attribute region but carrying one oversized extended-length attribute:

    mrt_test.go:156: parseRibEntry accepted an attribute overrunning the declared attribute length

parseRibEntry casts p.Len() to uint16 before the "exceeds remaining attribute length" check. An extended-length path attribute runs up to 65539 bytes, so the cast wraps (65539 -> 3), the guard passes, and the loop consumes the whole attribute past the declared region, then advances data into the following records. On-wire BGP messages cap at 65535 so the equivalent BGPUpdate path-attribute loop cannot reach this; MRT records are not capped, so a crafted TABLE_DUMP_V2 dump does. Compare and subtract in int so the bound holds for attributes larger than 65535.


fujita:
Thanks!

PR #3492, opened 2026-07-15
zebra: reject ZAPI header with Length below header size

Mounika2456:
ReceiveSingleMsg frames the ZAPI body straight from the on-wire Length:

    bodyBuf, err := readAll(conn, int(hd.Len-HeaderSize(version)))

`hd.Len` and `HeaderSize` are both uint16, so a message that declares Len below the header size (e.g. Len=4 on ZAPI v6 where HeaderSize is 10) wraps the subtraction to 65530 instead of producing a short read, and gobgp pulls ~64KiB of the following messages in as this one's body, desyncing the stream from zebra.

Validate `Len >= HeaderSize` while decoding the header so the body-length subtraction can no longer wrap. Added a regression case to Test_Header that fails before the change.

### DIFF
--- a/pkg/zebra/zapi.go
+++ b/pkg/zebra/zapi.go
@@ -1778,6 +1778,12 @@ func (h *Header) decodeFromBytes(data []byte) error {
 	default:
 		return fmt.Errorf("unsupported ZAPI version: %d", h.Version)
 	}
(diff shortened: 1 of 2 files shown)


fujita:
Thanks

PR #3494, opened 2026-07-16
packet/bgp: clamp flowspec components to declared NLRI length

Mounika2456:
go test, a flowspec NLRI whose declared length is shorter than the bytes its components occupy, followed by the first byte of the next NLRI:

    bgp_test.go: "4" is not less than or equal to "3"
    component read one byte past the declared NLRI length, into the following NLRI

`FlowSpecNLRI.decodeFromBytes` uses the declared NLRI length only as a loop counter and hands each component the whole remaining slice. A flowspec component's operator/value bytes are self-terminating, so a component whose bytes run past `length` keeps reading into the next NLRI packed in the same MP_REACH/MP_UNREACH and folds those bytes into the first NLRI, mis-framing the rest of the list. Clamp `data` to the declared length before the component loop.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -4525,6 +4525,12 @@ func (n *FlowSpecNLRI) decodeFromBytes(data []byte, options ...*MarshallingOptio
 		length -= 8
 	}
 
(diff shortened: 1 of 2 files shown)


fujita:
Thanks

PR #3495, opened 2026-07-17
packet/bgp, api, cmd/gobgp: update MUP SAFI support to draft-ietf-bess-mup-safi-01

higebu:
This PR updates the BGP MUP SAFI implementation from draft-mpmz-bess-mup-safi to [draft-ietf-bess-mup-safi-01](https://datatracker.ietf.org/doc/html/draft-ietf-bess-mup-safi-01).

- packet/bgp: add the optional TLVs of the Type 1/2 Session Transformed routes — 3gpp-5g Session Parameters (TEID + QFI), Interwork Endpoint and Source Address — with pass-through of unknown TLV types (Section 3.1.5). Type 1 ST routes with an invalid Source Address Length are now rejected.
- packet/bgp: restructure the BGP MUP Extended Community per Section 3.2 — Sub-Types 0x00-0x02 for Direct Segment and 0x03-0x05 for Interwork Segment, in the 2-octet AS, IPv4 address and 4-octet AS formats.
- api: expose the new TLVs and the per-format extended community messages over gRPC. The deprecated `MUPExtended` message is no longer accepted on unmarshal.
- cmd/gobgp: extend the extended community argument to `mup [direct|interwork] <global administrator>:<local administrator>` (defaults to `direct`; the existing syntax keeps working) and add the `t2st` arguments `session-teid`, `session-qfi`, `interwork-endpoint` and `source-address`.
- docs: update `docs/sources/srv6_mup.md`.

### DIFF
--- a/proto/api/extcom.proto
+++ b/proto/api/extcom.proto
@@ -98,10 +98,29 @@ message TrafficRemarkExtended {
   uint32 dscp = 1;
 }
 
-message MUPExtended {
+// BGP MUP Extended Communities as described in
+// draft-ietf-bess-mup-safi-01 Section 3.2. The sub_type distinguishes
+// both the value format and the semantics: 0x00, 0x01 and 0x02
+// identify a Direct Segment and 0x03, 0x04 and 0x05 an Interwork
+// Segment, in the 2-octet AS, IPv4 address and 4-octet AS formats
+// respectively.
+
+message MUPTwoOctetAsSpecificExtended {
   uint32 sub_type = 1;
-  uint32 segment_id2 = 2;
-  uint32 segment_id4 = 3;
+  uint32 asn = 2;
+  uint32 local_admin = 3;
+}
+
+message MUPIPv4AddressSpecificExtended {
+  uint32 sub_type = 1;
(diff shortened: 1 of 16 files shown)


proto/api/extcom.proto
```diff
@@ -146,10 +174,13 @@ message ExtendedCommunity {
     RedirectIPv4AddressSpecificExtended redirect_ipv4_address_specific = 18;
     RedirectFourOctetAsSpecificExtended redirect_four_octet_as_specific = 19;
     TrafficRemarkExtended traffic_remark = 20;
-    MUPExtended mup = 21;
+    MUPExtended mup = 21 [deprecated = true];
```
fujita: The mup field (MUPExtended, field 21) is no longer handled on unmarshal, so why keep it as deprecated instead of dropping it with reserved 21; (and reserved "mup";)?

higebu: Agreed, there is no reason to keep it. I will drop the field and the `MUPExtended` message, and reserve the field number and name instead.

fujita:
Pushed, thanks!

PR #3496, opened 2026-07-17
table, server: fix update suppression for paths differing only in nexthop or NLRI payload

higebu:
Re-adding a locally originated path with the same route key but a different nexthop or different NLRI payload fields outside the route key (e.g. the TEID of a MUP type-1 session transformed route) is not re-advertised to peers.

`Path.Equal` compares the path attributes hash, which deliberately excludes MP_REACH_NLRI because the hash doubles as the UPDATE batching key. Two paths differing only in nexthop or non-key NLRI content therefore compared as equal, and `Update.GetChanges` suppressed the re-advertisement. This affects all MP families; the same suppression also applies to paths received from peers, since `ProcessMessage` computes the hash with the same exclusion.

- `table: fix Path.Equal ignoring nexthop and NLRI payload` — compare the nexthop and the serialized NLRI explicitly after the hash check.
- `table: make lazy path attributes hash consistent with eager sites` — `updateHash` used fnv1a over all attributes while the eager sites use farm.Hash64 without MP_REACH_NLRI, so paths whose attributes were modified after creation (VRF import, policy) could never compare hash-equal to unmodified ones.
- `server: rely on lazy path attributes hash` — drop the two copies of the eager hash loop that duplicated `updateHash`.

Reproduced by the added tests: `TestPathEqual`, `TestGetChanges_NonKeyNlriOrNexthopOnlyChange`, and `TestWatchBestPathNexthopOnlyChange` fail on master without the fix.

### DIFF
--- a/internal/pkg/table/destination_test.go
+++ b/internal/pkg/table/destination_test.go
@@ -916,3 +916,43 @@ func TestUpdateGetMultiBestPathDiff(t *testing.T) {
 	assert.Equal(t, oldA.GetNexthop(), withdraw[0].GetNexthop())
 	assert.True(t, withdraw[0].IsWithdraw)
 }
+
+// TestGetChanges_NonKeyNlriOrNexthopOnlyChange reproduces update suppression
+// for a locally re-added path whose route key is unchanged but whose NLRI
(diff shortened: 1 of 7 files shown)


fujita:
Thanks!

PR #3497, opened 2026-07-18
packet/bgp: reject out-of-range prefix length in EVPNIPPrefixRoute

Mounika2456:
A crafted EVPN Type-5 (IP Prefix) NLRI decodes with any prefix length:

    [type:Prefix][rd:0:0][etag:0][prefix:1.2.3.4/255]

EVPNIPPrefixRoute.DecodeFromBytes reads the IP Prefix Length octet straight from the wire but never bounds it to the address family, unlike IPAddrPrefixDefault.decodePrefix and the sibling MAC/IP, Multicast and Ethernet-Segment routes (which reject IP lengths outside {0,32,128}). A peer can inject an illegal /33..255 (IPv4) or /129..255 (IPv6) length; the bogus value becomes part of the route key and is re-advertised to other peers. The prefix bytes are sliced with a fixed address length so there is no over-read, only a wrong stored length.

Add the same addrLen*8 check the other prefix decoders already use, and a regression test covering the v4 and v6 cases.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -3002,6 +3002,9 @@ func (er *EVPNIPPrefixRoute) DecodeFromBytes(data []byte) error {
 	er.ETag = binary.BigEndian.Uint32(data[18:22])
 
 	er.IPPrefixLength = data[22]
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks.

PR #3500, opened 2026-07-20
pkg/server: fix Loc-RIB BMP Peer Up to match RFC 9069

### LINKED ISSUE #3499: Loc-RIB BMP feed is unparseable: NLRIs use ADD-PATH path IDs the Peer Up never advertises, and the Peer Up identifies the instance as 0.0.0.0
When playwing with GoBGP to send BMP message to goBMP I find an issue with the loc-RIB BMP feed.

## Summary

With `route-monitoring-policy = "local-rib"` (or `"all"`), every prefix in the Loc-RIB BMP feed is decoded incorrectly by conforming BMP receivers. `10.101.0.0/24` arrives as `101.0.0.0/10`.

The Loc-RIB Route Monitoring messages are always marshalled with ADD-PATH path identifiers, but the Loc-RIB Peer Up that is supposed to describe that encoding disagrees with it in two independent ways. Both violate RFC 9069, and both must be fixed for the feed to be parseable.

This is not a receiver bug: gobgp's own RIB is correct, and gobgp's `pre-policy` / `post-policy` feeds parse perfectly through the same collector on the same TCP session. Only the Loc-RIB peer is affected.

**Version:** gobgp 4.7.0 (latest release) and master (`a06cf9d1`). Reproduced with goBMP as the collector (both `sbezverk/gobmp:latest` and goBMP built from master, identical behaviour).

## The bug at a glance

Both defects are in the **Peer Up**, in two different layers of it. The Route Monitoring messages are correct.

```text
Frame 9: BMP Peer Up
│
├── Per Peer Header              <- the BMP wrapper
│     Type: Loc-RIB Instance Peer (3)   OK
│     Address: 0.0.0.0                  OK   RFC 9069 5.1 zero-fills this one
│     ASN: 0                            BUG 2 - should be 65001
│     BGP ID: 0.0.0.0                   BUG 2 - should be 10.0.0.1
│
└── OPEN Message (x2)            <- the BGP payload inside it
      My AS: 65001                      OK   correct!
      BGP Identifier: 10.0.0.1          OK   correct!
      Optional Parameters Length: 0     BUG 1 - no ADD-PATH capability,
                                                although the Route Monitoring
                                                NLRIs carry path identifiers

Frame 11: BMP Route Monitoring         all OK
      Per Peer Header: ASN 65001, BGP ID 10.0.0.1
      NLRI: 10.101.0.0/24 PathId 1
```

Note what this means: the OPEN inside the Peer Up **already carries the correct AS and router-id**, so gobgp had both values in hand when it built the message and simply did not put them in the per-peer header. Bug 2 needs no RFC to justify it: the message contradicts itself.

Bug 1 lives in the OPEN's capabilities, bug 2 in the per-peer header wrapping it, and the two fixes are correspondingly independent.

## Minimal reproduction

Two containers, both public images, no data links, no BGP peer, one locally injected route. Full files in the attached repro, or inline:

`gobgp.conf`:

```toml
[global.config]
  as = 65001
  router-id = "10.0.0.1"

[[bmp-servers]]
  [bmp-servers.config]
    address = "172.20.30.10"
    port = 5000
    route-monitoring-policy = "local-rib"
```

`gobgp-locrib.clab.yaml` ([containerlab](https://containerlab.dev)):

```yaml
name: gobgp-locrib

mgmt:
  network: gobgp_locrib
  ipv4-subnet: 172.20.30.0/24

topology:
  nodes:
    # BMP collector: prints every parsed message as JSON to its container log.
    gobmp:
      kind: linux
      image: sbezverk/gobmp:latest
      mgmt-ipv4: 172.20.30.10
      cmd: --source-port=5000 --dump=console --logtostderr=true --v=5

    # gobgp exporting only its Loc-RIB over BMP.
    gobgp:
      kind: linux
      image: jauderho/gobgp:latest
      mgmt-ipv4: 172.20.30.11
      binds:
        - gobgp.conf:/config/gobgp.toml:ro
      # The image has no entrypoint, so the whole command line goes here.
      cmd: /usr/local/bin/gobgpd -f /config/gobgp.toml -t toml
```

```bash
clab deploy -t gobgp-locrib.clab.yaml

# inject one route into gobgp's Loc-RIB
docker exec clab-gobgp-locrib-gobgp gobgp global rib add 10.101.0.0/24 origin igp -a ipv4

# what gobgp actually holds
docker exec clab-gobgp-locrib-gobgp gobgp global rib -a ipv4

# what the BMP receiver decoded from the Loc-RIB feed
docker logs clab-gobgp-locrib-gobmp | grep -oE '"prefix":"[^"]*","prefix_len":[0-9]+'
```

**Expected:** the BMP feed reports `10.101.0.0/24`.

**Actual:**

```text
$ docker exec clab-gobgp-locrib-gobgp gobgp global rib -a ipv4
*> 10.101.0.0/24        0.0.0.0                                   00:00:08   [{Origin: i}]

$ docker logs clab-gobgp-locrib-gobmp | grep -oE '"prefix":"[^"]*","prefix_len":[0-9]+'
"prefix":"24.0.0.0","prefix_len":1
"prefix":"101.0.0.0","prefix_len":10
```

One injected `/24` is reported as two nonsense prefixes.

## Root cause

`pkg/server/bmp.go`. Loc-RIB Route Monitoring is **always** marshalled with add-path:

```go
options := bmpAddPathMarshallingOption(p)   // -> BGP_ADD_PATH_BOTH, unconditionally
u := table.CreateUpdateMsgFromPaths([]*table.Path{p}, options)[0]
payload, err := u.Serialize(options)
```

so every NLRI on the wire is prefixed with a 4-octet path identifier. But:

### Bug 1: the ADD-PATH capability is advertised only when global multipath is on

```go
func bmpLocRIBPeerUp(localAS uint32, routerID netip.Addr, tableName string, peerDist uint64, timestamp int64, addPathEnabled bool) *bmp.BMPMessage {
	...
	if addPathEnabled {                       // <-- conditional
		opts = append(opts, bgp.NewOptionParameterCapability(...NewCapAddPath(...)))
	}
```

and the caller gates it on an unrelated setting:

```go
write(bmpLocRIBPeerUp(
	b.s.bgpConfig.Global.Config.As,
	b.s.bgpConfig.Global.Config.RouterId,
	"global",
	0,
	time.Now().Unix(),
	table.UseMultiplePaths.Enabled,           // <-- addPathEnabled
))
```

With the default config the fabricated OPEN carries no optional parameters at all, so a receiver is told "no ADD-PATH" and parses the NLRIs 4 octets out of step.

RFC 9069 §5.2:

> This is a fabricated BGP OPEN message. Capabilities **MUST** include the 4-octet ASN and all necessary capabilities to represent the Loc-RIB Route Monitoring messages.

> Only include capabilities if they will be used for Loc-RIB monitoring messages. For example, if ADD-PATH is enabled for IPv6 and Loc-RIB contains additional paths, the ADD-PATH capability should be included for IPv6.

The 4-octet ASN capability is also missing: it is only added when `localAS > 0xffff`, whereas the RFC requires it unconditionally.

**The existing test suite already encodes this contradiction.** `TestBMPAddPathMarshallingOptionCarriesPathIDOnWithdraw` asserts the Loc-RIB encoding is always add-path (it checks the withdrawn-routes length is `8` = 4 path-id + 1 length + 3 prefix), while `TestBMPLocRIBPeerUpOmitsAddPathCapabilityWhenDisabled` asserts the capability may be omitted. Both pass today.

### Bug 2: the Peer Up contradicts itself about who the Loc-RIB instance is

Enabling multipath makes the Peer Up advertise ADD-PATH correctly, and the prefixes are **still wrong**.

The Peer Up disagrees with itself, in one message. Its per-peer header says the peer is `0.0.0.0` / AS 0, while the fabricated OPEN it carries says `10.0.0.1` / AS 65001:

Image

```text
BMP Peer Up Notification
    Per Peer Header
        Type: Loc-RIB Instance Peer (3)
        Address: 0.0.0.0
        ASN: 0                                  <-- says AS 0
        BGP ID: 0.0.0.0                         <-- says BGP ID 0.0.0.0
    Border Gateway Protocol - OPEN Message
        My AS: 65001                            <-- but the OPEN inside knows the AS
        BGP Identifier: 10.0.0.1                <-- and knows the router-id
        Optional Parameters Length: 0           <-- and advertises no capabilities (bug 1)
    Border Gateway Protocol - OPEN Message
        My AS: 65001
        BGP Identifier: 10.0.0.1
        Optional Parameters Length: 0
    Peer UP/Down TLV: (t=3,l=6) VRF/Table
        VRF/Table name: global
```

The values are right there: `bmpLocRIBPeerUp` passes `routerID` to `NewBGPOpenMessage` two lines before dropping it for `netip.IPv4Unspecified()` in the per-peer header.

```go
open, _ := bgp.NewBGPOpenMessage(myAS, 90, routerID, opts)   // routerID: correct here
ph := bmp.NewBMPPeerHeader(
	bmp.BMP_PEER_TYPE_LOCAL_RIB,
	0,
	peerDist,
	netip.Addr{},
	0,                                                       // Peer AS     = 0
	netip.IPv4Unspecified(),                                 // Peer BGP ID = 0.0.0.0
	float64(timestamp),
)
```

The Route Monitoring messages then use a third variant, the correct one, via `bmpPeerRoute`:

```go
// Route Monitoring: bmpPeerRoute
ph := bmp.NewBMPPeerHeader(t, flags, pd, peeri.Address, peeri.AS, peeri.ID, float64(timestamp))
//                                                      ^^^^^^^^  ^^^^^^^^
//                                          Peer AS = 65001, Peer BGP ID = 10.0.0.1 (router-id)
```

So the same Loc-RIB instance is announced with one identity and then reported with another. Straight from the capture:

```text
bmp.type | bmp.peer.type | bmp.peer.id | bmp.peer.asn
   3     |      3        |  0.0.0.0    |      0        <- Peer Up
   0     |      3        |  10.0.0.1   |    65001      <- Route Monitoring
```

A receiver keys per-session state, including negotiated capabilities, on the per-peer header identity. goBMP uses `PeerBGPID + PeerDistinguisher`, so it files the ADD-PATH capability under `0.0.0.0` and looks it up under `10.0.0.1`: a miss, and the capability is lost even when advertised.

RFC 9069 §5.1 zero-fills only the peer *address*:

> **Peer BGP ID:** Set the ID to the router-id of the VRF instance if VRF is used; otherwise, set to the global instance router-id.

> **Peer Address:** Zero-filled. The remote peer address is not applicable.

> **Peer Autonomous System (AS):** Set to the primary router BGP autonomous system number (ASN).

`bmpPeerRoute` already does this correctly; only `bmpLocRIBPeerUp` deviates.

## Wire evidence

The attached capture is three BMP messages, which is the entire session:

```text
frame   type                    peer type    BGP ID      ASN
  7     Initiation (4)
  9     Peer Up (3)             3 Loc-RIB    0.0.0.0       0
 11     Route Monitoring (0)    3 Loc-RIB    10.0.0.1  65001
```

Frames 9 and 11 describe the same Loc-RIB instance with different identities, which is bug 2 at a glance.

Frame 9, the Peer Up (`tshark -r gobgp-locrib-bug.pcap -d tcp.port==5000,bmp -Y "bmp.type==3" -V`):

```text
BGP Monitoring Protocol, Type Peer Up Notification
    Type: Peer Up Notification (3)
    Per Peer Header
        Type: Loc-RIB Instance Peer (3)
        Peer Distinguisher: 0:0
        Address: 0.0.0.0
        ASN: 0                                  <-- RFC 9069 5.1: must be the router ASN
        BGP ID: 0.0.0.0                         <-- RFC 9069 5.1: must be the router-id
    Border Gateway Protocol - OPEN Message
        Length: 29
        Type: OPEN Message (1)
        Optional Parameters Length: 0           <-- no capabilities at all
    Border Gateway Protocol - OPEN Message
        Optional Parameters Length: 0
```

### The path identifier, as Wireshark decodes it

Frame 11, the Route Monitoring message for that same Loc-RIB instance, expanded to the NLRI:

Image

In text:

```text
BGP Monitoring Protocol, Type Route Monitoring
    Type: Route Monitoring (0)
    Per Peer Header
        Type: Loc-RIB Instance Peer (3)
        Address: 0.0.0.0
        ASN: 65001                              <-- the Peer Up said 0
        BGP ID: 10.0.0.1                        <-- the Peer Up said 0.0.0.0
    Border Gateway Protocol - UPDATE Message
        Path attributes
            Path Attribute - ORIGIN: IGP
            Path Attribute - NEXT_HOP: 0.0.0.0
        Network Layer Reachability Information (NLRI)
            10.101.0.0/24 PathId 1
                NLRI path id: 1                 <-- path identifier, never announced
                Prefix Length: 24
                NLRI prefix: 10.101.0.0
```

`NLRI path id: 1` next to a Peer Up carrying `Optional Parameters Length: 0` is the contradiction in a nutshell.

Wireshark shows the prefix correctly here only because its BGP dissector *heuristically* detects add-path from the NLRI length arithmetic; it is a sniffer with no session state, so guessing is all it can do (verified: `tshark` with default preferences prints `10.101.0.0/24 PathId 1`, and there is no add-path preference in the defaults). A BMP receiver decodes according to the capabilities the Peer Up declared, which is what the Peer Up is for and what §5.2 requires it to describe. That a heuristic can paper over the bug is not evidence the feed is well-formed.

## Byte-level confirmation

Replaying the real wire bytes through a parser that was told "no ADD-PATH" reproduces the observed output exactly:

```python
wire = bytes([0x00,0x00,0x00,0x01, 0x18, 0x0A,0x65,0x00])   # path-id 1 + 10.101.0.0/24
i, out = 0, []
while i < len(wire):
    plen = wire[i]; i += 1
    nbytes = (plen + 7) // 8
    octets = list(wire[i:i+nbytes]) + [0]*(4-nbytes); i += nbytes
    out.append(f"{'.'.join(str(o) for o in octets)}/{plen}")
# -> ['0.0.0.0/0', '0.0.0.0/0', '0.0.0.0/0', '24.0.0.0/1', '101.0.0.0/10']
```

`24.0.0.0/1` and `101.0.0.0/10` are exactly what the collector reports, which confirms the path identifiers are on the wire and that the receiver is behaving correctly for a session that declared no ADD-PATH.

## Impact

Every prefix in the Loc-RIB feed is wrong, for any conforming receiver. The rest of each message (`peer_type`, `is_loc_rib`, next-hop, AS path, attributes) is correct, which makes the failure easy to miss: the feed looks alive and plausible, only the prefixes are silently garbage. Loc-RIB is the view most people reach for (it is the "looking glass" view), so this affects the headline use case of RFC 9069 support.

## Workarounds

- `route-monitoring-policy = "pre-policy"` or `"post-policy"` are unaffected.
- Enabling `[global.use-multiple-paths.config] enabled = true` fixes bug 1 only; bug 2 still breaks the feed.

## Fix

A patch fixing both is ready and verified end to end (correct prefixes and path-ids with the default config, matching `gobgp global rib`); `go test ./pkg/server/` passes in full. Happy to open a PR.

- always advertise the ADD-PATH capability, since the Loc-RIB encoder is unconditionally add-path, and drop the now-meaningless `addPathEnabled` parameter
- always advertise the 4-octet ASN capability, per the §5.2 MUST
- set the Peer Up's per-peer header Peer AS and Peer BGP ID to the router's, matching `bmpPeerRoute` and §5.1, and the same for `bmpLocRIBPeerDown`, which zero-fills them identically
- replace `TestBMPLocRIBPeerUpOmitsAddPathCapabilityWhenDisabled` with one asserting the capability is always present, and add tests for the 4-octet ASN capability and for the Peer Up / Peer Down header identity

## Related

- #3239 (BMP local-rib not sending PeerUp on new session) touches the same code path but is a different symptom.
- #1281 is the original Loc-RIB feature request.

antoinekh:
Fixes #3499

Loc-RIB Route Monitoring messages are always marshalled with `BGP_ADD_PATH_BOTH` (`bmpAddPathMarshallingOption`), but the Peer Up that is supposed to describe that encoding disagreed with it in two ways, which left the feed unparseable by any conforming receiver: an injected `10.101.0.0/24` is decoded as `101.0.0.0/10`.

### 1. The fabricated OPEN did not advertise the capabilities the encoder uses

The ADD-PATH capability was only advertised when `table.UseMultiplePaths.Enabled` was set, an unrelated global multipath setting. With the default config the OPEN carried no optional parameters at all, so a receiver was told "no ADD-PATH" and parsed every NLRI 4 octets out of step.

RFC 9069 §5.2 requires the fabricated OPEN's capabilities to "represent the Loc-RIB Route Monitoring messages", and requires the 4-octet ASN capability unconditionally (it was only sent when the ASN did not fit in the 2-octet My Autonomous System field).

Both capabilities are now always advertised, and the `addPathEnabled` parameter is removed, since the Loc-RIB encoder is unconditionally add-path.

### 2. The per-peer header did not identify the router

The Peer Up zero-filled the Peer BGP ID and the Peer AS, while the Route Monitoring messages carry the router-id and the local AS (`bmpPeerRoute`). Receivers key per-session state such as negotiated capabilities on that identity, so the capabilities were filed under `0.0.0.0` and looked up under the router-id, and were lost even once advertised.

RFC 9069 §5.1 zero-fills only the Peer Address: the Peer BGP ID is "the global instance router-id" and the Peer AS is "the primary router BGP autonomous system number". The Peer Up even contradicted itself, since the fabricated OPEN it carries already held the correct AS and router-id.

`bmpLocRIBPeerDown` zero-filled the same two fields and is fixed with it, so a receiver can tie the teardown to the instance it was told about.

### Tests

`TestBMPLocRIBPeerUpOmitsAddPathCapabilityWhenDisabled` asserted the first bug, and is replaced by `TestBMPLocRIBPeerUpAlwaysCarriesAddPathCapability`. That is consistent with `TestBMPAddPathMarshallingOptionCarriesPathIDOnWithdraw`, which already asserts the Loc-RIB encoding is always add-path: the two tests previously contradicted each other. Added `TestBMPLocRIBPeerUpAlwaysCarriesFourOctetASCapability`, `TestBMPLocRIBPeerUpHeaderIdentifiesTheRouter` and `TestBMPLocRIBPeerDownHeaderIdentifiesTheRouter`.

`go test ./pkg/server/` passes in full.

### Verification

Against goBMP as the receiver, with the **default** config (multipath off, the case that was broken):

before

```text
    prefix    | path_id
--------------+---------
 101.0.0.0/10 |
 24.0.0.0/1   |
```

after

```text
    prefix     | path_id
---------------+---------
 10.101.0.0/24 | 1
 10.102.0.0/24 | 1
 10.99.0.0/24  | 1
```

matching `gobgp global rib`. With multipath additionally enabled, multiple paths for the same prefix appear with distinct path-ids, as expected.

Each fix was also tested in isolation: fixing only the header still leaves the capability unadvertised with the default config, and fixing only the capability still leaves it filed under an identity the Route Monitoring messages never use. Both are required.

### DIFF
--- a/pkg/server/bmp.go
+++ b/pkg/server/bmp.go
@@ -199,7 +199,6 @@ func (b *bmpClient) loop() {
 					"global",
 					0,
 					time.Now().Unix(),
-					table.UseMultiplePaths.Enabled,
 				)); err != nil {
 					return false
 				}
@@ -309,36 +308,49 @@ func (b *bmpClient) loop() {
 	}
 }
 
-func bmpLocRIBPeerUp(localAS uint32, routerID netip.Addr, tableName string, peerDist uint64, timestamp int64, addPathEnabled bool) *bmp.BMPMessage {
+func bmpLocRIBPeerUp(localAS uint32, routerID netip.Addr, tableName string, peerDist uint64, timestamp int64) *bmp.BMPMessage {
 	const asTrans uint16 = 23456
 
 	myAS := asTrans
 	opts := []bgp.OptionParameterInterface{}
 	if localAS <= 0xffff {
 		myAS = uint16(localAS)
-	} else {
-		opts = append(opts, bgp.NewOptionParameterCapability([]bgp.ParameterCapabilityInterface{
-			bgp.NewCapFourOctetASNumber(localAS),
-		}))
 	}
-	if addPathEnabled {
-		opts = append(opts, bgp.NewOptionParameterCapability([]bgp.ParameterCapabilityInterface{
-			bgp.NewCapAddPath([]*bgp.CapAddPathTuple{
-				bgp.NewCapAddPathTuple(bgp.RF_IPv4_UC, bgp.BGP_ADD_PATH_BOTH),
-				bgp.NewCapAddPathTuple(bgp.RF_IPv6_UC, bgp.BGP_ADD_PATH_BOTH),
-			}),
-		}))
-	}
+	// RFC 9069 5.2: "Capabilities MUST include the 4-octet ASN and all necessary
+	// capabilities to represent the Loc-RIB Route Monitoring messages." The
+	// 4-octet ASN capability is therefore advertised unconditionally, not only
+	// when the ASN does not fit in the 2-octet My Autonomous System field.
+	opts = append(opts, bgp.NewOptionParameterCapability([]bgp.ParameterCapabilityInterface{
+		bgp.NewCapFourOctetASNumber(localAS),
+	}))
+	// Loc-RIB Route Monitoring messages are always marshalled with
+	// BGP_ADD_PATH_BOTH (see bmpAddPathMarshallingOption), so the receiver always
+	// needs the ADD-PATH capability to decode their NLRIs. Advertising it only
+	// when global multipath happened to be enabled left the fabricated OPEN
+	// describing an encoding that was not the one on the wire, and every NLRI was
+	// then parsed 4 octets out of step.
+	opts = append(opts, bgp.NewOptionParameterCapability([]bgp.ParameterCapabilityInterface{
+		bgp.NewCapAddPath([]*bgp.CapAddPathTuple{
+			bgp.NewCapAddPathTuple(bgp.RF_IPv4_UC, bgp.BGP_ADD_PATH_BOTH),
+			bgp.NewCapAddPathTuple(bgp.RF_IPv6_UC, bgp.BGP_ADD_PATH_BOTH),
+		}),
+	}))
 
 	open, _ := bgp.NewBGPOpenMessage(myAS, 90, routerID, opts)
 
+	// RFC 9069 5.1: for a Loc-RIB Instance Peer only the Peer Address is
+	// zero-filled. The Peer AS is "the primary router BGP autonomous system
+	// number" and the Peer BGP ID is "the global instance router-id". They must
+	// match the header the Route Monitoring messages carry (see bmpPeerRoute),
+	// otherwise a receiver cannot correlate this Peer Up with them and loses the
+	// capabilities negotiated above.
 	ph := bmp.NewBMPPeerHeader(
 		bmp.BMP_PEER_TYPE_LOCAL_RIB,
 		0,
 		peerDist,
 		netip.Addr{},
-		0,
-		netip.IPv4Unspecified(),
+		localAS,
+		routerID,
 		float64(timestamp),
 	)
 	return bmp.NewBMPPeerUpNotification(
@@ -353,13 +365,17 @@ func bmpLocRIBPeerUp(localAS uint32, routerID netip.Addr, tableName string, peer
 }
 
 func bmpLocRIBPeerDown(localAS uint32, routerID netip.Addr, tableName string, peerDist uint64, timestamp int64) *bmp.BMPMessage {
+	// RFC 9069 5.1: as in bmpLocRIBPeerUp, only the Peer Address is zero-filled.
+	// The Peer AS and Peer BGP ID identify the router, so that a receiver can tie
+	// this Peer Down to the Peer Up and the Route Monitoring messages for the
+	// same Loc-RIB instance.
 	ph := bmp.NewBMPPeerHeader(
 		bmp.BMP_PEER_TYPE_LOCAL_RIB,
 		0,
 		peerDist,
 		netip.Addr{},
-		0,
-		netip.IPv4Unspecified(),
+		localAS,
+		routerID,
 		float64(timestamp),
 	)
 	return bmp.NewBMPPeerDownNotification(
--- a/pkg/server/bmp_test.go
+++ b/pkg/server/bmp_test.go
@@ -96,18 +96,41 @@ func TestBMPAddPathMarshallingOptionCarriesPathIDOnWithdraw(t *testing.T) {
 	_ = update.WithdrawnRoutes[0].ID
 }
 
-func localRIBPeerUpOpen(t *testing.T, addPathEnabled bool) *bgp.BGPOpen {
+const (
+	locRIBTestAS       = uint32(65002)
+	locRIBTestRouterID = "100.1.1.102"
+)
+
+func localRIBPeerUp(t *testing.T) *packetbmp.BMPMessage {
 	t.Helper()
-	msg := bmpLocRIBPeerUp(
-		65002,
-		netip.MustParseAddr("100.1.1.102"),
+	return bmpLocRIBPeerUp(
+		locRIBTestAS,
+		netip.MustParseAddr(locRIBTestRouterID),
 		"global",
 		0,
 		time.Now().Unix(),
-		addPathEnabled,
 	)
-	up := msg.Body.(*packetbmp.BMPPeerUpNotification)
+}
+
+func localRIBPeerUpOpen(t *testing.T) *bgp.BGPOpen {
+	t.Helper()
+	up := localRIBPeerUp(t).Body.(*packetbmp.BMPPeerUpNotification)
 	return up.SentOpenMsg.Body.(*bgp.BGPOpen)
+}
+
+func findFourOctetASCapability(open *bgp.BGPOpen) *bgp.CapFourOctetASNumber {
+	for _, p := range open.OptParams {
+		param, ok := p.(*bgp.OptionParameterCapability)
+		if !ok {
+			continue
+		}
+		for _, c := range param.Capability {
+			if fourOctet, ok := c.(*bgp.CapFourOctetASNumber); ok {
+				return fourOctet
+			}
+		}
+	}
+	return nil
 }
 
 func findAddPathCapability(open *bgp.BGPOpen) *bgp.CapAddPath {
@@ -165,13 +188,12 @@ func TestRiboutWithdrawKeepsOtherPeersPath(t *testing.T) {
 	require.True(t, r.update(p1))
 }
 
-func TestBMPLocRIBPeerUpOmitsAddPathCapabilityWhenDisabled(t *testing.T) {
-	open := localRIBPeerUpOpen(t, false)
-	require.Nil(t, findAddPathCapability(open))
-}
-
-func TestBMPLocRIBPeerUpCarriesAddPathCapabilityWhenEnabled(t *testing.T) {
-	open := localRIBPeerUpOpen(t, true)
+// Loc-RIB Route Monitoring messages are always marshalled with add-path (see
+// TestBMPAddPathMarshallingOptionCarriesPathIDOnWithdraw), so the fabricated
+// OPEN must always advertise the capability. Omitting it left receivers parsing
+// the NLRIs 4 octets out of step. RFC 9069 5.2.
+func TestBMPLocRIBPeerUpAlwaysCarriesAddPathCapability(t *testing.T) {
+	open := localRIBPeerUpOpen(t)
 	addPath := findAddPathCapability(open)
 	require.NotNil(t, addPath)
 	require.Len(t, addPath.Tuples, 2)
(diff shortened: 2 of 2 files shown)


antoinekh:
I can remove comments related to RFC if needed

fujita:
Great, thanks!

PR #3501, opened 2026-07-20
packet/bmp: validate declared length in TLV16 value decoders

Mounika2456:
A BMP Termination "Reason" TLV that declares a 1-octet value:

    tlv[0] Reason16 value=0xaa00

The reason byte on the wire is 0xaa; the trailing 0x00 is read one octet
past the declared value, out of the following TLV's type field.

BMPTermTLV16.ParseValue and BMPRouteMirrTLV16.ParseValue read a fixed
Uint16(data[:2]) without checking the TLV length, unlike BMPStatsTLV32/64
which reject a mismatched Length. ParseBody only guarantees
len(data) >= Length, so a TLV shorter than 2 octets reads into the next
TLV and misframes the rest of the message; a Length of 0 also leaves the
loop advance at zero so it never terminates. Reject Length != 2 in both,
matching the Stats TLV parsers.

### DIFF
--- a/pkg/packet/bmp/bmp.go
+++ b/pkg/packet/bmp/bmp.go
@@ -810,6 +810,9 @@ func NewBMPTermTLV16(t uint16, v uint16) *BMPTermTLV16 {
 }
 
 func (s *BMPTermTLV16) ParseValue(data []byte) error {
+	if s.Length != 2 {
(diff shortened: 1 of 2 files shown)


fujita:
Thanks

Mounika2456:
Thanks for picking this up, glad it's in master.

PR #3502, opened 2026-07-20
packet/bgp: clamp SR Policy NLRI endpoint to declared length

Mounika2456:
go test, two SR Policy IPv4 NLRIs packed in one MP_REACH:

    endpoint absorbed sibling NLRI bytes: got 17 bytes (0a0000016000000001000000020a000002), want 4

SRPolicyNLRI.decodeFromBytes sets s.Endpoint = data[p:], but NLRIFromSlice hands it the whole remaining attribute buffer. With more than one SR Policy NLRI in an UPDATE, every NLRI except the last stores an endpoint that aliases the following siblings, so String()/MarshalJSON print the wrong endpoint (net.IP().To4()/To16() sees the over-length slice and returns <nil>). The declared length (data[0]/8) already fixes the endpoint at 4 bytes for IPv4 or 16 for IPv6, so slice to that instead of the rest of the buffer.

### DIFF
--- a/pkg/packet/bgp/sr_policy.go
+++ b/pkg/packet/bgp/sr_policy.go
@@ -50,7 +50,15 @@ func (s *SRPolicyNLRI) decodeFromBytes(data []byte, options ...*MarshallingOptio
 	p += 4
 	s.Color = binary.BigEndian.Uint32(data[p : p+4])
 	p += 4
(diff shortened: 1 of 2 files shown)


fujita:
Merged, thanks.

Issue #3493, opened 2026-07-15
TCP-AO Implementation


rastislavs:
This issue is for design & tracking of TCP-AO authentication in GoBGP ([RFC 5925](https://www.rfc-editor.org/info/rfc5925/)).

I plan to work on the TCP-AO implementation for GoBGP, and would like to ask for an early feedback for the API.

## Motivation
TCP-AO is an alternative for the existing TCP MD5 signature authentication, which is already considered weak and obsolete. TCP-AO allows use of various stronger algorithms, such as HMAC-SHA1 and AES-128 to create an internal traffic key and message digest. It is also better suited for long-lived connections, as it supports hitless rekeying without breaking TCP connection.

## Keychain API
Since TCP-AO allows hitless key-rollover, most of the existing router implementations keep the master keys (MKTs) in a keychain, that can be used among multiple peers. Each peer then can use a different key from the keychain, and can roll-over to a different key from the chain  independently. This concept matches with the [OpenConfig keychain YANG model](https://github.com/openconfig/public/blob/master/release/models/keychain/openconfig-keychain.yang) as well, so we will stick with it. 

```protobuf
enum TcpAoAlgorithm {
 TCP_AO_ALGORITHM_UNSPECIFIED = 0;
 TCP_AO_ALGORITHM_HMAC_SHA1_96 = 1;
 TCP_AO_ALGORITHM_AES_128_CMAC_96 = 2;
}

message TcpAoKey {
 uint32 send_id = 1;
 uint32 receive_id = 2;
 TcpAoAlgorithm algorithm = 3;

 // Write-only. Required when adding the key and cleared in every response.
 bytes master_key = 4 [debug_redact = true];

 bool exclude_tcp_options = 5; // all TCP options other than TCP-AO are excluded from MAC calculations so they can be changed on middleboxes
}


message TcpAoKeychain {
 string name = 1;
 repeated TcpAoKey keys = 2;
}

message AddTcpAoKeychainRequest {
 TcpAoKeychain keychain = 1;
}

message AddTcpAoKeychainResponse {
 TcpAoKeychain keychain = 1;
}

message UpdateTcpAoKeychainRequest {
 string name = 1;
 repeated TcpAoKey add_keys = 2;
 repeated TcpAoKey delete_keys = 2; // only send_id and receive_id is used
}

message UpdateTcpAoKeychainResponse {
 TcpAoKeychain keychain = 1;
}

message DeleteTcpAoKeychainRequest {
 string name = 1;
[...]

PR #3503, opened 2026-07-21
api,server: add TCP-AO keychain API and  keychain store implementation

rastislavs:
The is the first part of the TCP-AO implementation (https://github.com/osrg/gobgp/issues/3493) which adds the proto API for TCP-AO keychain configuration and implementation of the keychain store.

The configured keychains do not have any use for now, this will be added in follow-up changes as part of https://github.com/osrg/gobgp/issues/3493.

### DIFF
--- a/api/gobgp.pb.go
+++ b/api/gobgp.pb.go
@@ -647,6 +647,55 @@ func (BfdDiagnosticCode) EnumDescriptor() ([]byte, []int) {
 	return file_api_gobgp_proto_rawDescGZIP(), []int{10}
 }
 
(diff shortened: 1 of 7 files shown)


rastislavs:
Updated this PR to use the `netutils.TCPAOKey` type from https://github.com/osrg/gobgp/commit/08ee3960d4efa42d62062752df85676ea4fa40b8 instead of replicating it.

fujita:
Very sorry about the delay. I've just rebased and pushed this. I also made some changes related with this so please check out them.

PR #3504, opened 2026-07-21
server: resolve unnumbered NeighborInterface before validating neighbor address

### LINKED ISSUE #3023: Extract IPv6 link-local neighbor address from network interface inside AddPeer
It seems that via the CLI and conf file, the wrapper populates `State.NeighborAddress` from the network interface via `GetIPv6LinkLocalNeighborAddress` before passing in the peer config to `AddPeer`. I am wondering if `GetIPv6LinkLocalNeighborAddress` can just be called inside `AddPeer` so that the API properly supports BGP unnumbered in only passing the interface (since right now, it requires `Config.NeighborAddress` or `State.NeighborAddress` to be populated)? Was the way this was done an intentional design choice?

ecbaldwin:
Fixes #3023.

Problem

Adding an unnumbered (interface-only) neighbor via the gRPC AddPeer API — NeighborInterface set, NeighborAddress empty — is rejected up front with:

NeighborAddress is not configured

so gobgp's native unnumbered support is unreachable over gRPC. Only the config-file path works.

Cause

In (*BgpServer).addNeighbor the order was:

1. c.ExtractNeighborAddress() — errors on an addressless neighbor
2. peer-group lookup
3. oc.SetDefaultNeighborConfigValues(...)

But step 3 is what resolves the interface's link-local: SetDefaultNeighborConfigValues calls GetIPv6LinkLocalNeighborAddress(NeighborInterface) and stores the result in State.NeighborAddress (pkg/config/oc/default.go). ExtractNeighborAddress reads State.NeighborAddress first (pkg/config/oc/util.go) — but it ran before that field was populated. The config-file/CLI path defaults the neighbor config before reaching addNeighbor, which is why it doesn't hit this.

Fix

Reorder addNeighbor so peer-group lookup + SetDefaultNeighborConfigValues run first, then ExtractNeighborAddress sees the resolved zoned link-local. This makes the gRPC path behave identically to the config-file path, and reuses the existing defaulting routine rather than adding a second resolution call site.

Re #3023's question of whether the ordering was intentional: it reads as an ordering oversight — the resolution logic already lives in SetDefaultNeighborConfigValues; addNeighbor just validated before invoking it.

Testing

Added a unit test (TestAddPeerUnnumberedInterface) that adds an interface-only neighbor via AddPeer and asserts it reaches interface resolution rather than being rejected by the address-validation gate. Verified it fails on the pre-fix ordering with "NeighborAddress is not configured" and passes with the fix. A full end-to-end unnumbered session was validated separately against an FRR peer — interface resolution needs a real IPv6 link-local neighbor in the netlink table, so it isn't reproducible in a hermetic unit test (matching where gobgp's existing unnumbered coverage lives, the scenario_test integration suite).

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -3453,15 +3453,14 @@ func (s *BgpServer) addPeerGroup(c *oc.PeerGroup) error {
 }
 
 func (s *BgpServer) addNeighbor(c *oc.Neighbor) error {
-	addr, err := c.ExtractNeighborAddress()
-	if err != nil {
-		return err
-	}
-
-	if _, y := s.neighborMap[netip.MustParseAddr(addr)]; y {
-		return fmt.Errorf("can't overwrite the existing peer: %s", addr)
-	}
-
+	// Resolve config defaults BEFORE extracting/validating the neighbor address.
+	// For an unnumbered (interface-only) neighbor added via the gRPC AddPeer API
+	// - NeighborInterface set, NeighborAddress empty - SetDefaultNeighborConfigValues
+	// resolves the peer's IPv6 link-local from the interface into
+	// State.NeighborAddress and derives the local link-local as the transport
+	// source address. ExtractNeighborAddress would otherwise reject the addressless
+	// neighbor up front with "NeighborAddress is not configured". The config-file
+	// path already defaults before addNeighbor; this makes the gRPC path match.
 	var pgConf *oc.PeerGroup
 	if c.Config.PeerGroup != "" {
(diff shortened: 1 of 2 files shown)


fujita:
Thanks!

ecbaldwin:
I would love to get this into cilium. What is timeline on a `v4.8.0` that we could vendor over there?

fujita:
> I would love to get this into cilium. What is timeline on a `v4.8.0` that we could vendor over there?

I release a new version at the beginning of each month, so expect v4.8.0 this weekend or early next week.

PR #3505, opened 2026-07-23
fix(server): prevent watcher leaks during shutdown

frezbo:
Ensure queued management operations and watcher cleanup can complete after the server loop exits instead of leaving blocked goroutines.


fujita:
Merged, thanks.

PR #3506, opened 2026-07-23
netutils: add TCP-AO socket helpers

rastislavs:
Add helpers that will be used to attach and control TCP-AO keys to TCP sockets of BGP peers.

Right now they do not have any use except for tests, this will be added in a follow-up change.

(part of the TCP-AO implementation tracked in https://github.com/osrg/gobgp/issues/3493)

### DIFF
--- a/internal/pkg/netutils/tcp_ao_linux_test.go
+++ b/internal/pkg/netutils/tcp_ao_linux_test.go
@@ -1,0 +1,238 @@
+// Copyright (C) 2026 The GoBGP Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build linux
+
+package netutils
+
+import (
+	"context"
+	"errors"
+	"net"
+	"net/netip"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
+)
+
+func skipTCPAONotSupported(t *testing.T, err error) {
+	t.Helper()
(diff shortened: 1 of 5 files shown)


internal/pkg/netutils/tcp_ao_linux_test.go
```diff
@@ -0,0 +1,191 @@
...
+	peer := netip.MustParsePrefix("127.0.0.1/32")
+	config := TCPAOConfig{Keys: []TCPAOKey{{
+		SendID: 7, ReceiveID: 9, Algorithm: TCPAOAlgorithmHMACSHA1, MasterKey: []byte("secret"),
+	}}}
+	listenConfig := net.ListenConfig{Control: func(_, _ string, raw syscall.RawConn) error {
+		return AddTCPAOKeysSockopt(raw, peer, "", config)
+	}}
+	listenConfig.SetMultipathTCP(false)
+	listener, err := listenConfig.Listen(context.Background(), "tcp4", "127.0.0.1:0")
+	t.Cleanup(func() { listener.Close() })
```
fujita: We hit an error here, listener is nil?

internal/pkg/netutils/tcp_ao_uapi.go
```diff
@@ -0,0 +1,371 @@
...
+	expectedRecordSize := binary.Size(tcpAOGetKeyABI{})
+	if recordSize != uint32(expectedRecordSize) {
+		return nil, fmt.Errorf("TCP-AO key state record size is %d, want %d", recordSize, expectedRecordSize)
+	}
+
+	var first tcpAOGetKeyABI
+	if _, err := binary.Decode(buffer[:recordSize], binary.NativeEndian, &first); err != nil {
+		return nil, fmt.Errorf("failed to decode TCP-AO key count: %w", err)
+	}
+	if first.KeyCount > tcpAOMaxKeys {
```
fujita: A listen socket can have more than 256 keys?

internal/pkg/netutils/tcp_ao_linux_test.go
rastislavs: good catch, fixed

internal/pkg/netutils/tcp_ao_uapi.go
rastislavs: Good point, modified the code to allow returning keys from listen sockets as well. Not sure if we have an actual use case for it, but it's good to support it on this level.

fujita:
Pushed, thanks.

PR #3507, opened 2026-07-23
packet/bgp: reject a zero length next hop for families that require one

### LINKED ISSUE #3450: MP_REACH_NLRI with nexthop_len=0 accepted for non-FlowSpec families (RFC 4760 §3)
Found another spec compliance gap in the BGP parser, separate from the OPEN/header issues in #3449.

`PathAttributeMpReachNLRI.DecodeFromBytes` accepts MP_REACH_NLRI attributes with `nexthop_len=0` for any AFI/SAFI, not just FlowSpec. RFC 4760 §3 says nexthop_len=0 is only valid for FlowSpec (SAFI 134/135 per RFC 8956). For IPv4/IPv6 unicast, a zero-length nexthop is malformed.

The code in `pkg/packet/bgp/bgp.go:12227` has:
```go
switch nexthoplen {
case 0: // no nexthop, skip (FlowSpec)
```

The comment says "FlowSpec" but there's no actual `safi` check — it accepts nexthop_len=0 for every AFI/SAFI.

FRR 10.6.1 explicitly whitelists FlowSpec and rejects everything else:
```c
case 0:
    if (safi != SAFI_FLOWSPEC) {
        return BGP_ATTR_PARSE_ERROR_NOTIFYPLS;
    }
```

Tested all 3 parser entry points (`ParseBGPMessage`, `BGPUpdate.DecodeFromBytes`, `PathAttributeMpReachNLRI.DecodeFromBytes`) — all return `err=nil` for AFI=1/SAFI=1 with nexthop_len=0.

**Fix:** Add a family check before the `case 0` branch:
```go
case 0:
    if p.SAFI != SAFI_FLOWSPEC {
        return NewMessageError(...)
    }
```

Found through differential fuzzing (PathDiff v23), comparing gobgp vs FRR 10.6.1.

olasupo:
Fixes #3450

PathAttributeMpReachNLRI.DecodeFromBytes accepted a next hop length of zero for every AFI and SAFI. Per RFC 4760 section 3 that is only valid for families that carry no next hop, so for IPv4 or IPv6 unicast it is malformed. gobgp accepted it silently.

This rejects a zero length next hop except for the families that legitimately omit one: the FlowSpec families and gobgp's opaque key/value family. A flowspec only check would wrongly reject opaque (caught by Test_AddPath). The MRT only path is unchanged.

Testing: new test TestMpReachZeroNexthopFamilyValidation (unicast rejected, FlowSpec still round trips; passes with the fix, fails without). Full unit suite, gofmt, and go vet are clean.

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -12751,7 +12751,18 @@ func (p *PathAttributeMpReachNLRI) DecodeFromBytes(data []byte, options ...*Mars
 	}
 
 	switch nexthoplen {
-	case 0: // no nexthop, skip (FlowSpec)
+	case 0:
+		// A zero-length next hop is only valid for the families that do
+		// not carry a next hop: the FlowSpec families and the gobgp opaque
+		// key/value family. It is malformed for any other family (e.g.
+		// IPv4/IPv6 unicast). The AFI/SAFI is not decoded on the MRT-only
+		// next hop path, so that path is left unchanged.
+		if !onlyNexthop &&
+			p.SAFI != SAFI_FLOW_SPEC_UNICAST &&
+			p.SAFI != SAFI_FLOW_SPEC_VPN &&
+			p.SAFI != SAFI_KEY_VALUE {
+			return NewMessageError(eCode, eSubCode, eData, "mpreach nexthop length is zero for a family that requires a next hop")
+		}
 	case BGP_ATTR_NHLEN_IPV6_GLOBAL_AND_LL: // 16 bytes IPv6 Global + 16 bytes IPv6 Link Local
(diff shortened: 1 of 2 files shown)


pkg/packet/bgp/bgp.go
```diff
@@ -12751,7 +12751,19 @@ func (p *PathAttributeMpReachNLRI) DecodeFromBytes(data []byte, options ...*Mars
 	}
 
 	switch nexthoplen {
-	case 0: // no nexthop, skip (FlowSpec)
+	case 0:
+		// RFC 4760 section 3 (and RFC 8955): a zero-length next hop is only
```
fujita: RFC 4760 defines the MP format, doesn't mention a zero-length next hop. So remove it.

olasupo: Hi Fujita. Thank you for your review. I have now dropped the RFC 4760 reference from the comment.

fujita:
Merged, thanks.

PR #3509, opened 2026-07-24
packet/rtr: bound RTRIPPrefix prefix and max length to address family

Mounika2456:
A crafted RTR IPv4 Prefix PDU with Prefix Length 33 (RFC 8210 5.1 bounds it to
0..32) decodes without complaint, and the ROA built from it comes out as:

    ROA.Network = "<nil>"   // net.CIDRMask(33, 32) returned nil

RTRIPPrefix.DecodeFromBytes takes PrefixLen and MaxLen straight from the wire.
roaManager.handleRTRMsg hands the length to table.NewROA, and net.CIDRMask
returns a nil mask once the length is past the address size, so the ROA lands
in the critbit tree keyed as a /0 default route that matches every prefix
during origin validation. NewRTRIPPrefix already rejects these lengths on the
send side; apply the same bound in the decoder, and also reject Max Length
below Prefix Length per RFC 8210 5.1.

### DIFF
--- a/pkg/packet/rtr/rtr.go
+++ b/pkg/packet/rtr/rtr.go
@@ -229,11 +229,17 @@ func (m *RTRIPPrefix) DecodeFromBytes(data []byte) error {
 	m.PrefixLen = data[9]
 	m.MaxLen = data[10]
 	if m.Type == RTR_IPV4_PREFIX {
+		if m.PrefixLen > 32 || m.MaxLen > 32 {
+			return errors.New("prefix or max length out of range for IPv4 RTRIPPrefix")
+		}
 		m.Prefix, _ = netip.AddrFromSlice(data[12:16])
 		m.AS = binary.BigEndian.Uint32(data[16:20])
 	} else {
 		if len(data) < RTR_IPV6_PREFIX_LEN {
 			return errors.New("data too short for RTRIPPrefix")
+		}
+		if m.PrefixLen > 128 || m.MaxLen > 128 {
+			return errors.New("prefix or max length out of range for IPv6 RTRIPPrefix")
 		}
 		m.Prefix, _ = netip.AddrFromSlice(data[12:28])
 		m.AS = binary.BigEndian.Uint32(data[28:32])
--- a/pkg/packet/rtr/rtr.go
+++ b/pkg/packet/rtr/rtr.go
@@ -229,7 +229,7 @@ func (m *RTRIPPrefix) DecodeFromBytes(data []byte) error {
 	m.PrefixLen = data[9]
 	m.MaxLen = data[10]
 	if m.Type == RTR_IPV4_PREFIX {
-		if m.PrefixLen > 32 || m.MaxLen > 32 {
+		if m.MaxLen > 32 || m.PrefixLen > m.MaxLen {
 			return errors.New("prefix or max length out of range for IPv4 RTRIPPrefix")
 		}
 		m.Prefix, _ = netip.AddrFromSlice(data[12:16])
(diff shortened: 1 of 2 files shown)


pkg/packet/rtr/rtr.go
```diff
@@ -229,12 +229,18 @@ func (m *RTRIPPrefix) DecodeFromBytes(data []byte) error {
 	m.PrefixLen = data[9]
 	m.MaxLen = data[10]
 	if m.Type == RTR_IPV4_PREFIX {
+		if m.PrefixLen > 32 || m.MaxLen > 32 {
+			return errors.New("prefix or max length out of range for IPv4 RTRIPPrefix")
```
fujita: The RFC says
```
 Max Length:  An 8-bit unsigned integer denoting the longest prefix
 allowed by the Prefix element.  This MUST NOT be less than the
 Prefix Length element.
```

So please do here:

`if m.MaxLen > 32 || m.PrefixLen > m.MaxLen {`

pkg/packet/rtr/rtr.go
```diff
@@ -229,12 +229,18 @@ func (m *RTRIPPrefix) DecodeFromBytes(data []byte) error {
...
 		m.Prefix, _ = netip.AddrFromSlice(data[12:16])
 		m.AS = binary.BigEndian.Uint32(data[16:20])
 	} else {
 		if len(data) < RTR_IPV6_PREFIX_LEN {
 			return errors.New("data too short for RTRIPPrefix")
 		}
+		if m.PrefixLen > 128 || m.MaxLen > 128 {
+			return errors.New("prefix or max length out of range for IPv6 RTRIPPrefix")
+		}
 		m.Prefix, _ = netip.AddrFromSlice(data[12:28])
```
fujita, line 244: Ditto.

pkg/packet/rtr/rtr_test.go
```diff
@@ -137,6 +137,40 @@ func Test_ParseRTR_ErrorReportRejectsOversizedTextLen(t *testing.T) {
 	require.Error(t, err)
 }
 
+func Test_ParseRTR_IPPrefixRejectsOutOfRangeLength(t *testing.T) {
+	// RFC8210 5.7/5.8 bound Prefix Length and Max Length to the address
```
fujita: 5.1?

pkg/packet/rtr/rtr_test.go
```diff
@@ -137,6 +137,40 @@ func Test_ParseRTR_ErrorReportRejectsOversizedTextLen(t *testing.T) {
...
+	_, err = ParseRTR(ipv4PDU(24, 33))
+	require.Error(t, err)
+	_, err = ParseRTR(ipv6PDU(129, 128))
+	require.Error(t, err)
+
+	// A prefix within range still decodes.
+	_, err = ParseRTR(ipv4PDU(24, 32))
+	require.NoError(t, err)
+}
+
```
fujita, line 180: Please add a test for `m.PrefixLen > m.MaxLen`.

pkg/packet/rtr/rtr.go
Mounika2456: Makes sense, done. The ordering check also subsumes the old PrefixLen bound since PrefixLen <= MaxLen <= 32.

pkg/packet/rtr/rtr.go
Mounika2456: Done for the IPv6 branch as well.

pkg/packet/rtr/rtr_test.go
Mounika2456: Right, the field definitions are in 5.1. Fixed the comment (and the PR description, which had the same wrong reference).

pkg/packet/rtr/rtr_test.go
Mounika2456: Added ipv4(24,16) and ipv6(64,48) cases; both are rejected now.

fujita:
Merged, thanks.

PR #3513, opened 2026-07-28
bfd: discard received packets with an unbound discriminator

Mounika2456:
go test with the session in Down, fed a control packet whose sender never
observed our My Discriminator:

    state after blind Init packet: BFD_SESSION_STATE_UP

rxPacket compares Your Discriminator only when it is nonzero, so State Init
with a zero Your Discriminator walks the session Down -> Up with nothing tying
the packet to it. RFC 5880 6.8.6 accepts a zero Your Discriminator only from a
remote in Down or AdminDown, and separately requires discarding a zero My
Discriminator, which rxPacket currently stores as p.yourDiscriminator, the same
value setStateDown uses to mean "no session", and then echoes back on the wire.

With no BFD authentication and no receive-side TTL check, that echo is the only
evidence a packet came from the peer, so both checks are applied in rxPacket
next to the existing Detect Mult one.

### DIFF
--- a/pkg/server/bfd_peer.go
+++ b/pkg/server/bfd_peer.go
@@ -285,16 +285,42 @@ func (p *bfdPeer) startClient() {
 }
 
 func (p *bfdPeer) rxPacket(h *bfd.BFDHeader) {
-	if h.YourDiscriminator != 0 && h.YourDiscriminator != p.myDiscriminator {
-		p.stats.invalidDiscriminator.Add(1)
-		return
-	}
-
 	// RFC 5880 Section 6.8.6: if the Detect Mult field is zero, the packet
 	// MUST be discarded.
 	if h.DetectTimeMultiplier == 0 {
 		p.stats.invalidMultiplier.Add(1)
 		return
+	}
+
+	// RFC 5880 Section 6.8.6: if the My Discriminator field is zero, the
+	// packet MUST be discarded.
+	if h.MyDiscriminator == 0 {
+		p.stats.invalidDiscriminator.Add(1)
+		return
+	}
+
+	// RFC 5880 Section 6.8.6: a nonzero Your Discriminator selects the
+	// session and MUST match ours. A zero Your Discriminator carries no
+	// session binding, so it is only accepted from a remote system that has
+	// not learned our discriminator yet, i.e. one in Down or AdminDown.
+	if h.YourDiscriminator != 0 {
+		if h.YourDiscriminator != p.myDiscriminator {
+			p.stats.invalidDiscriminator.Add(1)
+			return
+		}
+	} else {
+		if h.State != bfd.StateDown && h.State != bfd.StateAdminDown {
+			p.stats.invalidDiscriminator.Add(1)
+			return
+		}
+
+		// Once the remote discriminator is bound, a packet that omits Your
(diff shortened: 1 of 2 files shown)


pkg/server/bfd_peer.go
```diff
@@ -285,7 +285,23 @@ func (p *bfdPeer) startClient() {
 }
 
 func (p *bfdPeer) rxPacket(h *bfd.BFDHeader) {
-	if h.YourDiscriminator != 0 && h.YourDiscriminator != p.myDiscriminator {
+	// RFC 5880 Section 6.8.6: if the My Discriminator field is zero, the
```
fujita, line 295: The RFC says that `When a BFD Control packet is received, the following procedure MUST be followed, in the order specified.`

So those changes should be after `h.DetectTimeMultiplier` checking?

pkg/server/bfd_peer.go
```diff
@@ -285,7 +285,23 @@ func (p *bfdPeer) startClient() {
...
+	// session and MUST match ours. A zero Your Discriminator carries no
+	// session binding, so it is only accepted from a remote system that has
+	// not learned our discriminator yet, i.e. one in Down or AdminDown.
+	if h.YourDiscriminator != 0 {
+		if h.YourDiscriminator != p.myDiscriminator {
+			p.stats.invalidDiscriminator.Add(1)
+			return
+		}
+	} else if h.State != bfd.StateDown && h.State != bfd.StateAdminDown {
 		p.stats.invalidDiscriminator.Add(1)
```
fujita: We need to do more if h.YourDiscriminator is zero so how about the following?

```
} else {
      if h.State != bfd.StateDown && h.State != bfd.StateAdminDown {
           p.stats.invalidDiscriminator.Add(1)
      }

      if p.yourDiscriminator != 0 && h.MyDiscriminator != p.yourDiscriminator {
			p.stats.invalidDiscriminator.Add(1)
			return
		}
}
```

fujita (commented):
As I commented, we need one more checking if `h.YourDiscriminator` is zero. If so, please add an unittest for it too.

pkg/server/bfd_peer.go
Mounika2456: Right, 6.8.6 is an ordered procedure and Detect Mult comes before the My Discriminator step. Moved both checks below it.

pkg/server/bfd_peer.go
Mounika2456: Added, with a return in the state branch as well, otherwise the packet still falls through to the state machine.

Test is Test_RxPacketZeroYourDiscriminatorForeignRemoteDiscarded. With the session Up and yourDiscriminator bound to 12345, a Down packet carrying MyDiscriminator 67890 and a zero Your Discriminator used to reach remoteDown() and reset the BGP peer; now it's discarded and the counter goes up. The second half checks the bound remote can still signal Down without echoing our discriminator.

Mounika2456:
Both addressed and pushed: the discriminator checks now run after the Detect Mult one, and the zero Your Discriminator path also rejects a My Discriminator that doesn't match the already-bound one, with a test covering it.

fujita:
Pushed, thanks.

PR #3515, opened 2026-07-28
server: withdraw filtered secondary routes on soft reset out

Akkanbe:
When all candidate paths for a destination become filtered by an export policy, soft reset out must withdraw the path that was previously advertised by the secondary-route selection.

PR #3434 added advertised-path tracking so that soft reset out can withdraw paths that become filtered. However, when `secondary-route` is enabled, `getBestFromLocalCallbackLocked` returns only the currently advertisable path and does not return a filtered candidate when all paths for a destination are rejected.

As a result, the advertised-path check added by #3434 is not performed for that destination, and the neighbor can continue using the stale secondary route.

Return a filtered candidate for secondary-route destinations that no longer have an advertisable path. Soft reset out can then check whether the destination was previously advertised and send an explicit withdrawal when necessary.

Add regression coverage for a previously advertised secondary route that becomes filtered after an export policy change.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1081,15 +1081,14 @@ func (s *BgpServer) getBestFromLocalCallbackLocked(peer *peer, rfList []bgp.Fami
 			dsts := tbl.GetDestinations()
 			dl := make([]*table.Update, 0, len(dsts))
 			for _, d := range dsts {
-				l := d.GetAllKnownPathList()
-				pl := make([]*table.Path, len(l))
-				copy(pl, l)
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks.

Are you using `secondary-route` in production, and is ADD-PATH not an option for you?

Akkanbe:
Thanks for pushing the changes.

We are using `secondary-route` in a proof of concept for a route server with some specific requirements. 
Since we need to advertise only a single path in the PoC, we were unable to use ADD-PATH...

PR #3522, opened 2026-07-29
config: add TCP-AO keychain model and load keychains from config

rastislavs:
This is a follow-up to https://github.com/osrg/gobgp/pull/3503

Extends the GoBGP YANG model with TCP-AO keychain configuration based on openconfig-keychain 0.5.0 model adapted to the fields and types supported by GoBGP. This is then used to load TCP-AO keychain configuration from file into the internal keychain store.

The configured keychains do not have any use for now, this will be added in follow-up changes as part of https://github.com/osrg/gobgp/issues/3493

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,7 +1,9 @@
 package config
 
 import (
+	"bytes"
 	"context"
+	"fmt"
 	"log/slog"
 
 	"github.com/osrg/gobgp/v4/api"
@@ -379,6 +381,15 @@ func InitialConfig(ctx context.Context, bgpServer *server.BgpServer, newConfig *
 
 	assignGlobalpolicy(ctx, bgpServer, &newConfig.Global.ApplyPolicy.Config)
 
+	// keychains and new keys must exist before peers can reference them
+	keychains, err := oc.NewTcpAoKeychainsFromConfigStruct(newConfig.Keychains)
+	if err != nil {
+		return nil, err
+	}
+	if err := addTcpAoKeychains(ctx, bgpServer, keychains); err != nil {
+		return nil, err
+	}
+
 	added := newConfig.Neighbors
 	addedPg := newConfig.PeerGroups
 	if isGracefulRestart {
@@ -404,6 +415,10 @@ func InitialConfig(ctx context.Context, bgpServer *server.BgpServer, newConfig *
 func UpdateConfig(ctx context.Context, bgpServer *server.BgpServer, c, newConfig *oc.BgpConfigSet) (*oc.BgpConfigSet, error) {
 	addedPg, deletedPg, updatedPg := oc.UpdatePeerGroupConfig(bgpServer.Log(), c, newConfig)
 	added, deleted, updated := oc.UpdateNeighborConfig(bgpServer.Log(), c, newConfig)
+	addedKeychains, deletedKeychains, addKeyUpdates, deleteKeyUpdates, err := tcpAoKeychainConfigChanges(c.Keychains, newConfig.Keychains)
+	if err != nil {
+		return c, err
+	}
 	updatePolicy := oc.CheckPolicyDifference(bgpServer.Log(), oc.ConfigSetToRoutingPolicy(c), oc.ConfigSetToRoutingPolicy(newConfig))
 
 	if updatePolicy {
@@ -427,8 +442,15 @@ func UpdateConfig(ctx context.Context, bgpServer *server.BgpServer, c, newConfig
 		updatePolicy = true
 	}
 
+	// keychains and new keys must exist before peers can reference them
+	if err := addTcpAoKeychains(ctx, bgpServer, addedKeychains); err != nil {
+		return c, err
+	}
+	if err := updateTcpAoKeychains(ctx, bgpServer, addKeyUpdates); err != nil {
+		return c, err
+	}
+
 	addPeerGroups(ctx, bgpServer, addedPg)
-	deletePeerGroups(ctx, bgpServer, deletedPg)
 	needsSoftResetIn := updatePeerGroups(ctx, bgpServer, updatedPg)
 	updatePolicy = updatePolicy || needsSoftResetIn
 	addDynamicNeighbors(ctx, bgpServer, newConfig.DynamicNeighbors)
@@ -436,6 +458,15 @@ func UpdateConfig(ctx context.Context, bgpServer *server.BgpServer, c, newConfig
 	deleteNeighbors(ctx, bgpServer, deleted)
 	needsSoftResetIn = updateNeighbors(ctx, bgpServer, updated)
 	updatePolicy = updatePolicy || needsSoftResetIn
+
+	// peer groups, keys, and keychains can only be removed after peers stop referencing them
+	deletePeerGroups(ctx, bgpServer, deletedPg)
+	if err := updateTcpAoKeychains(ctx, bgpServer, deleteKeyUpdates); err != nil {
+		return c, err
+	}
+	if err := deleteTcpAoKeychains(ctx, bgpServer, deletedKeychains); err != nil {
+		return c, err
+	}
 
 	if updatePolicy {
 		if err := bgpServer.ResetPeer(ctx, &api.ResetPeerRequest{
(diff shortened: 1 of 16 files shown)


rastislavs:
@fujita this one is now rebased and ready for review

pkg/config/config.go
```diff
@@ -449,3 +480,127 @@ func UpdateConfig(ctx context.Context, bgpServer *server.BgpServer, c, newConfig
...
+			added = append(added, desiredChain)
+			continue
+		}
+		addKeys, deleteKeys := tcpAoKeyChanges(currentChain.Keys, desiredChain.Keys)
+		if len(addKeys) == 0 && len(deleteKeys) == 0 {
+			continue
+		}
+		for _, addKey := range addKeys {
+			for _, currentKey := range currentChain.Keys {
+				if addKey.SendId == currentKey.SendId || addKey.ReceiveId == currentKey.ReceiveId {
```
fujita: The server side supports atomic key replacement, right?

pkg/config/config.go
```diff
@@ -427,8 +442,15 @@ func UpdateConfig(ctx context.Context, bgpServer *server.BgpServer, c, newConfig
 		updatePolicy = true
 	}
 
+	// keychains and new keys must exist before peers can reference them
+	if err := addTcpAoKeychains(ctx, bgpServer, addedKeychains); err != nil {
+		return c, err
+	}
```
fujita: This behavior isn't consistent with how UpdateConfig works. UpdateConfig keeps to continue with ignoring errors.

pkg/config/oc/util.go
```diff
@@ -689,6 +690,89 @@ func NewPeerFromConfigStruct(pconf *Neighbor) *api.Peer {
...
+	}
+	return result, nil
+}
+
+func tcpAoKeyIDToAPI(keyID string) (uint32, error) {
+	if keyID == "" {
+		return 0, fmt.Errorf("key-id is required")
+	}
+	value, err := strconv.ParseUint(keyID, 0, 8)
+	if err != nil {
```
fujita: Only base 10 and 16 should be allowed here?

tools/pyang_plugins/gobgp.yang
```diff
@@ -1692,4 +1692,179 @@ module gobgp {
     description "BFD configuration for neighbor";
     uses gobgp-bfd-config-set;
   }
+
+  // The keychain definitions below follow the structure of
+  // openconfig-keychain 0.5.0, limited to fields supported by GoBGP.
+
+  identity CRYPTO_TYPE {
```
fujita: kebab-case?

tools/pyang_plugins/gobgp.yang
```diff
@@ -1692,4 +1692,179 @@ module gobgp {
...
+    description
+      "Base identify for the cryptographic algorithm";
+  }
+
+  identity HMAC_SHA_1_96 {
+    base CRYPTO_TYPE;
```
fujita: ditto.

tools/pyang_plugins/gobgp.yang
```diff
@@ -1692,4 +1692,179 @@ module gobgp {
...
+    description
+      "HMAC-SHA-1-96 keyed hash algorithm constructed from SHA-1 hash
+      function and used as a HMAC, operating on 64-byte blocks of data.";
+    reference
+      "RFC 2404 - The Use of HMAC-SHA-1-96 within ESP and AH";
+  }
+
+  identity AES_128_CMAC_96 {
```
fujita: ditto.

tools/pyang_plugins/gobgp.yang
```diff
@@ -1692,4 +1692,179 @@ module gobgp {
...
+        "Identifier for the key within the keychain.  Note that the
+        hex-string type is deprecated and will be removed from a future
+        version of this model. Implementations should transition to using
+        the hex-string-prefixed type.";
+    }
+
+    leaf secret-key {
+      type string;
+      description
+        "Authentication key supplied as an encrypted value.  The system should store and
```
fujita: A base64 encoded key is expected?

pkg/config/config.go
```diff
@@ -437,6 +459,15 @@ func UpdateConfig(ctx context.Context, bgpServer *server.BgpServer, c, newConfig
...
+	// peer groups, keys, and keychains can only be removed after peers stop referencing them
+	deletePeerGroups(ctx, bgpServer, deletedPg)
+	if err := updateTcpAoKeychains(ctx, bgpServer, deleteKeyUpdates); err != nil {
+		return c, err
+	}
+	if err := deleteTcpAoKeychains(ctx, bgpServer, deletedKeychains); err != nil {
+		return c, err
+	}
+
 	if updatePolicy {
```
fujita, line 460: This is a fix? If so, please create a different pull request for this.

pkg/config/config.go
```diff
@@ -449,3 +480,127 @@ func UpdateConfig(ctx context.Context, bgpServer *server.BgpServer, c, newConfig
...
+func findAPITcpAoKeychain(chains []*api.TcpAoKeychain, name string) *api.TcpAoKeychain {
+	for _, chain := range chains {
+		if chain.Name == name {
+			return chain
+		}
+	}
+	return nil
+}
+
+func tcpAoKeyChanges(current, desired []*api.TcpAoKey) (add, delete []*api.TcpAoKey) {
```
fujita: deleted instead of delete.

pkg/config/oc/util.go
```diff
@@ -689,6 +690,89 @@ func NewPeerFromConfigStruct(pconf *Neighbor) *api.Peer {
...
+	case CRYPTO_TYPE_HMAC_SHA_1_96:
+		return api.TcpAoAlgorithm_TCP_AO_ALGORITHM_HMAC_SHA1_96, nil
+	case CRYPTO_TYPE_AES_128_CMAC_96:
+		return api.TcpAoAlgorithm_TCP_AO_ALGORITHM_AES_128_CMAC_96, nil
+	default:
+		return api.TcpAoAlgorithm_TCP_AO_ALGORITHM_UNSPECIFIED, fmt.Errorf("unsupported algorithm %q", algorithm)
+	}
+}
+
+func readTcpAoMasterKey(config *KeyConfig) ([]byte, error) {
```
fujita: You can simply use `string` type for the argument here?

pkg/config/config.go
rastislavs: Right, sorry, check forgotten from an earlier version - removed.

pkg/config/config.go
rastislavs: fixed - aligned with existing UpdateConfig logic

pkg/config/oc/util.go
rastislavs: fixed

tools/pyang_plugins/gobgp.yang
rastislavs: I have converted those to kebab-case to align with other GoBGP config, even though it deviates from the openconfig-keychain 0.5.0

tools/pyang_plugins/gobgp.yang
rastislavs: Updated the description to match our implementation (originally was verbatim copy from openconfig-keychain 0.5.0)

pkg/config/config.go
rastislavs: Yeah, will open a PR for this separately after this one is merged.

tools/pyang_plugins/gobgp.yang
```diff
@@ -1695,4 +1692,178 @@ module gobgp {
...
+    description "This grouping defines key-chain key parameters.";
+
+    leaf key-id {
+      type union {
+        type hex-string-prefixed {
+          length "3..66";
+        }
+        type uint64;
+      }
+      description
```
fujita, line 1748: I should have mentioned before. Do we need to use `hex-string-prefixed`? I know open config uses the type but we already use different types for some. Can't we simply `uint8`? It's consistent with receive-id.

Then the description would be simpler too, like "TCP-AO send key identifier associated with this key".

fujita:
It might be better to update the commit message; ```on openconfig-keychain 0.5.0 model copied over verbatim,
limited to fields supported by GoBGP.``` isn't true; we change some verbatim.

tools/pyang_plugins/gobgp.yang
rastislavs: yeah makes sense since we are already deviating from the OC model, updated

rastislavs:
> It might be better to update the commit message; 

updated the commit message

fujita:
Thanks!

PR #3523, opened 2026-07-29
table: don't infer a fixed ASN from an alternation community pattern

Mounika2456:
--- FAIL: TestCommunityMatcherAgreesWithRegexp
        pattern="^65100:666$|^65200:666$"  comm=65200:666   matcher=false regexp=true
        pattern="^65200:666$|^65100:\d+$"  comm=65200:9999  matcher=true  regexp=false

`extractLiteralASN` reads only the leading `^<digits>:` of the pattern, so one
with a top-level alternation still gets promoted to a fixed-AS matcher, which
tests `uint16(c>>16) == m.asn` and never looks at the other branches. A peer
that tags its route with the second alternative walks past a community-set
condition; the wildcard-local branch errs the other way and matches
communities the regexp rejects. `compileExtCommunityMatcher` shares the
helper, so `rt:` patterns behave the same way.

Declining the promotion when `regexp/syntax` parses the pattern to a top-level
`OpAlternate` leaves every other shape on the fast path it has today.

`FuzzCommunityCondition` and `FuzzExtCommunityCondition` compare the index
against the compiled matcher loop, so a matcher built on the wrong assumption
agrees with itself and stays invisible there. The two tests added here compare
each matcher against its own regexp instead.

### DIFF
--- a/internal/pkg/table/community_match_fuzz_test.go
+++ b/internal/pkg/table/community_match_fuzz_test.go
@@ -1,6 +1,7 @@
 package table
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/osrg/gobgp/v4/pkg/config/oc"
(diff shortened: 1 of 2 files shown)


fujita:
Thanks!

PR #3524, opened 2026-08-01
packet/bgp: bound the update attribute scan to its declared length

Mounika2456:
bound getBGPUpdateAttributes to TotalPathAttributeLen in BGPUpdate.DecodeFromBytes, otherwise the scan runs past the attributes into the NLRI field and a prefix in 40.0.0.0/8 registers a PREFIX_SID the peer never sent, which stops the label stack decoder after one label and turns a labeled 10.0.0.0/8 into a fabricated 0.2.17.10/32 that still passes the framing check.


fujita:
Thanks!

PR #3527, opened 2026-08-03
bfd: use transport local address for BFD

frezbo:
Bind outgoing BFD sessions to the neighbor transport local address so numbered peers use the same explicit source as BGP while preserving interface binding and kernel-selected defaults.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -372,7 +372,7 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
 		// register BFD for the dynamic neighbor too (explicit neighbors do this in addNeighbor): the
 		// BFD config is inherited from the peer group. Without this, BFD never runs for dynamic peers.
 		if s.bfdServer != nil && conf.Bfd.Config.Enabled {
-			if err := s.bfdServer.AddPeer(context.Background(), addr, conf.Bfd.Config, s.bgpConfig.Global.Config.BindToDevice); err != nil {
+			if err := s.bfdServer.addPeer(context.Background(), addr, conf.Bfd.Config, conf.Transport.Config.LocalAddress, s.bgpConfig.Global.Config.BindToDevice); err != nil {
 				s.logger.Warn("failed to add BFD peer for dynamic neighbor",
 					slog.String("Topic", "Peer"),
 					slog.String("Key", addr.String()),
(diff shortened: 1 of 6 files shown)


frezbo:
anything else needed to get this going forward?

pkg/server/server.go
```diff
@@ -3900,6 +3901,11 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
 		conf.Bfd.Config = c.Bfd.Config
 	}
 
+	if original.Transport.Config.LocalAddress != c.Transport.Config.LocalAddress {
+		peer.fsm.logger.Info("Update BFD local address")
+		bfdConfigChanged = true
```
fujita, line 3906: `NeedsResendOpenMessage()` below covers LocalAddress, so this returns through `deleteNeighbor()`/`addNeighbor()` before `bfdConfigChanged` is read. The BFD peer is recreated there, so this block can be dropped.

pkg/server/server.go
```diff
@@ -3952,6 +3958,7 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
 			err = s.updateBfdPeer(
 				addr,
 				original.Bfd.Config, c.Bfd.Config,
+				original.Transport.Config.LocalAddress, c.Transport.Config.LocalAddress,
```
fujita, line 3961: `original.Transport.Config.LocalAddress` is equal to `c.Transport.Config.LocalAddress` here.

pkg/server/server.go
```diff
@@ -372,7 +372,7 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
 		// register BFD for the dynamic neighbor too (explicit neighbors do this in addNeighbor): the
 		// BFD config is inherited from the peer group. Without this, BFD never runs for dynamic peers.
 		if s.bfdServer != nil && conf.Bfd.Config.Enabled {
-			if err := s.bfdServer.AddPeer(context.Background(), addr, conf.Bfd.Config, s.bgpConfig.Global.Config.BindToDevice); err != nil {
+			if err := s.bfdServer.addPeer(context.Background(), addr, conf.Bfd.Config, conf.Transport.Config.LocalAddress, s.bgpConfig.Global.Config.BindToDevice); err != nil {
```
fujita, line 375: This is a dynamic peer so `conn.Local()` should be used here?

fujita (changes requested):

pkg/server/server.go
frezbo: done

pkg/server/server.go
frezbo: done

pkg/server/server.go
frezbo: done

fujita:
Merged, thanks.

PR #3530, opened 2026-08-04
test(scenario): harden pytest orchestration and Docker cleanup

Ivan-Pokhabov:
- Add Docker IPv6 preflight, labeled managed cleanup, diagnostics, and a standalone cleanup helper
- Run the scenario orchestrator sequentially while preserving quiet and fail-fast reporting for individual pytest invocations
- Add timeout scaling and shared wait/retry helpers for scenario assertions and reachability checks
- Make Docker bridge/container setup more deterministic with IPv6 sysctls, network create retries, and link-local ping probing
- Limit Go build parallelism for scenario images and update runner/image build docs


fujita:
Please split the pull request; one commit includes too many changes.

I would recommend you to work on wait-loop cleanup and local() improvements first; both look nice and affect the CI on GitHub.

fujita:
Thanks, and sorry for the slow reply.

#3533 is merged, so I would like to close this and ask you to send the rest as
separate PRs. 16 of the 22 files here are already in master. This PR also
conflicts, and it still has the old counting `wait_for` that #3533 replaced, so
a rebase would put the conflict right there. Sending the rest again is easier.

Two parts are worth merging. Please send one PR each.

1. `probe_link_local_address` in `test/lib/utils.py`. The neighbor cache entry
   ages to STALE while packets still flow, so checking for `REACHABLE` is not
   reliable. A scoped ping is the right test.

2. The labeled Docker cleanup in `test/lib/base.py` and the conftest hooks.
   Cleaning only `label=gobgp-test` is right. I checked the teardown hook: every
   scenario file is a `unittest.TestCase`, so it cannot delete a topology that
   is still in use.

The retry in `Bridge._create_network` looks reasonable, but did you hit this?
`docker network create` returning non-zero after it made the network is not
something I have seen. A CI log would help.

The README paragraph for `--timeout-scale` is wanted, since the option went in
with #3533 undocumented. Please write it against current master. The diff here
is based on the README from before the Sep 1 rewrite.

Please drop the rest:

- `run_all_tests.py` and the `run_all_tests.sh` deletion. The script was already
  deleted in master on Sep 1. `parse_index_count` also takes `fields[-1]` of the
  last line matching `"invalid"`, which is the traceback source line, so it
  raises `ValueError`. That is the bug #3575 fixed.

- `cleanup.py`. It repeats the cleanup helper above and removes a DinD volume
  this repository does not use.

- `GOMAXPROCS=1` in `make-gobgp-ctn.sh`. `go build -p 1` as the default slows
  the image build for everyone. Please set it in your environment instead.

- The four IPv6 sysctls on `docker run`. Three already hold that value in a new
  network namespace. Only `default.accept_dad=0` changes anything, and it just
  removes about one second of tentative state, which the retry loop already
  absorbs.

Thanks again.

PR #3531, opened 2026-08-04
Return initial apply errors when parse config 

Ivan-Pokhabov:
Make InitialConfig return errors from failed config application instead of only
logging them, so gobgpd rejects invalid startup config such as missing peer-groups
or broken policy references. Keep UpdateConfig reloads non-fatal and covered by
tests.

Alse send sd_notify READY only after the startup config has been read and applied.
This prevents systemd from seeing gobgpd as ready right before it exits on an
invalid initial config, while preserving the no-config-file startup path.

### DIFF
--- a/cmd/gobgpd/main.go
+++ b/cmd/gobgpd/main.go
@@ -237,17 +237,8 @@ func main() {
 	prometheus.MustRegister(fsmTimingCollector)
 	go bgpServer.Serve()
 
-	if opts.UseSdNotify {
-		if status, err := daemon.SdNotify(false, daemon.SdNotifyReady); !status {
-			if err != nil {
-				logger.Warn("Failed to send notification via sd_notify()", slog.String("Error", err.Error()))
-			} else {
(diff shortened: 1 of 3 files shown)


fujita:
What we want to do is that InitialConfig should stop on the first error and UpdateConfig should continue, right?

Then why not just adding an argument like `stopOnError bool` to functions used by both (e.g., `addPeerGroups`)?

`deletePeerGroups`/`updatePeerGroups`/`deleteNeighbors`/`updateNeighbors` are used only `UpdateConfig` so they can always continue with warning.

Then we don't need a new wrapper like `configErrorHandler` so we can use Go's typical error handling style; `if err := f(); err != nil {}`

Ivan-Pokhabov:
> Then we don't need a new wrapper like `configErrorHandler` so we can use Go's typical error handling style; `if err := f(); err != nil {}`

I made it for future, because we probable would like to stop on error also UpdateConfig. But now I think you are right, i will rework it

fujita:
Pushed, thanks.

PR #3532, opened 2026-08-05
table: fix preserve flag in DeletePolicy

mcdoker18:
The DeletePolicy function incorrectly iterated over the policy's own statements when deleting individual policies with preserve=false.

The fix ensures that when deleting a specific policy (all=false), we iterate over the statements from the policy argument.


Lucky112:
Amazing! I was bothered with this issue for some time

fujita:
Merged, thanks!

PR #3533, opened 2026-08-05
Improve wait loops and local command

Ivan-Pokhabov:
Part 1 of #3530

### DIFF
--- a/test/lib/base.py
+++ b/test/lib/base.py
@@ -18,12 +18,15 @@ import os
 import subprocess
 import time
 import itertools
+import math
 
 import textwrap
 from colored import fg, attr
 
 from docker import APIClient as Client
 import netaddr
+
+from lib.noseplugin import parser_option
 
 
 DEFAULT_TEST_PREFIX = ''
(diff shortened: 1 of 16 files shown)


test/lib/base.py
```diff
@@ -110,21 +153,28 @@ def community_str(i):
     return ':'.join(reversed(values))
 
 
-def wait_for_completion(f, timeout=120):
-    interval = 1
+def wait_for(f, timeout=DEFAULT_WAIT_TIMEOUT, interval=1, timeout_message='timeout'):
+    timeout = scale_timeout(timeout)
```
fujita, line 157: Please use a wall-clock deadline.

test/lib/base.py
```diff
@@ -84,10 +87,47 @@
...
+        shell=True,
+        env=_env,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        timeout=scale_timeout(timeout),
+    )
+    stdout = result.stdout.decode('utf-8')
+    stderr = result.stderr.decode('utf-8')
+    if stderr:
+        print(stderr, end='')
```
fujita, line 120: The commit message says "stderr is preserved in test logs". But the code prints the child's stderr to stdout, because print() writes to sys.stdout by default. Is that intended?

test/scenario_test/route_server_test.py
```diff
@@ -217,21 +214,23 @@ def test_10_add_distant_relative(self):
...
-            return is_done
+                            return True
+                return False
+
+            wait_for(
+                _has_expected_nexthop,
+                timeout=self.retry_limit * self.wait_per_retry,
+                interval=self.wait_per_retry,
+            )
+            return True
```
fujita, line 233: Is this `return True` needed? wait_for raises on timeout, so check_nexthop can only return True. That makes `self.assertTrue(done)` below always pass.

test/lib/base.py
Ivan-Pokhabov: Done

test/scenario_test/route_server_test.py
Ivan-Pokhabov: It doesnt need

test/lib/base.py
Ivan-Pokhabov: Switch to stderr

fujita:
Merged, thanks.

PR #3534, opened 2026-08-09
packet/bgp: bound MUP Type 2 endpoint address length to address family

Mounika2456:
MUP Type 2 ST route, AFI=IPv4, with the Endpoint Address Length octet set below 32 in a peer-supplied RF_MUP UPDATE:

    [type:t2st][rd:100:100][endpoint-address-length:16][endpoint:10.10.10.1][teid:0.0.0.0]

MUPType2SessionTransformedRoute.DecodeFromBytes bounds Endpoint Address Length only from above (>64 for IPv4, >160 for IPv6). That field is the endpoint address plus the architecture-specific endpoint identifier, so it must be at least the endpoint address size. The endpoint is always consumed at its fixed 4/16 bytes and teidLen is EndpointAddressLength minus that size, so a value below it goes negative: the TEID is recorded as 0.0.0.0 and the octets the peer placed after the endpoint (its intended TEID) are handed to parseMUPTLVs. EndpointAddressLength and TEID are both part of the Type 2 route key, so the same endpoint decodes under several distinct keys, and other trailing layouts fail parseMUPTLVs and drop the whole UPDATE.

The sibling Type 1 decoder already restricts its length fields; this adds the matching lower bound (32 for IPv4, 128 for IPv6).

### DIFF
--- a/pkg/packet/bgp/mup.go
+++ b/pkg/packet/bgp/mup.go
@@ -810,7 +810,14 @@ func (r *MUPType2SessionTransformedRoute) DecodeFromBytes(data []byte, afi uint1
 		return malformedAttrListErr("invalid 3GPP 5G specific Type 2 Session Transformed Route length")
 	}
 	r.EndpointAddressLength = data[p]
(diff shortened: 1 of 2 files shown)


fujita:
Thanks!

Issue #2810, opened 2024-05-29
flowspec redirect next-hop ipv4 & ipv6


easternboyz:
Regarding flowspec redirect ip, is there a roadmap to support this feature

Reference: https://github.com/osrg/gobgp/issues/2310

channelsss:
I'm interested in this feature.
WANguard is starting to develop GoBGP support.
It looks excellent to replace ExaBGP.

channelsss:
Hey
Any news ?

ubinoob1:
Hi. Are you willing to implement this yet-draft https://datatracker.ietf.org/doc/draft-ietf-idr-flowspec-redirect-ip/ proposal? Many vendors already implemented it, as-well-as bgp software.

fujita:
If someone sends a pull request, I'm happy to merge it.

PR #3535, opened 2026-08-09
flowspec: add the redirect-to-IP action (draft-ietf-idr-flowspec-redirect-ip-16)

takehaya:
## Why
Implements the "Flow-spec Redirect to IP" action from [draft-ietf-idr-flowspec-redirect-ip-16]( https://datatracker.ietf.org/doc/html/draft-ietf-idr-flowspec-redirect-ip-16), requested in #2810, in both forms the draft defines: IPv4 (type 0x01, sub-type 0x0c) and IPv6 (type 0x000c, RFC 5701). 
The code points are IANA-assigned and the draft is in the RFC Editor queue. 

### Why not the existing `Redirect` types

Those implement redirect-to-VRF (RFC 8955 section 7.4), where the value is a **route target** naming a VRF. Here the value is the **address traffic is forwarded toward**. GoBGP previously decoded such communities into the generic `IPv4AddressSpecificExtended`, losing the action, and could only originate one by hand-crafting bytes.

The CLI gets its own verb for the same reason, `redirect` parses itsargument as a route target

```bash
$ gobgp global rib -a ipv4-flowspec add \
    match destination 192.0.2.1/32 destination-port '==61001' \
    then redirect-to-ip 198.51.100.11 [copy]
```

Trailing `copy` sets the C bit. `docs/sources/flowspec.md` documents the action with transcripts from a running `gobgpd`.

### DIFF
--- a/api/attribute.pb.go
+++ b/api/attribute.pb.go
@@ -5620,6 +5620,7 @@ type IP6ExtendedCommunitiesAttribute_Community struct {
 	//
 	//	*IP6ExtendedCommunitiesAttribute_Community_Ipv6AddressSpecific
 	//	*IP6ExtendedCommunitiesAttribute_Community_RedirectIpv6AddressSpecific
+	//	*IP6ExtendedCommunitiesAttribute_Community_FlowSpecRedirectToIpv6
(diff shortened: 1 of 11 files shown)


fujita:
Great, thanks!

PR #3536, opened 2026-08-10
table: skip rejected paths when building Drop withdrawals

### LINKED ISSUE #3455: Drop() in adj-RIB-In creates withdrawals for rejected paths, causing spurious "No matching path for withdraw" warnings
## Summary

The `AdjRib.Drop()` function in `internal/pkg/table/adj.go` does not filter out rejected paths (paths marked with `IsRejected=true`) when generating withdrawal paths. This causes the downstream `processWithdraw()` to fail to find these paths in the actual RIB (since rejected paths were never installed there), resulting in repeated `"No matching path for withdraw found, may be path was not installed into table"` warnings.

In large-scale BGP deployments — particularly HA transit topologies with many spokes — this can flood logs with tens of thousands of warnings per peer-flap event and waste memory storing rejected paths that never need to be withdrawn.

## Environment

- **GoBGP version:** v3.37.0 (issue exists on master as well)
- **Affected file:** `internal/pkg/table/adj.go`
- **Affected function:** `AdjRib.Drop()`

## Reproduction Scenario

1. Deploy two BGP routers (Router A and Router B) sharing the **same AS number** (HA pair scenario).
2. Connect ~200 BGP peers (spokes) to both routers.
3. Each spoke advertises routes to Router A.
4. Router A advertises routes to all other spokes; spokes re-advertise back to Router B.
5. Router B receives the routes, sees its own AS in the AS-path → **rejects them** (loop prevention) → marks them as `IsRejected=true` in `adj-RIB-In`.
6. With 200 spokes, ~200×199 ≈ ~40,000 rejected ghost paths accumulate in Router B's `adj-RIB-In`.
7. Trigger any BGP session flap on Router B.
8. `Drop()` is called → generates withdrawal entries for ALL paths (including rejected ones).
9. `processWithdraw()` cannot find the rejected paths in the actual RIB (they were never installed).
10. **Result:** ~40,000+ `"No matching path for withdraw"` warnings flood the log.

## Root Cause

In [`internal/pkg/table/adj.go` lines 158–173](https://github.com/osrg/gobgp/blob/v3.37.0/internal/pkg/table/adj.go#L158-L173):

```go
func (adj *AdjRib) Drop(rfList []bgp.RouteFamily) []*Path {
    l := make([]*Path, 0, adj.Count(rfList))
    adj.walk(rfList, func(d *Destination) bool {
        for _, p := range d.knownPathList {
            w := p.Clone(true)      // ← no IsRejected() guard
            w.SetDropped(true)
            l = append(l, w)
        }
        return false
    })
    ...
}
```

This iterates over all paths in `knownPathList` without checking `IsRejected()`. Rejected paths are included in the withdrawal list, but they were never installed into the RIB, so `processWithdraw()` cannot find them.

## Inconsistency with Existing Code

Other functions in the same file already correctly skip rejected paths:

- **`StaleAll()`** (lines 191–206) — skips rejected paths from the returned list:
  ```go
  if !n.IsRejected() {
      pathList = append(pathList, n)
  }
  ```

- **`MarkLLGRStaleOrDrop()`** (lines 208–231) — handles rejected paths separately:
  ```go
  if p.IsRejected() {
      d.knownPathList[i] = n
  } else {
      pathList = append(pathList, n)
  }
  ```

- **`PathList()`** (lines 125–137) — has an `accepted` flag that skips rejected paths.

The graceful restart code path also correctly handles rejected paths. `Drop()` is the outlier.

## Proposed Fix

Add an `IsRejected()` guard in `Drop()` to skip rejected paths, matching the pattern already established in `StaleAll()`, `MarkLLGRStaleOrDrop()`, and `PathList()`:

```go
func (adj *AdjRib) Drop(rfList []bgp.RouteFamily) []*Path {
    l := make([]*Path, 0, adj.Count(rfList))
    adj.walk(rfList, func(d *Destination) bool {
        for _, p := range d.knownPathList {
            if p.IsRejected() {
                continue
            }
            w := p.Clone(true)
            w.SetDropped(true)
            l = append(l, w)
        }
        return false
    })
    ...
}
```

## Impact

**Without the fix:**
- Log spam: ~40,000+ `"No matching path for withdraw"` warnings per peer flap in 200-spoke HA setups
- Wasted memory: ~19 MB per event at 200 spokes; potentially up to ~1.67 GB at 2000-spoke scale
- Difficult to debug actual issues due to log noise

**With the fix:**
- No spurious warnings
- Reduced memory footprint
- Cleaner, more reliable adj-RIB-In behavior on peer flaps

## Test Case

A test case can be added that:
1. Adds two paths to the AdjRib — one regular, one with `SetRejected(true)`.
2. Calls `Drop()`.
3. Asserts the returned path list contains only the non-rejected path.

## Willing to Contribute

I'm happy to submit a PR with the fix and a test case if maintainers agree with the approach.

locker95:
When a peer flaps, `AdjRib.Drop()` was cloning every adj-RIB-In path into a withdrawal list, including ones marked `IsRejected` (e.g. AS-loop rejects that never made it into the local RIB). Downstream withdraw processing then logs "No matching path for withdraw" for each of those.

Skip rejected paths when building the withdrawal list. The adj table is still cleared for the families being dropped.

Fixes #3455

### DIFF
--- a/internal/pkg/table/adj_test.go
+++ b/internal/pkg/table/adj_test.go
@@ -138,7 +138,6 @@ func TestStale(t *testing.T) {
 	assert.Equal(t, 1, len(adj.table[family].GetDestinations()))
 }
 
-
 func TestDropSkipsRejected(t *testing.T) {
 	pi := &PeerInfo{}
 	attrs := []bgp.PathAttributeInterface{bgp.NewPathAttributeOrigin(0)}
--- a/internal/pkg/table/adj_test.go
+++ b/internal/pkg/table/adj_test.go
@@ -138,6 +138,33 @@ func TestStale(t *testing.T) {
 	assert.Equal(t, 1, len(adj.table[family].GetDestinations()))
 }
 
+
+func TestDropSkipsRejected(t *testing.T) {
+	pi := &PeerInfo{}
+	attrs := []bgp.PathAttributeInterface{bgp.NewPathAttributeOrigin(0)}
+
+	nlri1, _ := bgp.NewIPAddrPrefix(netip.MustParsePrefix("20.20.10.0/24"))
+	p1 := NewPath(bgp.RF_IPv4_UC, pi, bgp.PathNLRI{NLRI: nlri1}, false, attrs, time.Now(), false)
+	nlri2, _ := bgp.NewIPAddrPrefix(netip.MustParsePrefix("20.20.20.0/24"))
+	p2 := NewPath(bgp.RF_IPv4_UC, pi, bgp.PathNLRI{NLRI: nlri2}, false, attrs, time.Now(), false)
+	p2.SetRejected(true)
+
+	family := p1.GetFamily()
+	families := []bgp.Family{family}
+
+	adj := NewAdjRib(slog.Default(), families)
+	adj.Update([]*Path{p1, p2})
+	assert.Equal(t, 2, adj.Count(families))
+	assert.Equal(t, 1, adj.Accepted(families))
+
+	// Drop must not emit withdrawals for rejected paths that never entered the RIB.
+	dropped := adj.Drop(families)
+	assert.Equal(t, 1, len(dropped))
+	assert.False(t, dropped[0].IsRejected())
+	assert.Equal(t, 0, adj.Count(families))
+	assert.Equal(t, 0, adj.Accepted(families))
+}
+
 func TestLLGRStale(t *testing.T) {
 	pi := &PeerInfo{}
 	attrs := []bgp.PathAttributeInterface{bgp.NewPathAttributeOrigin(0)}
--- a/internal/pkg/table/adj.go
+++ b/internal/pkg/table/adj.go
@@ -206,6 +206,11 @@ func (adj *AdjRib) Drop(rfList []bgp.Family) []*Path {
 	l := make([]*Path, 0, adj.Count(rfList))
 	adj.walk(rfList, func(d *destination) bool {
 		for _, p := range d.knownPathList {
+			// Rejected paths were never installed in the local RIB; do not
+			// emit withdrawals for them (avoids "No matching path for withdraw").
+			if p.IsRejected() {
+				continue
+			}
 			w := p.Clone(true)
 			w.SetDropped(true)
 			l = append(l, w)


internal/pkg/table/adj_test.go
```diff
@@ -138,6 +138,32 @@ func TestStale(t *testing.T) {
...
+
+	adj := NewAdjRib(slog.Default(), families)
+	adj.Update([]*Path{p1, p2})
+	assert.Equal(t, 2, adj.Count(families))
+	assert.Equal(t, 1, adj.Accepted(families))
+
+	// Drop must not emit withdrawals for rejected paths that never entered the RIB.
+	dropped := adj.Drop(families)
+	assert.Equal(t, 1, len(dropped))
+	assert.False(t, dropped[0].IsRejected())
```
fujita, line 162: Replace this assertion with a check on the NLRI. It passes whether or not the fix is there. Path.Clone() does not copy the rejected field, so dropped[0].IsRejected() is false for any path it was cloned from. The NLRI says which of the two paths survived.

fujita (changes requested):
The change itself is correct. One thing to fix in the new test.

Two more places have the same bug, DropStale and the HasNoLLGR() branch of MarkLLGRStaleOrDrop, so #3455 is not fixed yet for a peer that uses graceful restart. They need a different change. Both remove the Adj-RIB-In entries by passing the list they return to adj.Update(), so a rejected path left out of that list would stay in the Adj-RIB-In. I will do those in a follow-up.

The commit message says that the withdrawals produce log noise. On master that line is Debug, not Warn; the issue was filed against v3.37. The cost today is the clones and the work through the import path, not the log volume. Please reword it.

The commit message also says Fixes #3455, and so does the pull request description. That closes the issue on merge, which is too early while the graceful restart case is open. Please write "Part of #3455" in both.

Please rebase on master and squash the gofmt commit into the first one.

internal/pkg/table/adj_test.go
locker95: yeah, clone drops the rejected bit so that check was useless. looking at the prefix now.

fujita:
Merged, thanks.

PR #3537, opened 2026-08-11
zebra: reject a nexthop label count above the maximum

Mounika2456:
zapi6 / frr7.2 NEXTHOP_UPDATE, two nexthops, nexthop[0] declares label_num=20:

    nexthop[0]: gate=192.168.0.1 ifindex=2 labelnum=16   (clamped from 20)
    nexthop[1]: gate=0.0.0.0     ifindex=0  labelnum=3   (wire had gate=10.0.0.9 ifindex=9)
    err=<nil>

Nexthop.decode clamps label_num to maxMplsLabel (16) instead of rejecting the nexthop the way frr's zapi_nexthop_decode does, then the label loop consumes only 16*4 bytes while the sender wrote label_num*4. NexthopUpdateBody.decodeFromBytes ignores the length decodeNexthops consumed, so the cursor is left (label_num-16)*4 bytes short and the following nexthop is framed from the leftover label octets, with no error. The label branch is reachable on frr6-7.2 via the body MessageLabel flag and on frr7.3+/8.x via the per-nexthop label flag bit.

Reject label_num past the maximum in the decoder, matching frr.

### DIFF
--- a/pkg/zebra/zapi.go
+++ b/pkg/zebra/zapi.go
@@ -2465,8 +2465,12 @@ func (n *Nexthop) decode(data []byte, version uint8, software Software, family u
 		}
 		n.LabelNum = data[offset] // frr: STREAM_GETC(s, api_nh->label_num);
 		offset++
+		// frr rejects label_num > MPLS_MAX_LABELS instead of decoding the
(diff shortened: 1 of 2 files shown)


fujita:
Thanks!

PR #3538, opened 2026-08-12
fix(server): apply SetPolicies policy assignments

Ivan-Pokhabov:
Now, SetPolicy ignore new Assignements

### DIFF
--- a/internal/pkg/table/policy.go
+++ b/internal/pkg/table/policy.go
@@ -4646,27 +4646,23 @@ func (r *RoutingPolicy) Initialize() error {
 	return nil
 }
 
-func (r *RoutingPolicy) setPeerPolicy(id string, c oc.ApplyPolicy) {
+func (r *RoutingPolicy) setPeerPolicy(id string, c oc.ApplyPolicy) error {
 	for _, dir := range []PolicyDirection{POLICY_DIRECTION_IMPORT, POLICY_DIRECTION_EXPORT} {
 		ps, def, err := r.getAssignmentFromConfig(dir, c)
 		if err != nil {
-			r.logger.Error("failed to get policy info",
-				slog.String("Topic", "Policy"),
-				slog.String("Dir", dir.String()),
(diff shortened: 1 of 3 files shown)


internal/pkg/table/policy.go
```diff
@@ -4745,7 +4681,9 @@ func (r *RoutingPolicy) Reset(rp *oc.RoutingPolicy, ap map[string]oc.ApplyPolicy
 	}
 
 	for id, c := range ap {
-		r.setPeerPolicy(id, c)
+		if err := r.setPeerPolicy(id, c); err != nil {
+			return err
+		}
```
fujita, line 4686: reload() has already cleared assignmentMap at this point, so this loop is what restores it. Returning here leaves every id the loop has not reached yet with no assignment, including the global RIB. ap is a map, so which ids survive changes on every call.

internal/pkg/table/policy.go
```diff
@@ -4706,27 +4646,23 @@ func (r *RoutingPolicy) Initialize() error {
 	return nil
 }
 
-func (r *RoutingPolicy) setPeerPolicy(id string, c oc.ApplyPolicy) {
+func (r *RoutingPolicy) setPeerPolicy(id string, c oc.ApplyPolicy) error {
 	for _, dir := range []PolicyDirection{POLICY_DIRECTION_IMPORT, POLICY_DIRECTION_EXPORT} {
```
fujita, line 4650: If this returns an error, I think that it's better to update IMPORT and EXPORT atomically.

Firstly, calling r.getAssignmentFromConfig for both directions, only if both succeeds, then execute setDefaultPolicy and setPolicy for both.

Ivan-Pokhabov: Fixed

internal/pkg/table/policy.go
Ivan-Pokhabov: Fixed

Issue #2705, opened 2023-09-13
A data race occurs when a watcher handles path event while the related peer is closing down


OliverZeng:
We observe that a data race occurs when a watcher is handling path events while bgp server is handling the associated peer which is down.

```
==================	
WARNING: DATA RACE	

Read at 0x00c0019f5e08 by goroutine 60:	
  github.com/osrg/gobgp/v3/pkg/server.toPathAPI()	
      /usr/local/gobgp/pkg/server/grpc_server.go:182 +0xa6c
  github.com/osrg/gobgp/v3/pkg/server.toPathApi()
      /usr/local/gobgp/pkg/server/grpc_server.go:214 +0x215	
  github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).WatchEvent.func2()	
      /usr/local/gobgp/pkg/server/server.go:4065 +0x141b	

Previous write at 0x00c0019f5e08 by goroutine 15:	
  github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).handleFSMMessage()	
      /usr/local/gobgp/pkg/server/server.go:1413 +0x3209	
  github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).Serve.func1()	
      /usr/local/gobgp/pkg/server/server.go:432 +0xaa9	
  github.com/osrg/gobgp/v3/pkg/server.(*BgpServer).Serve()	
      /usr/local/gobgp/pkg/server/server.go:472 +0x870	
  main.main.func2()	

......

==================	
```

Version: 3.18.0

PR #3540, opened 2026-08-17
server: hold fsm.lock when reading fsm.recvOpen

gregwight:
## Problem

`fsm.recvOpen` is written by the FSM goroutine under `fsm.lock` (`pkg/server/fsm.go:1018`, and four other sites), but two readers accessed it without that lock:

- `pkg/server/server.go:989` in `newWatchEventPeer` — the read sat two lines below `peer.fsm.lock.Unlock()`
- `pkg/server/server.go:820` in `toConfig` — the lock was not taken at all

`newWatchEventPeer` is reached for every peer in `neighborMap` when a watcher registers via `WatchEvent`, and on each state transition via `broadcastPeerState`. Registering a watcher while a peer's FSM is starting up therefore races the FSM goroutine.

CONTRIBUTING states that `peer.fsm.lock` protects individual peer state, so these reads violate the documented invariant.

## Root cause

The read in `newWatchEventPeer` was inside the critical section from #2714 until f416268a, which merged two lock sections while moving to `atomic.Value` and left the single `Unlock()` above the read.

## Fix

Move the read above the `Unlock()` in `newWatchEventPeer`, and take the lock around the read in `toConfig`.

`toConfig` uses its own short critical section rather than extending the block above it, because `fsm.lock` is a plain `sync.Mutex` and `capabilitiesFromConfig`, `peer.State()` and `peer.AdminState()` sit in between.

## Test

`TestRace_NewWatchEventPeerRecvOpen` follows the existing `TestRace_*` pattern and reuses `newTestBgpServerWithPeer`. It fails under `-race` without the fix and passes with it. The test only writes `recvOpen` under the lock; the racing read arrives through the real FSM goroutine:

```
WARNING: DATA RACE
Write at 0x00c0001057a0 by goroutine 28:
  server.TestRace_NewWatchEventPeerRecvOpen.func2()
      pkg/server/fsm_test.go:1690

Previous read at 0x00c0001057a0 by goroutine 27:
  server.newWatchEventPeer()
      pkg/server/server.go:989
  server.(*BgpServer).broadcastPeerState()
      pkg/server/server.go:1018
  server.(*BgpServer).handleFSMMessage()
      pkg/server/server.go:1926
  server.(*fsmHandler).loop()
      pkg/server/fsm.go:2180
```

Verified on `linux/amd64`: `go test -race -timeout 900s ./...` passes for all packages, and `golangci-lint run` reports 0 issues.

Possibly related, though a different field and code path: #2705.

### DIFF
--- a/pkg/server/fsm_test.go
+++ b/pkg/server/fsm_test.go
@@ -1640,3 +1640,74 @@ func TestSendMessageloop_KillSignal(t *testing.T) {
 		t.Fatal("sendMessageloop did not exit after non-path message")
 	}
 }
+
+// TestRace_NewWatchEventPeerRecvOpen tests that newWatchEventPeer is race-free
+// when called concurrently with an FSM connection transition, which writes
+// fsm.recvOpen under fsm.lock.
+//
+// newWatchEventPeer reads fsm.recvOpen after releasing fsm.lock, so the read is
+// unsynchronised with respect to those writes. WatchEvent reaches it for every
+// peer in neighborMap when a watcher registers, and broadcastPeerState reaches
+// it on each state transition.
+//
+// Run with: go test -race -count=1 ./pkg/server/... -run TestRace_NewWatchEventPeerRecvOpen
+func TestRace_NewWatchEventPeerRecvOpen(t *testing.T) {
+	s, _, peerAddrIP := newTestBgpServerWithPeer(t)
+	defer s.StopBgp(context.Background(), &api.StopBgpRequest{})
+
+	var testPeer *peer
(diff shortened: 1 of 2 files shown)


floatingstatic:
@fujita any thoughts on this?

fujita:
Good catch, thanks!

I think that there was still a race about `fsm.recvOpen`. The pointer to the Open Message are passed to BMP clients and they call Serialize method. The problem is Serialize method modifies the Open Message. I fixed Serialize method not to modify an Open Message.

PR #3542, opened 2026-08-19
packet/bgp: reject VPLS NLRI with a length other than 17

Mounika2456:
reject a VPLS NLRI whose length field is not 17, since VPLSNLRI.Len() reports a fixed 19 bytes to the MP_(UN)REACH framing loop and any other length mis-frames the following NLRIs, with the 12-byte BGP-AD form additionally decoding into a route with a nil RD that panics on re-serialization.


fujita:
Thanks!

Mounika2456:
thanks for picking this up, glad it landed

PR #3543, opened 2026-08-20
server: expose raw NOTIFICATION error code/subcode on PeerState

Sreedhar-cp:
## Summary

`PeerState.DisconnectReason`/`DisconnectMessage` (added for peer disconnect reason reporting) describe why a session went down as an enum plus a formatted string, but don't expose the underlying RFC 4271 §6 NOTIFICATION error code/subcode as typed integers. A caller that needs the raw numeric values today has to fragile-parse the message string.

This PR adds `PeerState.notification_code` / `notification_subcode`, threaded through from the FSM's underlying `*bgp.BGPNotification` whenever one is present. This is intentionally not limited to a specific `DisconnectReason` (e.g. `NOTIFICATION_SENT`/`_RECEIVED`) — `fsmBadPeerAS`, `fsmInvalidMsg`, `fsmHoldTimerExpired`, and `fsmAdminDown` can all carry a real notification too, so any of them should surface a numeric code/subcode when one is actually present. Both fields are `0`/`0` when no notification was involved.

## Commits

1. **`server: fix wrong message wired into fsmStateReason on bad-peer-AS/invalid-msg`** — while adding a test for the feature above, found that `handleOpen()` builds the actual outgoing NOTIFICATION in `notif` and correctly sends it on the wire, but passes the *received* OPEN message (`m`, bound by the outer type switch) into `newfsmStateReason(...)` instead of `notif`, for both the bad-peer-AS and generic invalid-message paths. Every other FSM path in this file (hold-timer-expired, admin-down) correctly passes the actual notification. Besides being simply wrong, this also means `bmp.go`'s BMP Peer Down Notification embeds the wrong message content for a peer torn down via a bad-peer-AS or invalid-message OPEN rejection. Fixed to match the convention used everywhere else in the file.
2. **`api, server: expose raw NOTIFICATION error code/subcode on PeerState`** — the actual feature: new proto fields + regenerated Go, `apiutil.PeerState` fields, and the `extractNotificationCodeSubcode` helper wired into `WatchEvent`'s peer-event builder.

## Test plan

- New test `TestWatchEventPeerNotificationCodeSubcode` drives a genuine bad-peer-AS NOTIFICATION exchange between two live `*BgpServer` instances and asserts the resulting `PeerState` carries the real RFC 4271 OPEN Message Error / Bad Peer AS code/subcode.
- `go build ./...`, `go vet ./...` clean.
- `go test ./...` passes across every package (two pre-existing tests, `TestEBGPRouteStuck` and `TestRTCDeferralTimerRaceCondition`, require loopback IP aliases not configured on macOS by default and are unaffected by this change — verified they fail identically on unmodified `master`).


proto/api/gobgp.proto
```diff
@@ -869,6 +869,15 @@ message PeerState {
...
   BfdPeerState bfd_state = 23;
+  // Raw RFC 4271 §6 NOTIFICATION error code/subcode, extracted from the FSM's
+  // underlying BGP NOTIFICATION whenever one is present. This is not limited
+  // to DISCONNECT_REASON_NOTIFICATION_SENT/_RECEIVED -- fsmBadPeerAS,
+  // fsmInvalidMsg, fsmHoldTimerExpired, and fsmAdminDown can all carry a real
+  // notification too. 0/0 when no notification was involved. Needed because
+  // disconnect_reason/disconnect_message alone cannot expose these as typed
+  // integers without fragile string-parsing.
+  uint32 notification_code = 24;
+  uint32 notification_subcode = 25;
```
fujita, line 880:
```
// A BGP NOTIFICATION message, as defined in RFC 4271 section 4.5.
message BgpNotification {
  uint32 error_code = 1;
  uint32 error_subcode = 2;
  bytes data = 3;
}
```

Let's create a new type for notification because we might need data later.

fujita:
Thanks!

The first commit is a fix. So I've already merged it.
Please update the second commit as I commented and rebase against the current master.

PR #3544, opened 2026-08-22
server: withdraw ADD-PATH route when export policy starts rejecting it

YafimK:
### Problem

In `propagateUpdateToNeighbors`, the ADD-PATH send branch calls `filterpath` with `nil` as the `old` argument:

```go
newPath := s.filterpath(targetPeer, newPath, nil)
if newPath == nil {
    bestList = []*table.Path{}
}
```

Because `old` is always `nil`, a path that was previously advertised to a peer and then starts failing that peer's export policy is silently dropped instead of being withdrawn. This happens whenever an attribute change causes the path to cross a policy match condition (for example a community is added and a policy rejects that community).

The peer keeps a stale route indefinitely, with no further update to correct it.

The non-ADD-PATH sibling, `processOutgoingPaths`, threads a real `old` path into `filterpath` and handles this case correctly, so the bug is specific to peers with ADD-PATH send enabled.

Note that `filterpath` already contains the logic for exactly this case:

```go
// When 'path' is filtered (path == nil), check 'old' has been sent to this peer.
// If it has, send withdrawal to the peer.
if path == nil && old != nil {
```

but it is unreachable from this call site because `old` is hardcoded to `nil`. A correct per-path `old` is not available here: the `oldList` produced by `dstsToPaths` holds the old *best* path per destination, which is the right notion for the non-ADD-PATH branch but not for ADD-PATH, where a peer holds several paths per prefix and the path being replaced is usually not the old best.

### Fix

Use the `alreadySent` bookkeeping already computed on the preceding line to distinguish "never advertised" from "advertised, now filtered". In the latter case, clone the pre-filter path as a withdraw and send it.

An attribute-only replacement inherits the old path's local ADD-PATH ID, so the `alreadySent` bookkeeping is already accurate for this case, and `Clone(true)` preserves the local/remote path IDs the withdraw needs. No new struct field and no additional locking.

### Testing

- New regression test `TestAddPathWithdrawsPreviouslyAdvertisedPathOnPolicyFilter` in `pkg/server/server_test.go`. It fails without the fix (times out waiting for outbound paths) and passes with it.
- `go test -race ./...` — no data races; the only failures are two pre-existing environment-dependent tests (`TestEBGPRouteStuck`, `TestRTCDeferralTimerRaceCondition`) that need extra loopback address aliases and fail identically on master.
- `gofmt -l .` clean, `go vet` clean, `golangci-lint run` reports nothing attributable to these changes.
- `GOARCH=386 GOOS=linux go build ./...` passes.
- Scenario tests on an x86-64 Ubuntu 24.04 host: `addpath_test.py` (25 passed), `route_server_policy_test.py`, `global_policy_test.py` (18 passed) and `route_server_softreset_test.py` (3 passed) all pass.

### Known adjacent issue, not addressed here

When a path is withdrawn, its ADD-PATH send-max slot is freed, and a path previously suppressed by `send-max` should be promoted into it. The withdraw branch of this same function does that (its `toAdd` loop), but three other places that stop advertising a path do not: `softResetOut`, the RTC route-target-loss path, and the branch this PR touches.

This is pre-existing and independent of the change here — for example `softResetOut` withdraws export-policy-rejected routes with the same `hasPathAlreadyBeenSent`/`Clone(true)` pattern and has no `send-max` awareness at all. Fixing it consistently means extracting the promotion loop into a helper and calling it from every site that frees a slot, which is a separate change; I would rather not mix it into this bug fix. Happy to follow up with it if that is useful.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1547,10 +1547,22 @@ func (s *BgpServer) propagateUpdateToNeighbors(rib *table.TableManager, source *
 					}()
 				} else {
 					alreadySent := targetPeer.hasPathAlreadyBeenSent(newPath)
+					unfilteredPath := newPath
 					newPath := s.filterpath(targetPeer, newPath, nil)
-					// if the path is not filtered and the path has already been sent or land in the limit, we can send it
+					// if the path is not filtered and the path has already been sent or lands in the limit, we can send it.
+					// if the path IS filtered but was previously advertised, we must withdraw it explicitly.
 					if newPath == nil {
-						bestList = []*table.Path{}
+						if alreadySent {
+							// The path was previously advertised to this peer and no
+							// longer passes export policy (e.g. an attribute change
+							// crossed a policy match condition). It must be withdrawn
+							// explicitly, or the peer keeps a stale route indefinitely.
+							withdraw := unfilteredPath.Clone(true)
+							bestList = []*table.Path{withdraw}
+							targetPeer.updateRoutes(withdraw)
+						} else {
+							bestList = []*table.Path{}
+						}
 					} else if alreadySent || targetPeer.getRoutesCount(f, newPath.GetPrefix()) < targetPeer.getAddPathSendMax(f) {
 						bestList = []*table.Path{newPath}
 						if !alreadySent {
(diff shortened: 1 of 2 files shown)


YafimK:
Closing while I re-verify some adjacent behaviour around ADD-PATH send-max bookkeeping. I will reopen or resubmit once the analysis is settled.

PR #3555, opened 2026-08-26
config: fix deleting peer group and neighbor together

rastislavs:
When a peer and its peer group are deleted in the same configuration update, deleting the peer group before the peer fails because the peer still references it.

To fix this, move peer-group deletion after neighbor updates and removals.


fujita:
Thanks!

PR #3556, opened 2026-08-26
tcp-ao: Add support for HMAC-SHA256 profiles

rastislavs:
Extend the TCP-AO implementation to also support HMAC-SHA256 algorithm with variable MAC length (96 or 128 bits).

This is not defined in the RFC 5926, but supported by the Linux kernel as well as most router platforms.


fujita:
Thanks!

PR #3557, opened 2026-08-27
packet/bgp: keep the value of an unknown route distinguisher

Mounika2456:
Decode then re-serialize of an 8-byte RD whose type is not one of the three known values drops the value:

    in : 0003deadbeef0001
    out: 0003000000000000

GetRouteDistinguisher constructs RouteDistinguisherUnknown with only Type set, so Value stays nil and Serialize emits six zero octets. Any VPN, EVPN, MUP or VPLS route carrying such an RD is re-advertised with a rewritten RD, and two RDs differing only in the value share one table key. RouteDistinguisherUnknown.DecodeFromBytes already copies the value but has no caller; set it in the factory instead.


fujita:
Pushed, thanks.

PR #3558, opened 2026-08-28
apiutil: round-trip the SR-MPLS Binding SID label through the API

ryskn:
`MarshalSRBSID` copies the native Binding SID value into the API `Sid` field verbatim.

For an SR-MPLS Binding SID this value is a 4-octet MPLS label stack entry with the label in the high 20 bits (RFC 9830 Figure 6; the low 12 bits are reserved TC/S/TTL).
The API carries the plain label value, and `NewBSID` shifts it left by 12 into the label position when converting from the API.
The marshal direction never undoes that shift, so reading a route back through the API returns the encoded stack entry instead of the label: a label of 100 comes back as 409600.
Re-injecting that value shifts it again, so an API read/write round trip corrupts the label.

Undo the shift in `MarshalSRBSID` for the 4-octet SR-MPLS case, so the API round-trips the label value.
The 16-octet SRv6 case (an SRv6 SID carried in the Binding SID sub-TLV for backward compatibility) is copied verbatim as before.

Tests are included: an unmarshal/marshal round trip that checks the label survives, and a check that a 16-octet SRv6 SID is not shifted.

### DIFF
--- a/pkg/apiutil/attribute.go
+++ b/pkg/apiutil/attribute.go
@@ -3365,6 +3365,13 @@ func MarshalSRBSID(bsid *bgp.TunnelEncapSubTLVSRBSID) (*api.SRBindingSID, error)
 		Sid: make([]byte, len(bsid.BSID.Value)),
 	}
 	copy(s.Sid, bsid.BSID.Value)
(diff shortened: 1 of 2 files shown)


fujita:
Pushed, thanks.

PR #3559, opened 2026-08-28
tcp-ao: add TCP-AO peer configuration & state API

rastislavs:
Part of TCP-AO Implementation (https://github.com/osrg/gobgp/issues/3493) adds proto API and OC model for configuring TCP-AO for peers.

The actual implementation of per-peer config will be added in a follow-up PR.

### DIFF
--- a/proto/api/gobgp.proto
+++ b/proto/api/gobgp.proto
@@ -709,6 +709,7 @@ message Peer {
   repeated AfiSafi afi_safis = 10;
   TtlSecurity ttl_security = 11;
   BfdPeerConfig bfd = 12;
+  TcpAoPeerConfig tcp_ao = 13;
 }
 
 message PeerGroup {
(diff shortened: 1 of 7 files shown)


proto/api/gobgp.proto
```diff
@@ -1497,3 +1500,24 @@ message ListTcpAoKeychainRequest {
 message ListTcpAoKeychainResponse {
   TcpAoKeychain keychain = 1;
 }
+
+message TcpAoPeerConfig {
+  string keychain = 1;
+  uint32 preferred_send_id = 2;
```
fujita: "preferred" reads like a soft hint but it's used as a key to look up the keychain that has the that SendID. Simply `send_id` is better?

tools/pyang_plugins/gobgp.yang
```diff
@@ -1851,4 +1851,46 @@ module gobgp {
...
+    description
+      "TCP-AO keychain attachment for a BGP neighbor or peer-group.";
+
+    leaf keychain {
+      type keychain-ref;
+      description
+        "Reference to the TCP-AO keychain used by the peer.";
+    }
+
+    leaf preferred-send-id {
```
fujita: ditto.

proto/api/gobgp.proto
rastislavs: makes sense, renamed

fujita:
Thanks!

PR #3560, opened 2026-08-28
mrt: bound element counts read from the input before preallocating

tzh476:
## Description

`parsePeerIndexTable`, `parseRib` and `parseGeoPeerTable` each read a `uint16` element count out of the MRT data and use it as a capacity hint straight away:

```go
peerNum := binary.BigEndian.Uint16(data[:2])
data = data[2:]
t.Peers = make([]*Peer, 0, peerNum)
```

The per-element decoder rejects the first element that is short — `Peer.decodeFromBytes` returns `errNotAllPeerBytesAvailable` when `len(data) < 5` — so a count with no element data behind it gets reserved and then thrown away. MRT files come from disk or from a peer, so the count is not trusted input.

This bounds each count by what the remaining bytes could hold, using the minimum length each element decoder already enforces (9, 8 and 13 bytes respectively). Capacity is only a hint to `append`, so it cannot change what is parsed, and a real table is unaffected because its elements are longer than the minimum.

## Measurement

An 8 byte `PeerIndexTable` claiming 65535 peers. Apple M3 Pro, `-benchtime 100x`:

| | B/op | allocs/op | ns/op |
|---|---|---|---|
| before | **524,521** | 3 | 64,782 |
| after | **136** | 3 | 185.8 |
| Δ | **−99.97%** | | **−99.7%** |

The count is a `uint16`, so each call is capped around 512 KB rather than unbounded — but it is 8 bytes of input per 512 KB, and nothing stops it repeating.

## Why a proportional bound rather than a constant

I tried constant clamps first on similar code elsewhere and they were a regression: when a valid input has one element per separator, the count is an exact estimate rather than an overestimate, and `append` then has to regrow from the constant. Bounding by `len(data)/minElementLen` avoids that — hostile input is cut down while legitimate tables keep the exact hint they had before.

## Testing

`TestPeerIndexTableCapHint` added to `mrt_test.go`, using the `assert` import it already has:

- parses tables of 0 / 1 / 2 / 10 / 100 real peers and asserts the parsed count is unchanged;
- asserts a count of 65535 with no peer bytes behind it is still an error, so the bound cannot turn invalid input into a silent success.

`go test ./pkg/packet/...` passes (bfd, bgp, bmp, mrt, rtr).

## Disclosure

The three sites were found by a static checker I wrote that looks for allocations sized by a count decoded from untrusted bytes with no bounds check in between, and this change was prepared with AI assistance. The numbers are from benchmark runs on this diff, and I will answer review questions myself.

### DIFF
--- a/pkg/packet/mrt/mrt.go
+++ b/pkg/packet/mrt/mrt.go
@@ -227,6 +227,14 @@ type Peer struct {
 
 var errNotAllPeerBytesAvailable = errors.New("not all Peer bytes are available")
 
+// Minimum encoded length of one element, used to bound preallocation hints that
+// are read from the input. Each is the smallest length its own decoder accepts.
+const (
+	minPeerLen     = 5 + 4 // type, BGP id, then at least an IPv4 address
+	minRibEntryLen = 8     // peer index, originated time, attribute length
+	minGeoPeerLen  = 13    // type, BGP id, latitude, longitude
+)
+
 func (p *Peer) decodeFromBytes(data []byte) ([]byte, error) {
 	if len(data) < 5 {
 		return nil, errNotAllPeerBytesAvailable
@@ -340,7 +348,11 @@ func parsePeerIndexTable(data []byte) (*PeerIndexTable, error) {
 	}
 	peerNum := binary.BigEndian.Uint16(data[:2])
 	data = data[2:]
-	t.Peers = make([]*Peer, 0, peerNum)
+	// peerNum comes from the input, so it must not size an allocation on its own:
+	// decodeFromBytes below rejects the first peer that is short, and a peer needs
+	// at least minPeerLen bytes. Capacity is only a hint to append, so bounding it
+	// cannot change the result.
+	t.Peers = make([]*Peer, 0, min(int(peerNum), len(data)/minPeerLen+1))
 	var err error
 	for range peerNum {
 		p := &Peer{}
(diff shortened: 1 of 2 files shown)


tzh476:
Rebased onto current master (`eb2a727`) and squashed to a single commit — `bce14965`.

Worth flagging why: the branch no longer applied cleanly. The test file's import block on master lacks `encoding/binary` and `runtime`, both of which this test needs, so a merge would have produced a build failure rather than a conflict marker. Fixed and re-verified rather than left for you to hit.

Re-verified after the rebase:

```
go test ./pkg/packet/mrt/          ok
gofmt -l pkg/packet/mrt/           clean
```

And the control experiment, since a bound is only worth having if something notices its absence — removing the two `min(...)` bounds again:

```
--- FAIL: TestPeerIndexTableCapHint
    parsing a table that claims 65535 peers with no peer data allocated 524416 bytes
```

With the bounds in place that drops to the proportional figure and the same error is still returned, which is the point: capacity is a hint to append, so bounding it cannot change behaviour. Both tests assert the expected error is still produced.

One design note in case it matters to you: the bound is `len(data)/minElementLen+1`, derived from each element's own minimum encoded length, not a fixed ceiling. A constant cap would penalise a genuinely large table, which I wanted to avoid — I had that exact regression in an earlier version of this patch on a different repo.

Disclosure: this is human + LLM collaboration; the numbers above are from runs on this branch.

pkg/packet/mrt/mrt.go
```diff
@@ -227,6 +227,14 @@ type Peer struct {
 
 var errNotAllPeerBytesAvailable = errors.New("not all Peer bytes are available")
 
+// Minimum encoded length of one element, used to bound preallocation hints that
+// are read from the input. Each is the smallest length its own decoder accepts.
+const (
+	minPeerLen     = 5 + 4 // type, BGP id, then at least an IPv4 address
```
fujita, line 233: At least 2 byte AS is necessary so should be 11?

fujita:
In BGP operation an MRT file is not an untrusted source. And even if it were, a large valid RIB dump allocates far more than this, so there is no reason to craft a short invalid one. I do not think the allocation is worth three constants and a comment.

Issue #3449, opened 2026-06-10
ParseBGPMessage accepts BGP messages violating RFC 4271 MUST requirements


afldl:
While doing differential testing of BGP parsers (against bird-3.3.0, FRR 10.6.1, and OpenBGPD), I found that `gobgp/v4/pkg/packet/bgp.ParseBGPMessage` accepts several inputs that violate RFC 4271 MUST clauses.

To be clear: gobgp's own FSM calls `ValidateOpenMsg` separately, so the default `gobgp` daemon is not affected. This is about third-party code using the parser as a library.

**Inputs accepted by gobgp that bird/FRR/OpenBGPD reject:**

| Input | RFC clause | bird | FRR | openbgpd | gobgp |
|---|---|---|---|---|---|
| OPEN Hold Time = 1 | RFC 4271 §4.2 (must be 0 or ≥3) | reject | reject | reject | **accept** |
| OPEN Hold Time = 2 | RFC 4271 §4.2 | reject | reject | reject | **accept** |
| OPEN BGP ID = 0.0.0.0 | RFC 6286 §2 (must be nonzero) | reject | reject | reject | **accept** |
| OPEN Version = 3 | RFC 4271 §4.2 (must be 4) | reject | reject | reject | **accept** |
| Header Length = 4097+ | RFC 4271 §4.1 (max 4096) | reject | reject | reject | **accept** |
| KEEPALIVE Length ≠ 19 | RFC 4271 §4.4 (exactly 19) | reject | reject | reject | **accept** |

All return a non-nil `*BGPMessage` with `err == nil`.

**Root cause:** Validation logic exists in `ValidateOpenMsg` (`pkg/packet/bgp/validate.go`) but `DecodeFromBytes` doesn't call it. The header decoder (`BGPHeader.DecodeFromBytes`) only checks `Len < 19`, never `Len > 4096`. KEEPALIVE decoder just returns nil without checking length.

**Suggestion:** Either add RFC MUST checks directly in `DecodeFromBytes`, or document clearly that callers must run `ValidateOpenMsg` themselves to get spec-compliant validation.

Found through differential fuzzing (PathDiff v23), comparing gobgp, bird, FRR, and OpenBGPD.

PR #3567, opened 2026-08-30
packet/bgp: reject KEEPALIVE messages that are not 19 octets

Sonike:
### What

`BGPKeepAlive.DecodeFromBytes` returned `nil` without inspecting its input, so
both `ParseBGPMessage` callers and the FSM accepted a KEEPALIVE whose declared
length was not 19 octets. RFC 4271 Sections 4.4 and 6.1 require the length to be
exactly 19.

### How

`parseBody` passes the bytes declared after the 19-octet header to the body
decoder. A conforming KEEPALIVE therefore arrives as an empty slice; any
non-empty slice is rejected with Bad Message Length.

RFC 4271 Section 6.1 also requires the NOTIFICATION Data field for this error
to contain the erroneous two-octet Length field. The decoder now records that
big-endian length in `MessageError.Data`.

### Scope

This deliberately covers only the KEEPALIVE item from #3449:

- The OPEN checks for Hold Time, BGP Identifier and Version already live in
  `ValidateOpenMsg`. Moving them into the decoder would change validation
  boundaries and should be a separate maintainer decision.
- The general message-length ceiling is session-dependent because RFC 8654
  raises it for UPDATE, NOTIFICATION and ROUTE-REFRESH after Extended Message
  Capability negotiation. The session layer has that state; the standalone
  parser does not.

### Testing

- `go test ./pkg/packet/bgp -run '^TestParseBGPMessageKeepAliveLength$' -count=1`
- `go test -race -timeout 240s ./pkg/packet/bgp`
- `go build ./...`
- `go vet ./...`
- `golangci-lint run ./pkg/packet/bgp/...`
- `golangci-lint run --new-from-rev=upstream/master`

The full `go test -race -timeout 240s ./...` run passed outside `pkg/server`.
`TestEBGPRouteStuck` and `TestRTCDeferralTimerRaceCondition` cannot bind
127.0.0.100 and 127.0.0.201 on this macOS host; these are unchanged
environment limitations and unrelated to the packet decoder.

Refs #3449

### DIFF
--- a/pkg/packet/bgp/bgp.go
+++ b/pkg/packet/bgp/bgp.go
@@ -17025,6 +17025,13 @@ func ShouldHardReset(subcode uint8, hardResetOnAdminReset bool) bool {
 type BGPKeepAlive struct{}
 
 func (msg *BGPKeepAlive) DecodeFromBytes(data []byte, options ...*MarshallingOption) error {
+	// RFC 4271 Section 4.4: a KEEPALIVE consists of only the message
+	// header and has a length of exactly 19 octets, so the body that
+	// parseBody hands us must be empty.
+	if len(data) != 0 {
+		length := uint16(BGP_HEADER_LENGTH + len(data))
+		return NewMessageError(BGP_ERROR_MESSAGE_HEADER_ERROR, BGP_ERROR_SUB_BAD_MESSAGE_LENGTH, binary.BigEndian.AppendUint16(nil, length), fmt.Sprintf("KEEPALIVE length must be %d", BGP_HEADER_LENGTH))
+	}
 	return nil
 }
 
--- a/pkg/packet/bgp/extended_message_test.go
+++ b/pkg/packet/bgp/extended_message_test.go
@@ -2,6 +2,8 @@ package bgp
 
 import (
 	"bytes"
+	"encoding/binary"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


fujita:
I think that bird does not reject an over-long KEEPALIVE.

pkg/packet/bgp/bgp.go
```diff
@@ -17025,6 +17025,12 @@ func ShouldHardReset(subcode uint8, hardResetOnAdminReset bool) bool {
 type BGPKeepAlive struct{}
 
 func (msg *BGPKeepAlive) DecodeFromBytes(data []byte, options ...*MarshallingOption) error {
+	// RFC 4271 Section 4.4: a KEEPALIVE consists of only the message
+	// header and has a length of exactly 19 octets, so the body that
+	// parseBody hands us must be empty.
+	if len(data) != 0 {
+		return NewMessageError(BGP_ERROR_MESSAGE_HEADER_ERROR, BGP_ERROR_SUB_BAD_MESSAGE_LENGTH, nil, fmt.Sprintf("KEEPALIVE length must be %d", BGP_HEADER_LENGTH))
```
fujita: You can't use nil here. RFC 4271 Section 6.1 says that `The Data field MUST contain the erroneous Type field.`.

Sonike: Good catch on the empty Data field. I added a regression assertion and now populate it in e37c5a20. Because this is Bad Message Length, RFC 4271 Section 6.1 requires the Data field to carry the erroneous two-octet Length field; the following sentence about Type applies to Bad Message Type. The test now checks the code, subcode, and big-endian length for 20, 100, and 4096.

Sonike:
You are right about BIRD: its current receive path checks only the general 19-to-session-maximum range and dispatches KEEPALIVE without a type-specific length check. I removed that comparison from the PR body and amended commit message. The change is now justified directly by RFC 4271 Sections 4.4 and 6.1.

fujita:
Thanks.

PR #3575, opened 2026-08-31
Fix test/scenario_test/run_all_tests.sh script

SampleText2k77:
There are two problems and fixes:
1. Tests are using pytest instead of nose. Patched launch instructions to use pytest.
2. When running test scripts just to find out the amount of scenarios, there are two strings matching the awk pattern: the one we need and another one showing the traceback. Updated the awk pattern to match only required info.

### DIFF
--- a/test/scenario_test/run_all_tests.sh
+++ b/test/scenario_test/run_all_tests.sh
@@ -19,55 +19,55 @@ TESTDIR=`pwd`/test/scenario_test
 PIDS=()
 
 # route server test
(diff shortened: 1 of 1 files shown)


fujita:
Thanks for the fix, and the analysis is right.

The script has been completely broken for two months. It ran zero tests and still printed "all tests passed successfully". Nobody complained in that time.

It also never did what the name says. It runs 13 of the 30 scenario test files.

So I am going to delete run_all_tests.sh and document how to run the tests with pytest in the README instead.

PR #3578, opened 2026-09-01
server: drop stale notifications queued while a peer session was down

### LINKED ISSUE #3561: BFD: a stale ADMINISTRATIVE_RESET kills the next BGP session, causing a permanent flap loop
## What happens

A BFD session can go down while its BGP session is already down. When that happens, gobgp queues an
administrative reset instead of discarding it. The next BGP session then receives that reset and
shuts itself down, about 200 microseconds after it reaches `ESTABLISHED`.

Some peers destroy their BFD peer when the BGP session drops. FRR does this when the neighbour comes
from a `bgp listen range`. Against such a peer, the fault repeats forever. Every reconnect dies
immediately, BFD never gets time to come up, and the peering stays down until an operator intervenes.

We watched this run for 9 minutes and 19 consecutive resets before we broke the loop by hand.

## What we are asking for

We would like you to confirm which fix you prefer. We describe three under
[How to fix it](#how-to-fix-it), and we recommend the first.

We are happy to write the pull request once you tell us which approach you want.

BFD's server and peer state machine first shipped in **v4.6.0**. Every release since then is
affected, including **v4.8.0** and current `master`. Note that v4.5.0 contains only the packet codec
in `pkg/packet/bfd`, so it is not affected.

We checked v4.8.0 against master on 2026-08-28. `pkg/server/bfd_peer.go`, `pkg/server/fsm.go` and
`pkg/server/peer.go` are byte-identical. `pkg/server/server.go` has changed, but not on this code
path.

Line numbers below come from v4.8.0. Where master differs, we give both.

Our peer was FRR 7.5.1 with `bfdd` enabled.

## Why it happens

gobgp generates the reset correctly. It then delivers it to the wrong session.

Here is the path:

```
bfd_peer.go:394   expiry() calls resetPeer()
bfd_peer.go:415   resetPeer() sends ResetPeer{Soft: false, Communication: "BFD is down"}
server.go:4023    ResetPeer calls sendNotification("Neighbor reset", addr, ADMINISTRATIVE_RESET, ...)
server.go:3982    sendNotification calls peer.sendNotification(m) for each matched peer
peer.go:602       peer.sendNotification calls nonblockSendChannel(peer.fsm.notification, msg)
fsm.go:426        notification chan *bgp.BGPMessage    ← a field of the long-lived fsm struct
fsm.go:2046       case m := <-fsm.notification:        ← read in the ESTABLISHED loop
```

On master, `ResetPeer` is at `server.go:4064` and `sendNotification` at `server.go:4013`.

Nothing on this path asks whether the BGP session is up.

`fsm.notification` is the problem. `newFSM` creates it once per peer, with capacity 1
(`fsm.go:574`). The `fsm` struct outlives every session that peer ever builds. `fsm.go:2046` is the
only place in the file that reads from the channel, and nothing drains it when a session ends.

So a message posted while the peer is down just waits. The FSM reads it the next time it enters the
established loop, sends it, and ends the session it has only just built.

Two functions in the BFD code produce this reset:

- `expiry()` (`bfd_peer.go:381`) runs when the detect timer lapses.
- `remoteDown()` (`bfd_peer.go:405`) runs when the remote signals `AdminDown` or `Down`
  (`bfd_peer.go:340` and `:347`).

`expiry()` already guards against one case:

```go
func (p *bfdPeer) expiry() {
	if p.sessionState() == api.BfdSessionState_BFD_SESSION_STATE_DOWN {
		p.eventExpiry.Stop()
		return
	}
	...
	p.resetPeer()
	p.setStateDown()
}
```

This guard explains why a BFD session that never came up is harmless. It starts in `DOWN`, so
`expiry()` returns before it reaches `resetPeer()`. To hit the bug, BFD must reach `Init` or `Up` at
least once per cycle. A dynamic-neighbour peer provides exactly that.

## What we saw

The loop repeated every 32 seconds:

```text
21:19:34.263529 fsm.go:975  "starting outgoing connection manager"  Key="10.1.1.11"
21:19:36.214738 fsm.go:2152 "Peer Up"                               Key="10.1.1.11"
21:19:36.214790 fsm.go:719  "restart flag is not set"  State="BGP_FSM_OPENCONFIRM"
21:19:36.214921 fsm.go:784  "sent notification" State="BGP_FSM_ESTABLISHED" Code=6 Subcode=4 \
                              Communicated-Reason="BFD is down"
21:19:36.215088 fsm.go:2156 "Peer Down"  Reason="notification-sent code 6(cease) subcode 4(...)"
21:19:39.215929 bfd_peer.go:387 "Expired"  Topic="bfd"  Peer="10.1.1.11"
21:19:39.216049 server.go:3973  "Send operation notification"  Op="Neighbor reset"
21:20:06.216190 fsm.go:975  "starting outgoing connection manager"   ← next cycle
```

Look at the gap between `Peer Up` and the notification that kills it: **183 microseconds**. No BFD
implementation can finish a `Down → Init → Up` handshake that fast. The session cannot survive.

## Why the loop never ends

1. BGP reconnects. FRR creates the dynamic neighbour, and with it a BFD peer.
2. BFD starts its handshake and leaves `DOWN`.
3. BGP reaches `ESTABLISHED`. The queued notification fires and kills the session.
4. FRR destroys the dynamic neighbour and its BFD peer, then stops sending BFD packets.
5. gobgp's detect timer expires. The state is not `DOWN`, so `resetPeer()` runs and queues a **new**
   notification.
6. Return to step 1.

Every pass creates the conditions for the next one.

## What started it

A BFD parameter change started our loop:

```text
21:09:49.856258 server.go:3522 "Add a peer configuration"  Key="10.1.1.11"
21:09:49.856601 bfd_server.go:311 "Insert BFD peer"        Peer="10.1.1.11"
21:09:51.829557 fsm.go:2152 "Peer Up"                      Key="10.1.1.11"
                  ... healthy for ~3 minutes, BFD Up, confirmed from the FRR side ...
21:12:35.711023 server.go:3863 "Update BFD configuration"  Key="10.1.1.11"   ← timers changed
21:12:35.711178 bfd_server.go:341 "Remove BFD peer"        Peer="10.1.1.11"
21:12:35.711231 bfd_server.go:311 "Insert BFD peer"        Peer="10.1.1.11"
21:12:36.913630 fsm.go:2156 "Peer Down"  Reason="graceful-restart"
21:12:39.913977 bfd_peer.go:387 "Expired"                  Peer="10.1.1.11"
21:12:43.628538 fsm.go:2152 "Peer Up"
21:12:43.628697 fsm.go:784  "sent notification" ... "BFD is down"      ← the loop starts here
```

The parameter change is not special. Any event that drops BFD and BGP together reaches the same
state. A link failure would do it too.

One detail explains why a parameter change drops both. `updateBfdPeer` (`server.go:3569`, master
`:3598`) handles any change by deleting the BFD peer and adding it again. Changing an interval
therefore takes a healthy BFD session down and back up. We consider that a separate problem and will
raise it separately. We mention it here only because it gives operators a routine way into this bug.

## What we ruled out

We configured a second peer, `10.1.1.12`, identically on the gobgp side. It received the same
`Update BFD configuration`, with the same `Remove BFD peer` and `Insert BFD peer` pair, at the same
moment. We deliberately never enabled BFD for it on the FRR side.

It recorded **zero** resets while `10.1.1.11` recorded 19.

The two peers differ in one respect: whether the BFD session had ever left `DOWN`. This confirms that
the `expiry()` guard separates the safe case from the broken one. It also rules out the parameter
change as a sufficient cause on its own.

## How to reproduce it

Your peer must **destroy its BFD peer when the BGP session drops**. This is the one essential
ingredient. Without it, BFD stays `DOWN` after the first failure, the `expiry()` guard holds, and the
session settles after a single flap.

This rules out the obvious minimal setup of two gobgp instances and a firewall rule on UDP 3784. We
suggest you skip it.

Instead:

1. Configure FRR with `bfdd` enabled, `neighbor <group> bfd`, and
   `bgp listen range <prefix> peer-group <group>`.
2. Start a gobgp speaker inside that range with BFD enabled, connecting inbound. We used
   `RequiredMinimumReceive: 500000`, `DesiredMinimumTxInterval: 500000` and `DetectionMultiplier: 4`.
3. Wait until both the BGP session and the BFD session come up.
4. Change any BFD parameter on the gobgp side. Alternatively, force BFD down while BGP is also down.
5. Watch the `Peer Up` → `sent notification "BFD is down"` → `Peer Down` cycle repeat on the
   `IdleHoldTimeAfterReset` interval.

A fixed gobgp should flap at most once, then hold the session with BFD up.

## How to fix it

Each option below breaks the cycle. You can apply more than one.

**Option 1 — drop queued notifications when the FSM leaves `ESTABLISHED`.** A pending administrative
reset belongs to the session that was live when gobgp generated it. Delivering it to a later session
is always wrong, whether or not BFD is involved. Draining `fsm.notification` during teardown fixes the
delivery bug itself, and protects every producer on that path.

**Option 2 — call `resetPeer()` only when the BGP session is established.** This is the cheapest and
most local change, and it covers both `expiry()` and `remoteDown()`. It fixes what we reported, but it
leaves the stale-notification hazard for other callers.

**Option 3 — reset only on a real `Up → Down` transition.** Today `expiry()` fires on any detect-timer
lapse from a non-`DOWN` state. That includes `Init`, which means a session that was coming up for the
first time and never finished. Most implementations tear BGP down only after BFD was genuinely
established and then failed.

**We recommend Option 1.** It closes the loop on its own, and it is correct independently of BFD. We
suggest Option 3 as a follow-up, because it brings the trigger into line with conventional
BFD-for-BGP behaviour.

## How to check a fix works

Whichever option you choose, please confirm all three of these:

- The loop ends. gobgp resets at most once, then holds `ESTABLISHED`, and BFD reaches `Up`.
- A real BFD failure on an established session still tears BGP down promptly. This is the behaviour
  the fix must protect. A fix that suppresses it has broken the feature.
- A BFD session that never comes up still leaves BGP alone.

## One thing we could not test

We do not know whether the loop needs a dynamic neighbour.

We expect that it does. With a statically configured peer, FRR keeps its BFD peer alive across a BGP
flap. BFD can then converge while BGP is down, and the stale notification costs one extra flap before
the session settles. We have not tested this.

Given your recent work in [#3469](https://github.com/osrg/gobgp/pull/3469), you may know the answer
already.

## About the addresses

We scrubbed the addresses. The route reflector appears as `10.1.1.11`, the second peer as
`10.1.1.12`, and the `bgp listen range` prefix as `10.1.0.0/21`. Timestamps and source locations are
unchanged, because the sub-millisecond gaps between them are the evidence.

notsrch:
Fixes #3561.

### What was wrong

`newFSM` creates `fsm.notification` once per peer, and only the ESTABLISHED loop reads it, so the
channel outlives every session of the peer. `ResetPeer` and `ShutdownPeer` enqueue into it without
checking the session state. A NOTIFICATION queued while the session is down waits in the channel,
and the next session reads it the moment it reaches ESTABLISHED and tears itself down. BFD produces
exactly this sequence: its detect timer expires seconds after the BGP session already ended and
queues an administrative reset. Against a peer that destroys its BFD session when BGP goes down,
this repeats on every reconnect and the peering never recovers (#3561 has the full trace).

### What this changes

`handleFSMMessage` now drops whatever is still queued in `fsm.notification` on every state
transition, next to the existing `outgoingCh` drain.

This is the issue's Option 1 in spirit, with one correction to its wording. Draining when the FSM
*leaves* ESTABLISHED would not break the loop: in the reported cycle, BFD queues the reset about
three seconds *after* teardown, while the peer is IDLE. The drain has to run between that enqueue
and the next delivery point, so it runs on every transition instead.

Why this placement is race-free: the transition callback runs under the server's read lock, and the
producers (`ResetPeer`, `ShutdownPeer`, including BFD's calls) run under the write lock. An enqueue
therefore lands either before the transition into ESTABLISHED — and is dropped, correctly, because
the reset targeted a session that no longer exists — or after it, when the established loop is live
and delivers it. A BFD failure detected moments after Peer Up still resets the session.

Draining at the top of `established()` instead would race the producers: a legitimate reset issued
right after Peer Up could be destroyed, and BFD would not repeat it (`expiry()` stops itself once
its state is DOWN). That would suppress a real BFD failure, which is the one behavior the fix must
protect.

The drain also covers a rarer variant of the same bug: the receive goroutine queues NOTIFICATIONs
for parse errors, and one left behind when the session dies concurrently was likewise delivered to
the next session. The teardown joins that goroutine before the state function returns, so the next
transition's drain disposes of it deterministically.

Intended behavior change: a hard `ResetPeer` or `ShutdownPeer` against a peer whose session is down
is now a no-op, instead of a deferred kill of the next session.

### Checks from the issue

- The loop ends: `TestResetPeerWhileDownDoesNotResetNextSession` reproduces the exact BFD call
  (`ResetPeer` with `Soft: false` while the peer is down) and fails on master — the next session
  dies with a NOTIFICATION on its connection moments after establishing. With the fix it stays
  ESTABLISHED and the channel is empty.
- A real BFD failure on an established session still tears BGP down promptly:
  `TestResetPeerEstablishedSendsNotification` asserts the session leaves ESTABLISHED and a
  cease/administrative-reset NOTIFICATION reaches the wire. It passes with and without the fix.
- A BFD session that never comes up still leaves BGP alone: unchanged; the `expiry()` guard and the
  existing BFD tests cover it.

`go test -race ./pkg/server/` passes. Option 3 from the issue (reset only on a real Up → Down BFD
transition) is left for a follow-up PR as suggested there.

Assisted-by: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1799,6 +1799,12 @@ func (s *BgpServer) handleFSMMessage(peer *peer, e *fsmMsg) {
 		}
 
 		drainChannel(peer.fsm.outgoingCh.Out())
+		// Drop any queued notification. Either it raced with the end of the
+		// session it was meant for, or it was generated while no session was
+		// up, e.g. by BFD detecting a failure while the peer was down. Only
+		// the established loop reads this channel, so a leftover would kill
+		// the next session right after it establishes.
+		drainChannel(peer.fsm.notification)
 
 		if nextState == bgp.BGP_FSM_ESTABLISHED {
 			conf := peer.fsm.pConf.ReadOnly()
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -4781,3 +4781,163 @@ func TestDeletePeerDropsPolicyAssignment(t *testing.T) {
 	assert.NoError(addPeer("not-defined"))
 	assert.Empty(assignedPolicies())
 }
+
+// startServerWithPassivePeer starts a BgpServer without a TCP listener and
+// adds one passive ipv4-unicast neighbor, so tests can drive sessions by
+// injecting MockConnections into the peer's FSM.
+func startServerWithPassivePeer(t *testing.T, asn uint32, peerAddr string) (*BgpServer, *peer) {
+	t.Helper()
+
+	ctx := context.Background()
+
+	s := NewBgpServer()
+	go s.Serve()
+
+	err := s.StartBgp(ctx, &api.StartBgpRequest{
+		Global: &api.Global{
+			Asn:        asn,
+			RouterId:   "192.168.1.1",
+			ListenPort: -1,
+		},
+	})
+	require.NoError(t, err)
+	t.Cleanup(func() { s.StopBgp(ctx, &api.StopBgpRequest{}) })
+
+	neighbor := &oc.Neighbor{
+		Config: oc.NeighborConfig{
+			NeighborAddress: netip.MustParseAddr(peerAddr),
+			PeerAs:          asn,
+		},
+		Transport: oc.Transport{
+			Config: oc.TransportConfig{
+				PassiveMode: true,
+			},
+		},
+		AfiSafis: []oc.AfiSafi{
+			{
+				Config: oc.AfiSafiConfig{
+					AfiSafiName: oc.AFI_SAFI_TYPE_IPV4_UNICAST,
+					Enabled:     true,
+				},
+			},
+		},
+	}
+
+	w := newPeerStateWaiter(s, api.PeerState_SESSION_STATE_ACTIVE)
+	err = s.AddPeer(ctx, &api.AddPeerRequest{
+		Peer: oc.NewPeerFromConfigStruct(neighbor),
+	})
+	require.NoError(t, err)
+	w.Wait(t, 10*time.Second)
+
+	peer := s.neighborMap[netip.MustParseAddr(peerAddr)]
+	require.NotNil(t, peer)
+
+	return s, peer
+}
+
+// establishSession drives the passive peer to ESTABLISHED over a new
+// MockConnection and returns it.
+func establishSession(t *testing.T, s *BgpServer, peer *peer, asn uint32, peerAddr string) *MockConnection {
+	t.Helper()
+
+	m := NewMockConnection()
+	m.SetRemoteAddr(peerAddr)
+	t.Cleanup(func() { m.Close() })
+
+	peer.fsm.connCh <- m
+	openMsg, err := bgp.NewBGPOpenMessage(uint16(asn), 90, netip.MustParseAddr(peerAddr),
+		[]bgp.OptionParameterInterface{
+			bgp.NewOptionParameterCapability([]bgp.ParameterCapabilityInterface{
+				bgp.NewCapMultiProtocol(bgp.RF_IPv4_UC),
+			}),
+		})
+	require.NoError(t, err)
+	m.PushBgpMessage(openMsg)
+	m.PushBgpMessage(bgp.NewBGPKeepAliveMessage())
+
+	waitPeerState(t, s, api.PeerState_SESSION_STATE_ESTABLISHED, 10*time.Second)
+
+	return m
+}
+
+func sentNotification(m *MockConnection) *bgp.BGPNotification {
+	for _, buf := range m.GetSentMessages() {
+		msg, err := bgp.ParseBGPMessage(buf)
+		if err != nil || msg.Header.Type != bgp.BGP_MSG_NOTIFICATION {
+			continue
+		}
+		return msg.Body.(*bgp.BGPNotification)
+	}
+	return nil
+}
+
+// A hard ResetPeer against a peer whose session is down must not affect the
+// session the peer establishes later. Before the fix, the queued NOTIFICATION
+// survived in fsm.notification until the next session reached ESTABLISHED and
+// tore it down moments later, which looped forever against peers that destroy
+// their BFD session when BGP goes down.
+// https://github.com/osrg/gobgp/issues/3561
+func TestResetPeerWhileDownDoesNotResetNextSession(t *testing.T) {
+	const (
+		asn      = 65001
+		peerAddr = "10.0.0.1"
+	)
+
+	s, peer := startServerWithPassivePeer(t, asn, peerAddr)
+	m1 := establishSession(t, s, peer, asn, peerAddr)
+
+	// Bring the session down and wait until the peer settles in ACTIVE.
+	// Passive mode and no listener: it cannot progress on its own.
+	m1.Close()
+	waitPeerState(t, s, api.PeerState_SESSION_STATE_ACTIVE, 10*time.Second)
+
+	// Reset the peer while it is down: exactly what the BFD code does when
+	// its detect timer expires after the BGP session already ended.
+	err := s.ResetPeer(context.Background(), &api.ResetPeerRequest{
+		Address:       peerAddr,
+		Communication: "BFD is down",
+		Soft:          false,
+	})
+	require.NoError(t, err)
+	require.Len(t, peer.fsm.notification, 1)
+
+	m2 := establishSession(t, s, peer, asn, peerAddr)
+
+	// The stale reset must not reach the new session: it must stay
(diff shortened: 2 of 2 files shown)


fujita:
Fixed the lint error and pushed. Thanks!

PR #3579, opened 2026-09-01
server: bound the RTR message length before allocating the body

Mounika2456:
The RTR client read loop in (*roaClient).established reads the 4-byte Length field from an RTR PDU header and then does make([]byte, totalLen-RTR_MIN_LEN) for the body. totalLen is a 32-bit value taken straight off the wire and is only checked against the lower bound (RTR_MIN_LEN); nothing caps it. A cache that sends an 8-byte header declaring Length 0xffffffff drives a ~4 GiB allocation before a single body byte is read. On the 386 and arm builds that request is larger than the address space, so the runtime aborts with a fatal out-of-memory error and gobgpd dies.

Add an upper bound (RTR_MAX_LEN, 65535) next to the existing lower-bound check and move the framing into readRTRMessage so the declared length is validated before the body slice is sized. The ceiling sits well above every defined PDU and mirrors the 16-bit length limit already enforced on BGP and ZAPI messages. RTRErrorReport.DecodeFromBytes already bounds its own Length for the same reason, but that runs after the framer has allocated, so the cap belongs at the read.

### DIFF
--- a/pkg/packet/rtr/rtr.go
+++ b/pkg/packet/rtr/rtr.go
@@ -41,15 +41,17 @@ const (
 )
 
 const (
-	RTR_SERIAL_NOTIFY_LEN         = 12
-	RTR_SERIAL_QUERY_LEN          = 12
-	RTR_RESET_QUERY_LEN           = 8
-	RTR_CACHE_RESPONSE_LEN        = 8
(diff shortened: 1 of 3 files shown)


fujita:
Thanks, merged with a lint fix.

Issue #3564, opened 2026-08-28
BFD: gobgp does not slow transmission to one second while the session is down, as RFC 5880 Section 6.8.3 requires


notsrch:
## What happens

When a BFD session is not Up, RFC 5880 requires the transmit interval to drop to at least one second.
gobgp keeps transmitting at its configured interval instead.

A session that never comes up therefore keeps sending at full rate indefinitely.

## What we are asking for

Please confirm you want this fixed, and tell us how you want the Poll Sequence handled. We explain
the complication below. We are happy to write the change once that is settled.

## Which versions this affects

Every release with BFD, so **v4.6.0 onward**, including **v4.8.0** and current `master`.

We checked `pkg/server/bfd_peer.go` on 2026-08-28. It is byte-identical between v4.8.0 and `master`.
Line numbers below come from v4.8.0.

## What the RFC requires

RFC 5880 Section 6.8.3:

> the system MUST set bfd.DesiredMinTxInterval to a value of not less than one second
> (1,000,000 microseconds)

This applies whenever `bfd.SessionState` is not Up. The purpose is to stop a system from flooding a
peer that is not answering, and to keep the cost of a dead session low.

## What gobgp does

The transmit path does not consider session state at all. `tx` chooses which state to advertise, and
then sends (`bfd_peer.go:370`):

```go
func (p *bfdPeer) tx() {
	switch p.sessionState() {
	case api.BfdSessionState_BFD_SESSION_STATE_UP:
		p.sendPacket(bfd.StateUp, false, false, p.yourDiscriminator)
	case api.BfdSessionState_BFD_SESSION_STATE_INIT:
		p.sendPacket(bfd.StateInit, false, false, p.yourDiscriminator)
	default:
		p.sendPacket(bfd.StateDown, false, false, 0)
	}
}
```

The ticker that drives it is created once at construction (`bfd_peer.go:109`) and never reset. So a
peer configured with a 500 millisecond transmit interval sends twice a second forever, whether the
session is Up, Down, or has never come up.

## Why it matters

A configured BFD peer that never answers is not unusual. It happens when the far side has not been
configured yet, when BFD is enabled on one side only, and when UDP 3784 is blocked.

Our own deployment has the one-sided case: speakers with BFD enabled toward a route reflector that
deliberately does not run it. Each transmits twice a second, indefinitely, at a peer that discards
every packet. The same reflector's session table shows how common the state is: 31 of its 34 BFD
sessions sit permanently down.

[...]

Issue #3563, opened 2026-08-28
BFD: gobgp ignores the peer's Required Min RX Interval and transmits faster than the peer allows


notsrch:
## What happens

gobgp transmits BFD control packets at its own configured interval, always. It never slows down to
the interval the peer asked for.

RFC 5880 requires the transmit interval to be the larger of the two systems' values. When the peer
asks for a slower rate than gobgp is configured for, gobgp ignores the request and keeps sending at
its own rate.

## What we are asking for

Please confirm you want gobgp to honour the peer's advertised `Required Min RX Interval`. We are
happy to write the change.

## Which versions this affects

Every release with BFD, so **v4.6.0 onward**, including **v4.8.0** and current `master`.

We checked on 2026-08-28. `pkg/server/bfd_peer.go` and `pkg/packet/bfd/bfd.go` are byte-identical
between v4.8.0 and `master`. Line numbers below come from v4.8.0.

## What the RFC requires

RFC 5880 Section 6.8.7:

> a system MUST NOT transmit BFD Control packets at an interval less than the larger of
> bfd.DesiredMinTxInterval and bfd.RemoteMinRxInterval

`bfd.RemoteMinRxInterval` is the `Required Min RX Interval` the peer advertises in its own control
packets. The slower of the two systems sets the pace.

## What gobgp does

gobgp parses the field and then never uses it.

`UnmarshalBinary` reads it into `h.RequiredMinRxInterval` (`bfd.go:132`). In `pkg/server/bfd_peer.go`
that field appears exactly twice: once in a comment about detection time (`:330`), and once on the
transmit path where gobgp fills in *its own* value (`:444`). It is never read from a received packet.

The transmit ticker is set once, at construction (`bfd_peer.go:109`):

```go
p.eventTx = time.NewTicker(p.txInterval)
```

`p.txInterval` is written only in `NewBfdPeer`, at `:87` and `:105`. Nothing calls `eventTx.Reset`.
The only other operation on the ticker is `Stop`, during shutdown (`:401`).

So the transmit rate is fixed for the life of the peer, and no packet from the remote can change it.

## Why it matters

A peer that advertises a large `Required Min RX Interval` is asking gobgp to send less often, usually
because it cannot afford to process packets at gobgp's rate. gobgp overrides that request.

The consequence scales with the peer. A route reflector with many sessions asking each speaker to
slow down, and being ignored by all of them, is doing avoidable per-packet work in software.

[...]

PR #3581, opened 2026-09-01
server: jitter BFD control packet transmission

### LINKED ISSUE #3562: BFD: control packet transmission carries no jitter, as required by RFC 5880 Section 6.8.7
## What happens

gobgp transmits BFD control packets on a fixed period. RFC 5880 requires the period to be jittered on
every packet. gobgp applies no jitter at all.

## What we are asking for

Please confirm you want jitter added. We are happy to write the change. It is small, and the only
design question is which random source you want it to use.

Every release with BFD, so **v4.6.0 onward**, including **v4.8.0** and current `master`.

We checked `pkg/server/bfd_peer.go` on 2026-08-28. It is byte-identical between v4.8.0 and `master`.
Line numbers below come from v4.8.0.

## What the RFC requires

RFC 5880 Section 6.8.7:

> The periodic transmission of BFD Control packets MUST be jittered on a per-packet basis by up to
> 25%, that is, the interval MUST be reduced by a random value of 0 to 25%

There is a stricter rule when the detect multiplier is 1:

> If bfd.DetectMult is equal to 1, the interval between transmitted BFD Control packets MUST be no
> more than 90% of the negotiated transmission interval, and MUST be no less than 75%

That second rule is reachable in gobgp. `NewBfdPeer` accepts any non-zero multiplier
(`bfd_peer.go:96`), so a user can configure 1.

## What gobgp does

The transmit ticker is a plain `time.Ticker`, created once (`bfd_peer.go:109`):

```go
p.eventTx = time.NewTicker(p.txInterval)
```

The loop sends on every tick, with no per-packet variation (`bfd_peer.go:154`):

```go
case <-p.eventTx.C:
	p.tx()
```

`time.Ticker` fires on a fixed period by design, so every packet leaves on schedule.

## Why it matters

Jitter exists to stop independent sessions from synchronising. Without it, sessions that start at
similar times stay in lockstep and their transmissions bunch together.

The effect grows with session count. A gobgp speaker with many BFD peers configured at the same
interval will transmit them in a repeating burst rather than spreading them across the period. The
same applies in reverse to a peer receiving from many speakers that all behave this way.

This is a modest problem at three sessions and a real one at a few hundred.

## How to fix it

Replace the fixed ticker with a timer that is re-armed after each transmission, using a fresh
interval each time. Reduce the base interval by a random 0 to 25%, and apply the 75% to 90% window
instead when the detect multiplier is 1.

`randRange` already exists in `pkg/server/util.go` and is used to pick the BFD source port, so there
is a local precedent for the random source if you are happy with it here.

## How to check a fix works

- Capture a series of transmitted packets and confirm the intervals vary within the required window
  rather than sitting on a fixed period.
- With the detect multiplier at 1, confirm the intervals stay between 75% and 90% of the negotiated
  interval.
- Confirm the mean rate still satisfies the negotiated interval, and that sessions stay up.

## Related work in the same code

Two other transmit-path gaps sit in the same few lines, and we have raised them separately:

- gobgp does not raise the transmit interval to the peer's advertised `Required Min RX Interval`
  (Section 6.8.7).
https://github.com/osrg/gobgp/issues/3563
- gobgp does not reduce the transmit rate to at least one second while the session is down
  (Section 6.8.3).
https://github.com/osrg/gobgp/issues/3564

All three change how `eventTx` is driven, and this one changes its type. We suggest you take them in
sequence rather than in parallel. If you would prefer them as a single issue, say so and we will
consolidate.

notsrch:
Fixes #3562.

### What was wrong

The BFD transmit path drove transmission with a fixed `time.Ticker`, so every control packet left
exactly on the configured period. RFC 5880 section 6.8.7 requires the interval to be reduced on
every packet by a random value of 0 to 25%, and to stay between 75% and 90% of the transmission
interval when the detect multiplier is 1. Without jitter, sessions that start at similar times stay
in lockstep, and a speaker with many BFD peers on the same interval transmits them as a repeating
burst instead of spreading them across the period.

### What this changes

`eventTx` becomes a `*time.Timer` that the peer loop re-arms before each transmission with a fresh
interval: a uniformly random 75% to 100% of the configured desired minimum TX interval, narrowed to
75% to 90% when the detect multiplier is 1. Re-arming before the packet is written keeps the time
spent sending out of the gap between packets, which matters against the 90% ceiling in the
multiplier-1 case.

The issue asked which random source to use: this uses `randRange`, which already picks the BFD
source port. It draws whole percentage points rather than a continuous value (as the FSM's idle
hold-time jitter does), because 26 discrete values are directly assertable in a test and the RFC
only requires a random reduction within the window.

Jitter only shortens intervals, so the remote system's detection margin widens; it never rises
above the configured interval. Event-driven packets — poll replies and the poll sent on reaching
Up — are not periodic transmissions and are unchanged.

Out of scope, as proposed in the issue: raising the transmit interval to the peer's Required Min RX
Interval (#3563) and slowing transmission to one second while the session is down (#3564). Both
only change what base interval feeds the jitter, which now lives in one method; this PR changes the
type of `eventTx` first so those can follow in sequence.

### Checks from the issue

- Intervals vary within the required window: `Test_JitteredTxInterval` draws the interval 1000
  times per case and asserts the observed minimum and maximum land exactly on 75% and 100% of the
  configured interval. Endpoint equality also bounds every sample in between. The chance either
  endpoint never appears is about 1e-17, so the test is not timing-dependent and cannot flake the
  way a wall-clock packet capture in CI would.
- Detect multiplier 1 stays within 75% to 90%: the second half of the same test asserts the
  narrowed endpoints (75% and 90%) for a multiplier-1 peer.
- Mean rate and session health: the mean interval is 87.5% of the configured value and never
  exceeds it, so the negotiated rate is still satisfied. `Test_TxPacket` fails if the timer ever
  stops re-arming, and the pre-existing state-transition and expiry tests all pass unchanged.

`go test -race ./pkg/server/` passes.

Assisted-by: Claude [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/pkg/server/bfd_peer.go
+++ b/pkg/server/bfd_peer.go
@@ -58,7 +58,7 @@ type bfdPeer struct {
 
 	eventStart    *time.Ticker
 	eventRxPacket chan *bfd.BFDHeader
-	eventTx       *time.Ticker
+	eventTx       *time.Timer
 	eventExpiry   *time.Ticker
 	eventShutdown chan struct{}
 	shutdownOnce  sync.Once
@@ -106,7 +106,7 @@ func NewBfdPeer(ps peerState, logger *slog.Logger, peerAddress netip.Addr, confi
 	}
 
 	p.expiryInterval = time.Duration(p.multiplier) * p.rxInterval
-	p.eventTx = time.NewTicker(p.txInterval)
+	p.eventTx = time.NewTimer(p.jitteredTxInterval())
 
 	p.eventExpiry = time.NewTicker(p.expiryInterval)
 	p.eventExpiry.Stop()
@@ -152,6 +152,7 @@ func (p *bfdPeer) loop() {
 		case bfdPacket := <-p.eventRxPacket:
 			p.rxPacket(bfdPacket)
 		case <-p.eventTx.C:
+			p.eventTx.Reset(p.jitteredTxInterval())
 			p.tx()
 		case <-p.eventExpiry.C:
 			p.expiry()
@@ -365,6 +366,21 @@ func (p *bfdPeer) rxPacket(h *bfd.BFDHeader) {
 		p.sessionState() == api.BfdSessionState_BFD_SESSION_STATE_UP {
 		p.eventExpiry.Reset(p.expiryInterval)
 	}
+}
+
+// jitteredTxInterval returns the interval to wait before sending the next
+// BFD Control packet, based on the configured desired minimum TX interval.
+//
+// RFC 5880 Section 6.8.7: the transmit interval MUST be reduced per packet
+// by a random value of 0 to 25%; when the detect multiplier is 1, the
+// interval MUST be no more than 90% and no less than 75% of the negotiated
+// interval.
+func (p *bfdPeer) jitteredTxInterval() time.Duration {
+	maxPct := 100
+	if p.multiplier == 1 {
+		maxPct = 90
+	}
+	return p.txInterval * time.Duration(randRange(75, maxPct)) / 100
 }
 
 func (p *bfdPeer) tx() {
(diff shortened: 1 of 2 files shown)


fujita:
Merged, thanks!

PR #3584, opened 2026-09-02
server: apply TCP-AO to peer sockets

rastislavs:
Part of TCP-AO implementation (https://github.com/osrg/gobgp/issues/3493) this PR actually applies configured TCP-AO keys to peer sockets and makes TCP-AO work end-to-end.

Dynamic peers are not yet supported, will be covered in a later PR.


pkg/server/server.go
```diff
@@ -341,6 +341,19 @@ func (s *BgpServer) passConnToPeer(conn net.Conn) {
...
 			return
 		}
+		if keyBinding := peer.fsm.tcpAoKeyBinding.Load(); keyBinding != nil {
+			tcpAoKeys, err := keyBinding.socketKeys()
+			if err != nil {
+				peer.fsm.logger.Warn("could not load TCP-AO keychain", slog.String("Error", err.Error()))
+				conn.Close()
+				return
+			}
+			if err := setTcpAoConnectionPreferredKey(conn, tcpAoKeys); err != nil {
```
fujita: Should we use `setTcpAoConnectionRNext()` instead?  We should not update
`current`.  The peer's next segment moves it back anyway.
I think that we can remove `setTcpAoConnectionPreferredKey()`.

pkg/server/fsm.go
```diff
@@ -450,6 +464,16 @@ type fsm struct {
 	adminStateCh             chan adminStateOperation
 	outgoingConnCh           chan outgoingConn
 
+	// tcpAoKeyBinding holds resolved TCP-AO configuration for the peer:
+	// - Passive peers have all configured keys installed on matching listening sockets;
+	//   preferred key is not set on the listener until an inbound connection is accepted and matched to the peer.
+	// - Active peers install all keys in net.Dialer.Control in connectLoop() and mark the preferred key during installation.
+	// - Async keychain update causes update of keys directly on the listening sockets and fsm conn sockets of the affected peers.
+	//   This is handled on best effort; keychain update is not rolled back upon individual soket operation errors.
```
fujita: socket?

pkg/server/peer.go
```diff
@@ -93,7 +93,7 @@ func newDynamicPeer(g *oc.Global, neighborAddress string, pg *oc.PeerGroup, loc
 		return nil
```
fujita, line 99: If `oc.PeerGroup` has AO configurations, `newDynamicPeer` should fail?

pkg/server/server.go
```diff
@@ -5501,3 +5614,125 @@ func (s *BgpServer) ListTcpAoKeychain(ctx context.Context, r *api.ListTcpAoKeych
...
+}
+
+func (s *BgpServer) tcpAoKeyUsed(keychainName string, sendID uint8) bool {
+	for _, group := range s.peerGroupMap {
+		config := &group.Conf.TcpAo.Config
+		if config.Keychain != "" && string(config.Keychain) == keychainName && config.SendId == sendID {
+			return true
+		}
+	}
+	for _, peer := range s.neighborMap {
```
fujita, line 5648: `tcpAoKeyUsed()` can return false for a key that is in use. It only looks at `preferredSendID`, but the peer chooses our `Current_key` through RNextKeyID, and it can change that at any time?

Doing this properly needs care about races and partial updates, so I think key deletion can be left to a later PR.

fujita:
How about not supporting key deletion while a peer uses the keychain? There is no safe way to tell whether a key is in use. `Current_key` is chosen by the peer through RNextKeyID and can change at any time, so any check has a window.

Removing a whole keychain is a different story. `tcpAoKeychainUsed()` only reads our own state and the kernel never changes it. A keychain that no peer references has no keys on any socket, because `deleteNeighbor()` removed them.
That is why `DeleteTcpAoKeychain()` does not need to touch sockets. The same holds for a single key while no peer uses the keychain.

With that, the delete path in `updateTcpAoKeychainSockets()` is not needed at all, and neither is -EBUSY handling or a per peer replacement key. Adding keys to a used keychain still works, so key rotation still works.

The cost is that a key can only be removed after the peer stops using the keychain. That seems fine if key management lives outside gobgp.

pkg/server/server.go
rastislavs: I think we kind of already discussed this here: https://github.com/osrg/gobgp/issues/3493#issuecomment-5003536283

I implemented the discussed best-effort behavior: first perform config-level validation, then update the keychain as the source of truth, attempt to apply the deletion to live sockets, and let users inspect the peer state API to verify which MKTs remain installed. 

If a socket deletion returns EBUSY, the removed key may remain installed on that connection until it closes. This remains visible through the peer state API, allowing the user to react—for example, by resetting the peer.

I renamed `tcpAoKeyUsed()` to `tcpAoKeyConfigured()` to clarify that this is a configuration check and does not attempt to determine the kernel’s current key.

Preventing all key deletion while any peer references the keychain would be too restrictive operationally IMO. A keychain can contain at most 256 unique send IDs, and expired / invalid keys should be removable without first detaching the keychain or tearing down every affected peer.

fujita: Oops, you are right. This is what I asked for in that issue. The current behavior is more consistent.

fujita:
Merged, thanks!

PR #3588, opened 2026-09-05
server: apply AS path option updates without resetting sessions

### LINKED ISSUE #3582: GoBGP hard-resets the session upon AS path option update
The `NeedsResendOpenMessage` [returns `true` upon the AsPathOptions update](https://github.com/osrg/gobgp/blob/01c5c4c27f9a1ac3b5927f433b5115f9b0eee791/pkg/config/oc/util.go#L225) such as AllowOwnASN. This is in theory unnecessary, but maybe accidentally introduced in https://github.com/osrg/gobgp/commit/2e05695f683430bdf39573b19f51bac4dcbbc693 long ago.

sillage-hjj:
Updating `AllowOwnAs`, `ReplacePeerAs`, or `AllowAsPathLoopLocal` currently makes `NeedsResendOpenMessage` request a session restart. These options can be applied by reprocessing routes, so updating them unnecessarily interrupts established sessions.

Remove AS path options from the OPEN-message restart check. Report `NeedsSoftResetIn` for `AllowOwnAs` changes, reevaluate cached rejection state during inbound soft reset, and withdraw paths that are no longer accepted. Refresh outbound advertisements when `ReplacePeerAs` or `AllowAsPathLoopLocal` changes. Preserve peer-group inheritance and explicit member overrides, and update cached rejection state without growing path-parent history.

### Testing

Passed on Linux with Go 1.27.1 and `CGO_ENABLED=1`:

- `go test -v -race -count=1 -p=8 -timeout 240s ./...`
- `GOARCH=386 go test -v -count=1 -p=8 -timeout 240s ./...`
- `golangci-lint run` (v2.13.2)

The UNIX-socket tests used short `TMPDIR` and `GOTMPDIR` paths. Thirty CI scenarios completed successfully across local and native Linux runs; upstream skips were unchanged.

Closes #3582

### DIFF
--- a/internal/pkg/table/adj.go
+++ b/internal/pkg/table/adj.go
@@ -18,6 +18,7 @@ package table
 import (
 	"fmt"
 	"log/slog"
+	"slices"
 
 	"github.com/osrg/gobgp/v4/pkg/packet/bgp"
 )
@@ -38,6 +39,22 @@ func NewAdjRib(logger *slog.Logger, rfList []bgp.Family) *AdjRib {
 		accepted: make(map[bgp.Family]int),
 		logger:   logger,
 	}
+}
+
+// SetRejected replaces a cached path without retaining another layer of path
+// history. Attribute slices are copied so later changes cannot alter old views.
+func (adj *AdjRib) SetRejected(path *Path, rejected bool) *Path {
+	if path.IsRejected() == rejected {
+		return path
+	}
+	updated := path.Clone(false)
+	updated.parent = path.parent
+	updated.info = path.info
+	updated.pathAttrs = slices.Clone(path.pathAttrs)
+	updated.dels = slices.Clone(path.dels)
+	updated.SetRejected(rejected)
+	adj.Update([]*Path{updated})
+	return updated
 }
 
 func (adj *AdjRib) Update(pathList []*Path) {
--- a/pkg/config/oc/default.go
+++ b/pkg/config/oc/default.go
@@ -86,14 +86,13 @@ func setDefaultNeighborConfigValuesWithViper(v *viper.Viper, n *Neighbor, g *Glo
 	n.State.PeerType = n.Config.PeerType
 	if n.Config.PeerType == PEER_TYPE_EXTERNAL {
 		n.State.RemovePrivateAs = n.Config.RemovePrivateAs
-		n.AsPathOptions.State.ReplacePeerAs = n.AsPathOptions.Config.ReplacePeerAs
 	} else {
 		if string(n.Config.RemovePrivateAs) != "" {
 			return fmt.Errorf("can't set remove-private-as for iBGP peer")
 		}
-		if n.AsPathOptions.Config.ReplacePeerAs {
-			return fmt.Errorf("can't set replace-peer-as for iBGP peer")
-		}
+	}
+	if err := SetNeighborAsPathOptions(n, nil); err != nil {
+		return err
 	}
 
 	if !n.State.NeighborAddress.IsValid() {
@@ -101,8 +100,6 @@ func setDefaultNeighborConfigValuesWithViper(v *viper.Viper, n *Neighbor, g *Glo
 	}
 
 	n.State.PeerAs = n.Config.PeerAs
-	n.AsPathOptions.State.AllowOwnAs = n.AsPathOptions.Config.AllowOwnAs
-	n.AsPathOptions.State.AllowAsPathLoopLocal = n.AsPathOptions.Config.AllowAsPathLoopLocal
 
 	if !v.IsSet("neighbor.error-handling.config.treat-as-withdraw") {
 		n.ErrorHandling.Config.TreatAsWithdraw = true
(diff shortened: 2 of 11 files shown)


pkg/server/peer.go
```diff
@@ -657,6 +657,35 @@ func (peer *peer) updatePrefixLimitConfig(conf *oc.Neighbor, c []oc.AfiSafi) (bo
 	return reachLimit, nil
 }
 
+// isPathRejected applies the same loop checks to received and cached routes.
```
fujita, line 660: Master has this now, as peer.detectsLoop. Please drop this one and call that instead.

internal/pkg/table/adj.go
```diff
@@ -40,6 +41,22 @@ func NewAdjRib(logger *slog.Logger, rfList []bgp.Family) *AdjRib {
 	}
 }
 
+// SetRejected replaces a cached path without retaining another layer of path
+// history. Attribute slices are copied so later changes cannot alter old views.
+func (adj *AdjRib) SetRejected(path *Path, rejected bool) *Path {
```
fujita: Please return what Update gives you rather than dropping it. It now returns a withdrawal for every path that went from accepted to rejected, marked dropped so that the table releases the local path ID.

```
func (adj *AdjRib) SetRejected(path *Path, rejected bool) (*Path, []*Path) {
	...
	return updated, adj.Update([]*Path{updated})
}
```

pkg/config/oc/default.go
```diff
@@ -525,6 +522,23 @@ func setDefaultConfigValuesWithViper(v *viper.Viper, b *BgpConfigSet) error {
 	return nil
 }
 
+// SetNeighborAsPathOptions applies AS path option inheritance and state without
+// defaulting unrelated fields. The neighbor's peer type must already be set.
+func SetNeighborAsPathOptions(n *Neighbor, pg *PeerGroup) error {
```
fujita: This can go away too. It works around updatePeerGroup not re-inheriting, which is fixed on master. The iBGP replace-peer-as check you moved here is still in setDefaultNeighborConfigValuesWithViper, and updatePeerGroup runs it for every member now.

pkg/server/server.go
```diff
@@ -3978,6 +3997,10 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
 	if err == nil {
 		peer.fsm.pConf.Update(&conf)
 		peer.fsm.lock.Unlock()
+		if pgConf != nil {
+			// Retain current explicit member values for subsequent group updates.
+			s.peerGroupMap[conf.Config.PeerGroup].AddMember(conf)
```
fujita: Please delete this line and the SetNeighborAsPathOptions call above. peerGroup.members and AddMember are gone on master. Each peer keeps what the caller passed in peer.configuredConf, and updatePeerGroup merges that against the group.

pkg/server/server.go
```diff
@@ -3990,6 +4013,9 @@ func (s *BgpServer) updateNeighbor(c *oc.Neighbor) (needsSoftResetIn bool, err e
 				err = s.setAdminState(addr, "", adminStatePfxCt)
 			}
 		}
+		if err == nil && needsSoftResetOut {
+			err = s.softResetOut(addr, bgp.Family(0), false)
```
fujita, line 4017: Please delete this and report it instead. A configuration change that needs route processing to be re-run is reported through NeedsSoftResetIn and left to the client, an apply-policy change included. This should not be the one place that acts on its own.

pkg/server/server.go
```diff
@@ -2841,7 +2841,19 @@ func (s *BgpServer) softResetIn(addr string, family bgp.Family) error {
 		return err
 	}
 	for _, peer := range peers {
-		s.propagateUpdate(peer, peer.adjRibIn.PathList(familiesForSoftreset(peer, family), true))
+		paths := peer.adjRibIn.PathList(familiesForSoftreset(peer, family), false)
+		pathList := make([]*table.Path, 0, len(paths))
+		for _, path := range paths {
+			rejected := peer.isPathRejected(path)
```
fujita, line 2847: Call peer.detectsLoop(path, g) here, and build g once per peer with peer.loopCheckGlobals(s.rrClusterIDs) outside this loop. It reads gConf under fsm.lock, and keeping it outside is what stops the checks from taking that lock once per path.

Use the withdrawal SetRejected now returns instead of the path.Clone(true) below. Yours is not marked dropped, so the local path ID is never released and an emptied destination is left behind.

And withdraw only when the rejected state actually changes. The case the comment describes cannot happen on master any more: such an UPDATE now withdraws the installed route itself (31c8da1b). As it stands, every soft reset in emits a post-policy BMP withdrawal for routes that were never announced.

pkg/server/peer.go
```diff
@@ -657,6 +657,35 @@ func (peer *peer) updatePrefixLimitConfig(conf *oc.Neighbor, c []oc.AfiSafi) (bo
...
+func (peer *peer) isPathRejected(path *table.Path) bool {
```
fujita, line 661: Please drop this and call peer.detectsLoop instead. Master has it now, and it covers the CLUSTER_LIST check this one predates.

fujita (changes requested):
Sorry for the slow review.

Please rebase first. Master has moved a lot since this was written, and some of it takes work out of the patch: AdjRib.Update now reports the accepted-to-rejected transitions itself (31c8da1b), updatePeerGroup now re-merges every member against the group (05ba7350), and the three loop checks now live in peer.detectsLoop (c4d9e776).

fujita:
Thanks for the update. Everything is addressed except one comment:

https://github.com/osrg/gobgp/pull/3588#discussion_r4080918833

Please delete that code and I will merge.

sillage-hjj:
Thanks for the pointers. I've removed the redundant handling that you mentioned. Please take another look whenever you get a chance. Thanks again!

fujita:
Merged, thanks.

PR #3589, opened 2026-09-10
api: add multi-topology IDs to link descriptor API

watal:
Expose Multi-Topology IDs from link descriptors in the API, as defined in RFC 7752 Section 3.2.1.5.

Tests:

```bash
go test -count=1 ./pkg/apiutil/...
ok  	github.com/osrg/gobgp/v4/pkg/apiutil	0.572s
```

### DIFF
--- a/pkg/apiutil/attribute.go
+++ b/pkg/apiutil/attribute.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"net/netip"
+	"sort"
 
 	"github.com/osrg/gobgp/v4/api"
 	"github.com/osrg/gobgp/v4/pkg/packet/bgp"
@@ -704,6 +705,7 @@ func MarshalLsLinkDescriptor(n *bgp.LsLinkDescriptor) (*api.LsLinkDescriptor, er
 		NeighborAddrIpv4:  ipOrDefault(n.NeighborAddrIPv4),
 		InterfaceAddrIpv6: ipOrDefault(n.InterfaceAddrIPv6),
 		NeighborAddrIpv6:  ipOrDefault(n.NeighborAddrIPv6),
+		MultiTopoId:       marshalLsLinkMultiTopoIDs(n.MultiTopoIDs),
 	}, nil
 }
 
@@ -950,7 +952,30 @@ func UnmarshalLsLinkDescriptor(ld *api.LsLinkDescriptor) (*bgp.LsLinkDescriptor,
 		return nil, err
 	}
 
+	if multiTopoIDs := ld.GetMultiTopoId().GetMultiTopoIds(); len(multiTopoIDs) != 0 {
+		desc.MultiTopoIDs = make(map[uint16]struct{}, len(multiTopoIDs))
+		for _, id := range multiTopoIDs {
+			desc.MultiTopoIDs[uint16(id)] = struct{}{}
+		}
+	}
+
 	return desc, nil
+}
+
+func marshalLsLinkMultiTopoIDs(ids map[uint16]struct{}) *api.LsMultiTopologyIdentifier {
+	if len(ids) == 0 {
+		return nil
+	}
+
+	multiTopoIDs := make([]uint32, 0, len(ids))
+	for id := range ids {
+		multiTopoIDs = append(multiTopoIDs, uint32(id))
+	}
+
+	// Sort for deterministic output.
+	sort.Slice(multiTopoIDs, func(i, j int) bool { return multiTopoIDs[i] < multiTopoIDs[j] })
+
+	return &api.LsMultiTopologyIdentifier{MultiTopoIds: multiTopoIDs}
 }
 
 // parseLsLinkAddr parses one optional link descriptor address. An empty string
--- a/pkg/apiutil/attribute_test.go
+++ b/pkg/apiutil/attribute_test.go
@@ -2379,6 +2379,60 @@ func Test_MarshalLsLinkDescriptorLinkIDPresence(t *testing.T) {
 	assert.NotNil(t, present.LinkRemoteId)
 }
 
+func Test_MarshalLsLinkDescriptorMultiTopoID(t *testing.T) {
+	absent, err := MarshalLsLinkDescriptor(&bgp.LsLinkDescriptor{})
+	require.NoError(t, err)
+	assert.Nil(t, absent.MultiTopoId)
+
+	present, err := MarshalLsLinkDescriptor(&bgp.LsLinkDescriptor{
+		MultiTopoIDs: map[uint16]struct{}{3: {}, 1: {}, 2: {}},
+	})
+	require.NoError(t, err)
+	require.NotNil(t, present.MultiTopoId)
+	assert.Equal(t, []uint32{1, 2, 3}, present.MultiTopoId.GetMultiTopoIds())
+
+	boundary, err := MarshalLsLinkDescriptor(&bgp.LsLinkDescriptor{
+		MultiTopoIDs: map[uint16]struct{}{0: {}, 65535: {}},
+	})
+	require.NoError(t, err)
+	require.NotNil(t, boundary.MultiTopoId)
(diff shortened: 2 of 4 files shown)


pkg/apiutil/attribute.go
```diff
@@ -950,9 +952,32 @@ func UnmarshalLsLinkDescriptor(ld *api.LsLinkDescriptor) (*bgp.LsLinkDescriptor,
 		return nil, err
 	}
 
+	if multiTopoIDs := ld.GetMultiTopoId().GetMultiTopoIds(); len(multiTopoIDs) != 0 {
+		desc.MultiTopoIDs = make(map[uint16]struct{}, len(multiTopoIDs))
+		for _, id := range multiTopoIDs {
+			desc.MultiTopoIDs[uint16(id)] = struct{}{}
```
fujita: The MT-ID field is 12 bits wide. We should return an error if `id` > 0xfff?

proto/api/nlri.proto
```diff
@@ -262,6 +262,7 @@ message LsLinkDescriptor {
   string neighbor_addr_ipv4 = 4;
   string interface_addr_ipv6 = 5;
   string neighbor_addr_ipv6 = 6;
+  LsMultiTopologyIdentifier multi_topo_id = 7;
```
fujita: RFC 7752, Section 3.2.1.5 allows only one MT-ID in a Link Descriptor. Why `repeated` is necessary here?
MT-ID 0 is the standard topology, so absence must be distinguishable from zero.

`optional uint32 multi_topo_id = 7;`

is appropriate?

pkg/apiutil/attribute.go
```diff
@@ -950,9 +952,32 @@ func UnmarshalLsLinkDescriptor(ld *api.LsLinkDescriptor) (*bgp.LsLinkDescriptor,
...
+		return nil
+	}
+
+	multiTopoIDs := make([]uint32, 0, len(ids))
+	for id := range ids {
+		multiTopoIDs = append(multiTopoIDs, uint32(id))
+	}
+
+	// Sort for deterministic output.
+	sort.Slice(multiTopoIDs, func(i, j int) bool { return multiTopoIDs[i] < multiTopoIDs[j] })
```
fujita: `slices.Sort(multiTopoIDs)` doesn't work?

fujita (changes requested):

pkg/apiutil/attribute.go
watal: I've added a boundary check for the 12-bit MT-ID range.

pkg/apiutil/attribute.go
watal: Yes, slices.Sort works here. I've switched to it.

proto/api/nlri.proto
watal: You're right. 
A Link Descriptor can contain only a single MT-ID, so I've changed it to `optional uint32 multi_topo_id = 7;`.

pkg/apiutil/attribute_test.go
```diff
@@ -2470,6 +2470,76 @@ func Test_MarshalLsLinkDescriptorLinkIDPresence(t *testing.T) {
...
+	})
+	require.Error(t, err)
+	assert.Nil(t, outOfRange)
+}
+
+func Test_LsLinkDescriptorMultiTopoIDRoundTrip(t *testing.T) {
+	native := &bgp.LsLinkDescriptor{MultiTopoIDs: map[uint16]struct{}{1: {}, 2: {}, 3: {}}}
+
+	marshalled, err := MarshalLsLinkDescriptor(native)
+	require.NoError(t, err)
```
fujita: This failed. It needs to be updated.

proto/api/nlri.proto
```diff
@@ -269,6 +269,7 @@ message LsLinkDescriptor {
   string neighbor_addr_ipv4 = 4;
   string interface_addr_ipv6 = 5;
   string neighbor_addr_ipv6 = 6;
+  optional uint32 multi_topo_id = 7;
```
fujita, line 272: Sorry, I think that I was wrong. RFC 9552, Section 8.2.2 says that a Link-State NLRI must not be considered malformed based on the contents of its TLVs. GoBGP must accept a message with an NLRI including multiple Multi_topo_id TLVs even if it's incorrect. The GoBGP API should handle such an NLRI, so it should use a repeated type as the first version does.

Please also add a comment here. A Link Descriptor can carry only one MT-ID, so the reason for the repeated type is not obvious.

pkg/apiutil/attribute.go
```diff
@@ -699,13 +699,19 @@ func MarshalLsLinkDescriptor(n *bgp.LsLinkDescriptor) (*api.LsLinkDescriptor, er
 	// Both identifiers keep explicit presence: 0 is a valid Link Remote
 	// Identifier meaning "unknown" (RFC 5307, Section 1.1), so flattening an
 	// absent identifier to 0 would fabricate a Link Local/Remote Identifiers TLV.
+	multiTopoID, err := marshalLsLinkMultiTopoID(n.MultiTopoIDs)
+	if err != nil {
+		return nil, err
+	}
+
```
fujita, line 706: MarshalLsLinkDescriptor must not fail here. The NLRI is not malformed, so GoBGP accepts it, keeps it in the RIB and forwards it as it is. Marshal reports what is in the RIB, so it must be able to report this route too.

An error in this direction is for what GoBGP cannot represent, such as the unknown NLRI type in MarshalNLRI. It is not for what a peer did wrong. That check belongs in UnmarshalLsLinkDescriptor.

pkg/apiutil/attribute.go
```diff
@@ -699,13 +699,19 @@ func MarshalLsLinkDescriptor(n *bgp.LsLinkDescriptor) (*api.LsLinkDescriptor, er
 	// Both identifiers keep explicit presence: 0 is a valid Link Remote
```
fujita, line 699: This comment is about link_local_id and link_remote_id in the return statement below? Please move it back so that it sits next to them.

pkg/apiutil/attribute.go
```diff
@@ -982,9 +988,38 @@ func UnmarshalLsLinkDescriptor(ld *api.LsLinkDescriptor) (*bgp.LsLinkDescriptor,
 		return nil, err
 	}
 
+	if ld != nil && ld.MultiTopoId != nil {
+		id := ld.GetMultiTopoId()
```
fujita, line 992: Checking the 0x0FFF range is fine. A value above 4095 can only come from an API request, because LsTLVMultiTopoID.DecodeFromBytes masks it. But please do not reject more than one MT-ID here. This function is also used by the CLI:
makeShowRouteArgs calls GetNativeNlri, which reaches UnmarshalLsLinkDescriptor
because the CLI does not set EnableNlriBinary. An error here would print "?" again for the same route.

The "only one MT-ID in a Link Descriptor" rule belongs in NewLsLinkTLVs, which does not build the MT-ID TLV yet.

pkg/apiutil/attribute.go
```diff
@@ -982,9 +988,38 @@ func UnmarshalLsLinkDescriptor(ld *api.LsLinkDescriptor) (*bgp.LsLinkDescriptor,
...
+		}
+		desc.MultiTopoIDs = map[uint16]struct{}{
+			uint16(id): {},
+		}
+	}
+
 	return desc, nil
 }
 
+func marshalLsLinkMultiTopoID(ids map[uint16]struct{}) (*uint32, error) {
```
fujita, line 1004: The range check is not needed here. MultiTopoIDs is set either by LsLinkDescriptor.ParseTLVs, where LsTLVMultiTopoID.DecodeFromBytes has already masked the value with 0x0FFF, or by UnmarshalLsLinkDescriptor, which checks the range. This function also needs to accept more than one MT-ID.

After those changes it becomes very small, and it is used only from MarshalLsLinkDescriptor. You could drop it and inline the code.

Please use slices.Sort for the deterministic output. multiTopoIDsToString in pkg/packet/bgp/bgp.go does the same conversion from a map to a sorted slice.

fujita (changes requested):

PR #3591, opened 2026-09-10
api: add multiple adjacency-SID support to link descriptor API

watal:
Expose all Adjacency-SID TLVs from link descriptors in the API, preserving their Flags and Weight fields per RFC 8667 Section 2.2.1.

> An SR-capable router MAY allocate more than one Adj-SID to an adjacency.

Retain `sr_adjacency_sid` for backward compatibility.

```bash
go test -count=1 ./pkg/apiutil/...
ok  	github.com/osrg/gobgp/v4/pkg/apiutil	0.018s
go test -count=1 ./pkg/packet/bgp/...
ok  	github.com/osrg/gobgp/v4/pkg/packet/bgp	0.031s
```


fujita:
The title and the body say "link descriptor". However, it should be "link attribute"?

fujita:
Fixed the commit message and pushed, thanks.

PR #3592, opened 2026-09-10
metrics: fix the peer type values in the bgp_peer_type help text

fdomain:
The help text and inline comment documented internal as 0 and external as 1, but the metric reports the api.PeerType protobuf enum, where internal is 1 and external is 2. Unspecified (0) is never emitted: toPeerType maps every non-external value to internal.


fujita:
Merged, thanks.

PR #3593, opened 2026-09-10
config/oc: keep the neighbor config set through the API

fdomain:
OverwriteNeighborConfigWithPeerGroup decides which neighbor fields to keep by asking viper which keys the operator wrote. That record lives in configuredFields, which is only filled while parsing a configuration file, so a neighbor added through the gRPC API has no entry: every key reports as unset and the peer group overwrote everything the caller passed.

Fall back to treating a non-zero value as explicitly set when the neighbor has no config-file record. Neighbors read from a file keep the viper behavior, so a key configured to its zero value still wins over the peer group.


pkg/config/oc/default.go
```diff
@@ -528,29 +528,32 @@ func setDefaultConfigValuesWithViper(v *viper.Viper, b *BgpConfigSet) error {
...
+	overwriteConfig(&c.AsPathOptions.Config, &pg.AsPathOptions.Config, "neighbor.as-path-options.config", v, configured)
+	overwriteConfig(&c.AddPaths.Config, &pg.AddPaths.Config, "neighbor.add-paths.config", v, configured)
+	overwriteConfig(&c.GracefulRestart.Config, &pg.GracefulRestart.Config, "neighbor.gradeful-restart.config", v, configured)
+	overwriteConfig(&c.ApplyPolicy.Config, &pg.ApplyPolicy.Config, "neighbor.apply-policy.config", v, configured)
+	overwriteConfig(&c.UseMultiplePaths.Config, &pg.UseMultiplePaths.Config, "neighbor.use-multiple-paths.config", v, configured)
+	overwriteConfig(&c.RouteServer.Config, &pg.RouteServer.Config, "neighbor.route-server.config", v, configured)
+	overwriteConfig(&c.TtlSecurity.Config, &pg.TtlSecurity.Config, "neighbor.ttl-security.config", v, configured)
+	overwriteConfig(&c.Bfd.Config, &pg.Bfd.Config, "neighbor.bfd.config", v, configured)
 
 	if !v.IsSet("neighbor.afi-safis") {
```
fujita: Needs the similar change here too?

```
if configured {
    if !v.IsSet("neighbor.afi-safis") {
        c.AfiSafis = append([]AfiSafi{}, pg.AfiSafis...)
    }
} else if len(c.AfiSafis) == 0 {
    c.AfiSafis = append([]AfiSafi{}, pg.AfiSafis...)
}
```

fujita (changes requested):

pkg/config/oc/default.go
fdomain: indeed, added

fujita:
Merged, thanks.

PR #3594, opened 2026-09-10
config/oc: report the peer description in the peer state

fdomain:
NewPeerFromConfigStruct filled api.PeerConf.Description but left api.PeerState.Description empty, so a client reading the state of a peer never saw the description, however the peer had been configured. SetDefaultNeighborConfigValues already mirrors the value into State.Description; only the conversion to the API dropped it.


fujita:
Thanks.

PR #3595, opened 2026-09-10
Add the SR Policy Candidate Path NLRI

aronjohanns:
This PR adds support for SR policy candidate path NLRI. The implementation was verified against production network with Cisco ASR routers.

Assisted by claude fable 5.1


aronjohanns:
I can split this PR into a series of smaller PRs, one per logical change. It would end up as:

  1. `packet/bgp`: set the length of the IPv6 Local Router-ID TLV
  2. `apiutil`: return an error for an unhandled BGP-LS NLRI type
  3. `packet/bgp`: decode the Local Router-ID sub-TLVs in node descriptors (+ `api`)
  4. `packet/bgp`: add the BGP-LS Protocol-IDs for RSVP-TE and Segment Routing
  5. `packet/bgp`: add the SR Policy Candidate Path NLRI (RFC 9857)
  6. `packet/bgp`: add the SR Policy Binding SID, state and name TLVs
  7. `packet/bgp`: add the SR Segment List TLV and its sub-TLVs
  8. `packet/bgp`: add the SR Candidate Path Constraints TLV
  9. `api`: add the SR Policy Candidate Path NLRI and attribute
  10. `cmd/gobgp`: add `global rib add -a ls srpolicy`
  11. `test`: add a BGP-LS scenario test for SR Policy candidate paths

  The first four are small standalone fixes, but the rest depend on each other in that order. Would you prefer I close this PR and open them one at a time instead?

fujita:
Yes, please close this PR and open them one at a time.

Two rules for the split:

- Keep fixes and new features in separate PRs.
- Put at most three commits in one PR, and group the commits that belong
  together.

That gives four PRs:

1. Fixes: your 1 and 2. Both are standalone and do not depend on RFC 9857.
2. BGP-LS core: your 4 and 5, with 3 folded into 5. RFC 9552 does not list
   TLV 1028 and 1029 as node descriptor sub-TLVs. Only RFC 9857 Section 3
   does, and only for NLRI type 5. So the decoding must be limited to
   NLRI type 5, and then it is part of 5.
3. Attribute TLVs: your 6, 7 and 8.
4. API, CLI and test: your 9, 10 and 11.

The last three PRs depend on each other, so please post each one after the
previous one is merged.

PR #3598, opened 2026-09-14
server: reject local cluster loops on import

rastislavs:
Check the CLUSTER_LIST against all configured neighbor cluster IDs before importing a received path into the Loc-RIB. This prevents looped reflected routes from participating in best-path selection.

RFC4456 says: "If the local CLUSTER_ID is found in the CLUSTER_LIST, the advertisement received SHOULD be ignored" As GoBGP keeps CLUSTER_ID per-neighbor and there's no concept of single "local CLUSTER_ID", all configured neighbor cluster IDs are considered.

--

This fixes an issue with stale paths when two route reflectors originate the same prefix and reflect routes to one another. 

After they then both withdraw the local path, they can end up in a state with stale reflected routes on both RRs, even when none of them is advertising it anymore. See the attached reproducer:  [rr-self-reflect-repro.sh](https://github.com/user-attachments/files/31736767/rr-self-reflect-repro.sh)

--

**NOTE**: At the moment, we check for cluster loops, but on the sender side, see https://github.com/osrg/gobgp/blob/a1136eedbdeb384e4ff35736c9c32c3c20a960a3/pkg/server/server.go#L516-L527

This however is not the behavior that RFC describes, and has some flaws, like the one found with the reproducer script.
I believe we can remove this sender side check after merging the one proposed here, even though having both does not hurt.


pkg/server/server.go
```diff
@@ -1279,6 +1279,13 @@ func (s *BgpServer) propagateUpdate(peer *peer, pathList []*table.Path) {
 	}
 
 	for _, path := range pathList {
+		// RFC4456 requires a receiver to ignore a route whose CLUSTER_LIST contains a local cluster ID.
+		// GoBGP stores cluster IDs per neighbor, so check all configured IDs before importing the route into the Loc-RIB.
+		if s.hasLocalClusterLoop(path) {
+			s.logger.Debug("cluster list path attribute has a local cluster id, ignore", slog.Any("Path", path))
+			continue
+		}
+
```
fujita: The check should not be in `propagateUpdate`. That is too late. By the time `propagateUpdate` runs, `peer.handleUpdate` has already called `peer.adjRibIn.Update(pathList)`, so the path is in the Adj-RIB-In and is counted as accepted. Skipping the import here leaves the path accepted in the Adj-RIB-In but absent from the Loc-RIB. path. `SetRejected(true)` has to be called before adjRibIn.Update, so it has to happen in handleUpdate.

`peer.handleUpdate` already has the ORIGINATOR_ID check from the same RFC 4456 section 8:

```
// RFC4456 8. Avoiding Routing Information Loops
// A router that recognizes the ORIGINATOR_ID attribute SHOULD
// ignore a route received with its BGP Identifier as the ORIGINATOR_ID.
if path.GetOriginatorID() == routerId {
        ...
        path.SetRejected(true)
        continue
}
```
Please add the CLUSTER_LIST check next to it and call path.SetRejected(true) in the same way. Then the path stays in the Adj-RIB-In, gobgp neighbor <addr> adj-in still shows it, and the accepted counter remains correct.

Please also skip the check when the peer is a route server client.

For the cluster ID list, I think it is better to keep it in BgpServer and pass it to `handleUpdate` as an argument. `handleFSMMessage` is the only caller. Rebuild the list when a peer is added or deleted and when the configuration is updated. Reads happen under shared.mu.RLock() and writes under shared.mu.Lock(), so a plain field is enough. Build the list once before the loop over the path list, not per path.

pkg/server/server.go
```diff
@@ -1343,6 +1350,31 @@ func (s *BgpServer) propagateUpdate(peer *peer, pathList []*table.Path) {
...
+	}
+
+	clusterList := path.GetClusterList()
+	if len(clusterList) == 0 {
+		return false
+	}
+
+	for _, neighbor := range s.neighborMap {
+		conf := neighbor.fsm.pConf.ReadOnly()
+		clusterID := conf.RouteReflector.Config.RouteReflectorClusterId
```
fujita: This has to be State.RouteReflectorClusterId, not Config. A reflected route carries the State value: UpdatePathAttrs prepends info.RouteReflectorClusterID, which comes from State. Config only holds the value the operator set, and State falls back to the global router ID when it is not set. So with route-reflector-client = true and no route-reflector-cluster-id, which is the common case, Config is invalid and this check does nothing.

The test sets Config directly instead of going through oc.SetDefaultNeighborConfigValues(), so it does not catch this. The reproducer script sets the cluster ID, so it does not catch it either.

fujita (changes requested):

pkg/server/server.go
rastislavs: Thanks for the pointers, done

fujita:
Fixed the commit message and merged, thanks!

We use the cluster IDs of all route-reflector-client neighbors, not `all configured neighbor cluster IDs`.

PR #3601, opened 2026-09-14
bfd: pace transmission by RFC 5880's transmit interval

### LINKED ISSUE #3563: BFD: gobgp ignores the peer's Required Min RX Interval and transmits faster than the peer allows

### LINKED ISSUE #3564: BFD: gobgp does not slow transmission to one second while the session is down, as RFC 5880 Section 6.8.3 requires

notsrch:
Fixes #3563. Fixes #3564.

Builds on #3600 (jitter): the per-packet jitter is applied to the transmit interval this PR
computes.

### What was wrong

RFC 5880 Section 6.8.7 defines one transmit interval: the larger of `bfd.DesiredMinTxInterval`
and `bfd.RemoteMinRxInterval`. gobgp had neither term right.

- Section 6.8.3 requires `bfd.DesiredMinTxInterval` to be at least one second while the session
  is not Up. gobgp advertised and transmitted at the configured interval in every state, so a
  session that never came up kept sending at the full configured rate (#3564 has the deployment
  numbers).
- gobgp parsed the peer's `Required Min RX Interval` and never used it. The transmit ticker was
  armed once at construction and never re-armed, so a peer asking gobgp to slow down was ignored
  (#3563).

The two issues were filed separately, and we first fixed them on separate branches. They turned
out to rewrite the same few lines, the code that decides when the next packet goes out, and
each fix on its own implemented half of the Section 6.8.7 formula while explaining why the other
half was missing. One PR lets you review the scheduling model once.

### What this changes

`bfd.DesiredMinTxInterval` is now its own session variable, separate from the configured
interval. While the session is not Up it is the larger of the configured interval and one
second; reaching Up restores the configured value. It is what outgoing packets advertise. Every
accepted packet stores the peer's advertised value as `bfd.RemoteMinRxInterval` (Section 6.8.1,
initialized to 1 µs). The transmit interval is the larger of the two.

One function, `armTx`, does all scheduling. It keeps the transmit ticker armed at an absolute
deadline, the last transmission plus the transmit interval, and transmits at once if that
deadline has already passed. The deadline anchoring is deliberate, not a style choice. Simpler
designs fail against a misbehaving peer, and each failure was measured through the production
receive path during review:

- Re-arming a full period on every received packet lets any peer that transmits faster than the
  transmit interval postpone our next transmission forever (`time.Ticker.Reset` restarts the
  period). Zero transmissions in five seconds against a 20 packets/s stream.
- Guarding the re-arm ("only when the value changed") dies the same way against a peer that
  alternates between two advertised values above our interval.
- Re-arming an already-overdue deadline from the current instant plus a small epsilon, without
  sending, re-creates the same starvation above ~1 kHz, because the current instant is not an
  anchor. Sending the overdue packet at once does not have that problem: the send itself is the
  anchor. This is also what Section 6.8.3 requires once a reduced interval has already elapsed
  ("the local system MUST send the next periodic BFD Control packet as soon as practicable").

The per-packet jitter from #3600 is applied to this transmit interval, not to the
configured value: Section 6.8.7 words it as "the larger of bfd.DesiredMinTxInterval and
bfd.RemoteMinRxInterval, less applied jitter". `armTx` draws a fresh jitter on every re-arm, so
under a stream of received packets the deadline wanders inside the 75%-100% window and never
leaves it.

Section 6.8.3 also requires a Poll Sequence whenever the advertised `bfd.DesiredMinTxInterval`
changes, and gobgp had no Poll Sequence state: it sent one Poll packet on reaching Up and
discarded the Final that answered it. A change now sets the Poll bit on the scheduled periodic
packets (Section 6.5) until a packet with the Final bit set arrives (Section 6.8.6). The Final
check runs before the state switch in `rxPacket`, so a packet that both answers our poll and
signals Down terminates the old sequence and lets the Down transition start a new one. A change
to the peer's advertised value alters only the rate gobgp sends at, never what it advertises, so
it starts no Poll Sequence.

Interval changes take effect immediately. Section 6.8.3 defers an increase only when the
session is Up, and the only increase here happens on the transition out of Up.

### Visible behavior changes

- A session that is Down or Init transmits at one-second pace, so establishment takes up to a
  second per handshake step when the configured interval is faster. Failure detection on an Up
  session is unchanged. For a default-configured peer (one-second interval) nothing changes
  here.
- A peer advertising a `Required Min RX Interval` larger than gobgp's configured interval is
  sent packets at the peer's rate. The advertised interval does not change.
- The packet sent on reaching Up is no longer an extra transmission outside the schedule. It
  goes out at once when the restored interval has already elapsed since the last packet, which
  after a one-second Down cadence it nearly always has, and otherwise at the next deadline.
  Section 6.8.7 makes no exception to its rate limit for state changes. It carries the Poll bit
  only when the interval actually changed; with the default one-second interval nothing
  changes, so that bit is now clear where it was always set before (and the answering Final was
  discarded).
- The RFC puts no timeout on a Poll Sequence, so a peer that never answers with Final leaves
  the Poll bit set on every periodic packet until one arrives. Whether the sequence on the way
  down is wanted at all is the question #3564 leaves open; the RFC text has no exemption for
  non-Up states, so this PR follows it literally. Skipping it there is a one-line guard in
  `setStateDown` if you would rather not.

### What does not change

- The detection-time path (Section 6.8.4), which already followed the remote's timers.
- A zero `Required Min RX Interval` is stored verbatim; Section 6.8.7's meaning for zero (stop
  periodic transmission entirely) stays unimplemented, as before.
- The stored `bfd.RemoteMinRxInterval` survives a session going down, which the RFC prescribes
  (nothing resets it between session creations).

### Checks from the issues

- Not-Up floor and its restore (#3564): `Test_NewBfdPeerDesiredMinTxFloor`,
  `Test_RxPacketUpRestoresConfiguredIntervalAndStartsPollSequence`,
  `Test_RxPacketLeavingUpRefloorsAndStartsNewPollSequence`,
  `Test_RxPacketUpWithSlowConfigStartsNoPollSequence`. On the wire: `Test_TxPacketSlowWhileDown`
  reads real packets from a Down peer configured for 200 ms and checks the advertised 1 s and the
  gap; `Test_TxPacketPollBitAndFastRestoreOnUp` drives a session Up over a real socket and
  checks the Poll bit and the restored 200 ms cadence.
- Jitter on the right interval: `Test_JitteredTxInterval` (from #3600) now draws from a
  peer whose `desiredMinTx` is set and whose configured value is not, so it fails if the jitter
  is applied to the configured value again.
- Poll Sequence (#3564): `Test_RxPacketFinalTerminatesPollSequence`,
  `Test_RxPacketFinalThenDownOrdering`, `Test_PollReplyNeverSetsPollAndFinalTogether` (Section
  6.5's "MUST NOT have both bits set").
- Pacing by the peer (#3563): `Test_RxPacketTxPacingSlowsTransmission` feeds an advert through
  the production receive path and asserts a true quiet window (nothing sent inside the peer's
  new floor, the actual MUST) followed by the slowed cadence.
  `Test_RxPacketTxPacingReducedIntervalSendsPromptly` pins Section 6.8.3's "as soon as
  practicable" for a reduction that has already elapsed.
  `Test_RxPacketTxPacingStableUnderFastRxStream` alternates adverts every 50 ms;
  `Test_RxPacketTxPacingSurvivesRxFlood` does the same unthrottled for three seconds to pin the
  overdue path.
- Both terms at once: `Test_EffectiveTxIntervalAcrossStates` walks Down, Init, Up and back with
  a peer whose requirement sits between the configured rate and the floor, and checks at every
  step that what we send at follows the formula while what we advertise does not move with the
  peer's value. Neither issue's branch could express this test alone.
- The one path to Down that bypasses the receive path: `Test_ExpiryRefloorsTxPacing` lets an Up
  session expire and checks that the ticker is re-armed for the floor at once, rather than
  firing one more packet at the Up cadence.

The pacing tests bound cadence rather than exact timing, and the bounds allow for the 25%
jitter: quiet windows are shorter than the jittered minimum gap, and the flood and fast-stream
upper bounds use 300 ms (75% of the 400 ms requirement) as the cadence floor. They were
hardened by mutation: single-line mutants of the scheduling seam (overdue branch leaves the
timer alone, `tx()` does not advance the anchor, `setDesiredMinTx` does not re-arm, the remote
term dropped from the interval, `rxPacket` does not re-arm, jitter applied to the configured
value) each fail at least one test and none survives. `Test_StateUpDown` and `Test_ResetPeer` wait for Up
with a deadline instead of a fixed sleep, because establishment is now paced, and then wait for
one more packet before stopping the far end: the far end restores its 200 ms advertisement on
reaching Up, but the near end learns it, and with it the shorter detection time, only from the
first packet sent in that state.

`go test -race ./pkg/server/` passes. `docs/sources/bfd.md` gains both behaviors under
"Supported behavior" and its limitations bullet now describes what is and is not negotiated.

### Relationship to #3600

#3600 (jitter) introduced the timer this PR re-arms and the `jitteredTxInterval` helper this PR
points at the transmit interval. This PR is based on master with #3600 merged.


pkg/server/bfd_peer.go
```diff
@@ -532,8 +663,14 @@ func (p *bfdPeer) setStateUp(yourDiscriminator uint32) {
...
 	p.yourDiscriminator = yourDiscriminator
 
-	p.eventExpiry.Reset(p.expiryInterval)
+	// RFC 5880 Section 6.8.3: Up lifts the not-Up floor, restoring the configured
+	// interval — a change that starts a Poll Sequence unless it's a no-op. The first
+	// Up packet is not sent off-schedule here: armTx (through setDesiredMinTx) sends
+	// it at once when the shorter interval has already elapsed since the last packet,
+	// which after a one-second Down cadence it nearly always has, and otherwise waits
+	// out the remainder. Section 6.8.7's "MUST NOT transmit at an interval less than"
+	// has no exception for state changes.
```
fujita: There are exceptions in 6.8.7.

A BFD Control packet SHOULD be transmitted during the interval between periodic Control packet transmissions when the contents of that packet would differ from that in the previously transmitted packet (other than the Poll and Final bits) in order to more rapidly communicate a change in state.

?

pkg/server/bfd_peer.go
```diff
@@ -510,6 +632,11 @@ func (p *bfdPeer) setStateDown() {
 	p.state.Store(int32(api.BfdSessionState_BFD_SESSION_STATE_DOWN))
 	p.yourDiscriminator = 0
 
+	// RFC 5880 Section 6.8.3: "When bfd.SessionState is not Up, the system MUST set
+	// bfd.DesiredMinTxInterval to a value of not less than one second (1,000,000
+	// microseconds)."
+	p.setDesiredMinTx(max(p.txInterval, bfdSlowTxInterval))
```
fujita, line 693: It means sending a down packet with P bit set and zero your discriminator?
That should be avoided?

pkg/server/bfd_peer.go
```diff
@@ -285,6 +321,46 @@ func (p *bfdPeer) startClient() {
...
+// lastTx + jitteredTxInterval, or transmits at once if that deadline has
+// already passed. Every change to either term of the interval goes through here.
+// The jitter is drawn afresh on each call, so under a stream of received packets
+// the deadline wanders within the 75%-100% window Section 6.8.7 allows and never
+// leaves it.
+//
+// Anchoring on the last transmission rather than on the present instant is what
+// keeps the pacing safe against a peer that floods packets or flaps its
+// advertised value: repeated calls converge on the same deadline instead of
+// pushing it out, so the deadline moves only when a requirement changes, never
```
fujita: They do not converge. Each call draws a new random value, so a send happens at
the smallest draw seen since the last transmission. Under a stream of received
packets the mean gap falls toward the 75% floor. So the rate we send at depends
on the rate the peer sends at, which is what the jitter is there to prevent.

The fix is for armTx not to draw at all. Either keep the percentage drawn in
tx() and reuse it here, or redraw only when effectiveTxInterval() has changed
and return early otherwise. BIRD does the latter: it re-arms only on a change
(bfd_session_process_ctl) and at a fixed 75% (tm_set(s->tx_timer, s->last_tx +
tx_int_l)).

pkg/server/bfd_peer.go
```diff
@@ -56,6 +62,24 @@ type bfdPeer struct {
...
+	// peer's advertised Required Min RX Interval, the other term of the transmit
+	// interval. Same confinement as desiredMinTx.
+	remoteMinRxInterval time.Duration
+	// lastTx is when tx() last ran (peer creation time until the first run). It
+	// anchors the pacing deadline; see armTx. Same confinement as desiredMinTx.
+	lastTx time.Time
+	// pollSequence is true while a Poll Sequence is in flight: the P bit is set on
+	// periodic transmissions until a Final is received. Same confinement as desiredMinTx.
+	pollSequence bool
+
```
fujita, line 99: pollSequence is one bool with no sequence identity, so a Final that answers an
old Poll can terminate a newer one. The guard in setDesiredMinTx removes the
case this PR creates, because no new sequence then starts on the packet that
ended the old one. Please note the limitation here: it matters again once
RequiredMinRxInterval can change at runtime, which RFC 5880 6.8.3 also requires
a Poll Sequence for.

fujita (changes requested):

pkg/server/bfd_peer.go
notsrch: You're right, I had that backwards: the SHOULD in 6.8.7 is exactly the exception. Every state transition now transmits at once through `tx()`, so the packet is also the new pacing anchor and the periodic schedule restarts from it (no extra packet on top). Down→Init and Up→Down get this too, not only Up. `Test_StateChangeTransmitsAtOnce` pins it.

pkg/server/bfd_peer.go
notsrch: Yes, and agreed it should be avoided. A Poll Sequence now starts only while we hold a remote discriminator, and leaving Up ends any sequence in flight, so a Down packet never carries P next to a zero Your Discriminator. Same rule as BIRD's `rem_id == 0` check in `bfd_session_request_poll`. `Test_RxPacketLeavingUpRefloorsWithoutPollSequence` pins it.

pkg/server/bfd_peer.go
notsrch: Right, with a fresh draw per call the send lands at the smallest draw since the last transmission. The jitter is now drawn once per transmission in `tx()` and reused; `armTx` draws nothing, so the deadline is a pure function of `lastTx`, that draw and the current interval, and repeated calls converge for real. I took the reuse variant rather than return-early-on-no-change because it keeps `armTx` stateless about the previous interval. `Test_JitteredTxIntervalFixedBetweenDraws` pins it.

pkg/server/bfd_peer.go
notsrch: Noted on the field, with the two 6.8.3 cases where it would bite (a reduced RequiredMinRxInterval keeping the old Detection Time, an increased DesiredMinTxInterval while Up keeping the old transmit interval) and what a runtime change would need (sequence identity, or waiting for the in-flight sequence to end). The Final-ordering comment is reworded too: with no sequence starting on Down, the ordering matters for a Final riding on the Init→Up packet, and `Test_RxPacketFinalThenUpOrdering` replaces the Down variant.

fujita:
Merged, thanks.

PR #3603, opened 2026-09-15
tcp-ao: delete keys from listening sockets asynchronously

fujita:
The kernel documentation recommends it for a listening socket that holds many keys:

  A tip here for scaled TCP_LISTEN sockets, that may have some thousands
  TCP-AO keys, is: use filters in getsockopt(TCP_AO_GET_KEYS) and
  asynchronous delete with setsockopt(TCP_AO_DEL_KEY).

del_async is only valid on a listening socket. The kernel returns EINVAL on any other socket, so the caller has to pass it.

Assisted-by: Claude Opus 5 (1M context) <noreply@anthropic.com>


fujita:
@rastislavs

PR #3604, opened 2026-09-15
zebra: only encode nexthop label_num with MessageLabel before frr7.3

Dev-next-gen:
I was reading `Nexthop.encode` and `Nexthop.decode` in `pkg/zebra/zapi.go` side by side and noticed their label conditions don't group the same way. Decode reads `label_num` when `flags&zapiNexthopFlagLabel > 0 || message&MessageLabel > 0 && (version == 5 || version == 6 && frr 6 <= v < 7.3)`. Encode lost the inner parentheses in 157a7689e, so it evaluates `(message&MessageLabel > 0 && version == 5) || (version == 6 && frr 6 <= v < 7.3)`. For frr6, frr7 and frr7.2 that means every nexthop gets a `label_num` octet, even on routes without `MessageLabel`.

FRR 7.2 only puts and gets `label_num` under `CHECK_FLAG(api->message, ZAPI_MESSAGE_LABEL)` (lib/zclient.c, `zapi_route_encode` and `zapi_route_decode`). A plain unlabeled route from GoBGP therefore has one stray zero byte per nexthop, and zebra reads distance, metric and MTU shifted. Our own decoder rejects our own output. Before the fix, the added round-trip test fails for all three versions:

```
--- FAIL: Test_IPRouteBody_SerializeWithoutLabelsFrr6To7dot2/frr6
    Received unexpected error: message length invalid (last) pos:27 rest:28, message:0x17
--- FAIL: Test_IPRouteBody_SerializeWithoutLabelsFrr6To7dot2/frr7
    Received unexpected error: message length invalid (last) pos:27 rest:28, message:0x17
--- FAIL: Test_IPRouteBody_SerializeWithoutLabelsFrr6To7dot2/frr7.2
    Received unexpected error: message length invalid (last) pos:28 rest:29, message:0x17
```

With the fix it passes, and `go test ./pkg/zebra/` stays green. The change only restores the grouping, so encode matches decode again. `encode` is unexported. Nothing changes for zapi5 or for frr 7.3 and newer, including the default frr8.2. Routes that carry the label nexthop flag are still encoded exactly as before. The existing tests only build `NewSoftware(v, "")`, which is why the older frr variants never caught this.


pkg/zebra/zapi_test.go
```diff
@@ -824,6 +824,46 @@ func Test_IPRouteBody_IPv6(t *testing.T) {
 }
 
 // NexthopLookup exists in only quagga (zebra API version 2 and 3)
```
fujita: This puts the new test between the "NexthopLookup exists in only quagga"
comment and `Test_nexthopLookupBody`. The comment belongs to that function.
Please move the new test above the comment.

fujita (changes requested):

pkg/zebra/zapi_test.go
Dev-next-gen: You're right, the comment had ended up above the wrong function. It's back above `Test_nexthopLookupBody` in b0297ea8, with the new test placed before it. `gofmt -l pkg/zebra` is clean and `go test ./pkg/zebra/` still passes, including the three frr6/frr7/frr7.2 subtests.

fujita:
Squashed and merged, thanks.

PR #3606, opened 2026-09-15
fixes: correct BGP-LS NLRI marshalling and IPv6 Router-ID TLV length

aronjohanns:
Two small BGP-LS fixes split out of the work in #3595.


pkg/packet/bgp/bgp.go
```diff
@@ -7000,7 +7000,7 @@ func NewLsTLVLocalIPv6RouterID(l *netip.Addr) *LsTLVLocalIPv6RouterID {
 	return &LsTLVLocalIPv6RouterID{
 		LsTLV: LsTLV{
 			Type:   LS_TLV_IPV6_LOCAL_ROUTER_ID,
-			Length: 0,
+			Length: 16,
```
fujita, line 7003: `NewLsTLVRemoteIPv6RouterID` and  `NewLsTLVOpaquePrefixAttr`, and `NewLsTLVPrefixSID` have the same bug (LsTLV's wrong length). Please fix them together.

pkg/packet/bgp/bgp_test.go
```diff
@@ -4669,6 +4669,13 @@ func Test_LsNodeDescriptor(t *testing.T) {
 	}
 }
 
+func Test_LsTLVLocalRouterIDLength(t *testing.T) {
+	v4 := netip.MustParseAddr("10.0.0.1")
+	v6 := netip.MustParseAddr("2001:db8::1")
+	assert.EqualValues(t, 4, NewLsTLVLocalIPv4RouterID(&v4).Len()-tlvHdrLen)
+	assert.EqualValues(t, 16, NewLsTLVLocalIPv6RouterID(&v6).Len()-tlvHdrLen)
+}
```
fujita, line 4701: Better to test serializing `LsTLVLocalIPv6RouterID`?

```
b, err := NewLsTLVLocalIPv6RouterID(&v6).Serialize()
assert.NoError(t, err)
assert.Len(t, b, 20)
```

fujita (changes requested):

fujita:
The commit message says "Serialize writes the real value length, and the two
callers in NewLsAttributeTLVs only serialize, so the wire encoding is right
today." That is not right. LsTLV.Serialize does not write the value length;
it checks it:

	if len(value) != int(l.Length) {
		return nil, malformedAttrListErr("serialization failed: LS TLV malformed")
	}

So with Length 0 and a 16-octet address, Serialize fails. On master a BGP-LS
attribute that carries a Node IPv6 Local Router-ID cannot be serialized at
all:

	local:    Len()=4 err=serialization failed: LS TLV malformed
	pathattr: err=serialization failed: LS TLV malformed

The fix is right, but the bug is worse than the message says.

pkg/packet/bgp/bgp.go
aronjohanns: Fixed all three in the same commit and retitled it since it now covers all four constructors.

pkg/packet/bgp/bgp_test.go
aronjohanns: Done, replaced Len() checks with Serialize() for all four RouterID TLVs plus PrefixSID and OpaquePrefixAttr.

aronjohanns:
> The commit message says "Serialize writes the real value length, and the two callers in NewLsAttributeTLVs only serialize, so the wire encoding is right today." That is not right. LsTLV.Serialize does not write the value length; it checks it:
> 
> ```
> if len(value) != int(l.Length) {
> 	return nil, malformedAttrListErr("serialization failed: LS TLV malformed")
> }
> ```
> 
> So with Length 0 and a 16-octet address, Serialize fails. On master a BGP-LS attribute that carries a Node IPv6 Local Router-ID cannot be serialized at all:
> 
> ```
> local:    Len()=4 err=serialization failed: LS TLV malformed
> pathattr: err=serialization failed: LS TLV malformed
> ```
> 
> The fix is right, but the bug is worse than the message says.

Ok, I rewrote the commit message to state that

fujita:
Merged, thanks.

PR #3609, opened 2026-09-16
packet/mrt: write AFI/SAFI only for the RIB_GENERIC families

Dev-next-gen:
I was round-tripping MRT table dumps through `Rib.Serialize` and `parseRib` and found that only IPv4 unicast survives it. Everything else comes back either short or unparseable.

`Rib.Serialize` decides whether an entry starts with AFI and SAFI. RFC 6396 Section 4.3.2 gives `RIB_IPV4_UNICAST`, `RIB_IPV4_MULTICAST`, `RIB_IPV6_UNICAST` and `RIB_IPV6_MULTICAST` their own subtypes, whose entries carry the prefix directly; only `RIB_GENERIC` names the family on the wire. `parseRib` follows that rule: it reads AFI and SAFI only when `ParseBody` handed it the zero `Family`, which happens for `RIB_GENERIC` and `RIB_GENERIC_ADDPATH` alone.

f9997962 turned that switch inside out. The branch that listed the four dedicated families and did nothing became the branch that writes AFI and SAFI, and `RF_IPv4_UC` in the list became `RF_FS_IPv4_UC`. Both halves went wrong at once: `RF_IPv4_MC`, `RF_IPv6_UC` and `RF_IPv6_MC` now get three octets they must not have, and every family that genuinely is dumped as `RIB_GENERIC`, flowspec IPv4 unicast excepted, no longer gets them.

`pkg/server/mrt.go` picks the subtype from the same four families, so this reaches a plain table dump. With `gobgp mrt dump rib global` on an IPv6 unicast table, each entry goes out with three extra octets and reading the file back fails with `not all RibEntry bytes are available`. A VPNv4 entry goes out as `RIB_GENERIC` without its AFI and SAFI, and the reader takes the first three octets of the route distinguisher for them: `unknown route family. AFI: 28672, SAFI: 6`.

The fix restores the original polarity and keys the list on the four families that have a subtype of their own.

This does change what `Rib.Serialize` puts on the wire, so I want to be explicit about why I think it is safe rather than breaking. The output it changes is output that GoBGP itself cannot read back and that does not match RFC 6396; the output it leaves alone is IPv4 unicast, the one case the inverted switch still got right. A file written by an affected version stays unreadable either way, because the framing in it is wrong, not because the reader changed — `parseRib` is untouched here. No exported signature or field changes.

`TestMrtRib` only exercises `RF_IPv4_UC`, which is why the inversion went unnoticed, so I added `TestMrtRibFamilyHeaderMatchesSubtype`. It round-trips a message through `NewMRTMessage`, `ParseHeader` and `ParseBody` for the four dedicated subtypes and for a VPNv4 prefix under `RIB_GENERIC`. On master the IPv4 unicast case passes and the other four fail; with the patch all five pass. `go test ./...` is green.


pkg/packet/mrt/mrt_test.go
```diff
@@ -279,6 +279,80 @@ func TestMrtRibWithAddPath(t *testing.T) {
 	assert.Equal(t, reflect.DeepEqual(r1, r2), true)
 }
 
+func TestMrtRibFamilyHeaderMatchesSubtype(t *testing.T) {
+	// RFC 6396 4.3.2 puts AFI and SAFI at the start of a RIB_GENERIC entry
```
fujita: 4.3.3?

fujita (changes requested):

pkg/packet/mrt/mrt_test.go
Dev-next-gen: You're right, thanks for catching it: the AFI/SAFI entry header belongs to Section 4.3.3 (RIB_GENERIC Subtype), while 4.3.2 is the AFI/SAFI-Specific RIB Subtypes section. I fixed the test comment in 9dc5554 so it cites 4.3.3 for RIB_GENERIC and 4.3.2 for the four dedicated subtypes. The comment in `Rib.Serialize` already cites 4.3.2 for exactly those four families, so I left it as it was.

The change only touches the comment; `go test ./pkg/packet/mrt/ -run TestMrtRib -v` still passes all five subtests of `TestMrtRibFamilyHeaderMatchesSubtype` (ipv4-unicast, ipv4-multicast, ipv6-unicast, ipv6-multicast, generic-vpnv4), and `gofmt -l pkg/packet/mrt` prints nothing.

Dev-next-gen:
You're right, it's 4.3.3. Section 4.3.2 covers the AFI/SAFI-specific subtypes, the four families whose entries carry the prefix directly; it's 4.3.3, RIB_GENERIC Subtype, that puts AFI and SAFI at the head of the entry. The citation in `mrt.go` was pointing at the right section for those four families, the one in the test was not.

Fixed in 9dc55543 — the comment only, no change to the code or to the test.

fujita:
Squashed and merged, thanks!

PR #3613, opened 2026-09-16
netutils: do not panic on a non-IPv4 MD5 address on OpenBSD

Dev-next-gen:
On OpenBSD, `SetTCPMD5SigSockopt` installs the MD5 key through PF_KEY, and `rfkeyRequest` builds the `sockaddr_in` for each end with `newSockaddrIn`. That function does `net.ParseIP(addr).To4()` and indexes the result without checking it, so any address that is not a plain IPv4 address panics with `index out of range [3] with length 0`.

The server passes it two kinds of such addresses: an IPv6 neighbor with `auth-password` (through `addAuthKeysToListeners`), and a dynamic neighbor whose peer group has a password, where the prefix string such as `10.0.0.0/24` is passed as the address. Either one crashes gobgpd, even though every caller already logs a "failed to set md5" warning when an error comes back. With this change, `newSockaddrIn` returns an error when the address has no IPv4 form, and `rfkeyRequest` returns it before anything is written to the PF_KEY socket. `newSockaddrIn` is unexported and lives under `internal/`, so changing its signature affects nothing outside this file, and the empty-address case works as before.

CI only cross-builds openbsd, and I have no OpenBSD machine, so here is how I checked it. The new test is `//go:build openbsd`. `GOOS=openbsd go vet` and `GOOS=openbsd go test -c` both succeed. To run it, I copied `sockopt_openbsd.go` and the test into a scratch package on Linux under neutral file names with the build tag removed, next to the other untagged netutils files. The code under test is plain Go and fails before it touches the socket. With the current `sockopt_openbsd.go`, `TestSaAddRejectsNonIPv4Address` panics in `newSockaddrIn` (called from `rfkeyRequest` via `saAdd`). With the patched file, the test passes. `GOOS=openbsd`, `freebsd` and `linux` `go build ./...` are also clean.


fujita:
Thanks

PR #3614, opened 2026-09-16
packet/bgp: add the SR Policy Candidate Path NLRI

aronjohanns:
BGP-LS NLRI type 5 with the Protocol-ID values it needs. The headend descriptor decodes TLV 1028 and 1029 for this NLRI type only, as requested. The attribute TLVs follow in the next PR.


pkg/packet/bgp/ls_sr_policy.go
```diff
@@ -0,0 +1,407 @@
...
+				continue
+			}
+		case LS_TLV_SR_POLICY_CP_DESC:
+			if l.CandidatePathDesc == nil {
+				tlv = &LsTLVSrPolicyCandidatePathDescriptor{}
+			}
+		}
+
+		// Unknown TLVs and repeated instances of a TLV (RFC 9857: the first
+		// valid one is used) are skipped.
```
fujita: RFC 9552 5.1 says:

Unknown and unsupported types MUST be preserved and propagated within both the NLRI and the BGP-LS Attribute.

pkg/packet/bgp/bgp.go
```diff
@@ -10545,7 +10594,13 @@ func (l *LsTLVNodeDescriptor) DecodeFromBytes(data []byte) error {
 	_, lsTLVBgpRouterIDExists := m[LS_TLV_BGP_ROUTER_ID]
 	_, lsTLVAutonomousSystemExists := m[LS_TLV_AS]
 
-	if !lsTLVIgpRouterIDExists && (!lsTLVBgpRouterIDExists || !lsTLVAutonomousSystemExists) {
+	// RFC 9857 Section 3 identifies the headend by TLV 1028 or 1029.
+	// A PCE need not know the headend's IGP or BGP identifiers.
+	if srPolicy {
+		if !m[LS_TLV_IPV4_LOCAL_ROUTER_ID] && !m[LS_TLV_IPV6_LOCAL_ROUTER_ID] {
```
fujita: RFC 9552 8.2.2 says:

A Link-State NLRI MUST NOT be considered malformed or invalid based on the inclusion/exclusion of TLVs or contents of the TLV fields.

fujita (changes requested):

pkg/packet/bgp/bgp.go
aronjohanns: Fixed. Removed the RouterID check on the headend descriptor.

pkg/packet/bgp/ls_sr_policy.go
aronjohanns: Fixed. PathAttributeLs had the same problem, fixed that in 9d6d4d6.

fujita:
Merged, thanks.

PR #2816, opened 2024-06-12
Support `SetRouteOrigin` and `OriginEq` BGP Policy Predicates

wenovus:
This expands upon existing policy predicate support.

PR #3616, opened 2026-09-16
table, server: report the origin-eq condition from OriginEq

Dev-next-gen:
I was comparing the two copies of `toStatementApi` and noticed that the origin-eq condition does not survive being listed back through the API.

In `internal/pkg/table/policy.go`, the function checks `Conditions.BgpConditions.OriginEq`, but the switch that fills `Conditions.Origin` reads `Actions.BgpActions.SetRouteOrigin`. This has been the case since the origin predicates were added in #2816. `ListPolicy` goes through this copy, so it reports the set-route-origin action's origin as the condition. If the statement has no such action, the condition comes back unset. The copy in `pkg/server/grpc_server.go`, which `ListStatement` uses, has no conversion for the origin condition at all.

I added `TestListPolicyOriginCondition`, which adds a policy with two statements: one matches EGP and sets IGP, the other only matches INCOMPLETE. The test then reads both statements back through `ListPolicy` and `ListStatement`. Before the change, it fails on all four checks:

```
expected: 2 (EGP)         actual: 1 (IGP)   ListPolicy, statement egp-to-igp
expected: 3 (INCOMPLETE)  actual: 0         ListPolicy, statement incomplete
expected: 2               actual: 0         ListStatement, statement egp-to-igp
expected: 3               actual: 0         ListStatement, statement incomplete
```

With the change, it passes, and `go test ./pkg/server/ ./internal/pkg/table/ ./pkg/config/...` and `go vet` on the two touched packages pass too.

The fix doesn't change any function signature or API message. Only the value that `ListPolicy` and `ListStatement` put in `Conditions.Origin` changes, and it now matches what `AddPolicy`/`AddStatement` were given. That matters for anything that reads a policy back and adds it again: today, that round trip silently drops the origin condition or replaces it with the action's origin.


Dev-next-gen:
The zapi-v3 failure on this pull request does not come from the change. Rather than ask you to look into it, I ran the job here: `test/scenario_test/zapi_v3_test.py` against a `gobgp-oq` image built with `test/lib/make-gobgp-ctn.sh --tag gobgp-oq --from-image osrg/quagga:v1.0`, first at fb7e2fd7 and then at the base ee380df5, with nothing else changed between the two runs. Four tests passed both times, `test_02_vrf_routes_add` included.

That matches what the diff does. It only changes how an existing policy condition is reported over the API: `toStatementApi` in `internal/pkg/table/policy.go` switched on `Actions.BgpActions.SetRouteOrigin` where it meant `Conditions.BgpConditions.OriginEq`, and the copy in `pkg/server/grpc_server.go` did not report `OriginEq` at all. Neither path is involved in installing VRF routes over zapi.

What I cannot reproduce here is CI timing, and a netns scenario on a loaded runner is exactly where test isolation tends to bite. So if a re-run of that job comes back green, this comment is nothing you need to act on — I just thought you should have the result rather than the question.

fujita:
Thanks!

PR #3624, opened 2026-09-17
fix(table): ignore unmatched withdrawals to prevent RTC update loops

Ivan-Pokhabov:
Treat withdrawals that do not match an installed path by source and path ID as no-ops, and skip empty results in the table manager. Previously, these withdrawals produced updates with unchanged path lists, which RTC processing could propagate and cause update loops. Add regression tests for unmatched IPv4 and RTC withdrawals, preservation of matching withdrawal behavior, and suppression of outbound RTC updates when the RIB is unchanged.


fujita:
Nice, thanks!

PR #3628, opened 2026-09-17
feat(policy): add match-pg-set policy

Ivan-Pokhabov:


fujita:
Can you describe the use case?

I checked the other implementations. Cisco IOS, IOS-XR and Junos have no way to match a peer group name in a policy. BIRD has no peer groups at all. OpenConfig does not have this condition either. FRR is the only one with it, as "match peer WORD".

I would like to hear the problem you are solving before we add this.

Ivan-Pokhabov:
> Can you describe the use case?

Our use case is a controller-managed GoBGP deployment where regular, non-route-server peers share a global policy. Since per-peer or peer-group policy assignments are not applied to such peers, the global policy needs a way to distinguish their logical roles, such as customers, internal peers, or route reflectors.

Currently this requires duplicating peer-group membership in an IP-based neighbor-set. With dynamic peers or frequent peer updates, the controller must continuously synchronize that set as addresses appear, disappear, or change, creating two sources of truth and possible consistency gaps. Matching the peer group directly uses the classification already present in PeerInfo and keeps the policy independent of peer address lifecycle. We implement it as a GoBGP-specific extension because it is not part of OpenConfig.

fujita:
Thanks, that is clear.

So, if per-peer and per-peer-group policy assignment worked for regular peers, would you still need this condition?

I would rather add that instead. This patch adds a new defined-set type, a new condition and a YANG augment that only GoBGP has. The other way needs no new configuration at all. apply-policy already exists under neighbor and peer-group in OpenConfig, the GoBGP config structs already carry it, and the peer group apply-policy is already copied into each
member. Only the last step is missing. Junos, IOS-XR, IOS and FRR all let you attach an import or export policy to a peer group.

I would like to keep the configuration close to OpenConfig and add the general feature rather than a GoBGP-specific one. What do you think?

Ivan-Pokhabov:
> So, if per-peer and per-peer-group policy assignment worked for regular peers, would you still need this condition?

I would prefer to implement proper per-peer-group policy assignment instead of adding a peer-group match condition. However, I would not want to implement it merely by copying the peer-group assignment to every member and evaluating the same export policy separately for every peer.

For export, I would like the peer group to be an actual policy evaluation unit: evaluate the effective export policy and path transformations once, then reuse the result for all compatible members of the group. If a peer has different effective configuration or a per-peer-sensitive condition, it should transparently fall back to per-peer evaluation. Import policy can still use the assignment inherited from the peer group, since each received route already belongs to a single peer.

This is essentially the goal of #3161. We also extracted the receiver abstraction as #3423, but that design was not accepted because you preferred automatic grouping by effective policy rather than an explicit optimization mechanism.

This PR was intended to be the first change in a larger series. In the new design, a policy compiler determines the effective parts of the global policy for each peer and groups peers with identical compiled policy plans into a single evaluation unit. This should provide the automatic grouping you described while retaining per-peer evaluation where necessary. However, the complete implementation is fairly substantial, which is why I intended to split it into smaller, reviewable PRs.

Before I rework this PR, I think we should agree on the intended design and scope.
>  What do you think?

So I think we need to choose between two designs: should the peer group be an explicit policy assignment and sharing boundary, or should a dynamic policy compiler automatically group peers with identical effective policies, regardless of their configured peer groups? Which approach would you prefer?

PR #3630, opened 2026-09-17
cmd/gobgp: require protocol and identifier for BGP-LS routes

Dev-next-gen:
While going through the CLI route parsers after the I-PMSI route target fix (aac7dea4), I noticed that the four BGP-LS parsers in `cmd/gobgp/global.go` (`node`, `link`, `prefixv6`, `srv6sid`) index `m["protocol"][0]` and `m["identifier"][0]` straight away. `extractReserved` only checks the arity of the keywords that are present, so when one of them is left out the slice is nil and the command panics:

```
$ gobgp global rib add -a ls node identifier 1 local-asn 65000 local-bgp-ls-id 0
panic: runtime error: index out of range [0] with length 0
```

Both keywords are required by every syntax line in `docs/sources/bgp-ls.md`, so I added the same "specify %s" check the other route parsers already use, right after `extractReserved`. I also fixed the format comment above `parseLsSRv6SIDNLRIType`, which read `srv6sid bgp identifier ...` and left out `protocol <protocol>`; following it literally hit the same panic.

The new `Test_ParseLsArgsRequiresProtocolAndIdentifier` drops each keyword in turn for the four NLRI types. Before the change all 8 subtests fail with "should not panic"; after it they pass, and `go test ./cmd/gobgp/` and `go vet ./cmd/gobgp/` are clean.

This only changes what happens for commands that were already unusable: a command missing `protocol` or `identifier` now gets an error message instead of a crash. Commands that give both keywords, including all the examples in the docs, parse exactly as before, and nothing outside `cmd/gobgp` is touched.


fujita:
Merged, thanks.

PR #3631, opened 2026-09-17
server/mrt: record the peer AS in the TABLE_DUMPv2 peer index table

Dev-next-gen:
An MRT table dump starts with a PEER_INDEX_TABLE listing, for each peer a RIB entry can point at, its BGP identifier, its address and its AS number. I was reading a dump produced by gobgpd and every peer entry showed AS 0, so I went looking for where that number is filled in.

`dumpTable` used to build the list from the neighbor map and took the AS from `ocpeer.Config.PeerAs`. a2bc5a6d ("server/mrt: add only established-peers to dumptable entry") replaced that with a `peermap` filled in while the paths are walked, and the `dumpPeer` literal for a real peer sets `index`, `addr` and `id` but not `as`. The dummy entry for locally generated routes sets `as: 0` deliberately, which makes the omission easy to miss. So since that commit, every peer in the index table is written with AS 0 and a reader of the dump cannot attribute a RIB entry to a neighbor without guessing from the path attributes.

The value is already on the path: `PeerInfo.AS` is filled from `conf.State.PeerAs` when the session reaches established, which is the same value the old code read. The fix is that one field.

The test I added drives `dumpTable` over a one-path RIB whose source is a peer in AS 65001 and checks the three fields of the resulting peer entry. Before the fix it fails on the AS alone (`expected: 65001, actual: 0`); the address and the router ID were already correct. After it, the package builds and the test passes:

```
$ go test ./pkg/server/ -run TestMrtDumpTablePeerIndexTable -count=1
ok      github.com/osrg/gobgp/v4/pkg/server      0.008s
```

There was no `pkg/server/mrt_test.go`, so this adds one. I kept the fix and its test in a single commit since the test exists only to pin this behaviour; happy to split them if you would rather have two.


Dev-next-gen:
The single red check here, `protobuf lint and format check`, never reached a `.proto` file. The failure is in the toolchain install step, and the four steps that do the actual checking are recorded as skipped:

```
$ gh api repos/osrg/gobgp/actions/jobs/105384034525 --jq '[.steps[]|"\(.conclusion)\t\(.name)"]|.[]'
success	Set up job
success	Run actions/checkout@v6
success	Run actions/setup-go@v6
failure	Run go install github.com/bufbuild/buf/cmd/buf@latest
skipped	buf lint
skipped	buf format check
skipped	buf generate
skipped	buf generate check
skipped	Post Run actions/setup-go@v6
success	Post Run actions/checkout@v6
success	Complete job
```

The error inside that step is a transport failure against the checksum database, not a lint or format finding: `buf.build/gen/go/bufbuild/registry/connectrpc/go@v1.20.0-...: verifying module: reading https://sum.golang.org/tile/8/0/x240/307: stream error; stream ID 263; INTERNAL_ERROR; received from peer`. I want to be straight about the limit of that argument: I cannot point you at the same job red on master, because it is green there, and this is the only occurrence in the last 25 runs of the workflow, so I read it as a one-off rather than as something this branch causes. What I did verify is that the branch carries no proto change at all (`pkg/server/mrt.go` and `pkg/server/mrt_test.go`, 2 files changed, 71 insertions), that it sits directly on current master 2943c5ed, and that the added test still passes there in a clean container with go1.25.0: `--- PASS: TestMrtDumpTablePeerIndexTable (0.00s)`.

Re-running that one job would clear the board, and it is not something I can do from here (`gh run rerun 35275204283 --failed` answers `Must have admin rights to Repository`), so if one of you passes by with a spare moment it would save the next reader wondering what the red mark means.

fujita:
Merged, thanks.

PR #3640, opened 2026-09-18
config/oc: report the peer group in the peer state

fdomain:
NeighborState.PeerGroup was never populated, so api.PeerState.PeerGroup was always empty and a client reading the state of a peer never saw the group it belongs to. SetDefaultNeighborConfigValues now mirrors Config.PeerGroup into State.PeerGroup, as it already does for the description, and the conversion to the API reports it.

This also fixes the peer group filter of WatchUpdate and WatchPostUpdate: both compare Neighbor.State.PeerGroup, so a watch restricted to a peer group never matched an event.


fujita:
Thanks!

PR #3641, opened 2026-09-18
api: add the SR Policy Candidate Path NLRI and attribute

aronjohanns:
Expose the SR Policy candidate paths over gRPC, add "gobgp global rib
add -a ls srpolicy" to inject one, and add a scenario test that
advertises candidate paths between two gobgpd instances and checks
what arrives, attributes included.


cmd/gobgp/global.go
```diff
@@ -2820,6 +2821,548 @@ func parseLsSRv6SIDNLRIType(args []string) (bgp.NLRI, *bgp.PathAttributeLs, erro
...
+		metric := bgp.LsSrMetricConstraint{MetricType: uint8(metricType)}
+		if len(parts) > 1 {
+			set, err := lsParseFlagLetters("constraint-metric flags", parts[1], "OMAB")
+			if err != nil {
+				return nil, err
+			}
+			metric.Flags = bgp.LsSrMetricConstraintFlags{Optimization: set['O'], Margin: set['M'], Absolute: set['A'], Bound: set['B']}
+		}
+		values := []*uint32{&metric.Margin, &metric.Bound}
+		for i, s := range parts[2:] {
```
fujita, line 3084: `constraint-metric 2` (type only) panics: `strings.Split` returns a slice with cap 1, so `parts[2:]` is out of range. An error for `len(parts) < 2` looks right to me; the syntax above and in docs/sources/bgp-ls.md then needs to be `<type>:<flags>[:<margin>[:<bound>]]`.

pkg/apiutil/ls_sr_policy.go
```diff
@@ -0,0 +1,860 @@
...
+	if c.Algorithm > 0xff {
+		return nil, fmt.Errorf("invalid SR candidate path constraints algorithm %d", c.Algorithm)
+	}
+	n := &bgp.LsSrCandidatePathConstraints{
+		MTID:      uint16(c.Mtid),
+		Algorithm: uint8(c.Algorithm),
+		SRLGs:     slices.Clone(c.Srlgs),
+	}
+	if c.Flags != nil {
+		if c.Flags.ProtectedOnly && c.Flags.UnprotectedOnly {
```
fujita: The decoder accepts a TLV with both P and U set, as RFC 9552 section 8.2.2 requires. `GetNativePathAttributes` runs this code when the CLI displays a received route, so an error here drops every attribute of that path, not just this TLV. Please drop the check; the gobgp command already rejects `constraint-flags PU` when originating.

pkg/packet/bgp/ls_sr_policy.go
```diff
@@ -1472,6 +1472,52 @@ type LsSrSegment struct {
 	SIDStructure      *LsSrv6SIDStructure     `json:"sid_structure,omitempty"`
 }
 
+// Validate checks fields whose interpretation depends on the segment type,
+// so that an inconsistent segment taken in through the API is refused with
+// a specific error up front instead of a generic one at serialization.
+// Every value the wire decoder can produce passes, so a segment learned
+// from a peer always converts back through the API.
+func (s *LsSrSegment) Validate() error {
```
fujita: This is validation of API input, as the comment above says, so please make it a private function in apiutil. `UnmarshalLsSrSegment` is the only caller.

If you move it, please also fix the comment on `TestLsSrSegmentDecodedAlwaysReserializes` in ls_sr_policy_test.go: it says `LsSrSegment.Validate` would refuse an IPv4-mapped SID, but `!s.SID.Is6()` is false for one, so it does not.

pkg/apiutil/ls_sr_policy.go
```diff
@@ -0,0 +1,860 @@
...
+		}
+	} else if sid.IsValid() || specified.IsValid() {
+		return nil, fmt.Errorf("MPLS binding SID cannot carry IPv6 SIDs")
+	}
+	return n, nil
+}
+
+func validateLsSrv6Addrs(addrs ...netip.Addr) error {
+	for _, addr := range addrs {
+		if addr.IsValid() && (!addr.Is6() || addr.Is4In6() || addr.Zone() != "") {
```
fujita: Does any RFC prohibit an IPv4-mapped SRv6 SID? Even if one did, this code runs when the CLI displays a received route, so an error here drops every attribute of that path.

pkg/apiutil/attribute.go
```diff
@@ -909,6 +923,20 @@ func UnmarshalLsNodeDescriptor(nd *api.LsNodeDescriptor) (*bgp.LsNodeDescriptor,
...
+	}
+	if localRouterID.IsValid() && !localRouterID.Is4() {
+		return nil, fmt.Errorf("local_router_id_ipv4 must be an IPv4 address")
+	}
+	localRouterIDv6, err := lsParseAddr("local_router_id_ipv6", nd.LocalRouterIdIpv6)
+	if err != nil {
+		return nil, err
+	}
+	if localRouterIDv6.IsValid() && (!localRouterIDv6.Is6() || localRouterIDv6.Is4In6()) {
+		return nil, fmt.Errorf("local_router_id_ipv6 must be an IPv6 address")
```
fujita, line 940: Please drop the `Is4In6` term here too. TLV 1029 only checks that the value is 16 octets, so a peer can send a Router-ID that decodes to an IPv4-mapped address. This is on the NLRI side, so an error loses the whole route from the CLI output. `!Is6()` can stay, and the `!Is4()` check on `local_router_id_ipv4` above is fine because TLV 1028 is 4 octets.

proto/api/attribute.proto
```diff
@@ -572,12 +572,287 @@ message LsAttributeSrv6SID {
...
+  repeated LsSrMetricConstraint metrics = 9;
+}
+
+// BGP-LS attribute TLVs of an SR Policy Candidate Path NLRI (RFC 9857 Section 5).
+message LsAttributeSrPolicy {
+  LsSrBindingSID binding_sid = 1;
+  repeated LsSrv6BindingSID srv6_binding_sids = 2;
+  LsSrCandidatePathState state = 3;
+  string candidate_path_name = 4;
+  string policy_name = 5;
```
fujita: These two want `optional`. The native model holds them as `*string`, and a name TLV may be zero length (RFC 9857 sections 5.4 and 5.5 give no minimum), so an empty string here means both "absent" and "present and empty", and a zero-length TLV from a peer becomes an absent one after a round trip through the API. `Bandwidth` and `Identifier` already get presence from their wrapper messages, so only the names are missing it. The addresses do not need it because a valid address never renders as an empty string. Worth deciding now: adding `optional` later changes the generated field from `string` to `*string` and breaks every user of the api package. `LsLinkDescriptor` already uses `optional` in nlri.proto.

pkg/apiutil/ls_sr_policy.go
```diff
@@ -0,0 +1,860 @@
...
+		n.Bandwidth = &bw
+	}
+	if sl.Identifier != nil {
+		id := sl.Identifier.Identifier
+		n.Identifier = &id
+	}
+	// The TLV length is 2 octets; a long enough segment list can exceed it.
+	if wire, err := bgp.NewLsTLVSrSegmentList(n).Serialize(); err != nil {
+		return nil, err
+	} else if len(wire) > 4+0xffff {
```
fujita: `LsTLV.Serialize` checks `len(value) == int(l.Length)` before it builds the buffer, and `Length` is a uint16, so it never returns more than `4 + 0xffff` bytes. This branch is unreachable and can go; the error above already covers the case. The same applies at line 785.

fujita (changes requested):

cmd/gobgp/global.go
aronjohanns: Fixed. len(parts) < 2 errors now, and the syntax in the usage string and docs requires the flags.

pkg/apiutil/ls_sr_policy.go
aronjohanns: Dropped

pkg/packet/bgp/ls_sr_policy.go
aronjohanns: Moved to apiutil as a private function, and fixed the test comment.

pkg/apiutil/ls_sr_policy.go
aronjohanns: Dropped the check

pkg/apiutil/attribute.go
aronjohanns: Dropped

pkg/apiutil/ls_sr_policy.go
aronjohanns: Right, removed both.

proto/api/attribute.proto
aronjohanns: Done, both are optional now and a zero length TLV stays present through the API. I also added a test for it.

cmd/gobgp/global.go
```diff
@@ -2820,6 +2821,546 @@ func parseLsSRv6SIDNLRIType(args []string) (bgp.NLRI, *bgp.PathAttributeLs, erro
...
+			value, err := strconv.ParseUint(s, 10, 32)
+			if err != nil {
+				return nil, fmt.Errorf("invalid constraint-metric value: %v", err)
+			}
+			*values[i] = uint32(value)
+		}
+		c.Metrics = append(c.Metrics, metric)
+	}
```
fujita, line 3092: RFC 9857 section 5.6.6 says the O-flag "MUST NOT be set in more than one instance of this TLV for a given candidate path advertisement". Nothing checks that, so `constraint-metric 1:O 2:O` builds an advertisement that breaks it.

cmd/gobgp/global.go
```diff
@@ -2820,6 +2821,546 @@ func parseLsSRv6SIDNLRIType(args []string) (bgp.NLRI, *bgp.PathAttributeLs, erro
...
+	if !ok {
+		return nil, fmt.Errorf("invalid segment-list %q: expected <weight>:<sid>[,<sid>...]", arg)
+	}
+	weight, err := strconv.ParseUint(weightStr, 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("invalid segment-list weight %q: %v", weightStr, err)
+	}
+
+	sl := &bgp.LsSrSegmentList{
+		Flags:  bgp.LsSrSegmentListFlags{Explicit: true, Computed: true},
```
fujita: Please set the V-flag and the R-flag here and on the segment below. RFC 9857 sections 5.7 and 5.7.1 read a clear V-flag as "failed verification" and a clear R-flag as "failed resolution", so every path injected here is advertised as broken. For three of the four flags the RFC gives the reading this case needs, "its verification was not required" and "did not require resolution". The segment list R-flag has no such wording, but setting it is still better than claiming the first segment failed resolution. An option to clear them can wait until someone asks for it.

fujita (changes requested):

cmd/gobgp/global.go
aronjohanns: Added the check, only one metric may carry O now.

cmd/gobgp/global.go
aronjohanns: Done, both the list and the segments set V and R now.

docs/sources/bgp-ls.md
```diff
@@ -88,6 +103,21 @@ $ gobgp global rib -a ls
...
+
+# Show routes
+$ gobgp global rib -a ls
+   Network                                                                                                                                                        Next Hop             AS_PATH              Age        Attrs
+*> NLRI { SRPOLICY_CP { LOCAL_NODE: {ASN: 65001, BGP LS ID: 0, BGP ROUTER ID: 1.1.1.1, IPv4 ROUTER ID: 10.0.0.1} ENDPOINT: 10.0.0.2 COLOR: 100 ORIGIN: 3 ORIGINATOR: 65001/1.1.1.1 DISCRIMINATOR: 1 SR:0 } } 0.0.0.0                                   00:00:03   [{Origin: ?} {LsAttributes: {SR Binding SID: 24001 Specified: 0 Flags: B} {SR CP State: Priority:10 Preference:200 Flags:AEV} {SR CP Name: cp1} {SR Policy Name: blue} {SR Segment List: Weight:1 MTID:0 Algo:0 Flags:EC {Segment: Type:A Label:16002 Algo:0 Flags:SE} {Segment: Type:A Label:24006 Algo:0 Flags:SE}} {SR Segment List: Weight:2 MTID:0 Algo:0 Flags:EC {Segment: Type:A Label:16003 Algo:0 Flags:SE}} }]
```
fujita: The flags in this example are stale now that the V and R flags are set? Running the command above gives `Flags:ECVR` for the segment lists and `Flags:SEVR` for the segments.

fujita (changes requested):

docs/sources/bgp-ls.md
aronjohanns: Updated the example

fujita:
Merged, thanks.

aronjohanns:
> Merged, thanks.

Thank you for the careful reviews throughout the whole series of PRs for this feature.

PR #3642, opened 2026-09-18
Probe IPv6 link-local reachability with scoped pings

Ivan-Pokhabov:
Replace neighbor-cache REACHABLE checks with bounded, bidirectional scoped pings and shared polling. Dump diagnostics on failure with finite command timeouts, preserving the original error.


test/unit/test_link_local.py
```diff
@@ -0,0 +1,54 @@
+"""Link-local probing regressions; no Docker daemon required."""
```
fujita, line 1: No CI job runs test/unit/. .github/workflows/ci.yml runs pytest only on test/scenario_test/, and this pull request does not touch it. So this file never runs. pytest test/unit on its own also fails, because importing lib needs PYTHONPATH=test.

So the question is whether to add a job for it. I would rather drop the file.

Every test here runs against Mock(), so the test data decides the result, not the code. test_scoped_ping_does_not_require_reachable_neighbor_cache and test_diagnostics_preserve_original_timeout assert the command strings that the functions build a few lines above, so they cannot catch a bug. They only make the code harder to change. My two other comments in this review show it: changing -W 1 to -W 2 breaks one test, and dropping the second ping breaks three. Neither is a bug.

test_transient_ping_failure_retries_both_directions is the one that checks something real. A ping failure must not abort the probe, which is what the old code got wrong. But the CalledProcessError comes from the test's own side_effect. The real one comes from base.local(), which the mock replaces. The test stays green if local() stops raising it.

The test/lib/utils.py change itself looks good.

fujita (changes requested):

PR #3643, opened 2026-09-18
Clean labeled Docker resources around scenario tests

Ivan-Pokhabov:
Clean containers and networks labeled gobgp-test before and after pytest sessions and after each completed test topology. Preserve shared topology through tearDownClass, skip cleanup during collection and scenario discovery, and provide an opt-out for debugging. Document --timeout-scale.


test/lib/base.py
```diff
@@ -208,6 +208,24 @@ def get_containers():
...
+    ).split()
+    if containers:
+        local('docker rm -f {0}'.format(' '.join(containers)), capture=True)
+
+    networks = local(
+        'docker network ls -q -f label={0}'.format(TEST_NETWORK_LABEL),
+        capture=True,
+    ).split()
+    if networks:
+        local('docker network rm {0}'.format(' '.join(networks)), capture=True)
```
fujita, line 226: How about:
`local('docker network prune -f --filter label={0}'.format(TEST_NETWORK_LABEL), capture=True)`

`docker network rm` fails if the network is already gone or still has an endpoint attached, and `local()` turns that into `CalledProcessError`. `docker network prune` exits 0 in both cases.

Many scenario tests never create a `Bridge`, so they have no labeled network at all. One leftover network should not stop them.

test/unit/test_docker_cleanup.py
```diff
@@ -0,0 +1,78 @@
+"""Docker cleanup and pytest lifecycle regressions; no daemon required."""
```
fujita, line 1: No CI job runs `test/unit/`. `.github/workflows/ci.yml` runs pytest only on `test/scenario_test/`, and this pull request does not touch it. So this file never runs.

So the question is whether to add a job for it. I would rather drop the file.

The two `cleanup_docker_leftovers` tests assert the command strings that the function builds three lines above, so they cannot catch a bug. They only make the function harder to change.

`test_pytest_lifecycle` checks something real, but the scenario tests already cover it. `bgp_router_test.py` creates its containers once in `setUpClass` and shares them across 27 test methods. If the class boundary check broke and cleanup ran between methods, the second method would fail. The `tearDownClass` part guards nothing today: no scenario test defines one.

fujita (changes requested):

PR #3644, opened 2026-09-18
metrics: add the peer group and description labels

fdomain:
Adds `peer_group` and `description` to every peer-scoped metric exported by the BGP collector.

### Motivation

The peer/routes metrics are keyed only by the neighbor address, so any query that aggregates by peer group (or that needs to name a neighbor in a panel or an alert) has to bring the neighbor-to-group mapping in from outside Prometheus, through service discovery relabeling or a separate exporter, and then join on `peer`.

Exposing the metadata as its own info metric would remove the external dependency, but not the join:

```promql
# with an info metric
sum by (peer_group) (
  bgp_routes_received * on(peer) group_left(peer_group) bgp_peer_info
)

# with this change
sum by (peer_group) (bgp_routes_received)


fujita:
Merged, thanks.

PR #3646, opened 2026-09-19
table: make an AS_SET origin invalid when a ROA covers the prefix

Mounika2456:
`ROATable.Validate` with one VRP, 10.0.0.0/16-24 AS65000, and the `rpki-validation-result = "invalid"` condition from docs/sources/rpki.md, on master:

    10.0.0.0/24 as-path 65001          -> invalid   reject-invalid statement matches: true
    10.0.0.0/24 as-path 65001 {65000}  -> not-found reject-invalid statement matches: false
    10.0.0.0/24 as-path 65001 {65002}  -> not-found reject-invalid statement matches: false

`Validate` returns not-found as soon as the final AS_PATH segment is an AS_SET and never looks at the covering VRPs. RFC 6811 Section 2 gives such a route the origin "NONE", which no VRP matches, so a covered route is invalid; RFC 6907 Sections 7.1.9 to 7.1.11 list the same three cases `TestValidate7` had as not-found. A peer announcing a prefix it has no ROA for only needs to append an AS_SET to get past an import policy that rejects invalid routes.

The lookup now runs with the origin left at zero, which the existing `r.AS != 0 && r.AS == as` test never matches. A route with no covering VRP stays not-found (RFC 6907 Section 7.1.8).

AI tooling was used to help prepare this change.


fujita:
Thanks!

PR #3664, opened 2026-09-21
metrics: split bgp_peer_state into a session and an admin metric

fdomain:
bgp_peer_state reported 1 and carried both the session state and the administrative state in labels, although the two are independent. The value of a label is meant to stay as stable as it can over the life of a series, because changing it ends one series and starts another. Both of these states change by nature, so a transition of either one churned the series of the peer, which breaks changes() and plotting the state over time.

Report bgp_peer_session_state and bgp_peer_admin_state instead, each with the value of its API enum, so that a peer keeps one series whatever its state. Reporting a state as the value of the metric is also what bgp_peer_type does, so the collector becomes consistent.

The help text of both lists their values and is built from the enum names, so it cannot drift from what is reported, as it did for bgp_peer_type until 87e90f07 ("metrics: fix the peer type values in the bgp_peer_type help text").


pkg/metrics/metrics.go
```diff
@@ -386,17 +404,11 @@ func (c *bgpCollector) Collect(out chan<- prometheus.Metric) {
...
-			peerAddr,
-			peerPg,
-			peerDesc,
-			peerState.GetSessionState().String(),
-			peerState.GetAdminState().String(),
-		)
+		sendGauge := func(desc *prometheus.Desc, v float64) {
+			out <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, v, peerAddr, peerPg, peerDesc)
+		}
+		sendGauge(bgpPeerSessionStateDesc, float64(peerState.GetSessionState()))
```
fujita, line 409: Please add a test for these two values. TestMetrics already waits for the session to reach ESTABLISHED, and the file has a getMetric helper, so one Gather right after <-stateCh checks the metric names, the enum numbers and the gauge type at once.
```
	metrics, err := registry.Gather()
	assert.NoError(err)
	sessionState := getMetric(metrics, "bgp_peer_session_state")
	require.NotNil(sessionState)
	assert.Equal(float64(api.PeerState_SESSION_STATE_ESTABLISHED),
		sessionState.Metric[0].Gauge.GetValue())
```

The value now comes from the proto enum, and that numbering has changed once already. In 6b792afc AdminState was UP = 0, DOWN = 1, PFX_CT = 2. Today it is UNSPECIFIED = 0, UP = 1, DOWN = 2, PFX_CT = 3. If that happens again, the metric reports a number that is still valid but means something else. A rename makes a query return nothing, which is easy to notice. A renumbering is not visible at all. The test makes it visible.

pkg/metrics/metrics.go
fujita: While at it, enumValues works for this one too. This help text is the one 87e90f07 had to fix, it is still hand written, and it does not list UNSPECIFIED (0).
```
	"Type of the BGP peer: "+enumValues(api.PeerType_name, "PEER_TYPE_")
```

docs/sources/metrics.md
```diff
@@ -60,7 +60,8 @@ last column lists the labels a metric adds on top of those three.
 
 | **Metric**                         | **Description**                                                              | **Additional labels**                     |
 | ---------------------------------- | ---------------------------------------------------------------------------- |-------------------------------------------|
-| bgp_peer_state                     | State of the BGP session with peer and its administrative state              | `session_state`, `admin_state`            |
+| bgp_peer_session_state             | Session state of the peer                                                    |                                           |
```
fujita: The bgp_peer_type row below documents its values, "internal (1) or external (2)". These two rows do not. The HELP text has the list, but a reader of the docs cannot tell that 6 is ESTABLISHED. Please put the values here too.

The admin_state line you removed was also the only place that said PFX_CT means the prefix limit was reached. That should stay somewhere.

pkg/metrics/metrics.go
```diff
@@ -386,17 +404,11 @@ func (c *bgpCollector) Collect(out chan<- prometheus.Metric) {
...
-			bgpPeerStateDesc,
-			prometheus.GaugeValue,
-			1.0,
-			peerAddr,
-			peerPg,
-			peerDesc,
-			peerState.GetSessionState().String(),
-			peerState.GetAdminState().String(),
-		)
+		sendGauge := func(desc *prometheus.Desc, v float64) {
```
fujita: The commit message says that reporting the state as the value is what bgp_peer_type does, so the collector becomes consistent. The type does not match though. bgp_peer_type still goes through send, which always passes prometheus.CounterValue, so it is exported as "# TYPE bgp_peer_type counter"
while these two are gauges. Please move bgp_peer_type to sendGauge as well.

Five more metrics that are not counters go through send today. That is not your doing and not for this PR.

fujita (changes requested):

pkg/metrics/metrics.go
fdomain: Yes, it was in my visor as well, I wanted to do it in another PR but I can do it now.
UNSPECIFIED is never emitted because [toPeerType](https://github.com/osrg/gobgp/blob/31c8da1b99b38f917860aa9f6b2a0311d3fb25c9/pkg/config/oc/util.go#L474-L481) maps every non-external value to
internal, but I can still add it to the help text.

pkg/metrics/metrics.go
fdomain: Yes, I'll probably send another PR for the other metrics that are wrongly categorized as Counter.

pkg/metrics/metrics.go
fdomain: added test for session/admin state and peer type

fujita:
Thanks!

PR #3670, opened 2026-09-21
cmd: add TCP-AO commands

rastislavs:
Part of https://github.com/osrg/gobgp/issues/3493, adding CLI support for previously added TCP-AO implementation.

### Example usage

```bash
# List keychains
gobgp keychain
gobgp keychain fabric

# Add a keychain
gobgp keychain add fabric \
  --key '0,10,hmac-sha-1-96,emVybw==' \
  --key '1,11,aes-128-cmac-96,b25l,exclude-tcp-options'

# Update keys
gobgp keychain update fabric \
  --add-key '2,12,hmac-sha-256-96,dHdv' \
  --delete-key '1,11'

# Delete an unused keychain
gobgp keychain del fabric

# Configure TCP-AO on a neighbor
gobgp neighbor add 192.0.2.2 as 65002 \
  tcp-ao-keychain fabric \
  tcp-ao-send-id 0

# Show neighbor TCP-AO state
gobgp neighbor 192.0.2.2

BGP neighbor is 192.0.2.2, remote AS 65002
  BGP version 4, remote router ID 192.0.2.2
  BGP state = Established, up for 1d2h
  BGP OutQ = 0, Flops = 0
  Local address is 192.0.2.1, local ASN: 65001
  TCP-AO keychain is fabric, send ID: 0
  TCP-AO socket counters:
    Key not found: 0, AO required: 0, Dropped ICMP: 0
  TCP-AO socket key state:
    Send ID Receive ID Current Receive next Packets good Packets bad
          0         10    true        false        12450           0
          1         11   false         true            0           0
  Hold time is 90, keepalive interval is 30 seconds
  Configured hold time is 90, keepalive interval is 30 seconds
```

### Did you use an AI tool?

`Assisted-by: Codex (GPT-5) <noreply@openai.com>`
Modified manually and personally reviewed each line.


cmd/gobgp/tcp_ao.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+			keys, err := parseTcpAoKeys(addKeys, parseTcpAoKey)
+			if err != nil {
+				return err
+			}
+			_, err = client.AddTcpAoKeychain(ctx, &api.AddTcpAoKeychainRequest{Keychain: &api.TcpAoKeychain{Name: args[0], Keys: keys}})
+			return err
+		},
+	}
+	addCmd.Flags().StringArrayVar(&addKeys, "key", nil, "send-id,receive-id,algorithm,base64-master-key[,exclude-tcp-options] (repeatable)")
```
fujita: The master key goes on the command line here. That puts it in the shell history file, and in /proc/<pid>/cmdline, which is world-readable. I think that passing a secret as an argument is not a good interface in general.

BIRD keeps the TCP-AO keys in the config file and has no CLI command that takes a key. Its CLI only dumps the key state (`dump ao keys`). gobgp is the same today. The keychain can be set in the config file, and there is no CLI option for `auth-password` either.

If the CLI should be able to set a key, how about taking a file that holds the key, for example `--key 0,10,hmac-sha-1-96,file:/path/to/key`?

cmd/gobgp/tcp_ao.go
```diff
@@ -0,0 +1,264 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/osrg/gobgp/v4/api"
+	"github.com/spf13/cobra"
+)
+
+func parseTcpAoAlgorithm(value string) (api.TcpAoAlgorithm, error) {
+	switch strings.ToLower(value) {
+	case "hmac-sha-1-96", "hmac-sha1-96", "hmac_sha_1_96":
```
fujita: The config file takes only `hmac-sha-1-96`. This switch also takes `hmac-sha1-96` and `hmac_sha_1_96`, and `strings.ToLower` makes any case work.

So a spelling that works here fails in the config file:

    unsupported algorithm "hmac-sha1-96"

The new test uses `hmac-sha1-96`, which the config file rejects.

How about accepting only the exact `CryptoType` spellings (pkg/config/oc/bgp_configs.go:1075)? Adding a name later is easy, removing one is not.

docs/sources/cli-command-syntax.md
```diff
@@ -127,7 +128,9 @@ Also, refer to the following for the detail syntax of each address family.
 
 ```shell
 # add neighbor
-% gobgp neighbor add { <neighbor address> | interface <ifname> } as <as number> [ local-as <as number> | vrf <vrf-name> | route-reflector-client [<cluster-id>] | route-server-client | allow-own-as <num> | remove-private-as (all|replace) | replace-peer-as | ebgp-multihop-ttl <ttl>]
+% gobgp neighbor add { <neighbor address> | interface <ifname> } as <as number> [ local-as <as number> | vrf <vrf-name> | route-reflector-client [<cluster-id>] | route-server-client | allow-own-as <num> | remove-private-as (all|replace) | replace-peer-as | ebgp-multihop-ttl <ttl> | tcp-ao-keychain <name> | tcp-ao-send-id <0..255>]
+# update neighbor
```
fujita, line 132: `gobgp neighbor update` existed before this pull request. Only the docs were missing. That is a separate logical change. It does not need its own pull request, but please put it in its own commit, the first one here, and then add the TCP-AO options to both the add and the update line.

The new update line also drops `family` and `peer-group`, which `modNeighbor` accepts for update.

fujita (changes requested):

cmd/gobgp/tcp_ao.go
rastislavs: yeah good point, implemented loading from file as you suggested

cmd/gobgp/tcp_ao.go
rastislavs: makes sense, fixed

docs/sources/cli-command-syntax.md
rastislavs: moved to a separate commit and added missing family + peer-group

cmd/gobgp/tcp_ao.go
```diff
@@ -0,0 +1,279 @@
...
+		return nil, fmt.Errorf("TCP-AO master key must be specified as file:<path>")
+	}
+	masterKeyFile, err := os.Open(masterKeyPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open TCP-AO master key file %q: %w", masterKeyPath, err)
+	}
+	defer masterKeyFile.Close()
+	// Read one byte past the limit to distinguish an oversized key from one
+	// whose length is exactly TCPAOMaxKeyLen without reading the entire file.
+	masterKey, err := io.ReadAll(io.LimitReader(masterKeyFile, netutils.TCPAOMaxKeyLen+1))
```
fujita: The config file takes the master key as base64 (`secret-key`). This file is read as raw bytes, so a key cannot be moved between the two without an encode or a decode step.

How about reading the file as base64 too? It also removes a trap: `echo secret > key` makes a 7-byte key, because the trailing newline is part of it. Go's base64 decoder ignores `\n`, so `openssl rand -base64 32 > key` just works.

If you do this, the read limit has to move too. An 80-byte key encodes to 108 characters, so the current `TCPAOMaxKeyLen+1` cuts the file short and decodes to a 60-byte key with no error. A plain 256 is enough, and the length check on the decoded bytes still catches a file that is too long.

docs/sources/cli-command-syntax.md
```diff
@@ -629,3 +632,23 @@ If you want to remove one element(extended community) of ExtCommunitySet, to spe
...
+
+#### Syntax
+
+```shell
+# list all keychains or one named keychain
+% gobgp keychain [<name>]
+# create a keychain; --key is repeatable
+% gobgp keychain add <name> --key <send-id>,<receive-id>,<algorithm>,file:<master-key-path>[,exclude-tcp-options]
+# add and/or delete keys; both options are repeatable
+% gobgp keychain update <name> [--add-key <send-id>,<receive-id>,<algorithm>,file:<master-key-path>[,exclude-tcp-options>] [--delete-key <send-id>,<receive-id>]
```
fujita: Stray `>` and a missing `]`. It should be `[,exclude-tcp-options]]`?

fujita (changes requested):
The `docs: document neighbor update command` commit fixes something that is
missing on master today. A fix like that should come first, so it can be
merged even if the TCP-AO part needs more review. How about making it the
first commit, with the update line but without the TCP-AO options, and adding
those in the TCP-AO docs commit?

cmd/gobgp/tcp_ao.go
rastislavs: yeah makes much more sense indeed, updated

fujita:
Merged, thanks!

PR #3672, opened 2026-09-22
server: accept sessions on external listeners and dump BGP state

jagdishkumbhare:
Allow embedding GoBGP behind a non-host stack (for example gVisor) by taking caller-owned net.Listeners instead of binding TCP locally, and skip setsockopt on connections that are not OS sockets. Add DumpBGPServer for a JSON snapshot of global config, peers, and RIB.

# Did you use an AI tool?

If you did, say so here. See [Using AI tools](https://github.com/osrg/gobgp/blob/master/CONTRIBUTING.md#using-ai-tools).


fujita:
Thanks for the patch. Before going into the details, I want to understand the
use case.

1. What are you building? What runs GoBGP, who are the BGP peers, and why can
the host network stack not be used? Also, do you use the full gVisor sandbox
(runsc), or only the netstack library with gonet?

2. What about outgoing connections? connectLoop() in pkg/server/fsm.go uses
net.Dialer directly and cannot be replaced. With this patch a session comes up
only when the peer connects to us. Is that enough for you, or do you plan to
add the dial side later?

Accept is covered but dial is not, and the socket options are only skipped. If
all of them have to be covered, one transport abstraction may fit better than
adding listeners one by one. I have comments on the code itself, but they only
make sense once the direction is settled.

jagdishkumbhare:
Looks I raised this merge request by mistake for main, My intention was to merge it to our fork:

Use case. We are extending Nokia EDA’s testman: a process that simulates CE-side edge hosts attached to leaf/LAG ports so we can validate fabric intents (Router / VRF / BGPPeer) without extra hardware. Today those endpoints already do ping/ARP on simulated interfaces; we are adding BGP so the same endpoints can be real eBGP neighbors of the fabric.

Who runs GoBGP. GoBGP runs inside the testman process. Peers are the SRL/SROS nodes programmed by EDA BGPPeer / BGPGroup CRs. Testman listens; the NOS opens TCP/179 toward the simulated host.

Scale / why not the host stack. A deployment can have thousands of tagged EdgeInterfaces. Those VLAN/tagged endpoints are not created as Linux interfaces; they exist only as NICs in the userspace gVisor netstack. Host net.Listen cannot bind them, would not see fabric traffic on that path, and would collide with real host sockets. BGP must accept on the same gVisor NICs/IPs the fabric peers to.

gVisor. We use only the gVisor netstack Go library (pkg/tcpip + gonet), embedded in testman. We do not use runsc / a full gVisor sandbox.

Accept vs dial. Passive-only is intentional and enough for us. Peers use PassiveMode: true; we start with ListenPort: -1 and attach gVisor gonet listeners via AddExternalListener. The fabric dials us; we do not dial out and have no plan to add a dial side for this use case. Skipping host socket options on that path is fine — they do not apply to gVisor endpoints.

Current GoBGP limitation (why multiple instances today). GoBGP has VRF support, but the neighbor table is keyed by peer IP. The same neighbor IP in two VRFs cannot coexist in one server. Today we work around that with one GoBGP instance per VRF.

Future direction on our side. Move to a single GoBGP instance with multi-VRF: each VRF owns its listeners and BGP neighbors (and, if needed, we would push for neighbor keying that includes VRF, not IP alone). External listeners as in this patch remain the accept-side hook; a fuller transport abstraction would only matter if dial / full sockopt coverage became a requirement, which is out of scope for us.

